PR #11, opened 2018-11-29
Adding docs

gregkalapos:
Adding the very first doc files.

My goal is to have the infrastructure including the `configuration.asciidoc` in place. With this configs can be documented as they get implemented.

@bmorelli25 This builds locally. Can you please see if this is ok? 

Plus there is this sentence in `configuration.asciidoc`:
> To add multiple servers separate them with comma (`,`).

Not sure that's the right way to write it... Any feedback on that?

Thx!

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -22,7 +22,7 @@ To adapt the Elastic APM agent to your needs, you can configure it using environ
 | `http://localhost:8200` | List
 |============
 
-The URLs for your APM Servers. The URLs must be fully qualified, including protocol (`http` or `https`) and port. To add multiple servers separate them with comma (`,`).
+The URLs for your APM Servers. The URLs must be fully qualified, including protocol (`http` or `https`) and port. To add multiple servers, separate them with a comma (`,`).
 
 
 NOTE: Providing multiple URLs only works with APM Server v6.5+.
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -1,0 +1,28 @@
+[[configuration]]
+== Configuration
+
+To adapt the Elastic APM agent to your needs, you can configure it using environment variables.
+
+
+[[config-reporter]]
+=== Reporter configuration options
+[float]
+[[config-server-urls]]
+==== `server_urls`
+
+[options="header"]
+|============
+| Environment variable name
+| `ELASTIC_APM_SERVER_URLS`
+|============
+
+[options="header"]
+|============
+| Default                 | Type
+| `http://localhost:8200` | List
+|============
+
+The URLs for your APM Servers. The URLs must be fully qualified, including protocol (`http` or `https`) and port. To add multiple servers separate them with comma (`,`).
+
+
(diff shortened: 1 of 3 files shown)


docs/configuration.asciidoc
```diff
@@ -0,0 +1,28 @@
...
+| `ELASTIC_APM_SERVER_URLS`
+|============
+
+[options="header"]
+|============
+| Default                 | Type
+| `http://localhost:8200` | List
+|============
+
+The URLs for your APM Servers. The URLs must be fully qualified, including protocol (`http` or `https`) and port. To add multiple servers separate them with comma (`,`).
```
bmorelli25: > The URLs for your APM Servers.

Is this intended to be a title/heading?

docs/configuration.asciidoc
```diff
@@ -0,0 +1,28 @@
...
+
+
+NOTE: Providing multiple URLs only works if intake API v2 is enabled.
```
bmorelli25: I'm wondering if it might be better to mention 6.5 instead of v2:
```suggestion
NOTE: Providing multiple URLs only works with APM Server v6.5+.
```
Actually, is your agent even going to be compatible with less than 6.5? Might make more sense to put this elsewhere

docs/intro.asciidoc
```diff
@@ -0,0 +1,25 @@
...
+Elastic APM automatically measures the performance of your application and tracks errors.
+For example, it records spans for database queries and transactions for incoming HTTP requests.
+
+Spans are grouped in transactions - by default one for each incoming HTTP request.
+But it's possible to create custom transactions not associated with an HTTP request.
+
+By default the agent comes with support for ASP.NET Core out-of-the-box.
+To instrument other events, you can use custom spans.
+
+The agent is only one of multiple components you need to get started with APM.
```
bmorelli25: This is my mistake - forgot to mention that we just updated all of the APM Agents. Everything beyond this line should be changed to:

```
[float]
[[additional-components]]
=== Additional Components

APM Agents work in conjunction with the {apm-server-ref}/index.html[APM Server], {ref}/index.html[Elasticsearch], and {kibana-ref}/index.html[Kibana].
Please view the {apm-get-started-ref}/index.html[APM Overview] for details on how these components work together. 
```

docs/configuration.asciidoc
gregkalapos: Currently it works with less than 6.5 - or to be precise, it works with intake v1. I'd keep it with your suggestion here. If we kill v1 support in .NET (atm I don't plan to kill it) then this sentence is redundant, but still not incorrect, but if we keep v1 support then it's helpful.

docs/configuration.asciidoc
gregkalapos: No. Should it be? It's stolen from the [Python doc](https://github.com/elastic/apm-agent-python/blob/master/docs/configuration.asciidoc).

bmorelli25: My mistake, this is okay.

gregkalapos:
Thx for the help!

Issue #32, opened 2018-12-25
EntityFrameworkCore DiagnosticListener throws null ref 


josiahpeters:
We use IdentityServer4 for our Authentication provider. It is currently configured to use EntityFrameworkCore with SqlServer. A piece of functionality that comes out of the box with IdentityServer4 and EFCore as the backing store is a periodic task that runs in the background that deletes expired tokens from the database.

When this background task runs it triggers the OnNext method of the EfCoreDiagnosticListener. However because this task is not executed within the context of an HttpRequest, no transaction is created. Therefore when the background task executes a null reference exception is thrown because the TransactionContainer.Transactions.Value is null.

gregkalapos:
Thanks! This is indeed a bug.

Once [your PR](https://github.com/elastic/apm-agent-dotnet/pull/33) which fixes this gets merged, this can be closed.

gregkalapos:
Fixed by https://github.com/elastic/apm-agent-dotnet/pull/33 

Thanks!

Issue #3, opened 2018-11-08
Add Entity Framework Core telemetry through DiagnosticSource


gregkalapos:
Subscribe to "Microsoft.EntityFrameworkCore" events and create spans based on those events.

SergeyKleyman:
@gregkalapos Isn't this already implemented?

gregkalapos:
Yes, it's technically implemented, I just did not feel confident to close it, since the test coverage is not that great.

I'd like to spend some time on this before GA and make sure it's properly covered - (e.g. having tests with some more db providers, not only SQLite). Assigned it to myself and moved it to milestone GA - probably doesn't need much more effort to close this.

gregkalapos:
Done, we are good here.

PR #33, opened 2018-12-25
Prevent EfCoreDiagnosticListener Null Ref Exception #32

josiahpeters:
Added null check to verify that we have a reference to an active transaction before we attempt to create a new span from it. #32

### DIFF
--- a/src/Elastic.Apm.EntityFrameworkCore/EfCoreDiagnosticListener.cs
+++ b/src/Elastic.Apm.EntityFrameworkCore/EfCoreDiagnosticListener.cs
@@ -21,8 +21,8 @@ namespace Elastic.Apm.EntityFrameworkCore
         {
             switch (kv.Key)
             {
-                case string k when k == RelationalEventId.CommandExecuting.Name:
-                    if (TransactionContainer.Transactions.Value != null && kv.Value is CommandEventData commandEventData)
+                case string k when k == RelationalEventId.CommandExecuting.Name && TransactionContainer.Transactions.Value != null:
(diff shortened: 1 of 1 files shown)


gregkalapos:
Thanks for the PR @josiahpeters!

Very good point! By thinking about this I’d say we should abort as early as possible in each “case” block, so I moved the check directly to the case statement. 

With this in the `CommandExecuted` case we have a nullcheck instead of a failing `ConcurrentDictionary.TryRemove`, which is more preferable.

Are you ok with this?

Also, could you please sign [our CLA](https://www.elastic.co/contributor-agreement)?

2 comments:

- EF Core support is very much work-in-progress. It’s tracked [here](https://github.com/elastic/apm-agent-dotnet/issues/3). This isn’t even covered by tests at the moment - the only non tested part atm. It’ll get more love soon, my point is: things may change around it and please keep in mind that we are in a very pre-release stage.

- If you want to monitor that part of the code I recently committed the agent API that enables manual instrumentation, so you can start a transaction manually by calling `ElasticApm.StartTransaction(“MyTransactionName”, “MyTransactionType”)`. [This branch](https://github.com/elastic/apm-agent-dotnet/tree/AgentApi) has more chances, so that also may change later.

Let me know if you are happy with the suggested change.

elasticcla:
Hi @josiahpeters, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/apm-agent-dotnet/pull/33.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

PR #35, opened 2018-12-25
Feature/gh 34 dependency version

josiahpeters:
I changed the aspnet core dependency versions to 2.0.0 to allow projects that were freshly created when 2.0.0 aspnetcore was shipped the ability to work with Elastic.Apm.AspNetCore


elasticcla:
Hi @josiahpeters, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/apm-agent-dotnet/pull/35.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

Issue #26, opened 2018-12-13
Create API for manual instrumentation


gregkalapos:
Scope:
Create the public API. E.g. [this is the Java Agent API](https://www.elastic.co/guide/en/apm/agent/java/master/public-api.html).

gregkalapos:
Almost done, I'm thinking about moving a couple of types to other namespaces, and tags are missing, but other than that it's done. We have lots of [tests](https://github.com/elastic/apm-agent-dotnet/tree/master/test/Elastic.Apm.Tests/ApiTests) covering this and it's also [documented](https://github.com/elastic/apm-agent-dotnet/blob/master/docs/public-api.asciidoc).

gregkalapos:
Done, documentation is [here](https://github.com/elastic/apm-agent-dotnet/blob/master/docs/public-api.asciidoc) tests are [here](https://github.com/elastic/apm-agent-dotnet/tree/master/test/Elastic.Apm.Tests/ApiTests).

PR #37, opened 2018-12-31
Documenting the public agent API #26 + Contrib. guideline 

gregkalapos:
Catching up with documentation.

- Changes in `asciidoc` files: those document the newly added agent API.
- `.cs` files: Also added documentation directly in the code. This is a common practice in C# and that text is visible to users in the IDE when they use the API. I'd say this is very important, so I'd appreciate if we could also review those changes.
- `md` files: mainly helping contributors.

Couple of comments:
- Generally, the basis of the changes come from the java agent again.
- I see that some agents split the what I have in the 'public-api.asciidoc' into multiple files, and some don't. I think at this stage for .NET a single file is a better fit, otherwise we'd have some files with just a couple of lines. Later we can split.

Thx!


docs/public-api.asciidoc
```diff
@@ -0,0 +1,265 @@
...
+
+NOTE: The <<convenient-capture-transaction>> method does not start any `Task` for you, so you have to call it with the `await` keyword if you use one of the overloads that except `Task` or `Task<T>`.
+
+
+//----------------------------
+[float]
+[[api-transaction]]
+=== Transaction API
+//----------------------------
+A transaction is the data captured by an agent representing an event occurring in a monitored service
```
bmorelli25: I'm going to need to fix this paragraph in the java agent too - it's a bit difficult to understand. Here's what I'm thinking:

A transaction describes an event captured by an Elastic APM agent monitoring a service. Transactions help combine multiple <<api-span,`Spans`>> into logical groups, and they are the first <<api-span,`Span`>> of a service. More information on Transactions and Spans is available in the {apm-overview-ref-v}/apm-data-model.html[APM data model] documentation.

docs/public-api.asciidoc
```diff
@@ -0,0 +1,265 @@
...
+and groups multiple spans in a logical group.
+A transaction is the first <<api-span,`Span`>> of a service, and is also known under the term entry span.
+
+See <<api-current-transaction>> on how to get a reference of the current transaction.
+
+NOTE: Calling any of the transaction's methods after <<api-transaction-end>> has been called is illegal.
+You may only interact with transaction when you have control over its lifecycle.
```
bmorelli25:
```suggestion
You may only interact with a transaction when you have control over its lifecycle.
```

docs/public-api.asciidoc
```diff
@@ -0,0 +1,265 @@
...
+    span.End();
+}
+----
+
+
+[float]
+[[api-transaction-end]]
+==== `void End()`
+Ends the transaction and schedules it to be reported to the APM Server.
+It is illegal to call any methods on a span instance which has already ended.
```
bmorelli25: I think you should add a blank line before this sentence to maintain the layout consistency of the previous topics

docs/public-api.asciidoc
```diff
@@ -0,0 +1,265 @@
...
+
+[float]
+[[api-transaction-capture-exception]]
+==== `void CaptureException(Exception e)`
+Captures an exception and reports it to the APM server.
+
+[float]
+[[api-transaction-capture-error]]
+==== `void CaptureError(string message, string culprit, StackFrame[] frames)`
+Captures a custom error and reports it to the APM server. This method is typically used when you want to report an error, but you don't have an `Exception` instance.
```
bmorelli25: I think you should add a blank line after:

> Captures a custom error and reports it to the APM server.

docs/public-api.asciidoc
```diff
@@ -0,0 +1,265 @@
...
+** `Func<T>`
+** `Func<ITransaction,T>`
+** `Func<Task>`
+** `Func<ITransaction,Task>`
+** `Func<Task<T>>`
+** `Func<ITransaction,Task<T>>`
+* `supType` (optional): The subtype of the span
+* `action` (optional): The action of the span
+
+The following code is the equivalent of previous example from the section <<api-transaction-create-span>> with the convenient API. It automatically starts and ends the span and reports unhandled exceptions. The `s` parameter gives you access to the `ISpan` instance which represents the span that you just created.
```
bmorelli25:
```suggestion
The following code is the equivalent of the previous example from the <<api-transaction-create-span>> section with the convenient API. It automatically starts and ends the span and reports unhandled exceptions. The `s` parameter gives you access to the `ISpan` instance which represents the span that you just created.
```

docs/public-api.asciidoc
```diff
@@ -0,0 +1,265 @@
...
+----
+ITransaction transaction = Elastic.Apm.Agent.Api.CurrentTransaction;
+
+transaction.CaptureSpan("SampleSpan", Span.TYPE_DB, (s) =>
+{
+    //execute db query
+},Span.SUBTYPE_MSSQL, Span.ACTION_QUERY);
+----
+
+Similarly to the <<convenient-capture-transaction>> API this method also supports `async` methods with the `Func<Task>` overloads.
```
bmorelli25:
```suggestion
Similar to the <<convenient-capture-transaction>> API, this method also supports `async` methods with the `Func<Task>` overloads.
```

docs/public-api.asciidoc
```diff
@@ -0,0 +1,265 @@
...
+ITransaction transaction = Elastic.Apm.Agent.Api.CurrentTransaction;
+var asyncResult = await transaction.CaptureSpan("Select FROM customer", Span.TYPE_DB, async(s) =>
+{
+    //application code that is captured as a span
+    await Task.Delay(500); //sample async code
+    return 42;
+});
+----
+
+NOTE: The <<convenient-capture-span>> method does not start any `Task` for you, so you have to call it with the `await` keyword if you use one of the overloads that except `Task` or `Task<T>`.
```
bmorelli25:
```suggestion
NOTE: The <<convenient-capture-span>> method does not start any `Task` for you, so you have to call it with the `await` keyword if you use one of the overloads that accept `Task` or `Task<T>`.
```

docs/public-api.asciidoc
```diff
@@ -0,0 +1,265 @@
...
+//----------------------------
+[float]
+[[api-span]]
+=== Span API
+//----------------------------
+A span contains information about a specific code path, executed as part of a transaction.
+
+If for example a database query happens within a recorded transaction,
+a span representing this database query may be created.
+In such a case the name of the span will contain information about the query itself,
```
bmorelli25:
```suggestion
In such a case, the name of the span will contain information about the query itself,
```

docs/public-api.asciidoc
```diff
@@ -0,0 +1,265 @@
...
+
+[float]
+[[api-span-capture-exception]]
+==== `void CaptureException(Exception e)`
+Captures an exception and reports it to the APM server.
+
+[float]
+[[api-span-capture-error]]
+==== `void CaptureError(string message, string culprit, StackFrame[] frames)`
+Captures a custom error and reports it to the APM server. This method is typically used when you want to report an error, but you don't have an `Exception` instance.
```
bmorelli25: blank line after

> Captures a custom error and reports it to the APM server.

docs/public-api.asciidoc
```diff
@@ -0,0 +1,265 @@
...
+
+[float]
+[[api-span-end]]
+==== `void End()`
+Ends the span and schedules it to be reported to the APM Server.
```
bmorelli25: blank line after this sentence

docs/public-api.asciidoc
```diff
@@ -0,0 +1,265 @@
...
+
+[float]
+[[api-start-transaction]]
+==== `Transaction StartTransaction(string name, string type)`
+Use this method to create a custom transaction.
+
+
+Note that in case of auto-instrumentation the agent will automatically do this for you. For example in case of incoming HTTP calls in an ASP.NET Core application the agent automatically starts a transaction.
+
+You only need to use this method to create custom transactions.
```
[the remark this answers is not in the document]
gregkalapos: Sounds good, done.

docs/public-api.asciidoc
gregkalapos: Ok, I agree with the text, but I think something is not right here. I can build the docu, but in the generated html I don't see this part. Can you please look at this? Is it maybe this link: `{apm-overview-ref-v}/apm-data-model.html[APM data model] `?

docs/public-api.asciidoc
gregkalapos: done

docs/public-api.asciidoc
gregkalapos: done

docs/public-api.asciidoc
bmorelli25: This is really weird. The link and paragraph both build fine locally for me. Can you do me a favor and update your local copy of the `elastic/docs` repository and try building again?

gregkalapos: You are right, after I updated my local `elastic/docs` it works. Thank you! ...it was super strange, because before that, the paragraph just did not appear, but I haven't got any error message or warning.

gregkalapos:
Thank you @bmorelli25! 🎉🎉 As mentioned in the comment, now it also works for me locally - your suggestion helped.

docs/public-api.asciidoc
bmorelli25: Yeah, this is one of the many reasons we're moving from `asciidoc` to `asciidoctor`. In this instance, the docs repo wasn't updated, so the attribute for `{apm-overview-ref-v}` didn't exist locally. Asciidoc then _silently_ fails when you attempt to build the docs. The failure removes any text involved with the link. In other words, if you ever are missing text from an asciidoc file, it's likely due to an attribute being invalid or missing.

gregkalapos: Good to know, thanks for the info!

PR #41, opened 2019-01-03
[APM-CI] enable notifications

kuisathaverat:
* add the environment variables to access Google cloud service and send email notifications
* remove withEnvWrapper step

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -34,6 +34,7 @@ pipeline {
         */
         stage('Checkout') {
           steps {
+            deleteDir()
             gitCheckout(basedir: "${BASE_DIR}")
             stash allowEmpty: true, name: 'source', useDefaultExcludes: false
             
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -64,37 +66,36 @@ pipeline {
...
+            dir("${BASE_DIR}"){
+              sh '''#!/bin/bash
+              set -euxo pipefail
+              
+              # install tools
+              dotnet tool install -g dotnet-xunit-to-junit
+              for i in $(find . -name '*.??proj') 
+              do 
+                dotnet add "$i" package XunitXml.TestLogger --version 2.0.0
+                dotnet add "$i" package coverlet.msbuild
```
Mpdreamz: Would be better if these dependencies were not hidden away in the jenkins file but embedded conditionally in the projects through `<PackageReference .... Condition="" />` so that CI can do e.g `dotnet test /p:UseCover=true /p:XmlLogger=true` and use those msbuild properties in the condition for the package reference. `dotnet test` already does a build so `dotnet build` is not needed.

kuisathaverat: Those are packages used only on the CI, it is true that not be included in the project makes CI's scripts a little longer, if include those packages in the project does not affect to anything else, it would be really nice to have then inside the project.

PR #45, opened 2019-01-03
Fix sample application

gregkalapos:
Until this point we haven't created the database for the sample application. 

To make it easier we automatically do it from the C# code after this PR.


gregkalapos:
Master is currently failing with: 
`+ dotnet tool install -g dotnet-xunit-to-junit`.

Will wait with merging this one until #41  and #43 are merged. Those should make master green.

sample/SampleAspNetCoreApp/Controllers/HomeController.cs
```diff
@@ -21,6 +22,7 @@ public HomeController(SampleDataContext sampleDataContext)
 
         public async Task<IActionResult> Index()
         {
+            _sampleDataContext.Database.Migrate();
```
SergeyKleyman: Do we use _ prefix for fields?

PR #46, opened 2019-01-03
Enforced codestyles through editorconfig and resharper settings.

Mpdreamz:
The `.editorconfig` file settings are picked up by Roslyn and Resharper/Rider and thus pretty ubiquitous across all editors.
The `.DotSettings`file is picked up by Resharper/Rider. 

The `DotSettings` file mostly deals with reordering and is less opinionated version of the `.DotSettings` @russcam blogged about here: https://forloop.co.uk/blog/resharper-and-semanticmerge-equals-streamlined-development.

It basically makes sure fields/statics/consts are above constructors and properties follow afterwards. It doesn't sort methods at all. I would take a screenshot of RResharper visualizing it in its settings but I don't have VS installed (😱). Rider does not as of yet allow you to edit ordering rules sadly.

The `.editorconfig` settings are quite strict and most get marked as error (won't fail the build though).

* Tabs, currently the code base is spaces.
* No `this` keyword. I'm a fan of the `this` keyword but lost that to a quorum from @russcam @codebrain 😸 
* Preferring builtin types over class names e.g `string` over `String`.
* Very strong preference for expression bodied members.
* `var` everywhere 🌈 
* braces are set to [required_for_complex](https://www.jetbrains.com/help/resharper/EditorConfig_CSHARP_CSharpCodeStylePageImplSchema.html#resharper_csharp_braces_for_lock). This might mean no braces for single line statements if you run cleanup. It won't complain if single lines have braces though.
* Wrap at `150` characters.

These are just the first couple that stood out to me looking at the apm agent solution. 
We just did an automated cleanup excercise on our active integration branches as well:

https://github.com/elastic/elasticsearch-net/pull/3484
https://github.com/elastic/elasticsearch-net/pull/3483
https://github.com/elastic/elasticsearch-net/pull/3482

Might give a sense of the styles it enforced on the client codebase.

cc @elastic/dotnet


.editorconfig
```diff
@@ -0,0 +1,132 @@
...
+csharp_style_conditional_delegate_call = true:error
+
+csharp_prefer_braces = false:warning
+csharp_preferred_modifier_order = public,private,protected,internal,static,extern,new,virtual,abstract,sealed,override,readonly,unsafe,volatile,async:error
+
+# ---
+# formatting conventions https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference#formatting-conventions
+
+# Newline settings (Allman yo!)
+csharp_new_line_before_open_brace = all:error
```
gregkalapos: This does not seem to work with my Visual Studio (Version 15.8.9). It simply ignores this and actually puts the opening brace to the previous line (the opposite we want..) when I autoformat. I found [this](https://github.com/dotnet/roslyn/issues/18010) which says that specifying `error` is not valid. Once I remove it, like this, it works:
`csharp_new_line_before_open_brace = all`. Is it that maybe until this point no one noticed this, because this would be the default?

gregkalapos:
Also, another question: Do we maybe want to vote again about a couple of things? 

- I guess we can bring back `this` 😈
- I'm also wondering if everyone agrees on using `required_for_complex`. I'd always force braces. Once I spent almost a day (which I'll never get back :( ) hunting for a bug caused by not having braces, like:
```
if(mystatement)
  doThis();
  doThat();
```
So I'd prefer forcing braces.

But other than that I'd be ok with going with this. Once this is merged I'll do the cleanup.

gregkalapos:
Btw regarding the failing build: @kuisathaverat is working on CI, once #43 and #41 get merged we should be green again.

.editorconfig
Mpdreamz: Good catch! will amend.

codebrain:
Once merged will pull into ODBC installer.

Mpdreamz:
I've updated the PR to use `required_for_multiline` instead of `required_for_complex`. This will favor using braces in most cases where the body is actually complex ironically 😸

Mpdreamz:
I've made some changes to our default file layout:

* puts fields above the constructor only if they are readonly
* Properties/Indexers/assignable fields are now a single unsorted block after constructors (allows backing fields to appear closer to their usage). No longer sorts by name to accomplish this but I feel this a good trade off.
* Abstract members appear first thing after the constructor (makes it easy what is expected to be implemented, sorted by kind then name).
* Interfaces now have their own type simplified type pattern
   * properties sorted by name
   * the rest (methods) sorted by name 
* Removed NUnit type pattern
* Removed interface implementation members ordering before other members.


image

image

image

cc @elastic/dotnet

PR #47, opened 2019-01-03
Add global.json to encode the minimum required SDK to build the solution

Mpdreamz:
https://docs.microsoft.com/en-us/dotnet/core/tools/global-json

Communicates to the world what SDK the project is build with + will cause a build failure if the user does not have it (or equivalent later version) installed.


gregkalapos:
Hmm, I'm not totally convinced that we need this. 

This breaks my local build and Visual Studio gives a super misleading error message (which I already now, but some people may have trouble interpreting)
screenshot 2019-01-04 at 01 05 51

The reason is that I have `2.1.403`, but you specified `2.1.302`, which does not match (as it's described in the [docs](https://docs.microsoft.com/en-us/dotnet/core/tools/global-json). Also if we would pin the SDK to my version, then it'd fail on `2.1.3yz`. Of course we could pin it to a specific version (including patch levels), and force everyone to install it, but I don't really see the benefit. 

We aren't currently dependent on a specific min. sdk version, I think it'd even work with a super old one. So my point: by doing this I'm afraid we would cause more failing builds. 

What do you think @Mpdreamz? cc @SergeyKleyman

Mpdreamz:
I'm a stickler for repeatable builds and with .NET core being self contained indicating the intended SDK version is of very high value.

I got the number `2.1.302` from the Jenkins file: https://github.com/elastic/apm-agent-dotnet/blob/master/Jenkinsfile#L42 but global json should be the driver here in my opinion. The fact you build with `2.1.403` and I with `2.1.500` is fine during development but not when producing artifacts in my opinion.

In the .NET client we even prevent the rollover behaviour from our build scripts that produce artifacts to lock down repeatable builds:
https://github.com/elastic/elasticsearch-net/blob/master/build/scripts/Building.fsx#L33-L35

gregkalapos:
We discussed this and we decided to not have `global.json` for now. 

Reason: currently we don't really depend on any SDK version and we'd prefer to have an easy experience to start working on the code-base. By introducing `global.json` this would not be the case, because most of the people who'd work on the code probably'd have to install the specific SDK version we use.

On the other hand if we see that building with different SDKs can potentially cause problems (especially when 3.x and newer versions get released) we may consider adding a `global.json` file. 

@Mpdreamz would you be ok with closing this?

Mpdreamz:
Outvoted :) closing.

gregkalapos:
I'm reopening this, since https://github.com/elastic/apm-agent-dotnet/issues/527 could have been prevented by having a `global.json` file. After seeing #527 I think it was not the right decision to outvote this PR.

gregkalapos:
Since 3.0 is already released I'd suggest going with the latest 3.0 CLI version. Any other opinions?

Mpdreamz:
We just moved our other codebases over to, would be good to get all of our .net repos over to 3.0 and C# 8.0 across the board.

PR #53, opened 2019-01-05
Experimental - Add ASP.NET + Full Framework sample

gregkalapos:
Adding an ASP.NET (Classic) sample application that references the 'Elastic.APM' project and sends a transaction to the server.


josiahpeters:
Looking forward to trying out the sample. A majority of the apps where I work are still full framework asp.net.

Thanks for adding this.

gregkalapos:
Hey @josiahpeters,

Great to hear from you! This is currently very experimental, but one thing which may help you is the recently added Public Agent API. It's documented [here](https://github.com/elastic/apm-agent-dotnet/blob/master/docs/public-api.asciidoc). 

In order to use it you can just reference the [Elastic.Apm](https://github.com/elastic/apm-agent-dotnet/tree/master/src/Elastic.Apm) project. It's a .NET Standard 2.0 library, so it works on 4.6.1+. With that you can manually create transactions and spans. And the usual warning: we are still not even beta, so please don't put it into your prod yet :)

Issue #13, opened 2018-11-29
Implement Intake V2


gregkalapos:

gregkalapos:
PR: #136

gregkalapos:
Done in https://github.com/elastic/apm-agent-dotnet/pull/136

PR #57, opened 2019-01-08
Preview/naming conventions

Mpdreamz:
This PR serves as a preview of what #46 will enforce when it comes to naming.


src/Elastic.Apm/Model/Payload/Request.cs
```diff
@@ -14,7 +14,7 @@ public class Request
     public class Socket
     {
         public bool Encrypted { get; set; }
-        public String Remote_address { get; set; }
+        public String RemoteAddress { get; set; }
```
Mpdreamz: I take it this relies on convention to bind to: `remote_address` we might want to annotate this.

src/Elastic.Apm/Model/Payload/Span.cs
```diff
@@ -62,17 +62,17 @@ public void End()
         {
             if (!Duration.HasValue)
             {
-                this.Duration = (DateTimeOffset.UtcNow - start).TotalMilliseconds;
+                this.Duration = (DateTimeOffset.UtcNow - _start).TotalMilliseconds;
             }
 
-            transaction?.spans.Add(this);
+            Transaction?.SpanCollection.Add(this);
```
Mpdreamz: I renamed `spans` since it conflicted with `Spans` public property. Maybe transaction needs `.RegisterSpan()` so we can encapsulate `Spans` (should that property on `Transaction` be a `IReadOnlyCollection<ISpan>` instead of an array copy?

gregkalapos: With the Intake V2 API of the apm-server we probably won't need that property. Currently we need that field because within a transaction I send all the spans, but in V2 if I'm not mistaken I can just send a span with it's `transaction_id` separately from the transaction, so we won't need references from a transaction to its spans et all. So let's keep it, and i'll get rid of it in #13.

gregkalapos: But to also answer the question: yes `IReadOnlyCollection<ISpan>` would make the current situation better.

Issue #61, opened 2019-01-09
Structured logging


gregkalapos:
Currently the logging interface is very simple. It's basically 1 method, with it's string parameter (which contains the log level, all the parameters, etc). Users have access to the log, and they can redirect it wherever they want by doing this:

```
Agent.Config.Logger = new CustomLogCollector();

class CustomLogCollector : AbstractLogger
{
	protected override void PrintLogline(string logline)
	{
		//send the logline or process it...
	}
}
```

@SeanFarrow suggested [here](https://github.com/elastic/apm-agent-dotnet/issues/31#issuecomment-452500948) that: 

> Even if I derive from AbstractLogger to send the log somewhere else, I’m stuck with a string and would have to use something else to extract data using regex or an alternative input such as LogStash.

Let's discuss if we could make this a nicer interface! I'm totally open to provide an interface with more details.

@SeanFarrow feel free to suggest a specific interface if you have something in mind. 

Thanks

SeanFarrow:
Should we be logging any exceptions, or do they not apply?

Mpdreamz:
`TState` could very well be an `Exception`. `Microsoft.Extensions.Logging.ILogging.Log` has these separated not sure we need too.

SeanFarrow:
Makes sense.

PR #63, opened 2019-01-10
Proposal for Elastic.Apm.Agent instantiation

Mpdreamz:
This reworks `Elastic.Apm.Agent` making it static class holding a lazy singleton instance of `ApmAgent` which is `internal`.

There is now an `Elastic.Apm.Setup()` method that allows you to control the lazy instantiation which will throw if the singleton materialized already and can no longer be configured. This is in favor of having too many (internal) static property setters.

This also allows all other components to explicitly declare their dependencies in the constructor without referring to the static singleton directly. To me this was a `ServiceLocator` in disguise and so this PR introduces several constructors.

All the tests now instantiate a local `ApmAgent` and can test in isolation.

`Elastic.Apm.Agent` no longer has all the `CreateLogger` and `SetLoggerMethods()`. I refactored to make the tests pass but once #61 is settled on this could be cleaned up even further.

### DIFF
--- a/src/Elastic.Apm.AspNetCore/ApmMiddlewareExtension.cs
+++ b/src/Elastic.Apm.AspNetCore/ApmMiddlewareExtension.cs
@@ -1,7 +1,11 @@
 ﻿using System.Collections.Generic;
+using System.Reflection;
 using Elastic.Apm.AspNetCore.Config;
 using Elastic.Apm.AspNetCore.DiagnosticListener;
+using Elastic.Apm.Config;
 using Elastic.Apm.DiagnosticSource;
+using Elastic.Apm.Logging;
+using Elastic.Apm.Model.Payload;
 using Elastic.Apm.Report;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.Extensions.Configuration;
@@ -24,14 +28,28 @@ namespace Elastic.Apm.AspNetCore
 			this IApplicationBuilder builder, IConfiguration configuration = null, IPayloadSender payloadSender = null
 		)
 		{
-			if (configuration != null) Agent.Config = new MicrosoftExtensionsConfig(configuration);
+			var service = new Service
+			{
+				Agent = new Service.AgentC
+				{
+					Name = Consts.AgentName,
+					Version = Consts.AgentVersion
+				},
+				Name = Assembly.GetEntryAssembly()?.GetName().Name,
+				Framework = new Framework { Name = "ASP.NET Core", Version = "2.1" }, //TODO: Get version
+				Language = new Language { Name = "C#" } //TODO
+			};
+			var config = configuration != null ? new MicrosoftExtensionsConfig(configuration, service: service) : null;
+			Agent.Setup(config);
+			return UseElasticApm(builder, Agent.Instance);
+		}
 
-			if (payloadSender != null) Agent.PayloadSender = payloadSender;
-
+		internal static IApplicationBuilder UseElasticApm(this IApplicationBuilder builder, ApmAgent agent)
+		{
 			System.Diagnostics.DiagnosticListener.AllListeners
-				.Subscribe(new DiagnosticInitializer(new List<IDiagnosticListener> { new AspNetCoreDiagnosticListener() }));
+				.Subscribe(new DiagnosticInitializer(new[] { new AspNetCoreDiagnosticListener(agent.Config.Logger) }));
 
-			return builder.UseMiddleware<ApmMiddleware>();
+			return builder.UseMiddleware<ApmMiddleware>(agent.Tracer);
 		}
 	}
 }
--- a/src/Elastic.Apm.AspNetCore/DiagnosticListener/AspNetCoreDiagnosticListener.cs
+++ b/src/Elastic.Apm.AspNetCore/DiagnosticListener/AspNetCoreDiagnosticListener.cs
@@ -11,8 +11,7 @@ namespace Elastic.Apm.AspNetCore.DiagnosticListener
 	{
 		private readonly AbstractLogger _logger;
 
-		public AspNetCoreDiagnosticListener()
-			=> _logger = Agent.CreateLogger(Name);
+		public AspNetCoreDiagnosticListener(AbstractLogger logger) => _logger = logger;
 
 		public string Name => "Microsoft.AspNetCore";
 
--- a/src/Elastic.Apm.EntityFrameworkCore/EfCoreDiagnosticListener.cs
+++ b/src/Elastic.Apm.EntityFrameworkCore/EfCoreDiagnosticListener.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Data;
 using Elastic.Apm.Api;
 using Elastic.Apm.DiagnosticSource;
+using Elastic.Apm.Logging;
 using Elastic.Apm.Model.Payload;
 using Microsoft.EntityFrameworkCore.Diagnostics;
 
@@ -12,6 +13,11 @@ namespace Elastic.Apm.EntityFrameworkCore
 	internal class EfCoreDiagnosticListener : IDiagnosticListener
 	{
 		private readonly ConcurrentDictionary<Guid, ISpan> _spans = new ConcurrentDictionary<Guid, ISpan>();
+
+		public EfCoreDiagnosticListener(AbstractLogger logger) => Logger = logger;
+
+		private AbstractLogger Logger { get; }
+
 		public string Name => "Microsoft.EntityFrameworkCore";
 
 		public void OnCompleted() { }
--- a/src/Elastic.Apm.EntityFrameworkCore/ElasticEntityFrameworkCoreListener.cs
+++ b/src/Elastic.Apm.EntityFrameworkCore/ElasticEntityFrameworkCoreListener.cs
@@ -1,6 +1,7 @@
 ﻿using System.Collections.Generic;
 using System.Diagnostics;
 using Elastic.Apm.DiagnosticSource;
+using Elastic.Apm.Logging;
 
 namespace Elastic.Apm.EntityFrameworkCore
 {
@@ -12,8 +13,8 @@ namespace Elastic.Apm.EntityFrameworkCore
 		/// <summary>
 		/// Start listening for EF Core diagnosticsource events
 		/// </summary>
-		public void Start() => DiagnosticListener
+		public void Start(AbstractLogger logger) => DiagnosticListener
 			.AllListeners
-			.Subscribe(new DiagnosticInitializer(new List<IDiagnosticListener> { new EfCoreDiagnosticListener() }));
+			.Subscribe(new DiagnosticInitializer(new [] { new EfCoreDiagnosticListener(logger) }));
 	}
 }
--- a/src/Elastic.Apm/Agent.cs
+++ b/src/Elastic.Apm/Agent.cs
@@ -3,6 +3,7 @@ using System.Runtime.CompilerServices;
 using Elastic.Apm.Api;
 using Elastic.Apm.Config;
 using Elastic.Apm.Logging;
+using Elastic.Apm.Model.Payload;
 using Elastic.Apm.Report;
 
 //TODO: It'd be nice to move this into the .csproj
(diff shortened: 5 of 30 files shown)


src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListener.cs
```diff
@@ -117,8 +111,7 @@ private bool IsRequestFiltered(Uri requestUri)
 		{
 			switch (requestUri)
 			{
-				case Uri uri when uri == null:
-					return true;
+				case Uri uri when uri == null: return true;
 				case Uri uri when Agent.Config.ServerUrls.Any(n => n.IsBaseOf(uri)): //TODO: measure the perf of this!
```
gregkalapos: We have failing tests, and I think this line introduces a race condition. It can be fixed by passing an `AbstractConfig` to the constructor and using that instead of `Agent.Config.ServerUrls`. I already have the change locally on my machine.

Tests fail with: 
```
System.InvalidOperationException : Collection was modified; enumeration operation may not execute.
   at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source, Func`2 predicate)
   at Elastic.Apm.DiagnosticListeners.HttpDiagnosticListener.IsRequestFiltered(Uri requestUri) in /Users/..../Repos/apm-agent-dotnet/src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListener.cs:line 120

```

gregkalapos: Btw. it's quite unfortunate that those tests were green in the CI. I executed all tests locally on my machine and some were sporadically red, but only if no debugger was attached.

gregkalapos:
> This also allows all other components to explicitly declare their dependencies in the constructor without referring to the static singleton directly. To me this was a ServiceLocator in disguise and so this PR introduces several constructors.

👍 My only comment to this is that I think most of the components will end up having `AbstractAgentConfig` as the constructor parameter (will already happen [here](https://github.com/elastic/apm-agent-dotnet/pull/63#discussion_r246953419)). Mainly because later we will have a bunch of settings on `AbstractAgentConfig` (see [here](https://www.elastic.co/guide/en/apm/agent/java/current/configuration.html)), so components that now get these new constructors will typically need those (`AbstractAgentConfig.[SomeConfigSetting]`) and the `AbstractAgentConfig.Logger`. Should we maybe introduce a contract for this? Like adding an abstract `IDiagnosticListener` implementation that takes `AbstractAgentConfig` as parameter? (alternatively turned that into an abstract class)? That way we can avoid accidentally exploding the number of constructor parameters later.

gregkalapos:
Thanks to this change we can also remove the `[assembly: CollectionBehavior(DisableTestParallelization = true)]` lines [here](https://github.com/elastic/apm-agent-dotnet/blob/master/test/Elastic.Apm.AspNetCore.Tests/AspNetCoreMiddlewareTests.cs#L10) and [here](https://github.com/elastic/apm-agent-dotnet/blob/master/test/Elastic.Apm.Tests/ConfigTest.cs#L7). Maybe we would have seen the flaky tests with that change.

src/Elastic.Apm/DiagnosticSource/ElasticCoreListeners.cs
```diff
@@ -12,8 +13,8 @@ public class ElasticCoreListeners
 		/// <summary>
 		/// Start listening for diagnosticsource events. Only listens for sources that are part of the Agent.Core package.
 		/// </summary>
-		public void Start() => DiagnosticListener
+		public void Start(AbstractLogger logger) => DiagnosticListener
```
gregkalapos: Can we default to `Agent.Config.Logger` here and make that parameter optional? That way we would have a less complex public API.

src/Elastic.Apm.EntityFrameworkCore/ElasticEntityFrameworkCoreListener.cs
```diff
@@ -12,8 +13,8 @@ public class ElasticEntityFrameworkCoreListener
 		/// <summary>
 		/// Start listening for EF Core diagnosticsource events
 		/// </summary>
-		public void Start() => DiagnosticListener
+		public void Start(AbstractLogger logger) => DiagnosticListener
```
gregkalapos: Similarly to `ElasticCoreListeners.cs` I think we should make this optional and default to `Agent.Config.Logger` (or `Agent.Config` if we change it...).

src/Elastic.Apm.EntityFrameworkCore/EfCoreDiagnosticListener.cs
```diff
@@ -12,6 +13,11 @@ namespace Elastic.Apm.EntityFrameworkCore
 	internal class EfCoreDiagnosticListener : IDiagnosticListener
 	{
 		private readonly ConcurrentDictionary<Guid, ISpan> _spans = new ConcurrentDictionary<Guid, ISpan>();
+
+		public EfCoreDiagnosticListener(AbstractLogger logger) => Logger = logger;
```
gregkalapos: Another one: Similarly to `ElasticCoreListeners.cs` I think we should make this optional and default to `Agent.Config.Logger` (or `Agent.Config` if we change it...).

src/Elastic.Apm.AspNetCore/DiagnosticListener/AspNetCoreDiagnosticListener.cs
```diff
@@ -11,8 +11,7 @@ public class AspNetCoreDiagnosticListener : IDiagnosticListener
 	{
 		private readonly AbstractLogger _logger;
 
-		public AspNetCoreDiagnosticListener()
-			=> _logger = Agent.CreateLogger(Name);
+		public AspNetCoreDiagnosticListener(AbstractLogger logger) => _logger = logger;
```
gregkalapos: Similarly to ElasticCoreListeners.cs I think we should make this optional and default to Agent.Config.Logger (or Agent.Config if we change it...).

src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListener.cs
```diff
@@ -24,22 +24,16 @@ public class HttpDiagnosticListener : IDiagnosticListener
 		/// </summary>
 		internal readonly ConcurrentDictionary<HttpRequestMessage, ISpan> ProcessingRequests = new ConcurrentDictionary<HttpRequestMessage, ISpan>();
 
-		private readonly AbstractAgentConfig _agentConfig;
+		public HttpDiagnosticListener(AbstractLogger logger) => Logger = logger;
```
gregkalapos: Default to `Agent.Config`?

src/Elastic.Apm/Api/Tracer.cs
```diff
@@ -3,50 +3,39 @@
...
 	internal class Tracer : ITracer
 	{
 		private static AbstractLogger _publicTracerLogger;
 
-		public ITransaction CurrentTransaction
-			=> TransactionContainer.Transactions.Value;
+		public ITransaction CurrentTransaction => TransactionContainer.Transactions.Value;
 
-		public static AbstractLogger PublicTracerLogger => _publicTracerLogger ?? (_publicTracerLogger = Agent.CreateLogger("AgentAPI"));
+		public static AbstractLogger PublicTracerLogger => _publicTracerLogger;
```
gregkalapos: I think this remains `null`. This would be a logger that is used by the public API.

src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListener.cs
gregkalapos: By thinking further I'd say in the `AbstractAgentConfig` we need to introduce some locks for the initialization of the configsettings (btw. I think that's the ultimate solution to this problem). But that's unrelated to this change set, so I'd not focus on that one here.

Mpdreamz: I ended up splitting `AbstractAgentConfig` into `AgentComponents` and `IConfigurationReader` to separate the two concerns.

`AgentComponents`  and `ApmAgent` both implement `IApmAgent` and can be used where something relies on one of the agent components. Both are internal so `Elastic.Apm.Agent` is still the only public facing way to setup and interact with the agent.

src/Elastic.Apm/DiagnosticSource/ElasticCoreListeners.cs
Mpdreamz: I moved these listeners and start behind `IDiagnosticSubscriber` so the public contract folks can just provide us with an instance see e.g:

```diff
- app.UseElasticApm(Configuration);
- new ElasticCoreListeners().Start();
- new ElasticEntityFrameworkCoreListener().Start();
+ app.UseElasticApm(Configuration, new HttpDiagnosticsSubscriber(), new EfCoreDiagnosticsSubscriber());
```

or `IApmAgent` could also have `Subscribe()` to help hook this up even cleaner.

src/Elastic.Apm/Api/Tracer.cs
Mpdreamz: I removed this property all together in the subsequent commits, Tracer now gets its logger passed in like the rest of the components 👍

Mpdreamz:
>Thanks to this change we can also remove the [assembly: CollectionBehavior(DisableTestParallelization = true)] lines here and here. Maybe we would have seen the flaky tests with that change.

Can confirm this works! 🍰

src/Elastic.Apm/DiagnosticSource/ElasticCoreListeners.cs
gregkalapos: Not sure we are ok here. The downside I think we still have is that once this is not started in the `ApmMiddlewareExtension` then activating those is still more complex then it should be - I'm not sure it's even possible now. Example scenario: .NET Core console application that wants to monitor outgoing `HttpClient` calls. In that case it would be nice to have the option to subscribe to those diagnostic source events (aka turn on outgoing HTTP monitoring) with a 1 liner. And that was `new ElasticCoreListeners().Start();` before this change. I think we should aim for something that is as simple as this one because it's a very basic feature and we can't be sure that we can turn this on automatically in all scenarios. Same applies to other thing (EFCore, etc..)

src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListener.cs
gregkalapos: 👍

src/Elastic.Apm/Api/Tracer.cs
gregkalapos: 👍

src/Elastic.Apm.AspNetCore/ApmMiddlewareExtension.cs
```diff
@@ -20,18 +27,42 @@ public static class ApmMiddlewareExtension
...
+					Version = Consts.AgentVersion
+				},
+				Name = Assembly.GetEntryAssembly()?.GetName().Name,
+				Framework = new Framework { Name = "ASP.NET Core", Version = "2.1" }, //TODO: Get version
+				Language = new Language { Name = "C#" } //TODO
+			};
+			var configReader = configuration != null ? new MicrosoftExtensionsConfig(configuration) : null;
+			var config = new AgentComponents(null, configurationReader: configReader, service: service);
+			Agent.Setup(config);
+			return UseElasticApm(builder, Agent.Instance);
```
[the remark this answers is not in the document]
Mpdreamz: Nice catch! will amend 😸

src/Elastic.Apm/Config/AgentComponents.cs
```diff
@@ -0,0 +1,41 @@
...
+	public class AgentComponents : IApmAgent
+	{
+		public AgentComponents(
+			AbstractLogger logger,
+			IConfigurationReader configurationReader = null,
+			Service service = null,
+			IPayloadSender payloadSender = null)
+		{
+			Service = service ?? Service.Default;
+			Logger = logger ?? NoopLogger.Instance;
```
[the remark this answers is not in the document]
Mpdreamz: Going with your preference here and will make ConsoleLogger the fallback.

src/Elastic.Apm/Config/AgentComponents.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Reflection;
+using Elastic.Apm.Api;
+using Elastic.Apm.Logging;
+using Elastic.Apm.Model.Payload;
+using Elastic.Apm.Report;
+
+namespace Elastic.Apm.Config
```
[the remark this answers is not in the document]
Mpdreamz: Its not, will move it up the chain 😸

src/Elastic.Apm/Config/IAgentConfigReader.cs
```diff
@@ -0,0 +1,87 @@
...
+	{
+		public ConfigurationKeyValue(string key, string value, string readFrom) =>
+			(Key, Value, ReadFrom) = (key, value, readFrom);
+
+		public string Key { get; }
+		public string ReadFrom { get; }
+		public string Value { get; }
+	}
+
+	public abstract class AbstractConfigurationReader
```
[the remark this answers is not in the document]
Mpdreamz: I moved this into its own file now.

src/Elastic.Apm/Agent.cs
```diff
@@ -62,6 +63,17 @@ public static class Agent
...
+		/// </summary>
+		/// <param name="subscribers">
+		/// An array of <see cref="IDiagnosticSubscriber" /> that will set up <see cref="IDiagnosticListener" /> subscriptions
+		/// </param>
+		/// <returns>
+		/// A disposable referencing all the subscriptions, disposing this is not necessary for clean up only to unsubscribe if desired.
+		/// </returns>
+		public static IDisposable Subscribe(params IDiagnosticsSubscriber[] subscribers) => Instance.Subscribe(subscribers);
+
 		public static void Setup(AgentComponents agentComponents)
```
Mpdreamz: Could potentially also add subscribers to `Setup()`, maybe something for a follow up PR.

gregkalapos: 👍  But I think for now this is fine.

src/Elastic.Apm.AspNetCore/ApmMiddlewareExtension.cs
```diff
@@ -58,10 +56,7 @@ public static class ApmMiddlewareExtension
 		)
 		{
 			var subs = new List<IDiagnosticsSubscriber>(subscribers ?? Array.Empty<IDiagnosticsSubscriber>());
-			subs.Add(new AspNetCoreDiagnosticsSubscriber());
```
gregkalapos: This one should not be removed. - that's how we capture exceptions within the request.

Mpdreamz: 👍

gregkalapos:
🚀🚀🚀 LGTM, Thx @Mpdreamz! 👍

Issue #72, opened 2019-01-21
Bug: Unsubscribing from DiagnosticSource events does not work


gregkalapos:
This is more pressuring for writing tests.

Currently code like this does not unsubscribe the agent from receiving events and creating spans:

```
using (agent.Subscribe(new HttpDiagnosticsSubscriber()))
{
	//do stuff			
}
//we should not be subscribed here, but we are.
```

Therefore subsequent tests that assume the `HttpDiagnosticsSubscriber` is not registered will fail.

gregkalapos:
The root cause was in [DiagnosticInitializer.cs](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm/DiagnosticSource/DiagnosticInitializer.cs ):

```
public void OnNext(DiagnosticListener value)
{
   foreach (var listener in _listeners)
   {
      if (value.Name == listener.Name) value.Subscribe(listener);
   }
}
```

`value.Subscribe` returns `IDisposable` which wasn't disposed. Fix is on the way.

gregkalapos:
Fixed by #75

PR #75, opened 2019-01-22
Fix DiagnosticSource + Disposable issues + service name doc

gregkalapos:
- Added service name doc

- Some refactorings:

1.
~~The `ApmMiddlewareExtension` automatically subscribes to EFCore and HttpClient events. Additionally, the `Elastic.Apm.AspNetCore` now references the `Elastic.Apm.EntityFrameworkCore` project. 
With this, registering the agent in ASP.NET Core became simpler: `app.UseElasticApm(Configuration);`
This was the old code:
`app.UseElasticApm(Configuration, new HttpDiagnosticsSubscriber(), new EfCoreDiagnosticsSubscriber());~~

**Update**: 
Added `Elastic.Apm.All` project that references every other agent project and contains an `ApmMiddlewareExtension` that automatically subscribes to HttpClient and EFCore events. This way in typical ASP.NET Core apps (ASP.NET Core + EF Core) that reference `Microsoft.AspNetCore.All` users can just reference our `Elastic.Apm.All` package and enable the agent with all its features (including EF Core and HttpClient monitoring) with `app.UseElasticApm(Configuration);`. In other cases (e.g. ASP.NET Core without EF Core, or even no ASP.NET Core) the other packages can be referenced individually to avoid unnecessary dependencies. In those scenarios the diagnostic listeners must be enabled manually.

2. Moved `TransactionContainer` into `AgentComponents`. With that `TransactionContainer.Transactions` became `Agent.TransactionContainer`, which we can reset in tests.
 
3. Fix for #72.

4. As a consequence of fixing #72 the `ApmAgent` also became Disposable. With this in tests we can dispose the agent.

- Tests


src/Elastic.Apm.AspNetCore/Elastic.Apm.AspNetCore.csproj
```diff
@@ -9,6 +9,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\Elastic.Apm.EntityFrameworkCore\Elastic.Apm.EntityFrameworkCore.csproj" />
```
Mpdreamz: This pulls in `Microsoft.EntityFrameworkCore.Relational` as transient dependency which might not go down well for folks who use asp .net core without EF.

gregkalapos:
Just for reference: the coverage report is probably wrong. @kuisathaverat is on it...

src/Elastic.Apm/DiagnosticSource/DiagnosticInitializer.cs
```diff
@@ -18,7 +18,10 @@ public void OnNext(DiagnosticListener value)
 		{
 			foreach (var listener in _listeners)
 			{
-				if (value.Name == listener.Name) value.Subscribe(listener);
+				if (value.Name != listener.Name) continue;
+
+				var sourceSubscription = value.Subscribe(listener);
+				listener.SourceSubscription = sourceSubscription;
```
Mpdreamz: I think it would be better if `DiagnosticInitializer` implements `IDisposable` and manages its own state that way. 

`DiagnosticsInitializer` behaves a bit like `Subject<T>` in that its both an observer of `DiagnosticListener` while subscribing to `IObservable<KeyValuePair<string, string>>`. It might be worth exploring making that more explicit.

This would put `DiagnosticsInitializer` in the driving seat without "tagging" the observers with a disposable.

gregkalapos:
@bmorelli25 `docs/configuration.asciidoc` has a new section. Could you please take a look?

src/Elastic.Apm/ApmAgentExtensions.cs
```diff
@@ -23,7 +23,12 @@ public static class ApmAgentExtensions
 		public static IDisposable Subscribe(this IApmAgent agent, params IDiagnosticsSubscriber[] subscribers)
 		{
 			var subs = subscribers ?? Array.Empty<IDiagnosticsSubscriber>();
-			return subs.Aggregate(new CompositeDisposable(), (d, s) => d.Add(s.Subscribe(agent)));
+			var retVal = subs.Aggregate(new CompositeDisposable(), (d, s) => d.Add(s.Subscribe(agent)));
+
+			if (agent is ApmAgent apmAgent)
```
Mpdreamz: nitpick `is IDisposable d)` since that is all we care about?

src/Elastic.Apm.AspNetCore/Elastic.Apm.AspNetCore.csproj
gregkalapos: Yes, I was also thinking about this and that is a valid scenario. On the other hand the [Microsoft.AspNetCore.All](https://www.nuget.org/packages/Microsoft.AspNetCore.All) package already references EF Core, so I think most if the users already have that. And with this the benefit is that users don't have to explicitly enable the EF Core listener, which I think should be enabled by default. So the negative side of the change: for a small portion of users we add an unnecessary reference, the upside of the change is that for most of the users we avoid wondering how to enable db monitoring.

src/Elastic.Apm/ApmAgentExtensions.cs
gregkalapos: 👍

src/Elastic.Apm/DiagnosticSource/DiagnosticInitializer.cs
gregkalapos: Thx, I adapted this accordingly.

src/Elastic.Apm/ApmAgentExtensions.cs
gregkalapos: Update: unfortunately that doesn't work, because the `ApmAgent` class collects these disposables, so we have to cast to `ApmAgent`. Due to this change the `ApmAgent` collects the return values form the `Subscribe` method, and with that we can `Dispose` the `ApmAgent` instance in tests to unsubscribe from all diagnostic source events.

src/Elastic.Apm.AspNetCore/Elastic.Apm.AspNetCore.csproj
lafriks: Maybe it would be possible to also create additional csproj/nuget package `Elastic.Apm.All` that would contain helper that would add all subscribers and by default keep references to minimum. At least in case of running netcore apps in docker it is nice to have them small as possible

gregkalapos: This is a good idea!

docs/configuration.asciidoc
```diff
@@ -50,6 +50,30 @@ This is a typical `appsettings.json` file that contains some sample configuratio
 }
 ----
 
+[[config-reporter]]
```
bmorelli25: When I build these changes I get an `element section: validity error` on this line. That's because the ID `config-reporter` is defined twice. It can only be defined once. You'll need to change this ID for the documentation to build.

gregkalapos: Good catch, thx! I wonder why this was green in the CI 🤔 

If you don't mind I'd go with `[[config-core]]` to keep the naming aligned with the other ones. Already committed.

bmorelli25: Huh, that's a good point, I'm not sure why the CI was green.

Yes, that name works perfectly!

src/Elastic.Apm.AspNetCore/Elastic.Apm.AspNetCore.csproj
gregkalapos: I implemented the suggestion from @lafriks, I think that's a clean way to handle this. Also updated the description.

PR #76, opened 2019-01-22
Report Agent version and report ASP.NET Core version

gregkalapos:
First #75 should be merged.

- Implemented reporting agent version based on the version of the `Elastic.Apm` assembly. 
- Implemented reporting the correct ASP.NET Core version based on the `Microsoft.AspNetCore` assembly.


src/Elastic.Apm/Model/Payload/Service.cs
```diff
@@ -24,7 +25,7 @@ internal static Service GetDefaultService(IConfigurationReader configurationRead
 				Agent = new AgentC
 				{
 					Name = Consts.AgentName,
-					Version = Consts.AgentVersion
+					Version = Assembly.GetExecutingAssembly().GetName().Version?.ToString() ?? "n/a",
```
lafriks: Might be better to use:
```
typeof(AgentC).Assembly.GetName().Version?.ToString()
```

gregkalapos: 👍  ...ended up using the `Agent` type, since these model classes can easily end up in a shared library (with different versioning) and the `Agent` would be probably always in the core lib.

Issue #77, opened 2019-01-24
Public vs Internal - Make sure we limit the things that are public initially


gregkalapos:
By looking at the code-base I'd say there are too many things that are public. 

E.g. on `ITransaction` there is a public `Spans` array, and this is an example why this can be a problem in user code:
```
class MySpan : ISpan
{
  //...all the ISpan stuff
}
		
public static void MyMethod()
{
   Agent.Tracer.CurrentTransaction.Spans[0] = new MySpan { Id = -1 , Duration = -10};
  //leave everything else empty, and send id -1 and everything else as empty values...
}
```
The server won't like this...And there are many other examples.

I'd say generally for the first release we should make sure we only make things public that 100% should be public. Adding things later won't break, so if there is something missing we can add it later easily. On the other hand removing things would break.

Therefore: I'll go on and make sure that we only have things in public that will very-very likely remain public forever, and make things that are questionable internal, just to be on the safe side.

gregkalapos:
Fixed by #78

yhvicey:
Should `MicrosoftExtensionsConfig` and `EnvironmentConfigurationReader` be made public again? Users can only implement their own `IConfigurationReader` or use `EnvironmentConfigurationReader` by passing `null` into the agent constructor.

gregkalapos:
> Should `MicrosoftExtensionsConfig` and `EnvironmentConfigurationReader` be made public again? Users can only implement their own `IConfigurationReader` or use `EnvironmentConfigurationReader` by passing `null` into the agent constructor.

I think this is how it should be. By default we always use `EnvironmentConfigurationReader`, and in case you pass an `IConfiguration` instance to the Agent in ASP.NET Core we switch to `EnvironmentConfigurationReader`. I don't really see any need for users to know about this, or know the internals of those two - of course no secret there, I just don't see the need to add additional complexity.

Do you have any use-case where this does not work well?

yhvicey:
We'd like to split our integration with ElasticApm into different pieces - One for general console app, one for AspNetCore, and one for EFCore. Besides, we'll have a uber one for AspNetCore apps with EFCore. Basically we are doing the same thing you did in this packages, but we want to leave some space for OpenTracing/OpenTelemetry and design some generic pattern for future integration/switching. I guess my best option is to implement my own `IConfigurationReader` here.

gregkalapos:
[...]

PR #78, opened 2019-01-25
Public cleanup

gregkalapos:
The purpose of this change is to make sure that the agent initially only has a minimal public API surface. Therefore lots of public things became internal. The intention is to avoid braking changes later. #77 

- `IDiagnosticListener` and all its implementors became internal. `IDiagnosticsSubscriber` and its implementors remain public. This way from the user's perspective it's easier to know how to subscribe: just use the subscribers, and don't care about the listeners, that's handled internally.
- Lots of classes from `Elastic.Apm.Model.Payload` became internal.
- Introduced `IError` interface, mainly because the `Error` class leaked lots of internal information that is not relevant to users.
- Removed a couple of properties from `ITransaction` and `ISpan`, which could potentially cause problems when users set data that is not accepted by the server. Those properties are available on `Transaction` and `Span`.
- Due to the previous point I added `internal Span StartSpanInternal` to `Transaction` and `internal Transaction StartTransactionInternal` to `Tracer`. These can be used in agent dlls to access properties that are not on the public interface.

- Plus added code signing to `SampleAspNetCoreApp.csproj` to avoid build warnings. (It is referenced from a test project, which is signed, so it also should be signed, which it is from now on).

### DIFF
--- a/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
+++ b/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
@@ -16,12 +16,12 @@ using Microsoft.AspNetCore.Http;
 namespace Elastic.Apm.AspNetCore
 {
 	// ReSharper disable once ClassNeverInstantiated.Global
-	public class ApmMiddleware
+	internal class ApmMiddleware
 	{
 		private readonly RequestDelegate _next;
-		private readonly ITracer _tracer;
+		private readonly Tracer _tracer;
 
-		public ApmMiddleware(RequestDelegate next, ITracer tracer)
+		public ApmMiddleware(RequestDelegate next, Tracer tracer)
 		{
 			_next = next;
 			_tracer = tracer;
@@ -29,7 +29,7 @@ namespace Elastic.Apm.AspNetCore
 
 		public async Task InvokeAsync(HttpContext context)
 		{
-			var transaction = _tracer.StartTransaction($"{context.Request.Method} {context.Request.Path}",
+			var transaction = _tracer.StartTransactionInternal($"{context.Request.Method} {context.Request.Path}",
 				ApiConstants.TypeRequest);
 
 			transaction.Context.Request = new Request
--- a/src/Elastic.Apm/Logging/AbstractLogger.cs
+++ b/src/Elastic.Apm/Logging/AbstractLogger.cs
@@ -12,7 +12,7 @@
 		/// they need to write it (e.g. into a file, to the console)
 		/// </summary>
 		/// <param name="logline">This line that must be logged - it already contains the prefix and the loglevel</param>
-		protected abstract void PrintLogline(string logline);
+		protected abstract void PrintLogLine(string logLine);
 
 		private string GetPrefixString(LogLevel logLevel, string prefix) =>
 			string.IsNullOrWhiteSpace(prefix) ? $"{logLevel.ToString()} " : $"{logLevel.ToString()} {prefix}: ";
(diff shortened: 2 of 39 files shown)


src/Elastic.Apm/Model/Payload/Transaction.cs
```diff
@@ -33,6 +33,9 @@ public Transaction(AbstractLogger logger, string name, string type, IPayloadSend
 			Id = Guid.NewGuid();
 		}
 
+		/// <summary>
+		/// Any arbitrary contextual information regarding the event, captured by the agent, optionally provided by the user.
+		/// </summary>
 		public Context Context => _context.Value;
```
Mpdreamz: internal as well,  even though `Transaction` is already external this clearly indicates its intend.  Similar with other properties/methods not on `ITransaction`.

Mpdreamz: Also wondering why `_context`  and  `Context.tags` are `Lazy<T>`. The construction of either is not expensive in CPU nor memory.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
```diff
@@ -16,20 +16,20 @@
...
 		{
 			_next = next;
 			_tracer = tracer;
 		}
 
 		public async Task InvokeAsync(HttpContext context)
 		{
-			var transaction = _tracer.StartTransaction($"{context.Request.Method} {context.Request.Path}",
+			var transaction = _tracer.StartTransactionInternal($"{context.Request.Method} {context.Request.Path}",
 				ApiConstants.TypeRequest);
```
Mpdreamz: Not related to the changeset but more with initial exposore and naming/usage of `ApiConstants`. 

Instead of `type` being `string` we can have it accept `TransactionType`:

```
public struct TransactionType 
{
    public static readonly TransactionType Request = new TransactionType("request");

    private readonly string _name;
    public TransactionType(string name) => _name = name;
    
    public static implicit operator TransactionType(string name) => new TransactionType(name);

     public override string ToString() => _name;
}
```

If the last param to `StartTransaction` then takes `TransactionType` both of these will be valid:

```csharp
_.StartTransaction("a name", TransactionType.Request);
```

```csharp
_.StartTransaction("a name", "any type");
```

This would break the types we ship with down more clearly between `TransactionType` `SpanType` and `ActionType`.

src/Elastic.Apm/Logging/AbstractLogger.cs
```diff
@@ -12,7 +12,7 @@ public abstract class AbstractLogger
 		/// they need to write it (e.g. into a file, to the console)
 		/// </summary>
 		/// <param name="logline">This line that must be logged - it already contains the prefix and the loglevel</param>
-		protected abstract void PrintLogline(string logline);
+		protected abstract void PrintLogLine(string logLine);
```
Mpdreamz: Good catch! would still like to see https://github.com/elastic/apm-agent-dotnet/issues/61 land in the first alpha but understand the time pressure with v1 vs v2 of the intake API.

src/Elastic.Apm/Report/PayloadSender.cs
```diff
@@ -63,23 +59,20 @@ public async void StartWork()
 						new JsonSerializerSettings { ContractResolver = new CamelCasePropertyNamesContractResolver() });
```
Mpdreamz: `Json.NET` caches quite heavily per contract resolver, lets make this a field on the class and reuse it.

test/Elastic.Apm.AspNetCore.Tests/AspNetCoreMiddlewareTests.cs
```diff
@@ -88,10 +88,10 @@ public async Task HomeIndexSpanTest()
 			Assert.True(response.IsSuccessStatusCode);
 
 			var transaction = _capturedPayload.Payloads[0].Transactions[0];
-			Assert.NotEmpty(transaction.Spans);
+			Assert.NotEmpty(_capturedPayload.SpansOnFirstTransaction);
```
Mpdreamz: Threw me a bit thinking it was part of the public API but its only part of `MockPayloadSender` in tests which is much nicer to read in tests 👍

src/Elastic.Apm/Model/Payload/Transaction.cs
gregkalapos: > internal as well, even though Transaction is already external this clearly indicates its intend. Similar with other properties/methods not on ITransaction.

This is an interesting comment. I spent time thinking about this and after reading [this](https://ericlippert.com/2014/09/15/internal-or-public/) I thought I'd use a similar approach. Now the question is: what do we mean by public fields on internal types?

> public: this member is not an implementation detail at all; it is the public surface

From what perspective is it public surface? And I thought that from the server's perspective it's public... it's public because we send it thorough the network and the server works with it, so it's public within the APM solution. What do you think about this? Of course I can make it internal and mark it for serialization, so nothing stops us from doing that.

> Also wondering why _context and Context.tags are Lazy<T>.
There are lots of scenarios where transactions and spans don't contains tags, so I thought not allocating anything is a good idea. Furthermore its also realistic that transactions or spans don't have a context, and it that case I think we still save a couple of allocation. ....we should measure this! :)

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
gregkalapos: 👍 That sounds good! There are also `action` and `subtype` on span (https://github.com/elastic/apm/issues/8). We should also encapsulate those.

src/Elastic.Apm/Logging/AbstractLogger.cs
gregkalapos: I understand. #61 can really break the API, still I'd not consider it as high prio compared to other things we have now - plus replacing our logger is something that I think not many users would do. (Of course I'm not holding back anyone to propose a nicer API as early as possible) :)

src/Elastic.Apm/Model/Payload/Transaction.cs
gregkalapos: @Mpdreamz I merged, but I think we can still discuss this (I'd be happy to hear what you think) and I'm ok with changing it.

Mpdreamz: :thumbsup: on merge 

I'm happy with `public` coming around to it a second time. I have an irrational adversity against `internal` (and `sealed` but that's a separate discussion :smile:).  I actually never duplicate the `internal` modifier myself so not sure what triggered this comment on my part.

PR #79, opened 2019-01-26
PayloadSender improvements

Mpdreamz:
* Reuse http client (with DNS timeouts)
* Reuse json .NET ContractResolver
* Move to `BatchBlock<>` from `TPL Dataflow` over `BlockingCollection<>` (`System.Threading.Channels` also has similar constructs).

### DIFF
--- a/src/Elastic.Apm/Elastic.Apm.csproj
+++ b/src/Elastic.Apm/Elastic.Apm.csproj
@@ -21,5 +21,6 @@
     <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
     <!-- TODO: Are we ok with this reference/version? -->
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.0.0" />
+    <PackageReference Include="System.Threading.Tasks.Dataflow" Version="4.9.0" />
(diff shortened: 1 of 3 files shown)


src/Elastic.Apm/Elastic.Apm.csproj
```diff
@@ -21,5 +21,6 @@
     <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
     <!-- TODO: Are we ok with this reference/version? -->
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.0.0" />
+    <PackageReference Include="System.Threading.Tasks.Dataflow" Version="4.9.0" />
```
gregkalapos: Hmm... on the older versions this package targets .NET Standard 1.0 and .NET Framework 4.5. If we decide to target earlier versions (my classic example is WCF on Full Framework 3.5) then this can be a problem. I'd say we can go with this now since its usage will be only in `PayloadSender`, but if this spreads to the whole project it can limit us later.

gregkalapos:
LGTM, I let you decide if we keep `System.Threading.Tasks.Dataflow`.

PR #87, opened 2019-01-31
Initial stab at structured logging without dependencies

Mpdreamz:
Submitting this as I had it semi finished as per discussion on: elastic/apm-agent-dotnet#61

I like the abstraction and this will make it quite straight forward to have adapters (Apm.AspNet.All for instance can automatically forward to `Microsoft.Extensions.Logging`. 

The one downside is that we create a `LogValues` everytime we log as `TState` which obviously allocates more then just a `IFormattedString`. 

The upside is that the logs we emit play nice with structured logging solutions. 

Personally can't quite make the call if one is worth the other. 

The current abstraction would allow to fall back to `IFormattedString` logging as before.


src/Elastic.Apm/Logging/ConsoleLogger.cs
```diff
@@ -1,13 +1,61 @@
...
+			_errorOut = standardOut ?? Console.Error;
+		}
 
-		protected override void PrintLogLine(string logLine) => Console.WriteLine(logLine);
+		protected internal static LogLevel DefaultLogLevel { get; } = LogLevel.Error;
+		public static ConsoleLogger Instance { get; } = new ConsoleLogger(DefaultLogLevel);
+
+		public void Log<TState>(LogLevel level, TState state, Exception e, Func<TState, Exception, string> formatter)
+		{
+			//TODO: Discuss UtcNow?
```
gregkalapos: I'd prefer UtcNow... with logs from multiple processes on multiple machines in the cloud not using utc would make people's life harder.

src/Elastic.Apm/Logging/IApmLoggingExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+
+		public static void LogInfo(this IApmLogger logger, Exception exception, string message, params object[] args) =>
+			logger?.DoLog(LogLevel.Information, message, exception, args);
+
+		public static void LogDebug(this IApmLogger logger, string message, params object[] args) => logger?.LogDebug(null, message, args);
+
+		public static void LogDebug(this IApmLogger logger, Exception exception, string message, params object[] args) =>
+			logger?.DoLog(LogLevel.Debug, message, exception, args);
+
+		public static void LogDebugException(this IApmLogger logger, Exception exception) =>
```
gregkalapos: Nice! 👍

src/Elastic.Apm/Logging/ConsoleLogger.cs
SergeyKleyman: Yes, I will take a look at it today and update.

SergeyKleyman:
> First of all, sorry that this did not get any response for so long.
> 
> On the other hand after the preview release I think this is the right time to move forward with this. I think this is a nice abstraction, and I'd be ok with moving on with this.
> 
> @SergeyKleyman Do you wanna look at this?

Yes, I will take a look at it today and update.

src/Elastic.Apm/Logging/ScopedLogger.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+	internal class ScopedLogger : IApmLogger
+	{
+		public ScopedLogger(IApmLogger logger, string scope) => (Logger, Scope) = (logger, scope);
+
+		private ConcurrentDictionary<string, LogValuesFormatter> Formatters { get; } = new ConcurrentDictionary<string, LogValuesFormatter>();
+
+		public IApmLogger Logger { get; }
+
+		public string Scope { get; }
```
SergeyKleyman: I wonder if we can make Scope something more flexible so that we can scope not only by class name but the method name, transaction name, etc. Maybe we can make Scope in ScopedLogger to be a Map<string, string> and ScopedLogger 
 constructor will become ScopedLogger(IApmLogger logger, string scopedKey, string scopedValue) and a new contructor ScopedLogger(IApmLogger logger, Map<string, string> scopedKeyValues) to scope multiple (key. value) in one go. What do you think?

src/Elastic.Apm/Logging/ScopedLogger.cs
```diff
@@ -0,0 +1,29 @@
...
+
+		public LogValuesFormatter GetOrAddFormatter(string message, int expectedCount)
+		{
+			var formatter = Formatters.GetOrAdd(message, s => new LogValuesFormatter($"{{{{{{Scope}}}}}} {s}", Scope));
```
SergeyKleyman: So many curly braces - is it on purpose?

src/Elastic.Apm/Report/PayloadSender.cs
```diff
@@ -90,12 +91,12 @@ private async void StartWork()
 					switch (item)
 					{
 						case Payload p:
-							_logger.LogWarning($"Failed sending transaction {p.Transactions.FirstOrDefault()?.Name}");
-							_logger.LogDebug($"{e.GetType().Name}: {e.Message}");
+							_logger.LogWarning(nameof(PayloadSender), "Failed sending transaction {TransactionName}", p.Transactions.FirstOrDefault()?.Name);
```
SergeyKleyman: _logger is already scoped so I assume we don't need nameof(PayloadSender) here, correct?

src/Elastic.Apm/Logging/IApmLoggingExtensions.cs
```diff
@@ -0,0 +1,54 @@
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Net.Http;
+using System.Runtime.InteropServices;
+
+namespace Elastic.Apm.Logging
+{
+	internal static partial class LoggingExtensions
```
SergeyKleyman: Could you please explain why this class is partial? 10x :)

src/Elastic.Apm/Helpers/StacktraceHelper.cs
```diff
@@ -32,8 +32,8 @@ internal static List<Stacktrace> GenerateApmStackTrace(StackFrame[] frames, Abst
 			}
 			catch (Exception e)
 			{
-				logger.LogWarning($"Failed capturing stacktrace for {capturingFor}");
-				logger.LogDebug($"{e.GetType().Name}: {e.Message}");
+				logger?.LogWarning(e, nameof(StacktraceHelper), "Failed capturing stacktrace for {ApmContext}", capturingFor);
```
SergeyKleyman: IIUC "nameof(StacktraceHelper)" argument will be used as "message" parameter by LogWarning.

src/Elastic.Apm/Report/PayloadSender.cs
SergeyKleyman: Maybe we should define string constants for keys such as "TransactionName" to prevent spelling errors so the line above will become
`_logger.LogWarning($"Failed sending transaction {LogKey.TransactionName}", p.Transactions.FirstOrDefault()?.Name);`
and where `LogKey.TransactionName` value is `{TransactionName}`

src/Elastic.Apm/Logging/IApmLoggingExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+
+		public static void LogDebug(this IApmLogger logger, string message, params object[] args) => logger?.LogDebug(null, message, args);
+
+		public static void LogDebug(this IApmLogger logger, Exception exception, string message, params object[] args) =>
+			logger?.DoLog(LogLevel.Debug, message, exception, args);
+
+		public static void LogDebugException(this IApmLogger logger, Exception exception) =>
+			logger?.LogDebug(exception, "{ExceptionName}: {ExceptionMessage}", exception?.GetType().Name, exception?.Message);
+
+		public static void LogErrorException(this IApmLogger logger, Exception exception, string method) =>
```
SergeyKleyman: I think we can capture caller's method name automatically using https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/caller-information

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
```diff
@@ -8,31 +8,46 @@ namespace Elastic.Apm.Config
...
-			Enum.TryParse<LogLevel>(value, out var logLevel) ? logLevel : AbstractLogger.LogLevelDefault;
+
+		protected internal static LogLevel ParseLogLevel(string value)
+		{
+			switch (value.ToLowerInvariant())
+			{
+				case "trace": return LogLevel.Trace;
+				case "debug": return LogLevel.Debug;
+				case "information":
+				case "info": return LogLevel.Information;
```
SergeyKleyman: Could you clarify please why do we need both `information` and `info` - wouldn't `info` be enough (and naming the enum `LogLevel.Info`)?

src/Elastic.Apm/Logging/ScopedLogger.cs
Mpdreamz: Yeah. there is the named `{Scope}` placeholder wrapped in double `{}` so that the end result is `{<scope>}` in the message. The end result is one pair being emitted e.g ` {HttpDiagnosticListener} Exception in OnE....`

src/Elastic.Apm/Logging/ConsoleLogger.cs
Mpdreamz: Agreed, `UtcNow` all the way.

src/Elastic.Apm/Logging/IApmLoggingExtensions.cs
Mpdreamz: It shouldn't be! good catch.

src/Elastic.Apm/Logging/IApmLoggingExtensions.cs
Mpdreamz: Yep! should have done that from the get go. Good catch

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
Mpdreamz: I basically mirrored the enum from `Microsoft.Extensions.Logging` which lists `Information`. As a config option supporting both made sense to me

src/Elastic.Apm/Logging/ScopedLogger.cs
Mpdreamz: I personally don't feel we need to be that granular with scoping. Typically the class name should be sufficient. We can extend the extension methods to include Caller Information for all messages (not just exceptions) instead?

Mpdreamz:
Addressed PR comments and rebased. Thanks for the review @SergeyKleyman and @gregkalapos

src/Elastic.Apm/Report/PayloadSender.cs
Mpdreamz: +1, Ideally will do it in a new PR if that's okay. Rather pull this and #128  in first to keep overview.

src/Elastic.Apm/Logging/ScopedLogger.cs
SergeyKleyman: Agreed - we can wait and see what other needs arise.

SergeyKleyman:
LGTM

PR #89, opened 2019-02-04
Add NuGet metadata to .csproj files

gregkalapos:


src/Elastic.Apm.EntityFrameworkCore/Elastic.Apm.EntityFrameworkCore.csproj
```diff
@@ -3,17 +3,28 @@
...
     <AssemblyName>Elastic.Apm.EntityFrameworkCore</AssemblyName>
-    <SignAssembly>true</SignAssembly>
-    <AssemblyOriginatorKeyFile>..\..\elasticapm.snk</AssemblyOriginatorKeyFile>
     <AssemblyVersion>0.0.1.0</AssemblyVersion>
+    <InformationalVersion>0.0.1.0</InformationalVersion>
+    <FileVersion>0.0.1.0</FileVersion>
+    <SignAssembly>true</SignAssembly>
+    <AssemblyOriginatorKeyFile>..\..\build\elasticapm.snk</AssemblyOriginatorKeyFile>
     <DelaySign>false</DelaySign>
+    <PackageVersion>0.0.1.0</PackageVersion>
```
Mpdreamz: A nice technique here would be to move these common build props to e.g `Common.build.props` in `build` or `src`. 

Then including it like so:

```xml
<Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), .editorconfig))\src\Common.build.props" />
```

`GetDirectoryNameOfFileAbove` makes the csproj resiliant to being moved to any depth.

Then including `Common.build.props` as a solution item makes it easier to make changes in the future.

Mpdreamz: Example: https://github.com/elastic/elasticsearch-net/blob/master/src/Elasticsearch.Net/Elasticsearch.Net.csproj#L3

PR #95, opened 2019-02-07
[APM-CI] windows build

kuisathaverat:
* New windows build process
* Possibility to trigger the job with a comment
* Fixed lib versions

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -121,3 +121,11 @@ resharper_csharp_braces_for_fixed=required_for_multiline
 resharper_csharp_braces_for_ifelse=required_for_multiline
 
 resharper_csharp_accessor_owner_body=expression_body
+
+[Jenkinsfile]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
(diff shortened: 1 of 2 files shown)


.editorconfig
```diff
@@ -121,3 +121,11 @@ resharper_csharp_braces_for_fixed=required_for_multiline
 resharper_csharp_braces_for_ifelse=required_for_multiline
 
 resharper_csharp_accessor_owner_body=expression_body
+
+[Jenkinsfile]
```
gregkalapos: @kuisathaverat jfyi: we try to keep the `.editorconfig` files in sync across .NET repos within Elastic. @elastic/dotnet Are we ok with this change? Do you want to sync this to other repos? I think it's ok to have this...

Mpdreamz: Lgtm :+1:

kuisathaverat: I do not have a strong opinion, as you preference

gregkalapos: Great, then let's push it as it is now. We are ok with this. 👍

PR #99, opened 2019-02-13
Add readme to sample apps

gregkalapos:
Also
-Add GetNumberOfStars method in the HttpListenerSample
-Make transaction&span names&types const string in API tests


gregkalapos:
Build fails due to CI issue.

Issue #94, opened 2019-02-06
Bug: duration calculation


gregkalapos:
We have sometimes failing tests like this:

```
Error
Expected 9 to be greater or equal to: 10
Expected: True
Actual: False
Stacktrace
at Elastic.Apm.Tests.ApiTests.ConvenientApiTransactionTests.AssertWith1TransactionAsync(Func`2 func) in /var/lib/jenkins/workspace/tnet_apm-agent-dotnet-mbp_master/src/github.com/elastic/apm-agent-dotnet/test/Elastic.Apm.Tests/ApiTests/ConvenientApiTransactionTests.cs:line 487
at Elastic.Apm.Tests.ApiTests.ConvenientApiTransactionTests.AsyncTaskWithReturnType() in /var/lib/jenkins/workspace/tnet_apm-agent-dotnet-mbp_master/src/github.com/elastic/apm-agent-dotnet/test/Elastic.Apm.Tests/ApiTests/ConvenientApiTransactionTests.cs:line 265
--- End of stack trace from previous location where exception was thrown ---
```

This is the assert:

```
var duration = payloadSender.FirstTransaction.Duration;
Assert.True(duration >= SleepLength, $"Expected {duration} to be greater or equal to: {SleepLength}");
```

This is how we calculate the duration (very very bad!):
`if (!Duration.HasValue) Duration = (long)(DateTimeOffset.UtcNow - Start).TotalMilliseconds; `

We have a little bit of a data type chaos here and we should heavily overthink this! - also we should see what datatypes other agents use for duration and probably stick with that, but each time we convert we should be more careful.

gregkalapos:
We have 2 problems here:

- The `Duration` on `ITransaction` is a `long`, but the `Duration` on the `ISpan` is a `double`. This was clearly not intentional. The java agent uses `double` and the `.TotalMilliseconds` is also a `double`, so we stick to `double` for duration in the .NET Agent. 

- The other problem is that we use 10ms for these tests, but that's probably lower than the resolution of the timer used by `await Task.Delay()`, so it's impossible to await exactly 10ms (it can be 9,5ms). https://stackoverflow.com/a/31742754/1783306  To solve this I'll increase the waits to 20ms (I think 15ms is the resolution)

PR #104, opened 2019-02-14
Fix failing API tests - again

gregkalapos:
It looks like the timers on the VMs in CI have a very low resolution. ~~Therefore we wait even longer in the API tests.~~ 

Update: waiting longer does not help, we enable 10% error due to this inaccuracy in `await Task.Delay`. I think it's fine, we don't wanna measure in these tests with a super high accuracy, the point is that if we have a 10ms wait in the code than the duration should be around 10ms.. (and not 0 or 1..., but e.g. 9,12ms due to this inaccuracy is imo fine)

Solves #94


gregkalapos:
First build was green, all test pass. Re-triggered to give it another try, if it's still green I'll merge it.

gregkalapos:
Found 1 test which was not updated. Re-run again.

gregkalapos:
Tests green, we build fails because of our known "threshold" issue, which is WIP. I merge this.

PR #108, opened 2019-02-15
Add SecretToken configuration value

### LINKED ISSUE #107: APM Server secret token
APM Server hosted in ElasticCloud.  Seems to need `APM Server secret token`, but unclear on how to set that token.

Add support or understand how to insert this token

Not use for now.

**Additional context**

russcam:
This PR adds a SecretToken configuration value to allow
the agent to be used with Elasticsearch Service in Elastic Cloud.
The value assigned to the SecretToken configuration value will be
set as the value of a Bearer token Authorization header on
PayloadSender.

Closes #107

### DIFF
--- a/src/Elastic.Apm/Config/AbstractConfigurationReader.cs
+++ b/src/Elastic.Apm/Config/AbstractConfigurationReader.cs
@@ -18,6 +18,12 @@ namespace Elastic.Apm.Config
 		protected internal static LogLevel ParseLogLevel(string value) =>
 			Enum.TryParse<LogLevel>(value, out var logLevel) ? logLevel : AbstractLogger.LogLevelDefault;
 
+		protected string ParseSecretToken(ConfigurationKeyValue kv)
+		{
+			if (kv == null || string.IsNullOrEmpty(kv.Value)) return null;
+			return kv.Value;
+		}
+
 		protected LogLevel ParseLogLevel(ConfigurationKeyValue kv)
 		{
 			if (kv == null || string.IsNullOrEmpty(kv.Value)) return AbstractLogger.LogLevelDefault;
--- a/src/Elastic.Apm/Report/PayloadSender.cs
+++ b/src/Elastic.Apm/Report/PayloadSender.cs
@@ -2,6 +2,7 @@ using System;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
+using System.Net.Http.Headers;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks.Dataflow;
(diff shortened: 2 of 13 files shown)


src/Elastic.Apm/Report/PayloadSender.cs
```diff
@@ -33,7 +34,7 @@ internal class PayloadSender : IDisposable, IPayloadSender
 
 		static PayloadSender() => ServicePointManager.DnsRefreshTimeout = DnsTimeout;
 
-		internal PayloadSender(AbstractLogger logger, IConfigurationReader configurationReader)
+		internal PayloadSender(AbstractLogger logger, IConfigurationReader configurationReader, HttpMessageHandler handler)
```
gregkalapos: small nit: could we make the `handler` optional with `= null`? That way in `AgentComponents.cs` we don't have to pass `null`. This is the approach we usually used previously in the code-base.

test/Elastic.Apm.Tests/BasicAgentTests.cs
```diff
@@ -30,5 +38,30 @@ public void AgentVersion()
...
+			});
+
+			var secretToken = "SecretToken";
+			var logger = ConsoleLogger.Instance;
+			var payloadSender = new PayloadSender(logger, new TestAgentConfigurationReader(logger, secretToken: secretToken), handler);
+
+			using (var agent = new ApmAgent(new TestAgentComponents(secretToken: secretToken, payloadSender: payloadSender)))
+				agent.PayloadSender.QueuePayload(new Payload());
+
+			// ideally, introduce a mechanism to flush payloads
```
gregkalapos: Agree :) It's ok for now, in #13 there will be more changes in `PayloadSender`, I'll also add this functionality.

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
```diff
@@ -18,6 +18,12 @@ public abstract class AbstractConfigurationReader
 		protected internal static LogLevel ParseLogLevel(string value) =>
 			Enum.TryParse<LogLevel>(value, out var logLevel) ? logLevel : AbstractLogger.LogLevelDefault;
 
+		protected string ParseSecretToken(ConfigurationKeyValue kv)
+		{
+			if (kv == null || string.IsNullOrEmpty(kv.Value)) return null;
```
gregkalapos: I thought a little bit about maybe returning `String.Empty`, but I'm not sure... I bring this up, because prior to this change our configs never returned `null`, this would be the first one. Whether this is a real issue or not, I'm not sure... Maybe someone has a stronger opinion, I'm ok with this.

src/Elastic.Apm/Report/PayloadSender.cs
russcam: sure, will amend

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
russcam: I typically lean on evaluating `string.Empty` and `null` to `null`, but don't have strong opinions here.

gregkalapos: Yeah, I also think this is fine. 👍

Issue #109, opened 2019-02-16
Tags are missing from errors


skynet2:
**Describe the bug**
Tags are missing from the error object

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a new transaction
2. Add a tag example : apmTransaction.Tags.Add("response", JsonConvert.SerializeObject(result, JsonDefaults.JsonSerializerSettings));
3. Capture exception using CaptureException. example apmTransaction?.CaptureException(result.Exception
4. Check error in kibana (tags tab) is empty http://joxi.net/v296V1U385bqAG.png

**Expected behaviour**
Tags should not be empty

gregkalapos:
Thanks for reporting, and fixing this! I added a comment [here](https://github.com/elastic/apm-agent-dotnet/pull/111#discussion_r257651371). #111 will fix it. 

👍

PR #111, opened 2019-02-16
Include stack trace line

skynet2:
CI is failing to build for some reason, but looks like it`s not a code issue

### DIFF
--- a/src/Elastic.Apm/Config/AbstractConfigurationReader.cs
+++ b/src/Elastic.Apm/Config/AbstractConfigurationReader.cs
@@ -79,7 +79,7 @@ namespace Elastic.Apm.Config
 
 			if (string.IsNullOrEmpty(retVal))
 			{
-				var stackFrames = new StackTrace().GetFrames();
+				var stackFrames = new StackTrace(true).GetFrames();
 				foreach (var frame in stackFrames)
 				{
 					var currentAssembly = frame.GetMethod()?.DeclaringType.Assembly;
--- a/src/Elastic.Apm/Model/Payload/Error.cs
+++ b/src/Elastic.Apm/Model/Payload/Error.cs
@@ -18,6 +18,7 @@ namespace Elastic.Apm.Model.Payload
 				Id = Guid.NewGuid();
 			}
 
+			[JsonProperty]
 			internal Context Context { get; set; }
 			public string Culprit { get; set; }
 			public ICapturedException Exception { get; set; }
--- a/sample/SampleAspNetCoreApp/Controllers/HomeController.cs
+++ b/sample/SampleAspNetCoreApp/Controllers/HomeController.cs
@@ -84,7 +84,12 @@ namespace SampleAspNetCoreApp.Controllers
 			return Ok();
 		}
 
-		public IActionResult TriggerError() => throw new Exception("This is a test exception!");
+		public IActionResult TriggerError()
+		{
+			Agent.Tracer.CurrentTransaction.Tags["foo"] = "bar";
+			throw new Exception("This is a test exception!");
+		}
+
 
 		[ResponseCache(Duration = 0, Location = ResponseCacheLocation.None, NoStore = true)]
 		public IActionResult Error() => View(new ErrorViewModel { RequestId = Activity.Current?.Id ?? HttpContext.TraceIdentifier });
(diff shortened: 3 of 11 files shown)


src/Elastic.Apm/Config/AbstractConfigurationReader.cs
```diff
@@ -79,7 +79,7 @@ protected string ParseServiceName(ConfigurationKeyValue kv)
 
 			if (string.IsNullOrEmpty(retVal))
 			{
-				var stackFrames = new StackTrace().GetFrames();
+				var stackFrames = new StackTrace(true).GetFrames();
```
gregkalapos: This one here is just for the calculating the service name in case `Assembly.GetEntryAssembly()` returns `null`, so we don't use the line number info here. Could you please remove this one to save overhead?

src/Elastic.Apm/Model/Payload/Error.cs
```diff
@@ -18,6 +18,7 @@ public ErrorDetail()
 				Id = Guid.NewGuid();
 			}
 
+			[JsonProperty]
```
gregkalapos: Hahh, this is a very good catch! (and btw. totally my fault 😊 ). 

The strategy around this is to make properties that we want to send to the server within `internal` types to `public`. So, could you please remove `[JsonProperty]` and make it `public`. This is what we do in `Transaction` and `Span`. `Error` itself is still `internal`, so the property won't be visible to the outside world anyway, but Json.NET will serialize it that way. We consider these things `public` within the APM solution, so it's public to the server. 

Thx, this is a really good catch!

gregkalapos:
Update: also pushed another test to cover the issue with tags on error described in #109. It's not an ideal test, we should cover this with integration tests (which we will), but for now at least we have that.

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
skynet2: Hi @gregkalapos done 👍

src/Elastic.Apm/Model/Payload/Error.cs
skynet2: Hi @gregkalapos , done now it should be public, thanks

skynet2:
Hi @gregkalapos ,
Thanks for the review :)

0. sign our CLA - done
1. Yep, i can see that there is overhead. Probably we can do that based on some flag\log level in configuration or smth like that? Because as for us error stack trace without line numbers is complected to understand
2 . Yep, agree, line numbers will be visible only if you have pdb files. 

It`s Interesting to see if we`ll have that overhead with new StackTrace(true), but without having pdb files in the folder, what do you think @gregkalapos  ?

skynet2:
hi @gregkalapos ,
I`ve checked, there is no difference is there is .pdb file or no. The performance is the same for new StackTrace(true)

gregkalapos:
Thank you @skynet2 !

0. Did you use the same GitHub account to sign the CLA? I don’t see it 🤔 

1. Yes, it’ll be probably a configuration - we have already some of those. But I totally understand that line numbers in errors are important and I’d say we should always offer a way to get those, even if it’s more overhead.

2. Yeah, I think it doesn’t really matter where the pdbs are… the overhead is probably due to the work that the method names have to be found.

skynet2:
Hi @gregkalapos , 
Strange, i received a confirmation email, here is info which i used http://dl4.joxi.net/drive/2019/02/18/0000/2584/47640/40/4aa46972bb.png

gregkalapos:
Thanks @skynet2! That info looks exactly what we need, but the CLA checker is still not happy. I ask people who know more about this than I do. Once I get some info I'll let you know. Sorry about the confusion :) 

(unrelated: I pushed another change to the test that I added previously)

gregkalapos:
@skynet2 

Turns out we have some trouble with our CLA checker. Additionally, in [your commits](https://github.com/elastic/apm-agent-dotnet/commit/5978d620d87fd5f6f1983631a6ad2080e6b0c075.patch) you used a different mail address from the mail address of your [GitHub user](https://github.com/skynet2). Normally you’d get a notification and we’d ask you to resign with the e-mail address from your commits, but currently due to the troubles we have, even that would not work.

To make our life easy, I simply added your GitHub user with the mail address you used in your commit to our CLA database. Let me know if you are not ok with this.


If you are ok with this we are good to merge this.

Again, sorry for the trouble with the CLA.

skynet2:
Hi @gregkalapos ,
I`m ok with that, thanks.

PR #112, opened 2019-02-18
Merge AspNetFullFrameworkSampleApp to master (OLD)

SergeyKleyman:


sample/AspNetFullFrameworkSampleApp/MyModule1.cs
```diff
@@ -0,0 +1,53 @@
+﻿using Elastic.Apm.DiagnosticSource;
+using System;
+using System.Web;
+using Elastic.Apm;
+
+namespace AspNetFullFrameworkSampleApp
+{
+	public class MyModule1 : IHttpModule
```
gregkalapos: Could we maybe rename this to something more meaningful before we merge to master? (like ElasticApmModul, or something similar)

sample/AspNetFullFrameworkSampleApp/MyModule1.cs
```diff
@@ -0,0 +1,53 @@
...
+		/// see the following link: https://go.microsoft.com/?linkid=8101007
+		/// </summary>
+
+		#region IHttpModule Members
+
+		public void Dispose() => _httpDiagnosticsSubscription.Dispose();
+
+		public void Init(HttpApplication context)
+		{
+			//start listening for outgoing HTTP requests in the Elastic APM Agent.
```
gregkalapos: Aren't those comments outdated? I think those can be removed.

src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListener.cs
```diff
@@ -32,7 +32,7 @@ internal class HttpDiagnosticListener : IDiagnosticListener
 
 		private AbstractLogger Logger { get; }
 
-		public string Name => "HttpHandlerDiagnosticListener";
```
gregkalapos: This should be reverted. Otherwise it'll break our existing HttpClient monitoring stuff. I'd simply leave it as it is currently on master (so subscribe to this and move on, although as we saw this won't work on Full Framework due to different keys and types in the DiagnosticSource events.) Then in a later commit we can address `HttpClient` monitoring on Full Framework.

SergeyKleyman: I agree. So for now I should also remove call to Agent.Subscribe(new HttpDiagnosticsSubscriber()) in Init since it won't do anything anyway. And I'll open an issue to support monitoring HttpClient for Full Framework. How does it sound?

gregkalapos: 👍 Perfect, sounds good! If you want I can take that issue, I already looked into this and have something in mind to solve this.

SergeyKleyman: gr8 - here is the issue "Add support for tracing calls via HttpClient on Full Framework #115"

sample/AspNetFullFrameworkSampleApp/MyModule1.cs
SergeyKleyman: Done

SergeyKleyman:
This PR is superseded by #162

PR #118, opened 2019-02-18
Feature/add custom

skynet2:
Add Custom fields. Add the possibility to specify request_body

=============
p.s i need some help from you guys. There is a possible problem that Request and Response will be null on Context object when you create transaction manually. 
For now, i can see initialization for Request and Response only in internals libs such as Elastic.Apm.AspNetCore. The issue is that we need to have some mechanism of initialization for that objects outside of internal namespaces, for example for some custom rpc handlers or protocols. Any suggestions around that? We can pass some parameter to StartTransactionInternal, or pass some initial context object with request\response object or make that fields Lazy or smth like that. 

Thanks,
Stas

### DIFF
--- a/src/Elastic.Apm/Api/IRequest.cs
+++ b/src/Elastic.Apm/Api/IRequest.cs
@@ -1,0 +1,11 @@
+using Elastic.Apm.Model.Payload;
+
+namespace Elastic.Apm.Api
+{
+	public interface IRequest
+	{
+		string HttpVersion { get; set; }
+		string Method { get; set; }
+		string Body { get; set; }
+	}
+}
--- a/src/Elastic.Apm/Api/IRequest.cs
+++ b/src/Elastic.Apm/Api/IRequest.cs
@@ -6,6 +6,6 @@ namespace Elastic.Apm.Api
 	{
 		string HttpVersion { get; set; }
 		string Method { get; set; }
-		string Body { get; set; }
+		object Body { get; set; }
 	}
 }
--- a/src/Elastic.Apm/Api/ISpan.cs
+++ b/src/Elastic.Apm/Api/ISpan.cs
@@ -47,6 +47,11 @@ namespace Elastic.Apm.Api
 		Dictionary<string, string> Tags { get; }
 
 		/// <summary>
+		/// A list of user defined custom fields.
+		/// </summary>
+		Dictionary<string, object> Custom { get; }
+
+		/// <summary>
 		/// UUID of the enclosing transaction.
 		/// </summary>
 		Guid TransactionId { get; }
--- a/src/Elastic.Apm/Api/ISpan.cs
+++ b/src/Elastic.Apm/Api/ISpan.cs
@@ -47,11 +47,6 @@ namespace Elastic.Apm.Api
 		Dictionary<string, string> Tags { get; }
 
 		/// <summary>
-		/// A list of user defined custom fields.
-		/// </summary>
-		Dictionary<string, object> Custom { get; }
-
-		/// <summary>
 		/// UUID of the enclosing transaction.
 		/// </summary>
 		Guid TransactionId { get; }
--- a/src/Elastic.Apm/Api/ITransaction.cs
+++ b/src/Elastic.Apm/Api/ITransaction.cs
@@ -41,11 +41,6 @@ namespace Elastic.Apm.Api
 		Dictionary<string, string> Tags { get; }
 
 		/// <summary>
-		/// A list of user defined custom fields.
-		/// </summary>
-		Dictionary<string, object> Custom { get; }
-
-		/// <summary>
 		/// The type of the transaction.
 		/// Example: 'request'
 		/// </summary>
--- a/src/Elastic.Apm/Api/ITransaction.cs
+++ b/src/Elastic.Apm/Api/ITransaction.cs
@@ -35,6 +35,11 @@ namespace Elastic.Apm.Api
 		Dictionary<string, string> Tags { get; }
 
 		/// <summary>
+		/// A list of user defined custom fields.
+		/// </summary>
+		Dictionary<string, object> Custom { get; }
+
+		/// <summary>
 		/// The type of the transaction.
 		/// Example: 'request'
 		/// </summary>
--- a/src/Elastic.Apm/Api/ITransaction.cs
+++ b/src/Elastic.Apm/Api/ITransaction.cs
@@ -9,9 +9,14 @@ namespace Elastic.Apm.Api
 	public interface ITransaction
 	{
 		/// <summary>
-		/// Any arbitrary contextual information regarding the event, captured by the agent, optionally provided by the user.
+		/// If a log record was generated as a result of a http request, the http interface can be used to collect this information.
 		/// </summary>
 		IRequest Request { get; }
+
+		/// <summary>
+		/// User identification
+		/// </summary>
+		IUser User { get; }
 
 		/// <summary>
 		/// The duration of the transaction.
--- a/src/Elastic.Apm/Api/ITransaction.cs
+++ b/src/Elastic.Apm/Api/ITransaction.cs
@@ -2,11 +2,17 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Threading.Tasks;
+using Elastic.Apm.Model.Payload;
 
 namespace Elastic.Apm.Api
 {
 	public interface ITransaction
 	{
+		/// <summary>
+		/// Any arbitrary contextual information regarding the event, captured by the agent, optionally provided by the user.
+		/// </summary>
+		IRequest Request { get; }
+
 		/// <summary>
 		/// The duration of the transaction.
 		/// If it's not set (its HasValue property is false) then the value
--- a/src/Elastic.Apm/Elastic.Apm.csproj
+++ b/src/Elastic.Apm/Elastic.Apm.csproj
@@ -21,18 +21,18 @@
     <PackageTags>apm, monitoring, elastic, elasticapm, analytics, sdk</PackageTags>
   </PropertyGroup>
   <ItemGroup>
-    <Folder Include="Report\" />
-    <Folder Include="DiagnosticSource\" />
-    <Folder Include="DiagnosticListeners\" />
-    <Folder Include="Logging\" />
-    <Folder Include="Config\" />
-    <Folder Include="Helpers\" />
-    <Folder Include="Api\" />
+    <Folder Include="Report\"/>
+    <Folder Include="DiagnosticSource\"/>
+    <Folder Include="DiagnosticListeners\"/>
+    <Folder Include="Logging\"/>
+    <Folder Include="Config\"/>
+    <Folder Include="Helpers\"/>
+    <Folder Include="Api\"/>
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
+    <PackageReference Include="Newtonsoft.Json" Version="11.0.2"/>
     <!-- TODO: Are we ok with this reference/version? -->
-    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.0.0" />
-    <PackageReference Include="System.Threading.Tasks.Dataflow" Version="4.9.0" />
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.0.0"/>
+    <PackageReference Include="System.Threading.Tasks.Dataflow" Version="4.9.0"/>
   </ItemGroup>
 </Project>
--- a/src/Elastic.Apm/Elastic.Apm.csproj
+++ b/src/Elastic.Apm/Elastic.Apm.csproj
@@ -21,18 +21,18 @@
     <PackageTags>apm, monitoring, elastic, elasticapm, analytics, sdk</PackageTags>
   </PropertyGroup>
   <ItemGroup>
-    <Folder Include="Report\"/>
-    <Folder Include="DiagnosticSource\"/>
-    <Folder Include="DiagnosticListeners\"/>
-    <Folder Include="Logging\"/>
-    <Folder Include="Config\"/>
-    <Folder Include="Helpers\"/>
-    <Folder Include="Api\"/>
+    <Folder Include="Report\" />
+    <Folder Include="DiagnosticSource\" />
+    <Folder Include="DiagnosticListeners\" />
+    <Folder Include="Logging\" />
+    <Folder Include="Config\" />
+    <Folder Include="Helpers\" />
+    <Folder Include="Api\" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="11.0.2"/>
+    <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
     <!-- TODO: Are we ok with this reference/version? -->
-    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.0.0"/>
-    <PackageReference Include="System.Threading.Tasks.Dataflow" Version="4.9.0"/>
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.0.0" />
+    <PackageReference Include="System.Threading.Tasks.Dataflow" Version="4.9.0" />
   </ItemGroup>
 </Project>
--- a/src/Elastic.Apm/Helpers/StringExtensions.cs
+++ b/src/Elastic.Apm/Helpers/StringExtensions.cs
@@ -2,17 +2,14 @@ namespace Elastic.Apm.Helpers
 {
 	internal static class StringExtensions
 	{
-		internal static string TrimToLength(this string input, int maxLength)
+		private static string TrimToLength(this string input, int maxLength)
 		{
 			if (input.Length > maxLength)
-				input = $"{input.Substring(0, Consts.PropertyMaxLength-3)}...";
+				input = $"{input.Substring(0, Consts.PropertyMaxLength - 3)}...";
 
 			return input;
 		}
 
-		internal static string TrimToMaxLength(this string input)
-		{
-			return input.TrimToLength(Consts.PropertyMaxLength);
-		}
+		internal static string TrimToMaxLength(this string input) => input.TrimToLength(Consts.PropertyMaxLength);
 	}
 }
--- a/src/Elastic.Apm/Helpers/StringExtensions.cs
+++ b/src/Elastic.Apm/Helpers/StringExtensions.cs
@@ -1,10 +1,18 @@
 namespace Elastic.Apm.Helpers
 {
-	public static class StringExtensions
+	internal static class StringExtensions
 	{
-		public static string TrimToLength(this string input)
+		internal static string TrimToLength(this string input, int maxLength)
 		{
-			return "";
+			if (input.Length > maxLength)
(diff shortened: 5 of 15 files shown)


src/Elastic.Apm/Api/ISpan.cs
```diff
@@ -46,6 +46,11 @@ public interface ISpan
 		/// </summary>
 		Dictionary<string, string> Tags { get; }
 
+		/// <summary>
+		/// A list of user defined custom fields.
+		/// </summary>
+		Dictionary<string, object> Custom { get; }
```
gregkalapos: There is no `custom` on span. We only have this on the transaction's context. See [here](https://github.com/elastic/apm-server/blob/master/docs/spec/spans/span.json).

src/Elastic.Apm/Api/ITransaction.cs
```diff
@@ -2,11 +2,17 @@
...
 using System.Diagnostics;
 using System.Threading.Tasks;
+using Elastic.Apm.Model.Payload;
 
 namespace Elastic.Apm.Api
 {
 	public interface ITransaction
 	{
+		/// <summary>
+		/// Any arbitrary contextual information regarding the event, captured by the agent, optionally provided by the user.
```
gregkalapos: I think "arbitrary contextual information" isn't correct... this is not arbitrary information. Did you see this description of `Request` somewhere? I'd simply use the description from the [json spec](https://github.com/elastic/apm-server/blob/master/docs/spec/request.json#L4)

src/Elastic.Apm/Model/Payload/Span.cs
```diff
@@ -76,17 +79,24 @@ public void CaptureError(string message, string culprit, StackFrame[] frames)
...
 			private readonly Lazy<Dictionary<string, string>> _tags = new Lazy<Dictionary<string, string>>();
+			private readonly Lazy<Dictionary<string, object>> _custom = new Lazy<Dictionary<string, object>>();
 			public IDb Db { get; set; }
 			public IHttp Http { get; set; }
 			public Dictionary<string, string> Tags => _tags.Value;
+			public Dictionary<string, object> Custom => _custom.Value;
 		}
 	}
 
+	public interface IRequest
```
gregkalapos: Could you please move this into its own file into the `Api` folder? Thx!

skynet2:
Hi @gregkalapos ,
Thanks for the link to apm discussion, it`s sad about depreciation for custom, basically, the reason why i decided to use custom is that it can save big objects, for example, requests and response. 

Basically, for the request, there is another way, as I added Request.RequestBody property which doesn`t have a limit on the length, but for the response, there is no such property.

About IRequest
1. Are there any specific fields you want to have on that interface or i can take all fields from the Request?
 2. We need to decide about initialization for Request and Response, currently they are null by default, we have three ways: 1. make it get;set 2. initialize empty object on CreateTransactionInternal 3.  Lazy 
What do you think?

gregkalapos:
I see, thanks for the clarification. Yes, I see the problem - and `Tags` are indeed limited on length.

Btw. what exactly do you want to store in the `Custom`? Is it both the request body and the response body? There was an issue opened for that couple of hours ago: https://github.com/elastic/apm/issues/52. Would that help you?

To `IRequest`: 

1. We can add all the fields, if we decide to make this public, there is no reason to keep any of those private. There are some properties which are other types, for those I’d add interfaces similarly.

2. Since these are bigger types and in some cases they won’t be needed I’d just wrap them into `Lazy<T>` (so option 3)  - similarly to `Tags`.

skynet2:
Hello @gregkalapos ,
Thanks for the review. Currently, I`m storing response json object in custom, for one of the services we need to store that data. 
Yep it would help, currently there are 3 issues in which i`m interested https://github.com/elastic/apm/issues/52 https://github.com/elastic/apm/issues/53https://github.com/elastic/apm/issues/7  
So yep, while there is no final decision i`ll remove Custom from current pull request, and continue using in my own branch :). 

Going to refactor Request\Response to Lazy. 
Probably there is no reason to have IRequest which will have only 1 implementation and will not change, because the schema for apm server is fixed and there is no way to set an external object to that? what do you think @gregkalapos

src/Elastic.Apm/Api/ISpan.cs
skynet2: Reverted changes with custom

src/Elastic.Apm/Api/ITransaction.cs
skynet2: fixed

src/Elastic.Apm/Model/Payload/Span.cs
skynet2: fixed

src/Elastic.Apm/Elastic.Apm.csproj
```diff
@@ -21,18 +21,18 @@
     <PackageTags>apm, monitoring, elastic, elasticapm, analytics, sdk</PackageTags>
   </PropertyGroup>
   <ItemGroup>
-    <Folder Include="Report\"/>
```
gregkalapos: Do you know how you got these changes? I think Visual Studio does not want to add the new empty spaces before the closing `/>` and most of the .NET devs at Elastic uses JetBrains rider which specifically removes those empty spaces. Could we revert this please?

src/Elastic.Apm/Helpers/StringExtensions.cs
```diff
@@ -0,0 +1,18 @@
...
+	{
+		internal static string TrimToLength(this string input, int maxLength)
+		{
+			if (input.Length > maxLength)
+				input = $"{input.Substring(0, Consts.PropertyMaxLength-3)}...";
+
+			return input;
+		}
+
+		internal static string TrimToMaxLength(this string input)
```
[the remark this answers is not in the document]
skynet2: Hi @gregkalapos  i thought about making it private, but i think it can be useful in other parts of code, for that library, for example if we`ll need to trim to custom length or smth like that, that's why i leaved it as internal

src/Elastic.Apm/Model/Payload/Context.cs
```diff
@@ -5,11 +5,28 @@ namespace Elastic.Apm.Model.Payload
...
-		public Request Request { get; set; }
+		private readonly Lazy<Dictionary<string, string>> _tags = new Lazy<Dictionary<string, string>>();
+		private Request _request;
+		private Response _response;
+		private User _user;
 
-		public Response Response { get; set; }
+		public User User
+		{
+			get => _user ?? (_user = new User());
```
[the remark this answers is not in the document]
skynet2: done, thanks

src/Elastic.Apm/Model/Payload/Context.cs
```diff
@@ -5,11 +5,28 @@ namespace Elastic.Apm.Model.Payload
...
+		public Request Request
+		{
+			get => _request ?? (_request = new Request());
+			set => _request = value;
+		}
 
-		public Dictionary<string, string> Tags => tags.Value;
+		public Response Response
+		{
+			get => _response ?? (_response = new Response());
```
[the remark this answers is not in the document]
skynet2: done

src/Elastic.Apm/Model/Payload/Span.cs
```diff
@@ -81,7 +99,6 @@ private class ContextImpl : IContext
 			public Dictionary<string, string> Tags => _tags.Value;
 		}
 	}
-
```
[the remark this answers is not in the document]
skynet2: ok

src/Elastic.Apm/Elastic.Apm.csproj
skynet2: Hi i also use rider, not sure how i got that changes in the commit, probably missed them for some reason. will revert, thanks

src/Elastic.Apm/Model/Payload/Context.cs
```diff
@@ -5,11 +5,28 @@ namespace Elastic.Apm.Model.Payload
...
 
-		public Response Response { get; set; }
+		public User User
+		{
+			get => _user ?? (_user = new User());
+			set => _user = value;
+		}
+		public Request Request
+		{
+			get => _request ?? (_request = new Request());
```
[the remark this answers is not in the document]
skynet2: done

src/Elastic.Apm/Helpers/StringExtensions.cs
```diff
@@ -0,0 +1,18 @@
+namespace Elastic.Apm.Helpers
+{
+	internal static class StringExtensions
+	{
+		internal static string TrimToLength(this string input, int maxLength)
```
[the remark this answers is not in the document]
skynet2: Hi @gregkalapos i thought about making it private, but i think it can be useful in other parts of code, for that library, for example if we`ll need to trim to custom length or smth like that, that's why i leaved it as internal

src/Elastic.Apm/Api/IRequest.cs
```diff
@@ -0,0 +1,11 @@
+using Elastic.Apm.Model.Payload;
+
+namespace Elastic.Apm.Api
+{
+	public interface IRequest
+	{
+		string HttpVersion { get; set; }
+		string Method { get; set; }
+		object Body { get; set; }
```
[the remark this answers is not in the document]
skynet2: Hi, I'm also not sure about that, i changed to object according to schema https://github.com/elastic/apm-server/blob/master/docs/spec/request.json#L9 .
As for me, having that as an object is not a good idea, because we can have issues with serialization as you already mentioned, and the user can have possible issues with index mappings. So for me i will always use string or dictionary<string,string>. 
But I`m not sure that we should limit library users in some strict types however apm server supports to have it as an object.

src/Elastic.Apm/Helpers/StringExtensions.cs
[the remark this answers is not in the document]
gregkalapos: 👍 I don't have a strong opinion here... probably both options are fine. I see you made it private, let's keep it that way, unless you see it differently. We can make it internal once we need it later.

src/Elastic.Apm/Api/IRequest.cs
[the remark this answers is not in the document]
gregkalapos: Yeah, the problem is that `object` in the json schema is a json object and that's different from `object` in C#... we cannot map any C# `object` to a json object... or..well, we can, but we can't guarantee correctness.

Let's do the following: let's keep it `object` and if we see that this does not work we can simply change it before we GA the agent.

src/Elastic.Apm/Model/Payload/Context.cs
```diff
@@ -5,11 +5,12 @@ namespace Elastic.Apm.Model.Payload
 {
 	internal class Context
 	{
-		private readonly Lazy<Dictionary<string, string>> tags = new Lazy<Dictionary<string, string>>();
-		public Request Request { get; set; }
+		private readonly Lazy<Dictionary<string, string>> _tags = new Lazy<Dictionary<string, string>>();
+		public Lazy<User> User { get; set; }
+		public Lazy<Request> Request { get; set; }
```
gregkalapos: These remain `null`. You'll need a field and initialize that with `Lazy` and you need to return that. Like this: 
```
private readonly Lazy<Request> _request = new Lazy<Request>();
public Request Request => _request.Value;
```

Same for the other properties.

src/Elastic.Apm/Elastic.Apm.csproj
gregkalapos: Hmm.. interesting. I'll keep this in mind... maybe we need to add something to `.editorconfig`... I'm not surre, I was unable to repro this; on my machine Rider automatically removed those empty spaces. Anyway, thanks for reverting! I think we are good here.

src/Elastic.Apm/Model/Payload/Context.cs
gregkalapos: This way you'll also not need to change the user code... so the property itself will have the same type as before (e.g. `Request` and not `Lazy<Request>`), just the backing field will be `Lazy`.

src/Elastic.Apm/Api/IRequest.cs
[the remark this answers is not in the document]
skynet2: Cool, agree

skynet2:
Hi @gregkalapos ,

Thanks, my fail, fixed. 

Thanks,
Stas

src/Elastic.Apm/Model/Payload/Context.cs
skynet2: fixed thanks @gregkalapos

gregkalapos:
Jenkins test this please

skynet2:
Hi @gregkalapos , please don`t merge yet, looks like there are some cases which are not covered by tests. 

Currently, i`m getting that issue 
Error Failed sending transaction. data validation error: Problem validating JSON document against schema: I[#] S[#] doesn't validate with "error#"
  I[#/errors/0] S[#/properties/errors/items/allOf/0] allOf failed
    I[#/errors/0/context/request/method] S[#/properties/errors/items/allOf/0/properties/context/properties/request/properties/method/type] expected string, but got null
Error Failed sending transaction. data validation error: Problem validating JSON document against schema: I[#] S[#] doesn't validate with "transaction#"
  I[#/transactions/0] S[#/properties/transactions/items/allOf/0] allOf failed
    I[#/transactions/0/context/request/method] S[#/properties/transactions/items/allOf/0/properties/context/properties/request/properties/method/type] expected string, but got null
Error Failed sending transaction. data validation error: Problem validating JSON document against schema: I[#] S[#] doesn't validate with "transaction#"
  I[#/transactions/0] S[#/properties/transactions/items/allOf/0] allOf failed
    I[#/transactions/0/context/request/method] S[#/properties/transactions/items/allOf/0/properties/context/properties/request/properties/method/type] expected string, but got null
Error Failed sending transaction. data validation error: Problem validating JSON document against schema: I[#] S[#] doesn't validate with "transaction#"
  I[#/transactions/0] S[#/properties/transactions/items/allOf/0] allOf failed
    I[#/transactions/0/context/request/url] S[#/properties/transactions/items/allOf/0/properties/context/properties/request/properties/url/type] expected object, but got null

gregkalapos:
Och, good catch, I see!

The problem with `Lazy` is that once we do let's say this:

> Agent.Tracer.CaptureTransaction("TestTransaction", "TestType",
> t =>
> {
> 				t.Request.HttpVersion = "2";
> }

Then we initialized `Context.Request`, but we don't make sure that all required fields are filled. 

Hmm... this is an interesting problem. 🤔

skynet2:
I see two ways how we can solve that : 
1. We can remove Lazy and make it get;set properties
2. We can create a custom JSON resolver for json.net which will ignore uninitialized lazy objects

What do you think @gregkalapos

skynet2:
The second way will not work, as it`s auto property.
The first way will require exposing to public Request and Response classes.

gregkalapos:
Yes, and the problem with the 2. approach is that it can be partially initialized. E.g. in [`Request` `Method` and `Url` are required](https://github.com/elastic/apm-server/blob/master/docs/spec/request.json#L93). If we set e.g. `Url`, but leave `Method` empty then it's still not ok.

The "not so nice" thing about the 1. approach is that if those are properties, then if someone does not initialize it then there could ne nullref exceptions. 

I'm thinking about this... I'll try to come up with something soon :) Also feel free to let me know if you have something in mind.

gregkalapos:
What about utilizing ValueTuples? 

Instead of having an 'IRequest' property on `ITransaction` we could introduce properties that expose all `Request` related properties at once:

```
public interface ITransaction
{
   (string httpVersion, string method, object body, string fullUrl,
      string rawUrl, string hostName, string protocol) Request { get; set; }
 //...
}
```

This way the user code either sets everything we need for a `Request` (including required fields) or nothing. It's had to have partly initalized values this way (except users passing `null` on purpose, but not accidentally, so this does not count to me as a problem).

skynet2:
Hi @gregkalapos ,
Not sure that it`s a good idea. Named tuples are not well supported with json.net https://github.com/JamesNK/Newtonsoft.Json/issues/1230 so we`ll have problems during serialization. 

I think it`s easier to expose class with a constructor which requires a list of a parameter (for example url can be one of that parameters) . For example, in order to initialize Request, you should pass Url into a constructor

gregkalapos:
This property would not be serialized, this is just the interface that users use. We serialize the values on the `Context`. So the ValueTuple property would be marked with [JsonIgnore] on the implementor (`Transaction` in this case).

gregkalapos:
My problem with having a constructor is that since we have properties we can easily end up with null values. 

For example:

```
public interface ITransaction
{
   Request Request { get; set; }
  //...
}
```
```
internal class Request : IRequest
{
   public Request(string method, Url url)
   {
      _method = method;
      Url = url;
   }
//...
}
```

And then in the user code someone writes this:
```
Agent.Tracer.CurrentTransaction.Request.Method = "GET";
```

In this case Agent.Tracer.CurrentTransaction.Request can be null.... As a user each time you have to initalize properties. And I think that's not very nice.

skynet2:
As i understand, in some cases such properties such as Request and Response can be null. 
Tuples probably can work... need to think more, probably there are some other ways...

skynet2:
As i can see in golang apm they use pretty the same principle https://github.com/elastic/apm-agent-go/blob/master/context.go

gregkalapos:
@skynet2 In the Go code there is only 1 method specifically for the `http.Request` type, which is a framework type. It's like if we'd add a `SetRequest` method that accepts `Microsoft.AspNetCore.Http.HttpRequest`. The big difference is that in Go (at least this is what I was told) `http.Request` covers almost all the use cases, but in .NET `Microsoft.AspNetCore.Http.HttpRequest` is used by only 1 framework (ASP.NET Core), and e.g. it does not cover your use-case. So in Go you can't just pass a couple of strings and build your own `Request` instance.

Regarding the builder pattern: We can discuss that, but I don't see how you force the user on the API level that they pass all required fields. It could be that there is only 1 method and that expects all required properties, but then why do we need the builder pattern? It could be a simple property, and if it's not used we don't send the `Request` (same as with the builder pattern). 

Since this discussion is already drifted from the PR, could we solve this problem here? https://github.com/elastic/apm-agent-dotnet/issues/124 (it's work in progress, I'll add some potential solutions soon). Once we have a decision there we can continue with the PR here.

Mpdreamz:
> To give you an example: we consider renaming `Tags` to `Labels`, because we want to confirm to the [elastic common schema](https://github.com/elastic/ecs). By having an interface we can keep `Tags` and add a new `Labels` field and mark `Tags` deprecated but and simply redirect the values to `Labels`.

@gregkalapos mind explaining this a bit more concretely? This could also be done with a single `Request` implementation. This also depends on the versioning scheme of the agents vs the server and what the backward compatibility looks like. 

If the server renames a property that is a breaking change that should only happen on a major version.

gregkalapos:
> > To give you an example: we consider renaming `Tags` to `Labels`, because we want to confirm to the [elastic common schema](https://github.com/elastic/ecs). By having an interface we can keep `Tags` and add a new `Labels` field and mark `Tags` deprecated but and simply redirect the values to `Labels`.
> 
> @gregkalapos mind explaining this a bit more concretely? This could also be done with a single `Request` implementation. This also depends on the versioning scheme of the agents vs the server and what the backward compatibility looks like.
> 
> If the server renames a property that is a breaking change that should only happen on a major version.

True, this can be also done with a single `Request` implementation. But in this case the single `Request` implementation, which is public must conform to the model that the server expects. So we'd have 2 properties, 1 with [JsonIgnore], and then the new one. My goal is to keep the API and the model that the server expects separate.

PR #120, opened 2019-02-19
fixes #119 max span name + handle result

### LINKED ISSUE #119: Fix max length for span name
Max length for span name is 1024 (apm server configuration). 

In the case of entity, db queries span name = query content, and query length can be greater than 1024 characters.  http://joxi.net/5mdJeEHkbP4eA1

We need to limit max length for span name, a transaction cannot be recorded on APM server

skynet2:


src/Elastic.Apm/Report/PayloadSender.cs
```diff
@@ -83,7 +83,12 @@ private async void StartWork()
 							break;
 					}
 
-					// TODO: handle unsuccesful status codes
+					if (result != null && !result.IsSuccessStatusCode)
+					{
+						var str = await result.Content.ReadAsStringAsync();
+
+						throw new Exception(str);
```
gregkalapos: This would end up in the catch block below...so we'd throw an exception that we catch immediately. The general strategy as an APM agent is be as non-intrusive as possible, so we should avoid throwing exceptions. I'm working on the next version of the PayloadSender in #13 , and this class doesn't really have a long future, I think a debug log would be enough here.

src/Elastic.Apm/Model/Payload/Transaction.cs
```diff
@@ -95,6 +95,9 @@ public ISpan StartSpan(string name, string type, string subType = null, string a
 
 		internal Span StartSpanInternal(string name, string type, string subType = null, string action = null)
 		{
+			if (name.Length > 1024)
```
gregkalapos: Could we move this into the setter of 'Span.Name'? I'd also move `1024` into the `Consts` class. Who knows...maybe we change this :)

src/Elastic.Apm/Report/PayloadSender.cs
skynet2: Hi @gregkalapos , done

gregkalapos:
Thank you @skynet2! 🎉 

(I pushed a test to cover this.) 

-> merged 🚀

PR #125, opened 2019-02-25
speed up several tests that were needlessly slow

Mpdreamz:
Noticed this while developing using rider's continuous test feature.

### DIFF
--- a/test/Elastic.Apm.Tests/ApiTests/ConvenientApiSpanTests.cs
+++ b/test/Elastic.Apm.Tests/ApiTests/ConvenientApiSpanTests.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Diagnostics;
 using System.Threading;
 using System.Threading.Tasks;
 using Elastic.Apm.Api;
@@ -22,11 +23,21 @@ namespace Elastic.Apm.Tests.ApiTests
 	{
 		private const string ExceptionMessage = "Foo";
 		private const string SpanName = "TestSpan";
-		private const int SpanSleepLength = 45;
+
 		private const string SpanType = "TestSpan";
 		private const string TransactionName = "ConvenientApiTest";
-		private const int TransactionSleepLength = 30;
+
 		private const string TransactionType = "Test";
+		private int SpanSleepLength => TransactionSleepLength * 2;
+
+		private static int TransactionSleepLength
+		{
+			get
+			{
+				var frequency = Stopwatch.Frequency;
+				return 1000L / frequency < 1 ? 1 : (int)(1000L / frequency);
+			}
+		}
 
 		/// <summary>
 		/// Tests the <see cref="Transaction.CaptureSpan(string,string,System.Action,string,string)" /> method.
@@ -486,8 +497,7 @@ namespace Elastic.Apm.Tests.ApiTests
 			Assert.Equal(TransactionName, payloadSender.Payloads[0].Transactions[0].Name);
 			Assert.Equal(TransactionType, payloadSender.Payloads[0].Transactions[0].Type);
 
-			//`await Task.Delay` is inaccurate, so we enable 10% error in the test
-			var expectedTransactionLength = (TransactionSleepLength + SpanSleepLength) - (TransactionSleepLength * 0.1 + SpanSleepLength * 0.1);
+			var expectedTransactionLength = TransactionSleepLength + SpanSleepLength;
 			var duration = payloadSender.Payloads[0].Transactions[0].Duration;
 			Assert.True(duration >= expectedTransactionLength, $"Expected {duration} to be greater or equal to: {expectedTransactionLength}");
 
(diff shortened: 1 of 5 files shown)


test/Elastic.Apm.Tests/ApiTests/ConvenientApiSpanTests.cs
```diff
@@ -22,10 +22,10 @@ public class ConvenientApiSpanTests
 	{
 		private const string ExceptionMessage = "Foo";
 		private const string SpanName = "TestSpan";
-		private const int SpanSleepLength = 450;
```
gregkalapos: This was changed in https://github.com/elastic/apm-agent-dotnet/issues/94 

The final solution was to enable 10% error (e.g. [here](https://github.com/elastic/apm-agent-dotnet/blob/master/test/Elastic.Apm.Tests/ApiTests/ConvenientApiSpanTests.cs#L530) ). The resolution of the timer on the CI machine seems to be very bad, and I'm not sure it'll work... maybe 450 was an overkill from me.

Mpdreamz: Aye, 45 is still > min resolution of the timer. Maybe we should subtract min resolution of the timer in addition to your `10%` rule in the assert

gregkalapos: Is there a plain managed API to get the timer resolution? I suspect it's different on different OSs (that's why none of us was able to repro the original problem locally). I was unable to find anything...

Mpdreamz: [Stopwatch.Frequency](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch.frequency?view=netframework-4.7.2)

```csharp
if (Stopwatch.IsHighResolution)
{
	Console.WriteLine("Operations timed using the system's high-resolution performance counter.");
}
else
{
	Console.WriteLine("Operations timed using the DateTime class.");
}

long frequency = Stopwatch.Frequency;
Console.WriteLine("  Timer frequency in ticks per second = {0}",frequency);
long nanosecPerTick = (1000L * 1000L * 1000L) / frequency;
Console.WriteLine("  Timer is accurate within {0} nanoseconds",nanosecPerTick);

var dt1 = DateTime.UtcNow;
Thread.Sleep(1);
var dt2 = DateTime.UtcNow;
Console.WriteLine((dt2 - dt1).Milliseconds);
```

For me this outputs 

```
Operations timed using the system's high-resolution performance counter.
  Timer frequency in ticks per second = 10000000
  Timer is accurate within 100 nanoseconds
1
```

I image `Stopwatch.IsHighResolution` will be false on CI

gregkalapos: Nice! 

What about simply setting these sleeps and waits to exactly `nanosecPerTick` (or..well I guess in millisec, since `Thread.Sleep` and `Task.Delay` expect millisec) and assert that the duration is > `nanosecPerTick`? If we set it to something that is within the frequency of the timer then this assert should be correct and we don't need to add error to it.

gregkalapos: Update: if the resolution is less than 1 millisec I'd just go with 1ms...

test/Elastic.Apm.Tests/ApiTests/ConvenientApiTransactionTests.cs
```diff
@@ -21,10 +22,19 @@ namespace Elastic.Apm.Tests.ApiTests
 	public class ConvenientApiTransactionTests
 	{
 		private const string ExceptionMessage = "Foo";
-		private const int SleepLength = 45;
+
 		private const string TransactionName = "ConvenientApiTest";
 		private const string TransactionType = "Test";
 
+		private int SleepLength
```
Mpdreamz: Rather then re implementing this in each class maybe we can introduce a `Task.DelayMinimum()` extension method?

test/Elastic.Apm.Tests/ApiTests/ConvenientApiTransactionTests.cs
```diff
@@ -485,10 +495,8 @@ private async Task<MockPayloadSender> AssertWith1TransactionAsync(Func<ApmAgent,
 			Assert.Equal(TransactionName, payloadSender.Payloads[0].Transactions[0].Name);
 			Assert.Equal(TransactionType, payloadSender.Payloads[0].Transactions[0].Type);
 
-			//`await Task.Delay` is inaccurate, so we enable 10% error in the test
-			var expectedTransactionLength = SleepLength - (SleepLength * 0.1);
 			var duration = payloadSender.Payloads[0].Transactions[0].Duration;
-			Assert.True(duration >= expectedTransactionLength, $"Expected {duration} to be greater or equal to: {expectedTransactionLength}");
+			Assert.True(duration >= SleepLength, $"Expected {duration} to be greater or equal to: {SleepLength}");
```
Mpdreamz: Similar to previous comment, make this an extension method on Assert `Assert.MinimumDuration(duration)`

test/Elastic.Apm.Tests/ApiTests/ConvenientApiTransactionTests.cs
gregkalapos: I think there is no way to add static extension methods 🤔 

Nevertheless I agree that it'd make sense to implement this only once. I'll add a helper method.

test/Elastic.Apm.Tests/ApiTests/ConvenientApiTransactionTests.cs
gregkalapos: Same here... Assert is a class with static methods. I think we can't extend that.

gregkalapos:
Thx @Mpdreamz! 

Pushed some updates.

Mpdreamz:
Lgtm :+1:

Issue #59, opened 2019-01-09
Full Framework: support multiple AppDomains


gregkalapos:
We should think about how our service naming (and in general the Agent) works on Full Framework with multiple AppDomains.

Currently we assume 1 service per process, so probably we will send the data from each AppDomain with the same `service.name` with the same `service.agent` values.

I don't know if this is what we want. Either way, I think we should test this, think about what our goal is in such a (not so uncommon) scenario and document it.

gregkalapos:
https://github.com/elastic/apm-server/issues/1175 is relevant if we want to have e.g. different service name pro AppDomain.

gregkalapos:
Not relevant at this stage - closing.

PR #128, opened 2019-02-25
Fix #121 ConsoleLogger should pick up configuration

### LINKED ISSUE #121: Incorrect LogLevel in ConsoleLogger
Console logger always has a default value "Error", it doesn`t use LogLevel from a configuration.  The possible reason is that ConsoleLogger.Instance is called before config was read by the reader

Mpdreamz:
Pending elastic/apm-agent-dotnet#87 

Continuation of this PR that also fixes elastic/apm-agent-dotnet#121

### DIFF
--- a/src/Elastic.Apm/Agent.cs
+++ b/src/Elastic.Apm/Agent.cs
@@ -37,9 +37,9 @@ namespace Elastic.Apm
 
 		IPayloadSender PayloadSender { get; }
 
-		ITracer Tracer { get; }
-
 		Service Service { get; }
+
+		ITracer Tracer { get; }
 	}
 
 	internal class ApmAgent : IApmAgent, IDisposable
@@ -66,7 +66,6 @@ namespace Elastic.Apm
 	{
 		private static readonly Lazy<ApmAgent> Lazy = new Lazy<ApmAgent>(() => new ApmAgent(_components));
 		private static AgentComponents _components;
-
 
 		public static IConfigurationReader Config => Lazy.Value.ConfigurationReader;
 
--- a/src/Elastic.Apm/Agent.cs
+++ b/src/Elastic.Apm/Agent.cs
@@ -4,6 +4,7 @@ using Elastic.Apm.Api;
 using Elastic.Apm.Config;
 using Elastic.Apm.DiagnosticSource;
 using Elastic.Apm.Logging;
+using Elastic.Apm.Model.Payload;
 using Elastic.Apm.Report;
 
 //TODO: It'd be nice to move this into the .csproj
@@ -37,20 +38,22 @@ namespace Elastic.Apm
 		IPayloadSender PayloadSender { get; }
 
 		ITracer Tracer { get; }
+
+		Service Service { get; }
 	}
 
 	internal class ApmAgent : IApmAgent, IDisposable
 	{
 		internal readonly CompositeDisposable Disposables = new CompositeDisposable();
 
-		public ApmAgent(AgentComponents agentComponents) =>
-			Components = agentComponents ?? new AgentComponents(null, service: null);
+		public ApmAgent(AgentComponents agentComponents) => Components = agentComponents ?? new AgentComponents();
 
 		private AgentComponents Components { get; }
 		public IConfigurationReader ConfigurationReader => Components.ConfigurationReader;
 		public IApmLogger Logger => Components.Logger;
 		public IPayloadSender PayloadSender => Components.PayloadSender;
 		public ITracer Tracer => Components.Tracer;
+		public Service Service => Components.Service;
 
 		internal Tracer TracerInternal => Tracer as Tracer;
 
--- a/src/Elastic.Apm/Config/AbstractConfigurationReader.cs
+++ b/src/Elastic.Apm/Config/AbstractConfigurationReader.cs
@@ -50,8 +50,13 @@ namespace Elastic.Apm.Config
 		{
 			if (TryParseLogLevel(kv?.Value, out var level)) return level.Value;
 
-			Logger?.LogError("Failed parsing log level from {Origin}: {Key}, value: {Value}. Defaulting to log level '{DefaultLogLevel}'",
-				kv.ReadFrom, kv.Key, kv.Value, ConsoleLogger.DefaultLogLevel);
+			if (kv?.Value == null)
+				Logger?.LogDebug("No log level provided. Defaulting to log level '{DefaultLogLevel}'", ConsoleLogger.DefaultLogLevel);
+			else
+			{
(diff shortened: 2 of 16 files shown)


src/Elastic.Apm/Config/AbstractConfigurationReader.cs
```diff
@@ -17,22 +17,34 @@ public abstract class AbstractConfigurationReader
...
+					case "info": return LogLevel.Information;
+					case "warning": return LogLevel.Warning;
+					case "error": return LogLevel.Error;
+					case "critical": return LogLevel.Critical;
+					case "none": return LogLevel.None;
+					default: return null;
+				}
 			}
 		}
+		protected internal static LogLevel ParseLogLevel(string value)
```
gregkalapos: This is only used in test code if I see it correctly. I think it'd be better to only have 1 `ParseLogLevel` method.

src/Elastic.Apm/Agent.cs
```diff
@@ -36,21 +37,23 @@ public interface IApmAgent
 
 		IPayloadSender PayloadSender { get; }
 
+		Service Service { get; }
```
gregkalapos: What's the idea behind this? If I get this correctly this is actually unrelated to the logger. I'm asking this, because with this the `Service` will be tied to the agent. That is technically correct now, but may be a problem later. We are discussing to have multiple services in a single agent. See https://github.com/elastic/apm-agent-dotnet/issues/59 and https://github.com/elastic/apm-server/issues/1175

src/Elastic.Apm/Logging/ConsoleLogger.cs
```diff
@@ -18,14 +18,23 @@ public ConsoleLogger(LogLevel level, TextWriter standardOut = null, TextWriter e
...
+				return new ConsoleLogger(level.Value);
+			return Instance;
+		}
+
+		public LogLevel Level => _level;
+
 
 		public void Log<TState>(LogLevel level, TState state, Exception e, Func<TState, Exception, string> formatter)
 		{
 			//TODO: Discuss UtcNow?
```
gregkalapos: nit: we agreed on this, `UtfNow` FTW.

Mpdreamz:
> Added some comments.
> 
> Plus I have one comment to #87:
> Are we ok with the `switch` block in `ConsoleLogger.Log<TState>(LogLevel level, TState state, Exception e, Func<TState, Exception, string> formatter)` ?
> 
> Currently we basically print everything, it doesn't really matter what loglevel the user sets. I'd simply skip logs that are below the selected log level. So if the user sets the log level to e.g. `Error`, then logs with `LogLevel` `Debug` should not be visible et all. This is not the case now.


The reason for this is that most adapters provide ways to dynamically change the logging level: e.g https://nblumhardt.com/2014/10/dynamically-changing-the-serilog-level/. 

That wouldn't work if we filter prior. Happy to change it though :smile:

> 
> My other problem is the test code. We have tests testing what I describe above in `LoggerTest`. 
> The problem is that that class works with the `TestLogger` class, which I find very unfortunate, because 
> that mocks the prod code that we'd like to test (e.g. the switch block in `ConsoleLogger.Log` is something 
> we should test, but instead we mock that). So I think `TestLogger` is very dangerous and we should try to get rid of it.

Agreed. `TestLogger` exists as an easy way to record. We should look into wrapping ConsoleLogger instead, it accepts TextWriter for stdin and stderr so we should be able to do reroute to a `List<string>` in a subclass. 

The switch logic might need to be moved into an static method somewhere.

src/Elastic.Apm/Logging/ConsoleLogger.cs
Mpdreamz: 🐱 agreed

gregkalapos:
> > Added some comments.
> > Plus I have one comment to #87:
> > Are we ok with the `switch` block in `ConsoleLogger.Log<TState>(LogLevel level, TState state, Exception e, Func<TState, Exception, string> formatter)` ?
> > Currently we basically print everything, it doesn't really matter what loglevel the user sets. I'd simply skip logs that are below the selected log level. So if the user sets the log level to e.g. `Error`, then logs with `LogLevel` `Debug` should not be visible et all. This is not the case now.
> 
> The reason for this is that most adapters provide ways to dynamically change the logging level: e.g https://nblumhardt.com/2014/10/dynamically-changing-the-serilog-level/.
> 
> That wouldn't work if we filter prior. Happy to change it though 😄
> 
> > My other problem is the test code. We have tests testing what I describe above in `LoggerTest`.
> > The problem is that that class works with the `TestLogger` class, which I find very unfortunate, because
> > that mocks the prod code that we'd like to test (e.g. the switch block in `ConsoleLogger.Log` is something
> > we should test, but instead we mock that). So I think `TestLogger` is very dangerous and we should try to get rid of it.
> 
> Agreed. `TestLogger` exists as an easy way to record. We should look into wrapping ConsoleLogger instead, it accepts TextWriter for stdin and stderr so we should be able to do reroute to a `List<string>` in a subclass.
> 
> The switch logic might need to be moved into an static method somewhere.

Could we move the current logic with the switch block into a base class and then in the `ConsoleLogger` only log things according to the log level (so filter out things)? Additionally we could enable users to pass us a different implementation which then gets all the logs  - but by default in the console I'd not ignore the LogLevel setting, because that is a very strange behaviour. 

I think this would also solve our problem with testing. 

Does this sound reasonable?

Mpdreamz:
@gregkalapos mind having another look? `TestLogger` now implements `ConsoleLogger` and shares all of its logic. Also addressed your other comments 👍

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
```diff
@@ -42,12 +48,15 @@ protected string ParseSecretToken(ConfigurationKeyValue kv)
...
 		{
-			if (kv == null || string.IsNullOrEmpty(kv.Value)) return ConsoleLogger.DefaultLogLevel;
-
-			if (Enum.TryParse<LogLevel>(kv.Value, out var logLevel)) return logLevel;
+			if (TryParseLogLevel(kv?.Value, out var level)) return level.Value;
 
-			Logger?.LogError("Failed parsing log level from {Origin}: {Key}, value: {Value}. Defaulting to log level '{DefaultLogLevel}'",
-				kv.ReadFrom, kv.Key, kv.Value, ConsoleLogger.DefaultLogLevel);
+			if (kv?.Value == null)
+				Logger?.LogDebug("No log level provided. Defaulting to log level '{DefaultLogLevel}'", ConsoleLogger.DefaultLogLevel);
```
gregkalapos: This is a chicken-egg (🐔 🥚)  problem 🤣  We print with `LogDebug`, but set the `LogLevel` to `LogLevel.Error`, so this will never show up in the logs. One solution that comes to my mind is that we default to `LogLevel.Information` and print it as `LogInfo`. This is not high prio...

Issue #124, opened 2019-02-25
Expose fields and objects from `Transaction.Context` through the public agent API


gregkalapos:
# Why?

The current public agent API only exposes a small part of the model that the APM Server expects. These are mainly things directly on `Transaction` and `Span`. But `Transaction.Context` and `Span.Context` are not exposed. On those contexts we can set things like HTTP request related fields (HTTP method, URL, etc.) or Database related fields. 

Since there are lots of libraries that we currently don't support with auto instrumentation we should expose these fields and let users set those when they rely on the public agent API. 

This PR already contains a use case: https://github.com/elastic/apm-agent-dotnet/pull/118 

# Problem description
To avoid a too high-level discussion, I'll focus specifically on `Transaction.Context.Request`, but we should came up with a solution that works with other things on `Transaction.Context` and `Span.Context`. 

[`Request`](https://github.com/elastic/apm-server/blob/master/docs/spec/request.json) has 2 [required fields](https://github.com/elastic/apm-server/blob/master/docs/spec/request.json#L93): 
- Method
- Url

This means that both of these fields must be set, otherwise the the server would reject the `Transaction`. 

# Requirements 
#### (not set it stone, feel free to disagree)
- The main challenge is that we have to make sure that required fields are filled, otherwise the APM server would reject the request. This means we should not have partially initialized objects. E.g.: when the user does this: `Agent.Tracer.CurrentTransaction.Request.Method = "GET";` We could lazily create the `Request` object, but the problem is that the `Url` property is still null, so if the user does not set that property later then this `Request` object is invalid. 

- Avoid throwing exceptions. For example with the previous example (where the server'd reject the transaction, since not all required fields are filled on `Transaction.Context.Request`) we could build some verification before we send the data to the apm server (or at another point) and notify the user by throwing an exception and forcing them to avoid partly initialized data. As a monitoring tool one basic principle is to be as non-intrusive as possible, so I think throwing exceptions is not acceptable. 

- Avoid skipping data. Another option would be to not send that data and print a `warning` (or any other log message). I think this'd cause confusion. 

- Required fields should be forced on the API level. We should prefer solutions where it's not even possible to have partly initilized objects. This'd mean the user either sets all the required fields at once, or none of those.

[...]

PR #130, opened 2019-02-26
Expose Transaction.Context.Request through the public agent API

gregkalapos:
This is a potential solution for https://github.com/elastic/apm-agent-dotnet/issues/124 

Some general thinking:
- We force users to provide required fields - aka: added constructors. It's very hard to have data that'd make the APM server unhappy.
- I added a POCO class `Api.Request`. The idea behind this is that I'd like to avoid exposing the model classes. Those can be changes (yes, breaking changes happen only on major versions), still I think it's not a good idea to expose those - most agents don't do it either.
- The `Request` class flattens the object hierarchy, it does not have properties that are other classes (like `Socket` and `Url`). Idea behind this: I think it's simply confusing... With that we would have an API where we constantly force users to new up objects and nullcheck things.
- Some agents call this `Context`. Since `Context` will probably disappear I'd simply not use that word - I'd not expose the concept of `Context` in generqal. To me this is just an additional information on the `ITransaction` called `Request`. 

- I'd use a similar solution with other things that we'd like to expose on `Context`.

### DIFF
--- a/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
+++ b/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
@@ -6,6 +6,8 @@ using Elastic.Apm.Helpers;
 using Elastic.Apm.Model.Payload;
 using Microsoft.AspNetCore.Http;
 
+using Payload = Elastic.Apm.Model.Payload;
+
 [assembly:
 	InternalsVisibleTo(
 		"Elastic.Apm.Tests, PublicKey=002400000480000094000000060200000024000052534131000400000100010051df3e4d8341d66c6dfbf35b2fda3627d08073156ed98eef81122b94e86ef2e44e7980202d21826e367db9f494c265666ae30869fb4cd1a434d171f6b634aa67fa8ca5b9076d55dc3baa203d3a23b9c1296c9f45d06a45cf89520bef98325958b066d8c626db76dd60d0508af877580accdd0e9f88e46b6421bf09a33de53fe1")]
@@ -32,20 +34,19 @@ namespace Elastic.Apm.AspNetCore
 			var transaction = _tracer.StartTransactionInternal($"{context.Request.Method} {context.Request.Path}",
 				ApiConstants.TypeRequest);
 
-			transaction.Context.Request = new Request
+			var url = new Url
 			{
-				Method = context.Request.Method,
+				Full = context.Request?.Path.Value,
+				HostName = context.Request.Host.Host,
+				Protocol = GetProtocolName(context.Request.Protocol),
+				Raw = context.Request?.Path.Value //TODO
+			};
+			transaction.Context.Request = new Payload.Request(context.Request.Method, url)
+			{
 				Socket = new Socket
 				{
 					Encrypted = context.Request.IsHttps,
 					RemoteAddress = context.Connection?.RemoteIpAddress?.ToString()
-				},
-				Url = new Url
-				{
-					Full = context.Request?.Path.Value,
-					HostName = context.Request.Host.Host,
-					Protocol = GetProtocolName(context.Request.Protocol),
-					Raw = context.Request?.Path.Value //TODO
 				},
 				HttpVersion = GetHttpVersion(context.Request.Protocol)
 			};
--- a/src/Elastic.Apm/Api/Request.cs
+++ b/src/Elastic.Apm/Api/Request.cs
@@ -10,6 +10,7 @@ namespace Elastic.Apm.Api
 
 		public string HttpVersion { get; set; }
 		public string Method { get; }
+		public object Body { get; set; }
 
 		public bool SocketEncrypted { get; set; }
 		public string SocketRemoteAddress { get; set; }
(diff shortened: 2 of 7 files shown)


src/Elastic.Apm/Api/Request.cs
```diff
@@ -0,0 +1,22 @@
+namespace Elastic.Apm.Api
+{
+	/// <summary>
+	/// Encapsulates Request related information that can be attached to an <see cref="ITransaction" />.
+	/// See <see cref="ITransaction.Request" />
+	/// </summary>
+	public class Request
```
gregkalapos: We have another internal class called `Request`, but that does not affect users and I think we can live with this.

src/Elastic.Apm/Model/Payload/Transaction.cs
```diff
@@ -51,6 +53,21 @@ public Transaction(IApmLogger logger, string name, string type, IPayloadSender s
 
 		public string Name { get; set; }
 
+		public Api.Request Request
+		{
+			get => _request;
+			set
+			{
+				_request = value;
+				_context.Value.Request = new Request(value.Method,
```
gregkalapos: This can be also done during serialization... I'm not sure where to do this... I'm ok with moving this.

test/Elastic.Apm.AspNetCore.Tests/AspNetCoreMiddlewareTests.cs
```diff
@@ -115,7 +115,7 @@ public async Task FailingRequestWithoutConfiguredExceptionPage()
 			Assert.Single(_capturedPayload.Errors[0].Errors);
 
 			//also make sure the tag is captured
-			Assert.Equal(((_capturedPayload.Errors[0] as Error)?.Errors[0] as Error.ErrorDetail)?.Context.Tags["foo"], "bar");
```
gregkalapos: Unrelated change, sry about this.

test/Elastic.Apm.Tests/ApiTests/ConvenientApiTransactionTests.cs
```diff
@@ -380,7 +381,7 @@ public async Task CancelledAsyncTask()
 			Assert.Equal(TransactionType, payloadSender.Payloads[0].Transactions[0].Type);
 
 			//`await Task.Delay` is inaccurate, so we enable 10% error in the test
-			var expectedTransactionLength = SleepLength - (SleepLength * 0.1);
```
gregkalapos: Unrelated change, sry about this.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
```diff
@@ -32,21 +34,20 @@ public async Task InvokeAsync(HttpContext context)
...
 
-			transaction.Context.Request = new Request
+			var url = new Url
+			{
+				Full = context.Request?.Path.Value,
+				HostName = context.Request.Host.Host,
+				Protocol = GetProtocolName(context.Request.Protocol),
+				Raw = context.Request?.Path.Value //TODO
+			};
+			transaction.Context.Request = new Payload.Request(context.Request.Method, url)
```
gregkalapos: In internal code we can still directly set things on `Context`. This may be a little bit confusing, but from perf reasons I think we should do it this way - we avoid copying stuff this way.

src/Elastic.Apm/Model/Payload/Transaction.cs
gregkalapos: 🤦‍♂️  We have to... this won't work if we do it here.

gregkalapos: Done, moved this to `End()`

skynet2:
Hi @gregkalapos ,

Are you going to expose Response ?

src/Elastic.Apm/Model/Payload/Transaction.cs
```diff
@@ -78,6 +80,16 @@ public void End()
 		{
 			if (!Duration.HasValue) Duration = (DateTimeOffset.UtcNow - Start).TotalMilliseconds;
 
+			if (_context?.Value != null && _context.Value.Request == null && Request != null)
+			{
+				_context.Value.Request = new Request(Request.Method,
```
skynet2: Good idea

gregkalapos:
> Hi @gregkalapos ,
> 
> Are you going to expose Response ?

Yes, definitely! The concept would be the same: we create a POCO object for the API surface and expose that on `ITransaction`.  Assuming this PR is accepted... 

So this PR is about 2 thing: 1) the general concept about how we expose these things and 2) the specific implementation for `Request`. Once this is done we can move on and implement this for more things.

src/Elastic.Apm/Api/Request.cs
skynet2: Hi @gregkalapos , maybe you can add Body field to the object please ?

gregkalapos: Good catch, thanks! 

Fixed.

skynet2: Thanks @gregkalapos

skynet2:
Hi @gregkalapos ,

Is there any known time when will we have this code in master branch?

gregkalapos:
This got outvoted.

There will be a follow-up PR according to https://github.com/elastic/apm-agent-dotnet/issues/124, which will expose the internal server API.

PR #133, opened 2019-02-26
move assertions over to fluent assertions

Mpdreamz:
https://fluentassertions.com makes it easy to write assertions that have good exception messages and ships with a lot of useful asserts.

`Assert.IsTrue` `Assert.IsFalse` `Assert.Equal` all kind of hide the intend of the assertion. 

https://fluentassertions.com/tips/#improved-assertions lists quite an extensive list on how it can be utilized to write sane assertions 👍 

One particular thing I like is that it captures the variable under test (hence why `<DebugSymbols>` is now always true). 

Consider this xUnit assert:

```csharp
Assert.Equal("bar", payloadSender.SpansOnFirstTransaction[0].Context.Tags["foo"]);
```

```
Assert.Equal() Failure
             ↓ (pos 3)
Expected: bar
Actual:   bar2
             ↑ (pos 3)
```

vs FluentAssertions

```csharp
payloadSender.SpansOnFirstTransaction[0].Context.Tags.Should().Contain("foo","bar");
````

```
Expected payloadSender.SpansOnFirstTransaction[0].Context.Tags to contain value "bar" at key "foo", but found "bar2".
```

The nice thing here that it captures the `SUT` and will assert tags actually not being null or empty and having a key foo in one go 👍 

String assertions also include positional information:

```

Expected string to be "xya", but "xyz" differs near "z" (index 2).
```

Albeit it less nice then xUnit assertions.

### DIFF
--- a/test/Elastic.Apm.AspNetCore.Tests/AspNetCoreDiagnosticListenerTest.cs
+++ b/test/Elastic.Apm.AspNetCore.Tests/AspNetCoreDiagnosticListenerTest.cs
@@ -1,8 +1,10 @@
 ﻿using System;
 using System.Net.Http;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Elastic.Apm.Model.Payload;
 using Elastic.Apm.Tests.Mocks;
+using FluentAssertions;
 using Microsoft.AspNetCore.Mvc.Testing;
 using SampleAspNetCoreApp;
 using Xunit;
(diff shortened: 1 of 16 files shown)


Mpdreamz:
@gregkalapos rebased, also introduced duration should extension methods.

gregkalapos:
@Mpdreamz Thank you! Once the build goes through I'd say we can merge. 
👍

Mpdreamz:
Thanks for the quick looksie over @gregkalapos merged :)

PR #134, opened 2019-02-27
Expose Transaction.Context.Request and Transaction.Context.Response

gregkalapos:
Potential solution to elastic/apm-agent-dotnet#124 - 2. Version, follow-up from elastic/apm-agent-dotnet#130 

Types that we make public moved from `Elastic.Apm.Model.Payload` to `Elastic.Apm.Api`. 

This makes both `Context.Request` and `Context.Response` public.

~~I also made `Request`, `Response`, `Url`, and `Socket` to a `struct`.~~ Update: reverted, we use `class`.

~~With this dangerous code like this won't compile:~~
```
transaction.Context.Request.Method = "GET";
```

~~And code like this will work and won't throw, even if Request is not initialized:~~
```
var method = t.Context.Request.Method;
```

~~This'll also work and not throw:~~
```
var method = t.Context.Request.Url.Full;
```

~~I'll follow up on this, maybe we'll need to adjust our serialization to this, but I think it's worth it.~~


gregkalapos:
Reverted `struct` after discussing it. 

Reason: This value-behaviour and forcing users to set everything at once would be strange in C#. 

😢

test/Elastic.Apm.Tests/ApiTests/ConvenientApiTransactionTests.cs
```diff
@@ -467,6 +467,92 @@ public async Task TagsOnTransactionAsyncError()
...
+					{
+						WaitHelpers.SleepMinimum();
+						transaction.Context.Response = new Response() { Finished = true, StatusCode = 200 };
+					});
+				});
+
+			Assert.True(payloadSender.FirstTransaction.Context.Response.Finished);
+			Assert.Equal(200, payloadSender.FirstTransaction.Context.Response.StatusCode);
+		}
+
```
SergeyKleyman: Maybe it's worth adding a test that set a property of Request/Response object after creation.

gregkalapos: 👍 Added a test doing that.

Issue #129, opened 2019-02-26
Truncate strings 


gregkalapos:
Mostly triggered by https://github.com/elastic/apm-agent-dotnet/pull/118 

The initial suggestion was to do the truncation in the setters. 

We discussed this with @Mpdreamz and @SergeyKleyman and we found more advantages to do the truncation during serialization. 

- If it's serialization time we can most likely write a more generic way to trim/clean up these props while the dto's can stay simple getter/setters.
- If we do it in the setters we'd do it on a user-thread. By doing it during serialization we'd move this work to a non-user thread.
- In case we add new classes/string properties it's easy to forget if this is done in each setter

So the current agreement is to do this work during serialiazation.

FYI: @skynet2

skynet2:
HI @gregkalapos ,

My first intention was also to do that during serialization, but as I remember you had an idea to remove json.net serializer or am I missing something?

Thanks,
Stas

gregkalapos:
Hi @skynet2 , 

Do you mean [this comment](https://github.com/elastic/apm-agent-dotnet/pull/120#discussion_r259005099)?

By "change" I meant changing the number 1024. Whether we change json.net, I don't know... we talked about utf8json, but we don't really have a final decision. But I'd expect to have the option to do this during serialization with all major json serializers. 

Anyway, I think everyone is convinced already that it's better to move this to serliazation time. 

Btw. the [java agent also](https://github.com/elastic/apm-agent-java/blob/master/apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/DslJsonSerializer.java#L927) does it that way.

skynet2:
Hi @gregkalapos ,
Nope i`m talking about https://github.com/elastic/apm-agent-dotnet/issues/106  - Json.NET (the elasticsearch client already solves this, we should reuse this solution)

gregkalapos:
> Hi @gregkalapos ,
> Nope i`m talking about #106 - Json.NET (the elasticsearch client already solves this, we should reuse this solution)

I see, yes, that's what I referred to in my previous comment "we talked about utf8json, but we don't really have a final decision". 

But that should not change the concept regarding truncation.

Do we have any remaining objection against moving truncation to serialization time? If there is nothing I'd say we'll progress with this soon.

gregkalapos:
PR: https://github.com/elastic/apm-agent-dotnet/pull/135

gregkalapos:
PR #135 merged. Fixed.

PR #135, opened 2019-02-28
String truncation

gregkalapos:
Solves #129. 

This is very Json.NET specific. Maybe we move to another json serializer later, but still I think it's nice to have this now, so I'd say: let's solve it for the current serializer 🚀 

After this change:

- by default we trim every `string` property
- by default we also trim the values inside `Dictionary<string, string>`
- in case we don't want to trim something we can mark it with `[NoTruncationInJsonNet]`

+ added tests.


src/Elastic.Apm/Model/Payload/Span.cs
```diff
@@ -42,17 +43,7 @@ public Span(string name, string type, Transaction transaction)
 
 		public int Id { get; set; }
 
-		private string _name;
-		public string Name
-		{
-			get => _name;
-			set
-			{
-				if (value.Length > Consts.PropertyMaxLength)
```
gregkalapos: We don't truncate here anymore. This happens before serialization.

test/Elastic.Apm.Tests/BasicAgentTests.cs
```diff
@@ -40,29 +40,6 @@ public void AgentVersion()
...
-
-			for (var i = 0; i < 1030; i++) spanName.Append('a');
-
-			var payloadSender = new MockPayloadSender();
-			var agent = new ApmAgent(new TestAgentComponents(payloadSender: payloadSender));
-
-			agent.Tracer.CaptureTransaction("TestTransaction", "Test", (t) => { t.CaptureSpan(spanName.ToString(), "test", () => { }); });
-
-			Assert.NotNull(payloadSender.FirstSpan);
-			Assert.Equal(1024, payloadSender.FirstSpan.Name.Length);
```
gregkalapos: This isn't valid anymore. We can think about testing more in `PayloadSender`, but atm if we go though `MockPayloadSender` the trimming logic is not triggered. With Intake V2 there'll be a new `PayloadSender`, so I'd rather not focus on this now... we have good coverage by the other new tests I added.

skynet2:
HI @gregkalapos ,
Correct me if I'm wrong but your code will truncate all string and all dictionaries. 
According to json schema only some fields should be truncated, maybe it`s better to create attribute Truncate and truncate only that strings which are marked with that attribute?

gregkalapos:
> HI @gregkalapos ,
> Correct me if I'm wrong but your code will truncate all string and all dictionaries.
> According to json schema only some fields should be truncated, maybe it`s better to create attribute Truncate and truncate only that strings which are marked with that attribute?

By default we truncate everything. I basically have an inverse logic here: by default we truncate, and if we don't want to truncate something we add `[NoTruncationInJsonNet]` to it. I do this, because at first glance I think we have more fields with a maxLength limit than fields without limit and in case we don't truncate the server would reject the whole payload, so it's better to be on the safe side. 

I added the attribute to (I think) all fields where it's needed. It'd be nice to see #134 merged (since that moves some types that are effected here), and then I'd make `Span.Context` things also public and then I'll go through all fields again to make sure we don't miss anything we should not truncate.

test/Elastic.Apm.Tests/BasicAgentTests.cs
Mpdreamz: ++ `MockPayloadSender` is only used to assert messages get send not necessarily how. 

More isolated serialization tests would be helpful

src/Elastic.Apm/Model/Payload/Stacktrace.cs
```diff
@@ -1,9 +1,11 @@
-﻿using Newtonsoft.Json;
+﻿using Elastic.Apm.Report.Serialization;
+using Newtonsoft.Json;
 
 namespace Elastic.Apm.Model.Payload
 {
 	internal class Stacktrace
 	{
+		[NoTruncationInJsonNet]
```
Mpdreamz: Would it be cleaner to declare the reverse? `[Truncate(int chars)]` instead?

src/Elastic.Apm/Report/Serialization/StringTruncationValueResolver.cs
```diff
@@ -0,0 +1,88 @@
...
+using Newtonsoft.Json.Serialization;
+
+namespace Elastic.Apm.Report.Serialization
+{
+	/// <inheritdoc />
+	/// <summary>
+	/// Automatically applies <see cref="StringValueProvider"/> to <see cref="String"/> properties, and <see cref="StringDictionaryValueProvider"/>
+	/// to <see cref="Dictionary{String, String}"/> unless they are marked with <see cref="NoTruncationInJsonNetAttribute"/>.
+	/// </summary>
+	internal class StringTruncationValueResolver : CamelCasePropertyNamesContractResolver
```
Mpdreamz: Could potentially just be two `JsonConverter` e.g `TrimmedStringJsonConverter` and ` TagsJsonConverter` (for the `Dictionary<string, string>` case handled here as well. 

This would eschew the custom contract resolver and value providers.

src/Elastic.Apm/Model/Payload/Stacktrace.cs
Mpdreamz: I just spotted @skynet2 comment. I tend to agree, a default truncate feels implicit rather than explicit.

src/Elastic.Apm/Report/Serialization/StringTruncationValueResolver.cs
gregkalapos: 👍 Agreed, it makes things easier. Added the two converters, removed the contract resolver and the value provider.

src/Elastic.Apm/Model/Payload/Stacktrace.cs
gregkalapos: Ok, since my logic seems to be strange for basically everyone, we go with the reverse. 

Just for reference: the main reason for trimming by default was this:
If we trim by default and we forget to add `[DontTruncate]` -> We trim the string, but the first 1024 chars and everything else is visible.
If we don't trim and forget to add `[Truncate]` -> The server rejects the request (the whole transaction or span), nothing is visible. 

Of course we should not forget to add attributes in the first place :) 

Anyway, I accept that trimming by default would be strange. Reversed the logic.

gregkalapos:
@Mpdreamz thanks for the feedback! 

I pushed another version. 

I created the 2 converters, and I simply attach those to the properties that we want to trim. I think this is the easiest and simplest solution.

Another option would be to activate those globally through `JsonSerializerSettings`, but in that case we need additional attributes to mark things for trimming. I think the current solution is better, because this way we don't have to check for attributes within the converter. Once the converter fires we can trim, since we only add those to properties that we wanna trim.

....will rebase this PR soon....

sample/ApiSamples/Program.cs
```diff
@@ -9,11 +9,15 @@ internal class Program
...
 		{
+			Console.WriteLine("Start");
 			SampleCustomTransactionWithConvenientApi();
 
 			//WIP: if the process terminates the agent
 			//potentially does not have time to send the transaction to the server.
 			Thread.Sleep(1000);
+
+			Console.WriteLine("Done");
+			Thread.Sleep(Int32.MaxValue);
```
Mpdreamz:
```suggestion
		Console.ReadKey();
```

Issue #142, opened 2019-03-13
Add .Log<Level> overload that accepts lambda


SergeyKleyman:
It might be worth checking if debug log level is enabled or better yet to add .LogXYZ overload that accepts lambda and executes only when level XYZ is enabled.

_Originally posted by @SergeyKleyman in https://github.com/elastic/apm-agent-dotnet/pull/136_

gregkalapos:
Addressed in #146

PR #146, opened 2019-03-14
[Discussion #142] Logger optimization with `?.`- allocation free 

gregkalapos:
In #142 we had the idea to add overloads to our logger to log through lambdas. 

This is useful when we have to build the message and then log it. This way in case due to the log level we don't want to log the message we could avoid all the work needed. 

Example:
```	
catch (Exception e)
{
   _logger.LogWarning("Failed sending events. \nException: {ExceptionFullName}, Message: {ExceptionMessage} \nFollowing events were not transferred successfully to the server:\n{items}", e.GetType().FullName, e.Message, string.Join(",\n", queueItems.ToArray()));
}
```

If the `LogLevel` is higher than `Warning`, we do all the work, so all these code will be executed:
```
e.GetType().FullName, e.Message, string.Join(",\n", queueItems.ToArray()
```

But none of that will be used, because due to the loglevel, we don't print anything.

One approach would be to have `if` blocks in the code and check if the current loglevel would enable to log, and if not, then simply skip the work. This is not very nice, and we would always repeat this code. (Although I personally don't think it's very bad). 

By adding an overload to a logger that takes a `Func<sting>` or something similar we could only execute the `Func<string>` if the level enables that. This would save all the work. Problem is, that in this case we would allocate a delegate each time, regardless of the loglevel, so from overhead point of view still not as good as the if block.

This PR goes further and adds a nullable struct combined with `?.` to do the checking. Also by thinking about this I'd say we don't even need overloads that take lambdas. The `?.` already gives us what we want: with that we can prevent the execution of the expensive code when the loglevel would prevent us from printing the stuff we calculate expensively.

So lines like this (this is btw. the proposed syntax):
```
_logger.IfLevel(LogLevel.Debug)
	?.Log("Sent items to server: \n{items}", string.Join(",\n", queueItems.ToArray()));
```

Compile into this:

```
IL_03f2: nop
IL_03f3: ldarg.0
IL_03f4: ldfld class Elastic.Apm.Report.PayloadSenderV2 Elastic.Apm.Report.PayloadSenderV2/'<ProcessQueueItems>d__16'::'<>4__this'
IL_03f9: ldfld class Elastic.Apm.Logging.IApmLogger Elastic.Apm.Report.PayloadSenderV2::_logger
IL_03fe: ldc.i4.1
IL_03ff: call valuetype [netstandard]System.Nullable`1<valuetype Elastic.Apm.Logging.LoggingExtensions/MaybeLogger> Elastic.Apm.Logging.LoggingExtensions::IfLevel(class Elastic.Apm.Logging.IApmLogger, valuetype Elastic.Apm.Logging.LogLevel)
IL_0404: stloc.s 9
IL_0406: ldloca.s 9
IL_0408: dup
IL_0409: call instance bool valuetype [netstandard]System.Nullable`1<valuetype Elastic.Apm.Logging.LoggingExtensions/MaybeLogger>::get_HasValue()
IL_040e: brtrue.s IL_0413

IL_0410: pop
IL_0411: br.s IL_0446

IL_0413: call instance !0 valuetype [netstandard]System.Nullable`1<valuetype Elastic.Apm.Logging.LoggingExtensions/MaybeLogger>::GetValueOrDefault()
IL_0418: ldfld class Elastic.Apm.Logging.LoggingExtensions/MaybeLogger/LogDelegate Elastic.Apm.Logging.LoggingExtensions/MaybeLogger::Log
IL_041d: ldstr "Sent items to server: \n{items}"
IL_0422: ldc.i4.1
IL_0423: newarr [netstandard]System.Object
IL_0428: dup
IL_0429: ldc.i4.0
IL_042a: ldstr ",\n"
IL_042f: ldarg.0
IL_0430: ldfld object[] Elastic.Apm.Report.PayloadSenderV2/'<ProcessQueueItems>d__16'::queueItems
IL_0435: call !!0[] [netstandard]System.Linq.Enumerable::ToArray<object>(class [netstandard]System.Collections.Generic.IEnumerable`1<!!0>)
IL_043a: call string [netstandard]System.String::Join(string, object[])
IL_043f: stelem.ref
IL_0440: callvirt instance void Elastic.Apm.Logging.LoggingExtensions/MaybeLogger/LogDelegate::Invoke(string, object[])
IL_0445: nop

IL_0446: nop
```

And as you can spot in the IL code in case the `loglevel` is higher than `Debug` we don't allocate anything and we also don't run the expensive code, this part skips those instructions: 

```
IL_0409: call instance bool valuetype [netstandard]System.Nullable`1<valuetype Elastic.Apm.Logging.LoggingExtensions/MaybeLogger>::get_HasValue()
IL_040e: brtrue.s IL_0413
```

Ps: heavily inspired by @markrendle from [this presentation](https://youtu.be/QgL6vVyjQzw?t=2883).

### DIFF
--- a/src/Elastic.Apm/Logging/IApmLoggingExtensions.cs
+++ b/src/Elastic.Apm/Logging/IApmLoggingExtensions.cs
@@ -81,16 +81,15 @@ namespace Elastic.Apm.Logging
 			public void Log(string message, params object[] args) => _logger.DoLog(_level, message, null, args);
 
 			public void LogException(Exception exception, string message, params object[] args) =>
-				_logger.DoLog(_level, $"Exception: {{ExceptionType}}, Message: {{ExceptionMsg}} \n{message}", exception,
-					exception.GetType().FullName, exception.Message, args);
+				_logger.DoLog(_level, message, exception, args, exception.GetType().FullName, exception.Message);
 
 			public void LogExceptionWithCaller(Exception exception,
 				[CallerMemberName] string method = "",
-				[CallerFilePath] string filepath = "",
+				[CallerFilePath] string filePath = "",
 				[CallerLineNumber] int lineNumber = 0
 			)
 			{
-				var file = string.IsNullOrEmpty(filepath) ? string.Empty : new FileInfo(filepath).Name;
+				var file = string.IsNullOrEmpty(filePath) ? string.Empty : new FileInfo(filePath).Name;
 				_logger?.DoLog(_level, "{ExceptionName} in {Method} ({File}:{LineNumber}): {ExceptionMessage}", exception,
 					exception?.GetType().Name, method, file, lineNumber, exception?.Message);
 			}
--- a/src/Elastic.Apm/Logging/IApmLoggingExtensions.cs
+++ b/src/Elastic.Apm/Logging/IApmLoggingExtensions.cs
@@ -68,12 +68,16 @@ namespace Elastic.Apm.Logging
 		internal static MaybeLogger? IfLevel(this IApmLogger logger, LogLevel level) => logger.Level <= level ? new MaybeLogger(logger, level) : (MaybeLogger?)null;
 		internal readonly struct MaybeLogger
 		{
-			internal delegate void LogDelegate(string s, params object[] args);
+			private readonly IApmLogger _logger;
+			private readonly LogLevel _level;
 
-			internal readonly LogDelegate Log;
+			public MaybeLogger(IApmLogger logger, LogLevel level) => (_logger, _level) = (logger, level);
 
-			public MaybeLogger(IApmLogger logger, LogLevel level)
-				=> Log = (message, args) => logger.DoLog(level, message, null, args);
+			public void Log(string message, params object[] args) => _logger.DoLog(_level, message, null, args);
+
+			public void LogException(Exception exception, string message, params object[] args) =>
+ 				_logger.DoLog(_level, $"Exception: {{ExceptionType}}, Message: {{ExceptionMsg}} \n{message}", exception,
+					new object[]{ exception.GetType().FullName, exception.Message, args });
 		}
 	}
 }
--- a/src/Elastic.Apm/Logging/IApmLoggingExtensions.cs
+++ b/src/Elastic.Apm/Logging/IApmLoggingExtensions.cs
@@ -56,5 +56,33 @@ namespace Elastic.Apm.Logging
 			var logValues = formatter.GetState(args);
 			logger?.Log(level, logValues, e, (s, _) => formatter.Format(args));
 		}
+
+		/// <summary>
+		/// Depending on the two parameters it either returns a MaybeLogger instance or null.
+		/// By using this class with `?.` you can avoid allocating delegates that are not necessary to allocate
+		/// in case the log won't be printed because the loglevel would not allow it.
+		/// </summary>
+		/// <param name="logger"></param>
+		/// <param name="level"></param>
+		/// <returns>If the return value is not null you can call <see cref="MaybeLogger.Log"/> to log</returns>
+		internal static MaybeLogger? IfLevel(this IApmLogger logger, LogLevel level) => logger.Level <= level ? new MaybeLogger(logger, level) : (MaybeLogger?)null;
+		internal readonly struct MaybeLogger
+		{
+			private readonly IApmLogger _logger;
+			private readonly LogLevel _level;
+
+			public MaybeLogger(IApmLogger source, LogLevel level)
+			{
+				_logger = source;
+				_level = level;
+			}
+
+			public void Log(Func<String> getLog)
+			{
+				var lineToPrint = getLog();
+				//Do the logging
+				//WIP: must be extended to handle args, but that should be doable
+			}
+		}
 	}
 }
--- a/src/Elastic.Apm/Logging/IApmLoggingExtensions.cs
+++ b/src/Elastic.Apm/Logging/IApmLoggingExtensions.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.IO;
+using System.Runtime.CompilerServices;
 
 namespace Elastic.Apm.Logging
 {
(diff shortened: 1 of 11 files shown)


gregkalapos:
Update: I think we don't even need lambda here... 

Added a new version of `MaybeLogger` that uses a delegate instead of `Func` (The problem with `Func` and `Action` is that they don't support `params` parameters ). This simply forwards to `DoLog(this IApmLogger logger, LogLevel level, string message, Exception e, object[] args)`.

With the `?.` synctax we still wrap the whole thing into an `if` block once it goes through the compiler, so whatever we have after `?.` it simply won't run (and allocate) if the log level would prevent us from printing the log.

src/Elastic.Apm/Report/PayloadSenderV2.cs
```diff
@@ -166,29 +166,20 @@ private async Task ProcessQueueItems(object[] queueItems)
 
 				if (result != null && !result.IsSuccessStatusCode)
 				{
-					var str = await result.Content.ReadAsStringAsync();
-					_logger.LogError("Failed sending event. {ApmServerResponse}", str);
+					_logger?.IfLevel(LogLevel.Error)?.Log("Failed sending event. {ApmServerResponse}", await result.Content.ReadAsStringAsync());
```
Mpdreamz: Small suggestion what if `IfLevel()` is private and we expose the different levels as methods:

```suggestion
					_logger?.Debug()?.Log("Failed sending event. {ApmServerResponse}", await result.Content.ReadAsStringAsync());
```

IMHO we can then remove the other extensions methods on `IApmLogger` such as `LogWarning()` et all.

gregkalapos: 👍  Done, nice idea!

gregkalapos:
Updated the extension methods based on @Mpdreamz's suggestion to support this syntax:

```
_logger?.Debug()?.Log("Sent items to server: \n{items}",
 string.Join(",\n", queueItems.ToArray()));
```

Similarly we also have `_logger?.Warning()?.Log(...)`, `_logger?.Error()?.Log(...)`, etc.

Also added methods to `MaybeLogger` to support exceptions. 

Plus I adapted all logs in the codebase to use this. 

I measured the impact with BenchmarkDotNet:
Test: crate 100 `Transaction`s with 10 `Span`s on each `Transaction` and finish all of those with default log level, which is `Error` without any error or exception. This means during the test 0 log line was printed.

Before the `?.` syntax (aka running the code to gather stuff for logging and then not using it):
```
|                      Method |     Mean |     Error |    StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|---------------------------- |---------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
| M100TransactionsWith10Spans | 11.93 ms | 0.2591 ms | 0.7639 ms |   1656.2500 |    671.8750 |    203.1250 |             9.02 MB |
```

After adding the `?.` syntax (this is the current stage of the PR):
```
|                      Method |     Mean |     Error |    StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|---------------------------- |---------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
| M100TransactionsWith10Spans | 1.083 ms | 0.0146 ms | 0.0130 ms |    228.5156 |    113.2813 |           - |             1.37 MB |
```

🎉 🎉  Yupp, that is a 10x difference. 🚀 🚀

src/Elastic.Apm/Logging/IApmLoggingExtensions.cs
```diff
@@ -1,60 +1,99 @@
...
-		{
-			var formatter = logger is ScopedLogger sl
-				? sl.GetOrAddFormatter(message, args.Length)
-				: Formatters.GetOrAdd(message, s => new LogValuesFormatter(s));
+			public void Log(string message, params object[] args) => _logger.DoLog(_level, message, null, args);
 
-			var logValues = formatter.GetState(args);
-			logger?.Log(level, logValues, e, (s, _) => formatter.Format(args));
+			public void LogException(Exception exception, string message, params object[] args) =>
+				_logger.DoLog(_level, $"Exception: {{ExceptionType}}, Message: {{ExceptionMsg}} \n{message}", exception,
```
gregkalapos: We add the details of the exception to the log message here. Optionally we can move this to `ConsoleLogger`. My intention with this is that I think we should have the `Exception` also in a textual form and print that to the user. It's nice to have it in a structured form for sophisticated logging frameworks, but for a simple case like `ConsoleLogger` we should also give that info within the text that we print.

Mpdreamz: https://github.com/aspnet/Extensions/blob/master/src/Logging/Logging.Console/src/ConsoleLogger.cs

Adapters typically take care of this already when floating exception. I'm personally +1 to move this to our consolelogger as well

src/Elastic.Apm/Logging/IApmLoggingExtensions.cs
```diff
@@ -1,60 +1,99 @@
...
+					exception.GetType().FullName, exception.Message, args);
+
+			public void LogExceptionWithCaller(Exception exception,
+				[CallerMemberName] string method = "",
+				[CallerFilePath] string filepath = "",
```
SergeyKleyman: Small nitpick `filePath`

src/Elastic.Apm/Logging/IApmLoggingExtensions.cs
```diff
@@ -1,60 +1,99 @@
...
-				: Formatters.GetOrAdd(message, s => new LogValuesFormatter(s));
+			public void Log(string message, params object[] args) => _logger.DoLog(_level, message, null, args);
 
-			var logValues = formatter.GetState(args);
-			logger?.Log(level, logValues, e, (s, _) => formatter.Format(args));
+			public void LogException(Exception exception, string message, params object[] args) =>
+				_logger.DoLog(_level, $"Exception: {{ExceptionType}}, Message: {{ExceptionMsg}} \n{message}", exception,
+					exception.GetType().FullName, exception.Message, args);
+
+			public void LogExceptionWithCaller(Exception exception,
```
SergeyKleyman: I think we should always have a message explaining the meaning of log statement. For example, currently the only usage of `LogExceptionWithCaller` in `PayloadSenderV2.cs` should probably mention that `TaskCanceledException` is expected when Agent PayloadSender shuts down.

src/Elastic.Apm/Logging/IApmLoggingExtensions.cs
gregkalapos: I see. Agree, let's do it like `Microsoft.Extensions.Logging.Console.ConsoleLogger` - moved this part into our `ConsoleLogger`.

src/Elastic.Apm/Logging/IApmLoggingExtensions.cs
gregkalapos: 👍 fixed.

src/Elastic.Apm/Logging/IApmLoggingExtensions.cs
gregkalapos: For that we have the other method (`public void LogException(Exception exception, string message, params object[] args)`), and you are right, that's what we use most of the time. 

This one is intended when you only want to log the exception. We also use this in `HttpDiagnosticListener.OnError`, probably we should do it also in other listeners; in those cases I'm not really sure we can add a more meaningful error message than having the caller method + the exception details.

SergeyKleyman: I still think a useful message can be provided in almost any context and definitely for the ones you have mentioned.
Another reason for always having message as part of logging statement is that this way logging statement can serve two purposes for the price of one - with s message it's also a comment explaining why surrounding code does what it does. For example in `PayloadSenderV2.cs` a message explaining that TaskCanceledException is expected when Agent PayloadSender shuts down makes it clear to a reader of the code why this exception is swallowed.

SergeyKleyman: We discussed on zoom and agreed that we will keep a method to log exception without a message for just few use cases when it makes sense and if we see in the future that it's an anti-pattern we will reconsider removing it.

PR #147, opened 2019-03-18
[APM-CI] use current as pipeline shared library version

kuisathaverat:
* avoid having to change the Jenkinsfile every time we set a pipeline shared library as a stable version to use.
* Use buildDocs pipeline step to build docs
* limit max PR builds per hour to 60
* set quiet period to 10 seconds


kuisathaverat:
@gregkalapos yep, it is related to have the same build step for all agents and build all in the same way

Issue #27, opened 2018-12-13
Introduce ELASTIC_APM_CAPTURE_HEADERS option


gregkalapos:
If set to true, the agent will capture request and response headers, including cookies.
NOTE: Setting this to true can increase the network and disk requirements [and increases allocations].

If false the agent doesn't capture headers.

Default: true

gregkalapos:
Implemented by @se in #148.

PR #148, opened 2019-03-18
Added request and response headers to context.

se:
Maybe we can also add an environment variable to control it?

### DIFF
--- a/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
+++ b/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
@@ -40,9 +40,14 @@ namespace Elastic.Apm.AspNetCore
 				Raw = context.Request?.Path.Value //TODO
 			};
 
-			var requestHeaders = new Dictionary<string, object>();
-			foreach (var header in context.Request.Headers)
-				requestHeaders.Add(header.Key, header.Value.ToString());
+			Dictionary<string, object> requestHeaders = null;
+			if (Agent.Config.CaptureHeaders)
+			{
+				requestHeaders = new Dictionary<string, object>();
+
+				foreach (var header in context.Request.Headers)
+					requestHeaders.Add(header.Key, header.Value.ToString());
+			}
 
 			transaction.Context.Request = new Request(context.Request.Method, url)
 			{
@@ -62,12 +67,17 @@ namespace Elastic.Apm.AspNetCore
 			catch (Exception e) when (ExceptionFilter.Capture(e, transaction)) { }
 			finally
 			{
-				var responseHeaders = new Dictionary<string, object>();
-				foreach (var header in context.Response.Headers)
-					responseHeaders.Add(header.Key, header.Value.ToString());
+				Dictionary<string, object> responseHeaders = null;
 
-				transaction.Result =
-					$"{GetProtocolName(context.Request.Protocol)} {context.Response.StatusCode.ToString()[0]}xx";
+				if (Agent.Config.CaptureHeaders)
+				{
+					responseHeaders = new Dictionary<string, object>();
+
+					foreach (var header in context.Response.Headers)
+						responseHeaders.Add(header.Key, header.Value.ToString());
+				}
+
+				transaction.Result = $"{GetProtocolName(context.Request.Protocol)} {context.Response.StatusCode.ToString()[0]}xx";
 				transaction.Context.Response = new Response
 				{
 					Finished = context.Response.HasStarted, //TODO ?
--- a/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
+++ b/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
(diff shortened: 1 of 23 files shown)


gregkalapos:
Related issue regarding the setting: https://github.com/elastic/apm-agent-dotnet/issues/27

se:
I don't know why .editorconfig ignored by VS Code. I will look at it and fix. BTW, I will also add test code too.

se:
I hope these are enough @gregkalapos 😉

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
```diff
@@ -39,14 +40,24 @@ public async Task InvokeAsync(HttpContext context)
 				Raw = context.Request?.Path.Value //TODO
 			};
 
-			transaction.Context.Request = new Request( context.Request.Method, url)
+			Dictionary<string, object> requestHeaders = null;
+			if (Agent.Config.CaptureHeaders)
```
gregkalapos: We should pass an `IConfigurationReader` instance to `ApmMiddleware` here and not use the static `Agent.Config` property. With that we can inject the reader during testing. The problem with using the static instance is that we have tests running in parallel. On my machine when I run all the tests then `HomeSimplePageTransactionTest` fails on this assert:
```
if (_agent.ConfigurationReader.CaptureHeaders)
{
    transaction.Context.Response.Headers.Should().NotBeNull(); //<- this remains null
```

If I only run `HomeSimplePageTransactionTest` then it's green. (btw. I'm thinking about maybe getting rid of the `.ConfigurationReader` property... )

(e.g. `ReadConfingsFromEnvVarsViaIConfig` sets environment variables and the static instance defaults to reading those, if this happens in parallel, that can be problem, but once we dependency inject a new instance this problem is resolved).

So we can just add an `IConfigurationReader` constructor parameter to `ApmMiddleware` and use that one. [Here](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm.AspNetCore/ApmMiddlewareExtension.cs#L40) we already have one, we can just add a parameter to [this method](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm.AspNetCore/ApmMiddlewareExtension.cs#L51) where we can pass it to the `ApmMiddleware`. Additionally the previous method is used in tests, so each test can inject a new instance.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
```diff
@@ -56,12 +67,22 @@ public async Task InvokeAsync(HttpContext context)
 			catch (Exception e) when (ExceptionFilter.Capture(e, transaction)) { }
 			finally
 			{
-				transaction.Result =
-					$"{GetProtocolName(context.Request.Protocol)} {context.Response.StatusCode.ToString()[0]}xx";
+				Dictionary<string, object> responseHeaders = null;
+
+				if (Agent.Config.CaptureHeaders)
```
gregkalapos: Same as with the previous static instance -> use the injected instance.

test/Elastic.Apm.AspNetCore.Tests/AspNetCoreMiddlewareTests.cs
```diff
@@ -74,6 +74,13 @@ public async Task HomeSimplePageTransactionTest()
 
 			//test transaction.context.response
 			transaction.Context.Response.StatusCode.Should().Be(200);
+			if (_agent.ConfigurationReader.CaptureHeaders)
+			{
+				transaction.Context.Response.Headers.Should().NotBeNull();
+				transaction.Context.Response.Headers.Should().NotBeEmpty();
+			}
+			else
```
gregkalapos: Do we expect this `else` block to be executed? If I understand it correctly we don't parametrize this test, and by default we always capture headers, so we only need the two asserts from the `if` block.

test/Elastic.Apm.AspNetCore.Tests/AspNetCoreMiddlewareTests.cs
```diff
@@ -84,6 +91,14 @@ public async Task HomeSimplePageTransactionTest()
 			transaction.Context.Request.Url.HostName.Should().Be("localhost");
 			transaction.Context.Request.Url.Protocol.Should().Be("HTTP");
 
+			if (_agent.ConfigurationReader.CaptureHeaders)
```
gregkalapos: Same as with the other else block: do we have scenarios where it is actually tested?

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
se: You are right. I didn't want to do that because I tought it can be major. I'm changing it like that ;-) and try to put more test on it.

se:
Thank you for your reviews @gregkalapos. I appreciate it. Is it enough or do we need additional things?

gregkalapos:
Nice! Welcome on the contributor list @se.

se:
> Nice! Welcome on the contributor list @se.

My pleasure 😉. Thank you! More PR's are on the way.

gregkalapos:
Awesome! Feel free to open issues or comment if you want to sync in case you have any question before you put work into a PR.

PR #152, opened 2019-03-20
Add ToStringBuilder helper to support uniform ToString format

SergeyKleyman:
Added ToStringBuilder helper to support uniform ToString format

- Implemented ToStringBuilder helper
- Added tests to cover ToStringBuilder helper
- Replaced existing ToString ad-hoc implementations with ones using ToStringBuilder


src/Elastic.Apm/Elastic.Apm.csproj
```diff
@@ -22,9 +22,9 @@
     <LangVersion>latest</LangVersion>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="11.0.2"/>
```
gregkalapos: 🤔  I think there is some glitch in Rider: When I add new files I also always have this change, where an empty string is added to before the `/>`. But once you auto reformat the file it's gone, so according to the styling guidelines it shouldn't be added. Therefore I always revert those changes, I'd also suggest to do the same here. Maybe we should figure out why this happens.

gregkalapos:
Measured the difference between this and the current logging code on `master`. 

Test case (both ran on `master` and on the `pr branch`):
- 1. test: create 100transaction and 10span on each transaction and end those without any logging (loglevel set to `Error`, no log in this case)
- 2. test: same as 1. test, but with loglevel `Debug`, which utilizes the part of the code with the pr. 

`master` performance: 
```
|                              Method |      Mean |     Error |    StdDev |    Median | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|------------------------------------ |----------:|----------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
|           SimpleTransactions10Spans |  1.153 ms | 0.0157 ms | 0.0147 ms |  1.150 ms |    228.5156 |    113.2813 |           - |             1.37 MB |
| DebugLogSimple100Transaction10Spans | 24.687 ms | 2.8610 ms | 8.4359 ms | 27.703 ms |   1656.2500 |    531.2500 |    125.0000 |             6.43 MB |
```

PR performance: 
```
|                              Method |      Mean |     Error |    StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|------------------------------------ |----------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
|           SimpleTransactions10Spans |  1.100 ms | 0.0111 ms | 0.0092 ms |    228.5156 |    113.2813 |           - |             1.37 MB |
| DebugLogSimple100Transaction10Spans | 26.349 ms | 2.5303 ms | 7.4607 ms |   1468.7500 |    515.6250 |    140.6250 |             8.74 MB |
```

src/Elastic.Apm/Elastic.Apm.csproj
SergeyKleyman: I reverted Elastic.Apm.csproj to its original state (without space before `/>`) but when I tried to format in Rider it kept the  spaces before `/>`

gregkalapos: That's strange. When I format it in Rider it always removed the space before `/>`. Also tried to specify this in the `.editorconfig`, but it does not seem to work. https://developercommunity.visualstudio.com/content/problem/159834/editorconfig-doesnt-work-for-csproj-files.html 
Is maybe my setup wrong and the default is that there should be a space before the `/>`?

ElasticSearch-net also has those spaces: https://github.com/elastic/elasticsearch-net/blob/master/src/Elasticsearch.Net/Elasticsearch.Net.csproj

gregkalapos: Update: I think this was my fault... 👿   I found a setting that controls this:
Screenshot 2019-03-25 at 11 50 01

I suspect the default is `true`, but somehow in my case it was `false`. 

I'll try to figure out how to set this in `.editorconfig` and I'll try to update the whole repo accordingly. If I fail, then I'll simply also set it to `true` for me, update the repo, and we'll continue that way. For this PR I'd say we can go on with whatever option... space or not, doesn't matter, we'll probably udpate this long term.

SergeyKleyman:
I ran the benchmark project added by @gregkalapos on my machine (Windows 10 on MacBook Pro - Intel Core i9-8950HK CPU 2.90GHz, 1 CPU, 12 logical and 6 physical cores).

Results based on code in master on my machine:
```
|                              Method |        Mean |       Error |       StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|------------------------------------ |------------:|------------:|-------------:|------------:|------------:|------------:|--------------------:|
|           SimpleTransactions10Spans |    933.9 us |    18.34 us |     21.12 us |    228.5156 |    114.2578 |           - |             1.37 MB |
| DebugLogSimple100Transaction10Spans | 40,771.0 us | 3,499.04 us | 10,095.52 us |   1285.7143 |    500.0000 |           - |             6.43 MB |
```

Results based on code in this PR:
```
|                              Method |        Mean |       Error |      StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|------------------------------------ |------------:|------------:|------------:|------------:|------------:|------------:|--------------------:|
|           SimpleTransactions10Spans |    905.1 us |    13.65 us |    12.77 us |    228.5156 |    114.2578 |           - |             1.37 MB |
| DebugLogSimple100Transaction10Spans | 29,710.7 us | 3,318.01 us | 9,519.99 us |   1562.5000 |    500.0000 |           - |             8.05 MB |
```


Results based on code in this PR except for ToStringBuilder.Add method being generic:
```
|                              Method |        Mean |       Error |      StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|------------------------------------ |------------:|------------:|------------:|------------:|------------:|------------:|--------------------:|
|           SimpleTransactions10Spans |    912.4 us |    11.65 us |    10.90 us |    228.5156 |    114.2578 |           - |             1.37 MB |
| DebugLogSimple100Transaction10Spans | 31,196.1 us | 2,769.31 us | 8,078.21 us |   1625.0000 |    593.7500 |           - |             8.05 MB |
```

Issue #163, opened 2019-03-27
Capture user


gregkalapos:
`User` should be added to the API and ASP.NET Core should capture the authenticated user.

gregkalapos:
Implemented in #170.

PR #170, opened 2019-04-02
Capture user in ASP.NET Core + public API

gregkalapos:
Solves elastic/apm-agent-dotnet#163 

- Adds the `User` type to the public API (so users of the Agent API can capture the logged in user manually)
- Adds capturing users in ASP.NET Core automatically 
- Extends the sample application (`SampleAspNetCoreApp`) with 'Registration' and 'Login' features
- Adds an integration test that utilizes the new features from `SampleAspNetCoreApp` and makes sure that after the login the user is automatically captured
- (Update) Also looks for OpenID claims to get the user id and email - covered by a unit test.
- Also adds a public API test. 

A minor comment to everyone, who use the `SampleAspNetCoreApp`: I managed to mess up the EF Core migrations, so I removed those and added a fresh one (the sample previously created a table called `Users`, which interferes with ASP.NET Core Identity, to clean this up properly I needed to add a breaking change). In case the sample app does not start due to db issues just remove the `blogging.db` from the disk and send an HTTP GET to `/Home/Index`, that'll create a fresh db. Sorry about this.


src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
```diff
@@ -88,6 +86,22 @@ public async Task InvokeAsync(HttpContext context)
 					Headers = responseHeaders
 				};
 
+				if (context.User?.Identity != null && context.User.Identity.IsAuthenticated && context.User.Identity != null
+					&& transaction.Context.User == null)
+				{
+					transaction.Context.User = new User { UserName = context.User.Identity.Name };
+
+					var idQuery = context.User.Claims.Where(n => n.Type == "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/nameidentifier");
```
lafriks: would be nice if this could be configurable to other claim type or at least also check OAuth/OpenID claims in this case `sub`

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
```diff
@@ -88,6 +86,22 @@ public async Task InvokeAsync(HttpContext context)
...
+					&& transaction.Context.User == null)
+				{
+					transaction.Context.User = new User { UserName = context.User.Identity.Name };
+
+					var idQuery = context.User.Claims.Where(n => n.Type == "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/nameidentifier");
+					var idClaims = idQuery.ToList();
+					if (idClaims.Any())
+						transaction.Context.User.Id = idClaims.FirstOrDefault()?.Value;
+
+					var mailQuery = context.User.Claims.Where(n => n.Type == "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress");
```
lafriks: Same here, in case of OAuth/OpenID - `email`

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
gregkalapos: Great point, thanks! 👍  I think an extra config would be an overkill, so I added a fallback and we also look for OpenID claims.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
gregkalapos: Same as above.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
```diff
@@ -88,6 +87,31 @@ public async Task InvokeAsync(HttpContext context)
 					Headers = responseHeaders
 				};
 
+				if (context.User?.Identity != null && context.User.Identity.IsAuthenticated && context.User.Identity != null
+					&& transaction.Context.User == null)
+				{
+					transaction.Context.User = new User { UserName = context.User.Identity.Name };
+
+					var nameId = GetClaimValue(ClaimTypes.NameIdentifier);
```
lafriks: This could be more optimal
```cs
					transaction.Context.User.Id = GetClaimWithFallbackValue(ClaimTypes.NameIdentifier, "sub");
					transaction.Context.User.Email = GetClaimWithFallbackValue(ClaimTypes.Email, "email");
				}

				string GetClaimWithFallbackValue(string claimType, string fallbackClaimType)
				{
					var idQuery = context.User.Claims.Where(n => n.Type == claimType || n.Type == fallbackClaimType);
					var idClaims = idQuery.ToList();
					return idClaims.Any() ? idClaims.First().Value : string.Empty;
				}

Comment (gregkalapos) [reply src/Elastic.Apm.AspNetCore/ApmMiddleware.cs]: Ok, let's do it this way. Pushed an update.

**Comment by gregkalapos** [comment]
I'd also add logging to this feature in case we need to investigate something related to it.

In #168 I already added a logger to the middleware. To avoid even more conflicts between those 2 PRs I wait until #168 is merged and will add logging after that.

File: src/Elastic.Apm.AspNetCore/ApmMiddleware.cs [hunk 4]
```diff
@@ -88,6 +85,24 @@ public async Task InvokeAsync(HttpContext context)
 					Headers = responseHeaders
 				};
 
+				if (context.User?.Identity != null && context.User.Identity.IsAuthenticated && context.User.Identity != null
+					&& transaction.Context.User == null)
+				{
+					transaction.Context.User = new User
+					{
+						UserName = context.User.Identity.Name,
+						Id = GetClaimWithFallbackValue(ClaimTypes.NameIdentifier, "sub"),
```
SergeyKleyman: I would define a const for `"sub"` - IIUC it's OAuth/OpenID claim type.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
```diff
@@ -88,6 +85,24 @@ public async Task InvokeAsync(HttpContext context)
...
 				};
 
+				if (context.User?.Identity != null && context.User.Identity.IsAuthenticated && context.User.Identity != null
+					&& transaction.Context.User == null)
+				{
+					transaction.Context.User = new User
+					{
+						UserName = context.User.Identity.Name,
+						Id = GetClaimWithFallbackValue(ClaimTypes.NameIdentifier, "sub"),
+						Email = GetClaimWithFallbackValue(ClaimTypes.Email, "email")
```
SergeyKleyman: The same comment as for `"sub"` above

test/Elastic.Apm.AspNetCore.Tests/CaptureUserTest.cs
```diff
@@ -0,0 +1,137 @@
...
+	/// The ASP.NET Core auto instrumentation is able to capture logged in users.
+	/// This class tests that feature.
+	/// </summary>
+	[Collection("DiagnosticListenerTest")]
+	public class CaptureUserTest : IDisposable
+	{
+		private readonly MockPayloadSender _payloadSender = new MockPayloadSender();
+		private ApmAgent _agent;
+
+		private void SetUpSud()
```
SergeyKleyman: Did you mean `SetUpSut` (SUT - System Under Test)?

gregkalapos: Yes, fixed. Thx!

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
gregkalapos: Yes, those are the OpenID claim types. I moved that to a const type.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
gregkalapos: 👍 , same as above.

src/Elastic.Apm.AspNetCore/Consts.cs
```diff
@@ -0,0 +1,8 @@
+namespace Elastic.Apm.AspNetCore
+{
+	internal static class Consts
+	{
+		internal const string Sub = "sub";
```
SergeyKleyman: I would give it a more informative name. For example
```
internal static class Consts
{
    internal static class OpenIdClaimTypes
    {
        internal const string UserId = "sub";
        internal const string Email = "email";
    }
}
```

gregkalapos: Ok, let's do it like that.

Adapted it.

Issue #166, opened 2019-03-29
Group by Route


bladefist:
We're using the pre-release and seeing that every request is being logged uniquely by it's route parameters, for example:

/user/profile/32
/user/profile/456

Where it would be more interesting to group these by:
/user/profile/{userid}

Is there anyway to do this currently, or is this something that just have to wait for a newer release?

Thanks!

gregkalapos:
Hi @bladefist,

great to see you already use the preview! 

Unfortunately currently there is no way to do this. This is a missing feature. We already have an issue for it: https://github.com/elastic/apm-agent-dotnet/issues/84.

I think that covers exactly what you'd like to have - and we'll do it, we just haven't had a chance to implement it yet. 

Could you please verify that https://github.com/elastic/apm-agent-dotnet/issues/84 covers all what you need?

Thanks!

bladefist:
@gregkalapos thanks for the response.  #84 looks like the same issue, I'll go subscribe to that. Thanks for all the hard work on this!

Issue #84, opened 2019-01-30
Transaction name should not contain request specific data (user ID, name, etc.)


SergeyKleyman:
**Motivation**: Having request specific data (user ID, name, etc.) in transaction name makes metrics aggregation very problematic. For example, let's say we saw two GET requests with URL paths: `/user/123/info` and `/user/456/info` and we would like to calculate average transaction latency - if we just use URL path as the transaction name for each request we will calculate average latency separately for transaction name `GET /user/123/info` and separately for `GET /user/456/info` - it would be much more useful to deduce transaction name without request specific data (in this case without user ID). For example, if the transaction name is `GET /users/{id}/info` then average latency will be calculated using both requests.

PR #182, opened 2019-04-13
ASP.NET Core - create transaction name based on route and group by optional values

gregkalapos:
Addresses #84 

Very much inspired by what [Application insights is doing](https://github.com/Microsoft/ApplicationInsights-aspnetcore/blob/883589c8a99fded1b754c5ed688734274c9243e9/src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/MvcDiagnosticsListener.cs#L41).

Prior to this PR we simply used the URL as the transaction name in ASP.NET Core. As #84 and #166 describe this is not optimal in scenarios where part of the URL is a parameter like `/user/info/42`, because our UI creates new transaction groups for each URL.

This PR uses the ASP.NET routing information to create the transaction name. We resolve the `controller` and `action` names, and also the `page` in case of razor pages, so those will be in the transaction name, everything else is not resolved, instead, in case of optional route parameters the name of the parameter will be captured.

Example:
`/home/sample/1`, where the route map is the default `"{controller=Home}/{action=Index}/{id?}"` becomes `GET home/sample {id}`.

Constants are not part of the transaction name:
```
[Route("api/[controller]")]
[ApiController]
public class ValuesController : ControllerBase
{
	
  // GET api/values/5
  [HttpGet("{id}")]
  public ActionResult<string> Get(int id)
  {
     return "value" + id;
  }
```

In this case `api/values/42` will be `GET Values/Get`, since that is the name of the controller/action. This is ideal for developers, because the URL is captured and shown in Kibana for each transaction and the transaction name reveals the name of the Controller/Action that served the request.

This will be a braking change in the sense that grouping on transaction name will be different after this PR, in fact transaction names in some cases could look very different. I think this is acceptable, since this solves an important issue, and there is no reason to maintain compatibility with the previous version. Plus we are in alpha, so let's use it as an excuse 😇 .

### DIFF
--- a/sample/SampleAspNetCoreApp/Controllers/HomeController.cs
+++ b/sample/SampleAspNetCoreApp/Controllers/HomeController.cs
@@ -113,6 +113,10 @@ namespace SampleAspNetCoreApp.Controllers
 			throw new Exception("This is a test exception!");
 		}
 
+		public IActionResult Sample(int id)
+		{
+			return Ok(id);
+		}
 
 		[ResponseCache(Duration = 0, Location = ResponseCacheLocation.None, NoStore = true)]
 		public IActionResult Error() => View(new ErrorViewModel { RequestId = Activity.Current?.Id ?? HttpContext.TraceIdentifier });
--- a/sample/SampleAspNetCoreApp/Controllers/HomeController.cs
+++ b/sample/SampleAspNetCoreApp/Controllers/HomeController.cs
@@ -113,6 +113,7 @@ namespace SampleAspNetCoreApp.Controllers
 			throw new Exception("This is a test exception!");
 		}
 
+		//Used as test for optional route parameters
 		public IActionResult Sample(int id)
 		{
 			return Ok(id);
--- a/sample/WebApiSample/Controllers/ValuesController.cs
+++ b/sample/WebApiSample/Controllers/ValuesController.cs
@@ -27,5 +27,12 @@ namespace WebApiSample.Controllers
 
 			return new[] { "value1", "value2", lastValue };
 		}
+
+		// GET api/values/5
+		[HttpGet("{id}")]
+		public ActionResult<string> Get(int id)
+		{
+			return "value" + id;
+		}
 	}
 }
--- a/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
+++ b/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
@@ -11,6 +11,7 @@ using Elastic.Apm.Helpers;
 using Elastic.Apm.Logging;
 using Elastic.Apm.Model.Payload;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
 
 [assembly:
 	InternalsVisibleTo(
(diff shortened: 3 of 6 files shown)


SergeyKleyman:
I have two comments regarding resulting name format:

1. Regarding using Controller/Action instead of original URL path with argument values replaced their names. In the example above the current implementation is to give the name `GET Values/Get` instead of `api/values/{id}`. Is this aligned with the other agents? I understand your argument that it might have some advantage for developers but on the other hand it's an implementation detail. For example let's say monitored application got upgraded and controllers and/or actions changed a little bit while URL paths stayed the same. The agent will report different transaction names for the upgraded version which might make it harder to compare/see trend before/after the upgrade. Do you think we should consider this use case?

2. The way parameters names are combined with Controller/Action part of the name (for example `
GET home/sample {id}`) - is this aligned with the other agents? I find it a little bit confusing but as long as this is what other agents do...

gregkalapos:
@SergeyKleyman I talked to other devs.

> In the example above the current implementation gives the name GET Values/Get instead of GET api/values/{id}. Is this aligned with the other agents? 

Yes, other agents do the same... If there is a routing concept (or something like controller/action or view) in the corresponding framework, they use that to create the `Transaction.Name`. The `url` is captured separately, so to me this seems like the best way to go, because `url` is captured anyway.

Regarding scenarios where a controller or action get renamed: I think that's just a consequence of this approach and that would have the same consequences in other agents. Nevertheless, I think users can also query transactions separately, based on `url`. So if this happens they still can aggregate the data as they want, but of course the grouping on the main Transaction page in Kibana will by default use the new grouping. I'd say this can be an issue, but I don't see any way to fix it with the current approach - and for the standard case the current approach seems to be better than using the `url`.

> The way parameters names are combined with Controller/Action part of the name (for example GET home/sample {id}) - is this aligned with the other agents? 

Agents use formats that are native to the platform/framework. 
- Node.js: `/foo/:id`
- Java: `ControllerName#method` 

The approach is that basically every agent should use whatever feels more natural for the given language/framework/library.

>  I find it a little bit confusing and URL path approach (GET api/values/{id}) has another advantage of integrating parameters names more naturally. 

That's a good point, but I'm not sure it's feasable to do that. The order of the parameters is not defined (at least that's what I experienced - it's stored in [`RouteValueDictionary`](https://docs.microsoft.com/en-us/dotnet/api/system.web.routing.routevaluedictionary?view=netframework-4.7.2) - that's a `IDictionary<String,Object>`, the `Keys` are the route parameters, but those are not ordered), and if the mapping is something like: `{controller=Home}/{id?}/{action=Index}` then I'm not sure we can reconstruct this format when we iterate through the route parameters. Another thing I also noticed is that attribute routing also changes the order of the parameters. The only way to do this I see, is that we get all the matched routings (can be both the route mapping with the string, plus all the matching attributes and combine that with the parameters we get. But this sounds like lots of work, very hard to test, and potential overhead for a very little improvement.  If someone has an idea to solve this I'm open for it; otherwise I'd just leave it as it is.

gregkalapos:
> The only change to the current implementation I would propose is to use Application Insights' format for parameters. Judging by examples at http://apmtips.com/blog/2015/02/23/request-name-and-url/ and [implementation](https://github.com/Microsoft/ApplicationInsights-aspnetcore/blob/883589c8a99fded1b754c5ed688734274c9243e9/src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/MvcDiagnosticsListener.cs#L41) Application Insights' format is `VERB controller/action [parameter1/parameter2/...]` so instead of `GET home/sample {id}` it would be GET home/sample [id]`

Yeah, I saw that in the Application Insights code. The reason I changed `[]` to `{}` is because that is how you define it in the route mapping. I thought it's better to follow what ASP.NET Core does and not what Application Insights does :)

SergeyKleyman:
But AFAICS if there is more than one argument the current implementation will produce `GET home/sample {id1/id2}` while I assume you intended to have `GET home/sample {id1}/{id2}`, am I correct?

gregkalapos:
> But AFAICS if there is more than one argument the current implementation will produce `GET home/sample {id1/id2}` while I assume you intended to have `GET home/sample {id1}/{id2}`, am I correct?

Yes, it will create `GET home/sample {id1/id2}` and that was also my intention...  Btw. none of those are really correct, because the order can be different - but I think that's fine.

SergeyKleyman:
If that is intentional then it's okay.

PR #203, opened 2019-04-25
Unit test helper to write Agent log to Xunit output

SergeyKleyman:


SergeyKleyman:
Obsolete - the same feature was implemented as part #273.

PR #218, opened 2019-05-02
[ci] add files for JJBB jobs

jonahbull:
Add JJBB job definitions to the repo so they can be managed here.

`check_paths_for_matches` is a utilty script that lets us work around
some undesirable behavior by the `git` plugin, which requires an
existing workspace to check the `included/excluded-regions` directive.

We'll configure a multijob that runs this script on every push to master
for the repo and when the script detects a match in the changed paths
(default is to look for paths that begin with `^.ci/(jobs|views)`),
it'll launch the JJBB job to update job definitions. Otherwise the JJBB
job would get triggered every push to master, which is undesirable for a
variety of reasons.


gregkalapos:
Added @v1v and @kuisathaverat. I think they are the best to review this. 

@jonahbull excuse my ignorance, but could you please shortly describe what this all means on a more higher level? (...so that .NET devs like me also understand this..) What is this JJBB and why is it better than the current system?

.ci/jobs/defaults.yml
```diff
@@ -0,0 +1,54 @@
+---
+
+##### GLOBAL METADATA
+
+- meta:
+    cluster: apm-ci
+
+##### JOB DEFAULTS
+
+- job:
```
kuisathaverat: @jonahbull I miss a reference to the location of the Jenkinsfile

`script-path: 'Jenkinsfile'`

jonahbull:
> @jonahbull excuse my ignorance, but could you please shortly describe what this all means on a more higher level? (...so that .NET devs like me also understand this..) What is this JJBB and why is it better than the current system?

Fair question, @gregkalapos I should have included a little more information! JJBB is a wrapper around JJB which aims to simplify CI job configuration. The main benefit of JJBB is that it moves the job configuration files out of the Infra repo and into each team's repo(s). So there's no more need to open PRs against the Infra repo when a job change needs to be made. Another benefit is that each job corresponds to a single file rather than being spread around multiple files with templates, macros, etc. You can read more about the project in the original [issue](https://github.com/elastic/infra/issues/5293) and in the [README](https://github.com/elastic/infra/blob/master/flavortown/jjbb/README.md) Hope this information is helpful, let me know if you have more questions!

.ci/jobs/defaults.yml
jonahbull: Good catch, @kuisathaverat I'll push a fix shortly.

jonahbull: Added that item and a few others that got missed when flattening the jobs originally and corrected the flattening script to warn us when there are items that were present originally missing in the flattened job.

gregkalapos:
Thanks for the explanation @jonahbull! 👍 👍

Issue #215, opened 2019-04-30
Failing concurrent integration test


gregkalapos:
Sending requests in parallel to the agent fails - some of them aren't captured.

``` AssertionError: queried for [('processor.event', 'transaction'), ('service.name', ['dotnetapp'])], expected 1000, got 832```

Logs from the agent: 
[agentlogs_failingconcurrentTest.txt](https://github.com/elastic/apm-agent-dotnet/files/3132407/agentlogs_failingconcurrentTest.txt)

**Comment by gregkalapos** [comment]
I suspect this is a bug in `PaylodSender`. 

Number of `[Debug] - {Tracer} Starting Transaction` logs: 1000. (also 100 for ending `Ending Transaction`)
Number of `{PayloadSenderV2} Sent items to server` logs: 768. (some of those batches send more than 1 transaction! )

Now, the way this works is that every `Transaction.End` forces the `PayloadSender` to send data (spans are batched until 20 items). 

But I suspect that due to the parallelism some Transactions remain in the queue. I planned to add some timeout and send those... maybe that's the way to go.

**Comment by gregkalapos** [comment]
The problem is that we can't queue the Transansactions:

```
public void QueueTransaction(ITransaction transaction)
{
   _eventQueue.Post(transaction)) //<-- this returns false and the `transaction` won't be in the `BatchBlock`, so we don't send it to the server 
   _eventQueue.TriggerBatch();
}
```

gregkalapos:
Fixed in https://github.com/elastic/apm-agent-dotnet/pull/219

PR #219, opened 2019-05-02
Make integration tests work

gregkalapos:
Solves  #215.
Triggered by https://github.com/elastic/apm-integration-testing/pull/424 

There were multiple problems:
- The concurrent tests failed, because the number of captured transactions was less than the number of requests sent to the sample app. The fix was to remove the upper bound on the number of messages on the `BatchBlock<object> _eventQueue`.
- The tests used the `url.path` field, which was not captured by the agent, this PR adds this field and the ASP.NET Core sensor also fills it.
- The tests used the `url.search` field (the query string), which was not captured by the agent, this PR adds this field and the ASP.NET Core sensor also fills it.

### DIFF
--- a/sample/SampleAspNetCoreApp/appsettings.json
+++ b/sample/SampleAspNetCoreApp/appsettings.json
@@ -6,7 +6,7 @@
   },
   "AllowedHosts": "*",
   "ElasticApm": {
-    "LogLevel": "Error",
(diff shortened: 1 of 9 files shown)


gregkalapos:
`PayloadSentWithBearerToken` got re-enabled in #220.

Issue #239, opened 2019-05-27
Use proper fileName function for stacktrace


skynet2:
Hi,
Currently, when error occur we can see a stack trace like that : 
![image](https://user-images.githubusercontent.com/3065048/58433618-b6c5c000-80c0-11e9-84d8-9acd7957469a.png)

The problem here that value which we capture is not a real file name, it`s more like a namespace where that file exists. 
Another problem is that the function name is invalid (MoveNext) it`s because the function is async.

My suggestion is to capture real fileName and functionName. I made some test and here is a working example

I made some research and here are my results : 
![image](https://user-images.githubusercontent.com/3065048/58434645-7288ee80-80c5-11e9-95cf-5a6f34fbb97c.png)

I will create a pull request with those changes for the review.

gregkalapos:
Addressed and discussed in #240.

PR #240, opened 2019-05-27
feat: real stack #239

skynet2:


src/Elastic.Apm/Helpers/StacktraceHelper.cs
```diff
@@ -23,11 +27,11 @@ internal static List<CapturedStackFrame> GenerateApmStackTrace(StackFrame[] fram
 			try
 			{
 				retVal.AddRange(from item in frames
-					let fileName = item?.GetMethod()?.DeclaringType?.Assembly?.GetName()?.Name
+					let fileName = item?.GetFileName()
```
gregkalapos: The problem with this change is that this is the value that we show on the UI, and as a result, we show things like this:
image
...and this can be very long and very hard to read. 

Also, the field names are here somewhat misleading, because the server side was kind of designed for JavaScript. So I know that the original code looked confusing, still I think it's better to show the assembly name on the UI and not the filename, so I still think we should "misuse" the filename field and store the assembly name there.

skynet2: Hi @gregkalapos ,

Thanks for the review, but I think having here assembly name is not the best idea, because it can be confusing.
===
Let`s pretend that we have a library with Name "Services" and we have class BaseProductService with virtual or abstracts methods, and we have a set of classes which inherits BaseProductService and implements method GetProduct. 
In the scenario where we capture assembly name, in stack trace, we`ll see that an error happened in assembly "Services", in method "GetProduct" and line 123. My concern is that without file name we don`t know which implementation we should check.

Thanks,
Stas

gregkalapos: We also capture the reflected type name:
`Module = item?.GetMethod()?.ReflectedType?.Name,`
You can always jump to discover and I believe with that you can unequally identify what type the method belongs to.

What I'd prefer to show on the UI would be this:
[Fullnamespace].[Classname] in [MethodName] at line [linenumber] 

I think showing the filename and the method name is not natural in .NET. It's ok in JavaScript, but for .NET devs I think it'd look very strange. 

Again, the main problem here is that it's very hard to map the C# info to the schema what the server expects. 

So, new proposal:
- `FileName`:  Make it capture `[Fullnamespace].[Classname]` (Again, I know that the name of the field is totally confusing, but in the .NET World, tools usually show fully qualified classnames and not filenames, and this is what we show on the UI)
- `Module`: There is a [module concept](https://stackoverflow.com/questions/9271805/net-module-vs-assembly) in .NET, but I think it'd be better just store the assembly here. As an alternative we could maybe discuss to store the filename here. My only concern with this it that will already be total chaos :) 

So, what about the proposal?

- `FileName` -> Make it capture `[Fullnamespace].[Classname]` 
- `Module` -> captures the assembly

With that you should be able to identify everything.

gregkalapos: This is how it would look like; I think this look very ".NET Friendly": 

image

skynet2: Sounds good to me, thanks @gregkalapos .

If you have that code ready, can you please commit?

Thanks,
Stas

gregkalapos: 👍 on it.

gregkalapos: Done, I also added a test covering this. Let me know what you think about the code. I'll also go through it once again before I merge it.

skynet2: Looks good to me, thanks @gregkalapos

SergeyKleyman: @gregkalapos Please note that there are integrations planned between APM and parts of Elastic stack used to analyze source code. Unfortunately I don't remember the name of the relevant product in Elastic stack. For example one of the planned integrations is to allow jump from method references in APM UI to source code. So it's possible that file name and line number will be used for that integration. I assume having class name where file name is expected might present a problem. @axw Do you think the planned integrations will use information sent by agent as part of error stack trace?

gregkalapos: Code search? :)
I also think that when we integrate with that there will be challenges - it would be nice to know what we exactly need for that (what fields, what are the fallbacks if we don't have those, etc.)

I think we should show stacktraces in a way that it is natural for .NET devs (e.g. what we also see in Visual Studio), so fully qualified classname - method, but I think we don't even have a field for the class name in the [current `stacktrace_frame.json`](https://github.com/elastic/apm-server/blob/master/docs/spec/stacktrace_frame.json). Maybe we should gather all the requirements from all the supported agent languages and rethink `stacktrace_frame.json` a little bit based on that. Another thing that could help would be to somehow dynamically select what fields we use on the UI - e.g. filename is I think totally fine in JavaScript, but in Object Oriented languages class name is definitely a better fit.

SergeyKleyman: Let's see if we can find more from inside Elastic and we will be in better position to make a decision on what to do in short term. One approach I thought about is to leave file name in `filename` but set method name to <full class name>.<method name> until we have dedicated fields for every interesting property.
Regarding the issue that the way it's customary to display stack frames varies between languages - maybe agent for each language will also attach some template so that UI knows how to display stack frames. We can create elastic/apm issue to get other agents devs feedback.

SergeyKleyman: I've created #265 to follow up on integration with Code search. Let's assume that it's not relevant as far as this PR is concerned.

PR #246, opened 2019-05-31
[APM-CI][.NET] GH Checks for the PR

v1v:
## Highlights
- Create GH checks with the most important stages in the CI Pipeline.
- This particular change might help to debug when there are errors in the CI and identify them easily within then PR itself.

## Example

image


Jenkinsfile
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env groovy
 
-@Library('apm@current') _
+@Library('apm@feature/overrideStage') _
```
v1v: TODO: this particular change should be removed before merging this PR

Jenkinsfile
```diff
@@ -130,7 +134,9 @@ pipeline {
                       cleanDir("${WORKSPACE}/${BASE_DIR}")
```
kuisathaverat: if something happened here the error would not be marked inside the correct stage, isn't it?

v1v: You are right, I'll ensure it is handled correctly

Issue #228, opened 2019-05-11
Integration with .Net Core logging infrastructure


vhatsura:
**Is your feature request related to a problem? Please describe.**
Current logging implementation inside Elastic.Apm for ASP.Net Core has some limitations.
1. Limited support of logging providers. At the moment, only Console is supported.
2. Unable to modify format of log messages which causes some issues in case of hosting in k8s. in k8s we're writing log messages directly into console in json format. Then such messages are fetched by filebeats and shipped to logstash via kafka. In case of enabling apm, our filebeats generate a huge amount of errors during parsing messages, which are produced by `ConsoleLogger`. It doesn't broke out logging infrastructure, however, it would be nice to have logs from apm into our logging storage.

**Describe the solution you'd like**
It would be nice for `Elastic.Apm.AspNetCore` project to add support of .Net Core logging infrastructure (Microsoft.Extensions.Logging). For compatibility with `IApmLogger` it can be written adapter which accepts `ILogger` as dependency and implement `IApmLogger` interface. In such case will be added eability to support various logging providers ([built-in providers](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-2.2#built-in-logging-providers)) with richer configuration capabilities.

**Describe alternatives you've considered**
1. We can implement custom implementation of `IApmLogger` interface, however we cannot override usage of default console logger
https://github.com/elastic/apm-agent-dotnet/blob/2447ecb7212d3659dcae91bf9d3e39370040f59b/src/Elastic.Apm.AspNetCore/ApmMiddlewareExtension.cs#L32-L41
2. `LogLevel` can be set to `None` to not produce log messages from APM, however, we lost all messages from apm agent

gregkalapos:
Thanks for opening this @vhatsura, we definitely plan to do more work in this area; `ConsoleLogger` won't be the only logger long term.

vhatsura:
@gregkalapos, thanks for reply.
Do you have in plans to have integration with `Microsoft.Extensions.Logging` in `Elastic.Apm.AspNetCore`?

gregkalapos:
@vhatsura Yes. 

I don't have a final action plan, but what I can think of is that we implement this in `Elastic.Apm.AspNetCore` in a way that you can pass the logger to it and from that point the agent will log through it as any other component (like ASP.NET Core, or EFCore).

vhatsura:
@gregkalapos, what do you mind if I try to implement this?

[...]

PR #249, opened 2019-06-01
use logger created from logger factory instead of console logger

vhatsura:
related to #228.

What was done:
1. Use logger created from `ILoggerFactory` inside `Elastic.Apm` infrastructure.
2. Additional dependency to `Microsoft.Extensions.Logging.Abstractions`

Open questions:
1. `IApmLogger` has `Level` property in interface definition. In case of asp.net core philosophy, configuration per logger category is splitted from `ILogger` functionality.
2. `Elastic.Apm.AspNetCore` project doesn't have extensions for `IServiceCollection`. Instead of directly creation of some instances, there are can be configured through dependency injection. It's allow to simplify configuration in the future and replace some parts of `Elastic.Apm` by users.

@gregkalapos, what do you think about my changes? Is it aligned with your thoughts?

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -36,7 +36,34 @@ With this you can use any configuration source that you configured on the `IConf
 
 NOTE: By simply calling `app.UseElasticApm()` without the overload, the agent will read configurations only from environment variables.
 
-This is a typical `appsettings.json` file that contains some sample configuration. The part below `ElasticApm` is fetched by the agent if the corresponding `IConfiguration` is passed to the agent.
+[float]
+[[sample-config]]
+==== Sample configuration file
+
+Below is a sample `appsettings.json` configuration file for a typical ASP.NET Core application. There are two important takeaways:
+1. The part below `ElasticApm` is fetched by the agent if the corresponding `IConfiguration` is passed to the agent.
+2. With ASP.NET Core, you must set `LogLevel for the internal APM logger in the standard `Logging` section with the `ElasticApm` category name.
+
+[source,js]
+----
+{
+  "Logging": {
+    "LogLevel": {
+      "Default": "Warning",
+      "Elastic.Apm": "Debug"
+    }
+  },
+  "AllowedHosts": "*",
+  "ElasticApm":
+    {
+      "ServerUrls":  "http://myapmserver:8200",
+      "TransactionSampleRate": 1.0
+    }
+}
+----
+
+In certain scenarios -- like when you're not using ASP.NET Core -- you wont activate the agent with the `UseElasticApm()` method.
+In this case, you can set the log level of the agent with `ElasticApm:LogLevel`, as shown in the following `appsettings.json` file:
 
 [source,js]
 ----
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -41,6 +41,7 @@ NOTE: By simply calling `app.UseElasticApm()` without the overload, the agent wi
 ==== Sample configuration file
 
 Below is a sample `appsettings.json` configuration file for a typical ASP.NET Core application. There are two important takeaways:
+
 1. The part below `ElasticApm` is fetched by the agent if the corresponding `IConfiguration` is passed to the agent.
 2. With ASP.NET Core, you must set `LogLevel for the internal APM logger in the standard `Logging` section with the `ElasticApm` category name.
 
--- a/sample/SampleAspNetCoreApp/appsettings.json
+++ b/sample/SampleAspNetCoreApp/appsettings.json
@@ -1,12 +1,12 @@
 {
   "Logging": {
     "LogLevel": {
-      "Default": "Warning"
+      "Default": "Warning",
+      "Elastic.Apm": "Error"
     }
   },
   "AllowedHosts": "*",
   "ElasticApm": {
-    "LogLevel": "Error",
     "ServerUrls": "http://localhost:8200",
     "TransactionSampleRate": 1.0
   }
--- a/src/Elastic.Apm.AspNetCore/ApmMiddlewareExtension.cs
+++ b/src/Elastic.Apm.AspNetCore/ApmMiddlewareExtension.cs
@@ -10,6 +10,7 @@ using Elastic.Apm.DiagnosticSource;
 using Elastic.Apm.Logging;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
 
 namespace Elastic.Apm.AspNetCore
 {
@@ -42,12 +43,13 @@ namespace Elastic.Apm.AspNetCore
 			params IDiagnosticsSubscriber[] subscribers
 		)
 		{
-			var logger = ConsoleLogger.Instance;
+			var logger = builder.ApplicationServices.GetApmLogger();
+
 			var configReader = configuration == null
 				? new EnvironmentConfigurationReader(logger)
 				: new MicrosoftExtensionsConfig(configuration, logger) as IConfigurationReader;
 
-			var config = new AgentComponents(configurationReader: configReader);
+			var config = new AgentComponents(configurationReader: configReader, logger: logger);
 			UpdateServiceInformation(config.Service);
 
 			Agent.Setup(config);
@@ -68,6 +70,11 @@ namespace Elastic.Apm.AspNetCore
 			agent.Subscribe(subs.ToArray());
 			return builder.UseMiddleware<ApmMiddleware>(agent.Tracer, agent);
 		}
+
+		internal static IApmLogger GetApmLogger(this IServiceProvider serviceProvider) =>
+			serviceProvider.GetService(typeof(ILoggerFactory)) is ILoggerFactory loggerFactory
+				? (IApmLogger)new AspNetCoreLogger(loggerFactory)
+				: ConsoleLogger.Instance;
 
 		internal static void UpdateServiceInformation(Service service)
 		{
--- a/src/Elastic.Apm.AspNetCore/AspNetCoreLogger.cs
+++ b/src/Elastic.Apm.AspNetCore/AspNetCoreLogger.cs
@@ -1,0 +1,39 @@
+using System;
+using Elastic.Apm.AspNetCore.Extensions;
+using Elastic.Apm.Logging;
+using Microsoft.Extensions.Logging;
+using LogLevel = Elastic.Apm.Logging.LogLevel;
+
+namespace Elastic.Apm.AspNetCore
+{
+	internal class AspNetCoreLogger : IApmLogger
+	{
+		private readonly ILogger _logger;
+
+		public AspNetCoreLogger(ILoggerFactory loggerFactory)
+		{
+			_logger = loggerFactory?.CreateLogger("Elastic.Apm") ?? throw new ArgumentNullException(nameof(loggerFactory));
+			Level = _logger.GetMinLogLevel();
+		}
+
+		public LogLevel Level { get; }
+
+		public void Log<TState>(LogLevel level, TState state, Exception e, Func<TState, Exception, string> formatter) =>
+			_logger.Log(Convert(level), new EventId(), state, e, formatter);
+
+		private static Microsoft.Extensions.Logging.LogLevel Convert(LogLevel logLevel)
+		{
+			switch (logLevel)
+			{
+				case LogLevel.Trace: return Microsoft.Extensions.Logging.LogLevel.Trace;
+				case LogLevel.Debug: return Microsoft.Extensions.Logging.LogLevel.Debug;
+				case LogLevel.Information: return Microsoft.Extensions.Logging.LogLevel.Information;
+				case LogLevel.Warning: return Microsoft.Extensions.Logging.LogLevel.Warning;
+				case LogLevel.Error: return Microsoft.Extensions.Logging.LogLevel.Error;
+				case LogLevel.Critical: return Microsoft.Extensions.Logging.LogLevel.Critical;
+				case LogLevel.None:
+				default: return Microsoft.Extensions.Logging.LogLevel.None;
+			}
+		}
+	}
+}
--- a/src/Elastic.Apm.AspNetCore/Extensions/LoggerExtensions.cs
+++ b/src/Elastic.Apm.AspNetCore/Extensions/LoggerExtensions.cs
@@ -1,0 +1,26 @@
+using System;
+using Microsoft.Extensions.Logging;
+using LogLevel = Elastic.Apm.Logging.LogLevel;
+
+namespace Elastic.Apm.AspNetCore.Extensions
+{
+	internal static class LoggerExtensions
+	{
+		internal static LogLevel GetMinLogLevel(this ILogger logger)
+		{
+			if (logger == null) throw new ArgumentNullException(nameof(logger));
+
+			if (logger.IsEnabled(Microsoft.Extensions.Logging.LogLevel.Trace)) return LogLevel.Trace;
+
+			if (logger.IsEnabled(Microsoft.Extensions.Logging.LogLevel.Debug)) return LogLevel.Debug;
+
+			if (logger.IsEnabled(Microsoft.Extensions.Logging.LogLevel.Information)) return LogLevel.Information;
+
+			if (logger.IsEnabled(Microsoft.Extensions.Logging.LogLevel.Warning)) return LogLevel.Warning;
+
+			if (logger.IsEnabled(Microsoft.Extensions.Logging.LogLevel.Error)) return LogLevel.Error;
+
+			return logger.IsEnabled(Microsoft.Extensions.Logging.LogLevel.Critical) ? LogLevel.Critical : LogLevel.None;
+		}
+	}
+}
--- a/test/Elastic.Apm.AspNetCore.Tests/AspNetCoreLoggerTests.cs
+++ b/test/Elastic.Apm.AspNetCore.Tests/AspNetCoreLoggerTests.cs
@@ -1,0 +1,48 @@
+using System;
+using Microsoft.Extensions.Logging;
+using Moq;
+using Xunit;
+
+namespace Elastic.Apm.AspNetCore.Tests
+{
+	/// <summary>
+	/// Tests the <see cref="Elastic.Apm.AspNetCore.AspNetCoreLogger"/> type.
+	/// </summary>
+	public class AspNetCoreLoggerTests
+	{
+		[Fact]
+		public void AspNetCoreLoggerShouldThrowExceptionWhenLoggerFactoryIsNull()
+			=> Assert.Throws<ArgumentNullException>(() => new AspNetCoreLogger(null));
+
+		[Fact]
+		public void AspNetCoreLoggerShouldGetLoggerFromFactoryWithProperCategoryName()
+		{
+			var loggerFactoryMock = new Mock<ILoggerFactory>();
+			loggerFactoryMock.Setup(x => x.CreateLogger(It.IsAny<string>()))
+				.Returns(() => Mock.Of<ILogger>());
+
+			// ReSharper disable UnusedVariable
+			var logger = new AspNetCoreLogger(loggerFactoryMock.Object);
+			// ReSharper restore UnusedVariable
+
+			loggerFactoryMock.Verify(x => x.CreateLogger(It.Is<string>(s => s.Equals("Elastic.Apm"))), Times.Once);
+		}
+
+		[Fact]
+		public void AspNetCoreLoggerShouldCalculateMinLogLevelOnCreation()
+		{
+			var loggerFactoryMock = new Mock<ILoggerFactory>();
+			var loggerMock = new Mock<ILogger>();
+			loggerFactoryMock.Setup(x => x.CreateLogger(It.IsAny<string>()))
+				.Returns(() => loggerMock.Object);
+
+			var logger = new AspNetCoreLogger(loggerFactoryMock.Object);
+
+			loggerMock.Verify(x => x.IsEnabled(It.Is<LogLevel>(l => l == LogLevel.Trace)), Times.Once);
+			// ReSharper disable UnusedVariable
+			var level = logger.Level;
+			// ReSharper restore UnusedVariable
+			loggerMock.Verify(x => x.IsEnabled(It.Is<LogLevel>(l => l == LogLevel.Trace)), Times.Once);
+		}
+	}
+}
--- a/sample/WebApiSample/appsettings.json
+++ b/sample/WebApiSample/appsettings.json
@@ -1,12 +1,12 @@
 {
   "Logging": {
     "LogLevel": {
-      "Default": "Warning"
+      "Default": "Warning",
+      "Elastic.Apm": "Error"
     }
   },
   "AllowedHosts": "*",
   "ElasticApm": {
-    "LogLevel": "Error",
     "ServerUrls": "http://localhost:8200"
   }
 }
--- a/src/Elastic.Apm.AspNetCore/Elastic.Apm.AspNetCore.csproj
+++ b/src/Elastic.Apm.AspNetCore/Elastic.Apm.AspNetCore.csproj
@@ -29,6 +29,7 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="2.0.0" />
     <PackageReference Include="Microsoft.AspNetCore.Routing.Abstractions" Version="2.0.0" />
     <PackageReference Include="Microsoft.AspNetCore.Http.Extensions" Version="2.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.0.0" />
   </ItemGroup>
   <ItemGroup>
     <Folder Include="Config\" />
--- a/test/Elastic.Apm.AspNetCore.Tests/ApmMiddlewareExtensionTest.cs
+++ b/test/Elastic.Apm.AspNetCore.Tests/ApmMiddlewareExtensionTest.cs
@@ -1,0 +1,33 @@
+using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+
+namespace Elastic.Apm.AspNetCore.Tests
+{
+	/// <summary>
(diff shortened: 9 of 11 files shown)


src/Elastic.Apm.AspNetCore/ApmMiddlewareExtension.cs
```diff
@@ -35,7 +36,9 @@ public static class ApmMiddlewareExtension
 			params IDiagnosticsSubscriber[] subscribers
 		)
 		{
-			var logger = ConsoleLogger.Instance;
+			var loggerFactory = builder.ApplicationServices.GetService(typeof(ILoggerFactory)) as ILoggerFactory;
+
+			var logger = new AspNetCoreLogger(loggerFactory);
```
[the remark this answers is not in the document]
vhatsura: fixed

src/Elastic.Apm.AspNetCore/ApmMiddlewareExtension.cs
```diff
@@ -35,7 +36,9 @@ public static class ApmMiddlewareExtension
 			params IDiagnosticsSubscriber[] subscribers
 		)
 		{
-			var logger = ConsoleLogger.Instance;
+			var loggerFactory = builder.ApplicationServices.GetService(typeof(ILoggerFactory)) as ILoggerFactory;
```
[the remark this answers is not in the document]
vhatsura: fixed

src/Elastic.Apm.AspNetCore/AspNetCoreLogger.cs
```diff
@@ -0,0 +1,39 @@
...
+	public class AspNetCoreLogger : IApmLogger
+	{
+		public AspNetCoreLogger(ILoggerFactory loggerFactory)
+		{
+			_logger = loggerFactory?.CreateLogger("apm") ?? throw new ArgumentNullException(nameof(loggerFactory));
+		}
+
+		private readonly ILogger _logger;
+
+		public LogLevel Level { get; } = LogLevel.Error;
```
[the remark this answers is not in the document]
vhatsura: Agreed with you. Error was a simple solution for draft pull request. Also, I agreed with the 2nd option (use log level from standard asp.net core config section). However, fetch logger configuration isn't a simple task in asp.net core due to it can be configured in various way. By this reson, I iterate over `IsEnabled` method to find out min level which is enabled for logger.

vhatsura:
@gregkalapos, thank you for review, all coments are relevant and was fixed. Could you look again, please? If all is ok, I will start to cover changes by unit tests

src/Elastic.Apm.AspNetCore/AspNetCoreLogger.cs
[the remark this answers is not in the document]
gregkalapos: > However, fetch logger configuration isn't a simple task 

Yes, I also tried to come up with something, but that is kind of not part of the API. One thing comes to my mind which may be a problem: Is the loglevel changeable on the fly while the application is running? Atm the code caches the value in the `AspNetCoreLogger` construction, and I wonder if this can change. 
On the other hand reading it always, when the `LogLevel` property is accessed sounds dangerous in terms of overhead (if we go there, we can just measure this).

But overall I think we are good to processed here, this should not block you, we may rethink this part, but it's ok for now.

gregkalapos:
Thank you @vhatsura, 
I think it's ok to proceed here, feel free to continue and also feel free to ping me if you have other questions. 

>  Also, I will appreciate if you specify which docs need to updated with configuration info for asp.net core apm agent

[This file](https://github.com/elastic/apm-agent-dotnet/blob/master/docs/setup.asciidoc) is I think a good place to describe this. Feel free to simply add it to the end.

vhatsura:
Hi, @gregkalapos. I almost done with unit tests (tests AspNetCoreLogger are left), however, have a simple question: what do you think about using [`Moq`](https://github.com/moq/moq4) library?

src/Elastic.Apm.AspNetCore/AspNetCoreLogger.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace Elastic.Apm.AspNetCore
+{
+	internal class AspNetCoreLogger : IApmLogger
+	{
+		private readonly ILogger _logger;
+
+		public AspNetCoreLogger(ILoggerFactory loggerFactory)
+		{
+			_logger = loggerFactory?.CreateLogger("Apm") ?? throw new ArgumentNullException(nameof(loggerFactory));
```
gregkalapos: Could we maybe make it `ElasticApm` instead of `Apm`? That's be the same as the category we use in `appsettings.json`, and also we'd not take up the word that is used industry wide for ourself.

docs/configuration.asciidoc
```diff
@@ -36,20 +36,22 @@ With this you can use any configuration source that you configured on the `IConf
 
 NOTE: By simply calling `app.UseElasticApm()` without the overload, the agent will read configurations only from environment variables.
 
+NOTE: For better user experience, `LogLevel` for internal apm logger is used from standard `Logging` section inside `appsettings.json` with `Apm` category name.
```
gregkalapos: These changes only apply to ASP.NET Core (since we activate the new logger in the `ApmMiddlewareExtension`), so I'd try to make it clear in the docs.

I propose to add this hint to the beginning: `For better user experience in case of ASP.NET Core, `LogLevel` for internal apm …`

docs/configuration.asciidoc
```diff
@@ -36,20 +36,22 @@ With this you can use any configuration source that you configured on the `IConf
...
+
 This is a typical `appsettings.json` file that contains some sample configuration. The part below `ElasticApm` is fetched by the agent if the corresponding `IConfiguration` is passed to the agent.
```
gregkalapos: Similarly, here I'd make clear that this is an `appsettings.json` sample for ASP.NET Core.

docs/configuration.asciidoc
```diff
@@ -36,20 +36,22 @@ With this you can use any configuration source that you configured on the `IConf
...
     "LogLevel": {
-      "Default": "Warning"
+      "Default": "Warning",
+      "Apm": "Debug"
     }
   },
   "AllowedHosts": "*",
   "ElasticApm":
     {
-      "LogLevel":  "Debug",
```
gregkalapos: Additionally I'd add another block below this one to explain how to use `ElasticApm:LogLevel` when ASP.NET Core is not involved. 

Something like this:

In case you don't activate the agent with the `UseElasticApm()` method (typically non-ASP.NET Core scenario) then you can set the loglevel of the agent with the `ElasticApm:LogLevel` setting, as shown in the following `appsettings.json` file:

[and then use the original file]

(_feel free to adapt or improve the text if you feel something is not ok_)

vhatsura: added

docs/configuration.asciidoc
vhatsura: added

src/Elastic.Apm.AspNetCore/AspNetCoreLogger.cs
vhatsura: changed

docs/configuration.asciidoc
vhatsura: added

vhatsura:
Hey, @gregkalapos !
Thanks for review. I updated `configuration` document in documentation and changed category name for logger. Also, I wrote unit tests for `AspNetCoreLogger` with `Moq`. It is what I needed from Moq. If you decide that it would be better to have such unit tests with own mocks, then I will update such tests, but IMHO, with `Moq` such tests is more elegant

docs/configuration.asciidoc
```diff
@@ -36,7 +36,29 @@ With this you can use any configuration source that you configured on the `IConf
 
 NOTE: By simply calling `app.UseElasticApm()` without the overload, the agent will read configurations only from environment variables.
 
-This is a typical `appsettings.json` file that contains some sample configuration. The part below `ElasticApm` is fetched by the agent if the corresponding `IConfiguration` is passed to the agent.
+NOTE: For better user experience in case of ASP.NET Core, `LogLevel` for internal apm logger is used from standard `Logging` section inside `appsettings.json` with `Apm` category name.
```
bmorelli25: I don't love the idea of having two _NOTEs_ in a row:
Screen Shot 2019-07-01 at 2 17 11 PM

Does this new information need to be in a note? Also, it could be lack of .NET knowledge, but I'm not exactly sure what this is trying to say? @gregkalapos do you?

docs/configuration.asciidoc
```diff
@@ -36,7 +36,29 @@ With this you can use any configuration source that you configured on the `IConf
...
+
+This is a typical `appsettings.json` file for ASP.NET Core application that contains some sample configuration. The part below `ElasticApm` is fetched by the agent if the corresponding `IConfiguration` is passed to the agent.
```
```suggestion
Below is a sample `appsettings.json` configuration file for a typical ASP.NET Core application. The part below `ElasticApm` is fetched by the agent if the corresponding `IConfiguration` is passed to the agent.
```

docs/configuration.asciidoc
```diff
@@ -36,7 +36,29 @@ With this you can use any configuration source that you configured on the `IConf
...
+  "AllowedHosts": "*",
+  "ElasticApm":
+    {
+      "ServerUrls":  "http://myapmserver:8200",
+      "TransactionSampleRate": 1.0
+    }
+}
+----
+
+In case you don't activate the agent with the `UseElasticApm()`` method (typically non-ASP.NET Core scenario) then you can set the log level of the agent with the `ElasticApm:LogLevel` setting, as shown in the following `appsettings.json` file:
```
```suggestion
In certain scenarios -- typically when you're not using ASP.NET Core -- you may choose not to active the agent with the `UseElasticApm()` method.
In this case, you can set the log level of the agent with `ElasticApm:LogLevel`, as shown in the following `appsettings.json` file:
```

docs/configuration.asciidoc
gregkalapos: Yes, I do. This changes how to set the loglevel for the agent. In case of ASP.NET Core, here is how to do it:
```
{
	  "Logging": {
	    "LogLevel": {
	      "Default": "Warning",
	      "ElasticApm": "Debug"
	    }
	  },
	  "AllowedHosts": "*",
	  "ElasticApm":
	    {
	      "ServerUrls":  "http://myapmserver:8200",
	      "TransactionSampleRate": 1.0
	    }
	}
```

The `Logging` part is standard and also used by the ASP.NET Core logger - so we (the APM Agent) log to the same logger.

Now, if you don't use ASP.NET Core (e.g. you have a command line application) we can't register our logger with the default logger (since there isn't really one), so in that case here is how you set the log level (it's moved to the `ElasticApm` section:

```
{
	  "Logging": {
	    "LogLevel": {
	      "Default": "Warning",
	    }
	  },
	  "AllowedHosts": "*",
	  "ElasticApm":
	    {
	      "ServerUrls":  "http://myapmserver:8200",
	      "TransactionSampleRate": 1.0,
              "LogLevel" : "Debug"
	    }
	}
```

Probably there is a better way to phrase this; happy to take suggestions :)

gregkalapos: Btw. now that I see the text again, we should change 
> `appsettings.json` with `Apm` category name 

 to 

> `appsettings.json` with `ElasticApm` category name` 

We changed that after one of my comments.

bmorelli25: This definitely helps, thanks Greg! I'm wondering if we should put this under a "sample config" section... here's what I'm thinking:

```asciidoc
[float]
[[sample-config]]
==== Sample configuration file

Below is a sample `appsettings.json` configuration file for a typical ASP.NET Core application. There are two important takeaways:
1. The part below `ElasticApm` is fetched by the agent if the corresponding `IConfiguration` is passed to the agent.
2. With ASP.NET Core, we recommend setting `LogLevel for the internal APM logger in the standard `Logging` section with the `ElasticApm` category name.

[source,js]
----
{
  "Logging": {
    "LogLevel": {
      "Default": "Warning",
      "ElasticApm": "Debug"
    }
  },
  "AllowedHosts": "*",
  "ElasticApm":
    {
      "ServerUrls":  "http://myapmserver:8200",
      "TransactionSampleRate": 1.0
    }
}
----

In certain scenarios -- typically when you're not using ASP.NET Core -- you may choose not to active the agent with the `UseElasticApm()` method.
In this case, you can set the log level of the agent with `ElasticApm:LogLevel`, as shown in the following `appsettings.json` file:

[source,js]
----
{
  "Logging": {
    "LogLevel": {
      "Default": "Warning"
    }
  },
  "AllowedHosts": "*",
  "ElasticApm":
    {
      "LogLevel":  "Debug",
      "ServerUrls":  "http://myapmserver:8200",
      "TransactionSampleRate": 1.0
    }
}
----
```

sample/SampleAspNetCoreApp/appsettings.json
```diff
@@ -1,12 +1,12 @@
 {
   "Logging": {
     "LogLevel": {
-      "Default": "Warning"
+      "Default": "Warning",
+      "Apm": "Error"
```
gregkalapos: Should be `ElasticApm`

docs/configuration.asciidoc
gregkalapos: Yeah, I think the structure is very good! 

Some comments:

> With ASP.NET Core, we recommend setting `LogLevel` for the internal APM logger in the standard `Logging`

It's not a recommendation, it's the only way to set it, if you do it the other way (like the 2. example above, it'll be ignored). 

> In certain scenarios -- typically when you're not using ASP.NET Core -- you may choose not to active the agent with the `UseElasticApm()` method.

This is very good, I'd just change a small thing: the part with "you may choose not to active the agent with..."  should be a different. It's not really a choice, if you are not on ASP.NET Core you can't really call the `UseElasticApm` method, so it should be something like "you typically don't activate the agent with..."

Overall I think with this structure the doc is good and easy to understand. On the other hand I'm still not sure this is a good move, it's still hard to explain this, it can be confusing. I think it'd be better to just use the same in both cases. Unfortunately I think there is no way to fetch the loglevel from `"ElasticApm":  { "LogLevel":  "..."}` when we use `ILoggerFactory`.

So I see 2+1 options:

- We go with the improved docs 
- We continue thinking about this
- (and +1: we go with the improved docs and reiterate on this and see if there is a better way to always configure the loglevel the same way)

I'd choose the last one. 

@vhatsura let us know what you think.

bmorelli25: Thanks again Greg! I've added your recommendations and updated the docs below. If you and @vhatsura revisit this and are able to fetch the loglevel in the same place, let me know and I can update the docs accordingly.

```asciidoc
[float]
[[sample-config]]
==== Sample configuration file

Below is a sample `appsettings.json` configuration file for a typical ASP.NET Core application. There are two important takeaways:
1. The part below `ElasticApm` is fetched by the agent if the corresponding `IConfiguration` is passed to the agent.
2. With ASP.NET Core, you must set `LogLevel for the internal APM logger in the standard `Logging` section with the `ElasticApm` category name.

[source,js]
----
{
  "Logging": {
    "LogLevel": {
      "Default": "Warning",
      "ElasticApm": "Debug"
    }
  },
  "AllowedHosts": "*",
  "ElasticApm":
    {
      "ServerUrls":  "http://myapmserver:8200",
      "TransactionSampleRate": 1.0
    }
}
----

In certain scenarios -- like when you're not using ASP.NET Core -- you wont activate the agent with the `UseElasticApm()` method.
In this case, you can set the log level of the agent with `ElasticApm:LogLevel`, as shown in the following `appsettings.json` file:

[source,js]
----
{
  "Logging": {
    "LogLevel": {
      "Default": "Warning"
    }
  },
  "AllowedHosts": "*",
  "ElasticApm":
    {
      "LogLevel":  "Debug",
      "ServerUrls":  "http://myapmserver:8200",
      "TransactionSampleRate": 1.0
    }
}
----
```

kuisathaverat:
jenkins run the test please

docs/configuration.asciidoc
gregkalapos: Thanks @bmorelli25. I think this is a fairly important decision - I thought about it and opened https://github.com/elastic/apm-agent-dotnet/issues/338 with a suggestion - we'll decide there. Will ping you once we came up with something.

src/Elastic.Apm.AspNetCore/Extensions/LoggerExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Elastic.Apm.AspNetCore.Extensions
+{
+	internal static class LoggerExtensions
+	{
+		internal static LogLevel GetMinLogLevel(this ILogger logger)
+		{
+			if (logger == null) throw new ArgumentNullException(nameof(logger));
+
+			if (logger.IsEnabled(Microsoft.Extensions.Logging.LogLevel.Trace)) return LogLevel.Trace;
```
SergeyKleyman: I think we should replace `Level` property in `IApmLogger` by `IsEnabled` with the same semantics as `Microsoft.Extensions.Logging`'s `ILogger.IsEnabled`. This is the only way `IApmLogger.Level` is currently used anyway and it will avoid adding code like this. @gregkalapos WDYT?

gregkalapos: I don't really see the end result of it - we log the `LogLevel`, so we should know that (it's part of `void Log<TState>(LogLevel level, TState state, Exception e, Func<TState, Exception, string> formatter)`). But maybe it's just I don't see the whole thing here.

SergeyKleyman: I think there is some misunderstanding: log level that is included in the log record is derived from the level of the logging statement that emitted the record. For example, log records emitted by the following statement `logger.Debug()?.Log(...)` will have level `Debug`. OTOH `Level` property in `IApmLogger` controls which records are allowed to be emitted based on their level so it functions exactly like Microsoft.Extensions.Logging's `ILogger.IsEnabled`. Other logging libraries do it the same way as Microsoft.Extensions.Logging: for example [Serilog's `ILogger.IsEnabled`](https://github.com/serilog/serilog/blob/dev/src/Serilog/ILogger.cs#L190) and [NLog's `ILogger.IsTraceEnabled`](https://github.com/NLog/NLog/blob/master/src/NLog/ILogger.cs#L49)
Because of this discrepancy between the way Microsoft.Extensions.Logging does it and the way we currently specified `IApmLogger` interface (by forcing `Level` property), the bridge between `IApmLogger` and Microsoft.Extensions.Logging's `ILogger`, implemented in this PR, has to "guess" the minimal level that is not filtered out by repeatedly calling `ILogger.IsEnabled` as this code does. This is both wasteful from performance point of view and very hackish code because `IApmLogger.Level` is used only by [`LoggingExtensions.IfLevel`](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm/Logging/IApmLoggingExtensions.cs#L32) and it used exactly like `ILogger.IsEnabled`
```
logger.Level <= level ? new MaybeLogger(logger, level) : (MaybeLogger?)null;
```
It's like parsing a number from a string just to print it. But in this case it's even worse because it also prevents supporting dynamically changing the level. If we replace `IApmLogger`'s `Level` by `IsEnabled` then `AspNetCoreLogger.IsEnabled` from this PR can just call `_logger.IsEnabled` and automatically support more use cases (dynamically changing the level in particular) with less code - no need to guess level anymore and thus `LoggerExtensions.GetMinLogLevel` is not necessary.

gregkalapos: Ok, I see... so you want an `bool IsEnabled(LogLevel)` method (Or `bool IsDebugEnabled()`) methods - not an `bool IsEnabled` property.

So we'd have this:

```
public interface IApmLogger
{
	bool IsEnabled(LogLevel level);		
}
```

and this:
```
internal class AspNetCoreLogger : IApmLogger
{
   private readonly ILogger _logger;

   public bool IsEnabled(LogLevel level)
	=> _logger.IsEnabled(Convert(level));
//...more coee
}
```

Right?

SergeyKleyman: Yes, exactly. Although I'm not sure what you mean by `bool IsEnabled` property - do you mean NLog's approach with `bool IsDebugEnabled` etc. I'm not sure why they did that way (performance ?) - I would just go with Microsoft.Extensions.Logging and Serilog's `ILogger.IsEnabled` approach. It's hidden behind `LoggingExtensions.Trace()`, `LoggingExtensions.Debug()` etc. so even if in the future we change our minds and decide to go with NLog's `IsDebugEnabled` approach we will just need to change the way `LoggingExtensions.Trace()` is implemented but the rest of the agent won't be affected.

gregkalapos: > Although I'm not sure what you mean by bool IsEnabled property 

No.. that was my original interpretation of you original message, like I thought you want a property called `IsEnabled` that returns a bool - which was obviously not what you meant :) I think we understand each other now - C# code snippet helps ;)

src/Elastic.Apm.AspNetCore/ApmMiddlewareExtension.cs
```diff
@@ -62,6 +64,13 @@ public static class ApmMiddlewareExtension
 			return builder.UseMiddleware<ApmMiddleware>(agent.Tracer, agent);
 		}
 
+		internal static IApmLogger GetApmLogger(this IServiceProvider serviceProvider)
+		{
```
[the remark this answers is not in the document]
vhatsura: fixed

test/Elastic.Apm.AspNetCore.Tests/AspNetCoreLoggerTests.cs
```diff
@@ -0,0 +1,50 @@
...
+namespace Elastic.Apm.AspNetCore.Tests
+{
+	/// <summary>
+	/// Tests the <see cref="Elastic.Apm.AspNetCore.AspNetCoreLogger"/> type.
+	/// </summary>
+	public class AspNetCoreLoggerTests
+	{
+		[Fact]
+		public void AspNetCoreLoggerShouldThrowExceptionWhenLoggerFactoryIsNull()
+		{
```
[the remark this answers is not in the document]
vhatsura: fixed

sample/SampleAspNetCoreApp/appsettings.json
```diff
@@ -1,12 +1,12 @@
...
   "Logging": {
     "LogLevel": {
-      "Default": "Warning"
+      "Default": "Warning",
+      "Apm": "Error"
     }
   },
   "AllowedHosts": "*",
   "ElasticApm": {
-    "LogLevel": "Error",
```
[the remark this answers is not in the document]
vhatsura: added

sample/SampleAspNetCoreApp/appsettings.json
vhatsura: fixed

docs/configuration.asciidoc
vhatsura: updated

src/Elastic.Apm.AspNetCore/AspNetCoreLogger.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace Elastic.Apm.AspNetCore
+{
+	internal class AspNetCoreLogger : IApmLogger
+	{
+		private readonly ILogger _logger;
+
+		public AspNetCoreLogger(ILoggerFactory loggerFactory)
+		{
+			_logger = loggerFactory?.CreateLogger("ElasticApm") ?? throw new ArgumentNullException(nameof(loggerFactory));
```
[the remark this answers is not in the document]
vhatsura: changed

src/Elastic.Apm.AspNetCore/Extensions/LoggerExtensions.cs
vhatsura: It would be nice to have consistent API with `Microsoft.Extensions.Logging`, however, I think it's not a goal of this PR. For me, it would be better to have another PR with replacing `Level` property to `IsEnabled`. @gregkalapos, @SergeyKleyman, what do you think about it?

SergeyKleyman: @vhatsura I would recommend to make this change in this PR but it's not a blocker for me so the final decision is yours.

The reason I recommend to make this change in this PR is because it's a classic case for refactoring - you want to implement a new feature but the current design is too rigid to accommodate it. Of course you can hack around the current design but I prefer the approach of refactoring the existing code and then implementing the new feature on top thus avoiding hacks. 

But in this case the refactoring required is very localized so postponing it doesn't matter that much that is why it's just a recommendation from me and the final decision is yours.

gregkalapos: I'd prefer to do it in another PR, mainly because it's a breaking change. I think having a breaking change as part of a feature PR isn't optimal - And since we can avoid it, it take that route.

vhatsura: Got it. Will do changes in another PR

vhatsura:
@gregkalapos, @SergeyKleyman, @bmorelli25, it seems I fixed all your comments. Can you look into changes once again?

gregkalapos:
> @gregkalapos, @SergeyKleyman, @bmorelli25, it seems I fixed all your comments. Can you look into changes once again?

Yes, will go over it soon one last time. I think we'll be good to merge 👍  

Thanks for all the work @vhatsura, very much appreciated!

docs/configuration.asciidoc
```diff
@@ -36,7 +36,34 @@ With this you can use any configuration source that you configured on the `IConf
 
 NOTE: By simply calling `app.UseElasticApm()` without the overload, the agent will read configurations only from environment variables.
 
-This is a typical `appsettings.json` file that contains some sample configuration. The part below `ElasticApm` is fetched by the agent if the corresponding `IConfiguration` is passed to the agent.
+[float]
+[[sample-config]]
+==== Sample configuration file
+
+Below is a sample `appsettings.json` configuration file for a typical ASP.NET Core application. There are two important takeaways:
+1. The part below `ElasticApm` is fetched by the agent if the corresponding `IConfiguration` is passed to the agent.
```
```suggestion

1. The part below `ElasticApm` is fetched by the agent if the corresponding `IConfiguration` is passed to the agent.
```

PR #250, opened 2019-06-02
Update CONTRIBUTING.md

gregkalapos:
- Added some general advice to communicate how we usually work and what our preferences are when someone opens a PR
- Added a paragraph about performance tests with BenchmarkDotNet
- Describe the strategy about the dependencies of the agent and also about the versions that we depend on - addressing a real issue here, since lots of people want to introduce new dependencies, or increase the version of our dependencies to use new stuff - in all those cases I have to be the party pooper and explain how that would kill support for older versions. I think it'd make sense to have some formal description about it and explain why that is the case.

Happy to change things in case something is not ok. Also open for discussion here; we don't have to merge this quickly, I'd rather prefer a broad agreement on this stuff.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,7 +25,7 @@ Talk about what you would like to do.
 It may be that somebody is already working on it,
 or that there are particular issues that you should know about before implementing the change.
 
-We aim to maintain a high code quality, therefore every PR goes through the same review process, no matter who opened the PR.
+We aim to maintain high code quality, therefore every PR goes through the same review process, no matter who opened the PR.
 
 ### General advice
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -17,14 +17,41 @@ In order to build the source code of the .NET APM Agent you need
...
 
+We aim to maintain high code quality, therefore every PR goes through the same review process, no matter who opened the PR.
+
+### General advice
+
+Please make sure that your PR addresses a single issue and its size is reasonable to review. There are no hard rules, but please keep in mind that every line of your PR will be reviewed and it's not uncommon that a longer discussion evolves in case of a sensitive change.
+
+Therefore it's preferred to have multiple smaller PRs over a single big PR.
+
+This makes sure that obvious changes get merged quickly and don't get blocked by unrelated changes. Additionally, the repository will also have a useful and searchable history by doing so.
```
SergeyKleyman: I'm not sure "obvious changes" is the right term - I would go with something like "changes that have consensus".

CONTRIBUTING.md
```diff
@@ -17,14 +17,41 @@ In order to build the source code of the .NET APM Agent you need
...
+
+Please don't:
+- Create a PR where you address multiple issues at once.
+- Create a giant PR with a huge change set. There is no hard rule, but if your change grows over 1000 lines, it's maybe worth thinking about making it self contained and submit it as a PR and address follow-up issues in a subsequent PR. (of course there can be exceptions)
+- Actively push code to a PR until you haven't received any feedback. Of course if you spot some minor things after you opened a PR, it's perfectly fine to push a small fix. But please don't do active work on a PR that haven't received any feedback. It's very possible that someone already looked at it and is about to write a detailed review. If you actively add new changes to a PR then the reviewer will have a hard time to provide up to date feedback. If you just want to show work-in-progress code, feel free to use the draft feature of github, or indicate in the title, that the work is not 100% done yet. Of course, once you have feedback on a PR, it's perfectly fine, or rather encouraged to start working collaboratively on the PR and push new changes and address issues and suggestions from the reviewer.
```
SergeyKleyman: > Actively push code to a PR until you haven't received any feedback

This is how I would phrase it in Russian and Hebrew but I think in English "until" is followed by a _positive_ condition so it should be

> Actively push code to a PR until you have received feedback

@bmorelli25 Am I correct here?

CONTRIBUTING.md
```diff
@@ -91,6 +118,38 @@ Once approved, the author of the pull request,
 or reviewer if the author does not have commit access,
 should "Squash and merge".
 
+### Design considerations
+
+#### Performance
+
+The agent is designed to monitor production applications. Therefore it's very important to keep the overhead of the agent as low as possible.
```
SergeyKleyman: "the overhead" -> "the performance overhead"

CONTRIBUTING.md
```diff
@@ -91,6 +118,38 @@ Once approved, the author of the pull request,
...
+
+If this is the case then a perf. test should be added to the `test\Elastic.Apm.PerfTests` project which proves that the new code does not make the performance of the agent worse than it was before your PR.
+
+We care both about memory and CPU overhead and both should be measured. The `test\Elastic.Apm.PerfTests` is configured to measure both.
+
+#### Compatibility
+
+We aim to support the broadest set of .NET flavors and versions possible. This includes different OSs, typically Windows, Linux, and macOS.
+
+This is especially true for versions currently officially supported by Microsoft. In case of specific libraries and frameworks (e.g. ASP.NET Core) we aim to support every version which is supported by Microsoft or the library/framework author.
```
SergeyKleyman: "currently officially" sounds a little strange - how about "officially supported by Microsoft at the moment"

CONTRIBUTING.md
```diff
@@ -91,6 +118,38 @@ Once approved, the author of the pull request,
...
+
+This means that the .NET APM Agent can end up in very different environments.
+
+There are 2 categories of projects in the `src` folder:
```
SergeyKleyman: I would a add some transition to tie "Agent can end up in very different environments" and "There are 2 categories of projects in the `src` folder" parts. How about: "To allow agent to be used in all these different environments there are 2 categories of projects in the `src` folder"

CONTRIBUTING.md
bmorelli25: Good catch @SergeyKleyman, you're correct.

CONTRIBUTING.md
gregkalapos: Changed it accordingly.

CONTRIBUTING.md
gregkalapos: Interesting, none of my languages work this way.

Fixed it, thanks!

CONTRIBUTING.md
gregkalapos: Fixed

CONTRIBUTING.md
gregkalapos: Yeah, that sounds better, changed it.

CONTRIBUTING.md
gregkalapos: Ok, done 👍

gregkalapos:
Thanks for the reviews!

Issue #251, opened 2019-06-04
Run Elastic.Apm.DockerTests in Docker


gregkalapos:
elastic/apm-agent-dotnet#252 implements reading container id and also adds a test covering that - this test is in the `test/Elastic.Apm.DockerTests/Elastic.Apm.DockerTests` project. We need to run this project in docker in order for it to work (if it's not running inside docker the test just skips all the asserts, so the PR is ok for merge).

Happy to create `Dockerfile` or get involved with this, but I'll need help from @elastic/observablt-robots. Let me know if there are any steps I can prepare.

kuisathaverat:
I am testing a solution in this PR https://github.com/elastic/apm-agent-dotnet/pull/256

kuisathaverat:
in review

Issue #238, opened 2019-05-27
Send container.id in system.


gregkalapos:
Preparing a talk and ran into this. This is very little work for a very useful feature:

Once we send this, users can jump from infrastructure/container to APM data.

This is the missing part: https://github.com/elastic/apm-server/blob/master/docs/spec/system.json#L23

gregkalapos:
Also worth noting: this must be the full container id.

Something like this would be I think a good start: 

```
using (var sr = new StreamReader("/proc/self/cgroup"))
{
	var line = sr.ReadLine();
	var parts = line?.Split('/');
	if (parts != null)
	{
		var lastpart = parts.Last();
		_system.Container = new Container { Id = lastpart };
	}
}
```

axw:
> We should also come up with something for Windows containers.

@gregkalapos If you do come up for something for Windows containers, please create an issue on elastic/apm. I guess it'll be significantly more common for .NET, but still would be useful to have this documented somewhere central.

gregkalapos:
Since the linux support is fairly simple, I implemented it in #238, I think that's a good start, I don't have data, but I feel Linux containers are still more popular among .NET devs than Windows containers - you can also run those on a Windows host, so there is no giant real need for windows containers - and I think people prefer windows containers.

If #238 gets merged I'll close this, open another one just for Windows, once we come up with a solution for that, I'll document it in elastic/apm.

axw:
By the way, in case you haven't seen it: there's a description of how we should go about extracting k8s and container info in https://github.com/elastic/apm/issues/21#issuecomment-444035232, as well as some test cases. You don't have to do the k8s bit straight away, but it might be useful to run through the test cases to make sure the container ID is extracted correctly.

gregkalapos:
> By the way, in case you haven't seen it: there's a description of how we should go about extracting k8s and container info in [elastic/apm#21 (comment)](https://github.com/elastic/apm/issues/21#issuecomment-444035232), as well as some test cases. You don't have to do the k8s bit straight away, but it might be useful to run through the test cases to make sure the container ID is extracted correctly.

👍I added a unit test to test through all the values from the referenced comment - the algorithm itself is a copy-paste from Java with some C#-ification. We only capture the `containerId`, we cover all those k8s cases as well - but don't capture k8s specific things, just the container id.

PR #252, opened 2019-06-04
Implement sending docker container id

gregkalapos:
Solves elastic/apm-agent-dotnet#238 

The algorithm to parse the container id is basically the C#-ified version of the code from the Java agent.

Currently linux-only, but I think that's fine for this first iteration.

~Testing coverage is in `test/Elastic.Apm.DockerTests/Elastic.Apm.DockerTests`. Problem is currently non of our tests run inside docker. For this I opened #251. Nevertheless I tested it locally, works fine, and I'd merge this, so we can start working on #251.~ Update: #251 done, should be fine in CI.


gregkalapos:
After reading https://github.com/elastic/apm-agent-dotnet/issues/238#issuecomment-498916757 I will add some more tests and change the parsing of the id according to the referenced link.

gregkalapos:
> After reading [#238 (comment)](https://github.com/elastic/apm-agent-dotnet/issues/238#issuecomment-498916757) I will add some more tests and change the parsing of the id according to the referenced link.

Done.

@SergeyKleyman this is ready to review. Please have a look if you have time. Thanks.

src/Elastic.Apm/Helpers/SystemInfoHelper.cs
```diff
@@ -0,0 +1,106 @@
...
+				if (matcher.Success)
+				{
+					for (var i = 1; i <= matcher.Groups.Count; i++)
+					{
+						var podUid = matcher.Groups[i].Value;
+						if (string.IsNullOrWhiteSpace(podUid)) continue;
+
+						//logger.debug("Found Kubernetes pod UID: {}", podUid);
+						// By default, Kubernetes will set the hostname of the pod containers to the pod name. Users that override
+						// the name should use the Downward API to override the pod name.
```
SergeyKleyman: What is "Downward API"?

src/Elastic.Apm/Helpers/SystemInfoHelper.cs
```diff
@@ -0,0 +1,106 @@
...
+			var idPart = Path.GetFileName(cGroupPath);
+
+			if (string.IsNullOrWhiteSpace(idPart)) return null;
+
+			// Legacy, e.g.: /system.slice/docker-<CID>.scope
+			if (idPart.EndsWith(".scope"))
+				idPart = idPart.Substring(0, idPart.Length - ".scope".Length).Substring(idPart.IndexOf("-", StringComparison.Ordinal) + 1);
+
+			// Looking for kubernetes info
+			var dirPathPart = Directory.GetParent(cGroupPath); // Path.get (cGroupPath);  cGroupPath Path.GetPathRoot(cGroupPath);
```
SergeyKleyman: I wouldn't recommend using file system API on strings that are not related to file system. For example on Windows for `cGroupPath` with value `/kubepods/besteffort/pode9b90526-f47d-11e8-b2a5-080027b9f4fb/15aa6e53-b09a-40c7-8558-c6c31e36c88a`, `Directory.GetParent(cGroupPath)` is `C:\kubepods\besteffort\pode9b90526-f47d-11e8-b2a5-080027b9f4fb` and of course the whole test fails.

src/Elastic.Apm/Helpers/SystemInfoHelper.cs
```diff
@@ -0,0 +1,106 @@
...
+			if (dirPathPart != null)
+			{
+				var dir = dirPathPart.ToString();
+
+				var pattern = new Regex(PodRegex);
```
SergeyKleyman: In this case it's not that important because this code runs only once but in general it's better to construct `Regex` only once (so I would make it a field) because it needs to parse the pattern and build state machine.

test/Elastic.Apm.DockerTests/ContainerIdCalculationTests.cs
```diff
@@ -0,0 +1,44 @@
...
+	{
+		[Theory]
+		[InlineData("13:name=systemd:/docker/ae8d0ef9c8757a42d01e994b0ffe0e93e26ece256f68cd850467836c83202367",
+			"ae8d0ef9c8757a42d01e994b0ffe0e93e26ece256f68cd850467836c83202367")]
+		[InlineData("12:devices:/docker/051e2ee0bce99116029a13df4a9e943137f19f957f38ac02d6bad96f9b700f76",
+			"051e2ee0bce99116029a13df4a9e943137f19f957f38ac02d6bad96f9b700f76")]
+		[InlineData("1:name=systemd:/system.slice/docker-cde7c2bab394630a42d73dc610b9c57415dced996106665d427f6d0566594411.scope",
+			"cde7c2bab394630a42d73dc610b9c57415dced996106665d427f6d0566594411")]
+		[InlineData("1:name=systemd:/kubepods/besteffort/pode9b90526-f47d-11e8-b2a5-080027b9f4fb/15aa6e53-b09a-40c7-8558-c6c31e36c88a",
+			"15aa6e53-b09a-40c7-8558-c6c31e36c88a")]
```
SergeyKleyman: I would also add the following test mentioned at https://github.com/elastic/apm/issues/21#issuecomment-444410552

    "1:name=systemd:/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod90d81341_92de_11e7_8cf2_507b9d4141fa.slice/crio-2227daf62df6694645fee5df53c1f91271546a9560e8600a525690ae252b7f63.scope": {
        "container": {
            "id": "2227daf62df6694645fee5df53c1f91271546a9560e8600a525690ae252b7f63",
        },
        "pod": {
            "uid": "90d81341_92de_11e7_8cf2_507b9d4141fa",
        }
    }

gregkalapos:
@SergeyKleyman Thanks for finding the failing test on Windows (I haven't ran this on Win, and the CI seem to have trouble, builds are scheduled but never run).

All the new tests should only run within a Linux container. The `BasicDockerTests` has 1 check for this, to other test which fails doesn't. I think best would be to just add the same check, or at least make sure it runs on Linux. This feature only works within a Linux container, so I suggest that's what we focus on. There will be a follow up issue for Windows, which I think is less important at the moment (https://github.com/elastic/apm-agent-dotnet/issues/238#issuecomment-498862085).

SergeyKleyman:
Sounds good - if you would like me to re-run the tests on Windows after the fix please let me know.

src/Elastic.Apm/Helpers/SystemInfoHelper.cs
gregkalapos: `Path.GetDirectoryName(cGroupPath);` would probably a better choice here - we use `Path` everywhere, so with this we rely on that for parsing. What do you think? Tested it on Windows and that works. 

Or by saying 
> I wouldn't recommend using file system API on strings 
you also mean `Path` and not only `Directory`?

SergeyKleyman: I would not use file system API here at all because relation to file system is purely coincidental - just because `/` is used in the string. My concern is that file system API is much heavier than just parsing strings and I would avoid bringing heavier API than the necessary minimum. Like I mentioned above I would just implement it by searching backwards for the last `/`. But it's up to you to decide - I'm fine with your proposal if it passes all the tests.

src/Elastic.Apm/Helpers/SystemInfoHelper.cs
```diff
@@ -0,0 +1,106 @@
...
+using Elastic.Apm.Api;
+using Elastic.Apm.Logging;
+
+namespace Elastic.Apm.Helpers
+{
+	internal class SystemInfoHelper
+	{
+		private const string ContainerUidRegex = "^[0-9a-fA-F]{64}$";
+
+		private const string PodRegex = "(?:^/kubepods/[^/]+/pod([^/]+)$)|"
```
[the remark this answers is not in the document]
gregkalapos: Done :)

test/Elastic.Apm.DockerTests/ContainerIdCalculationTests.cs
gregkalapos: 👍 added

src/Elastic.Apm/Helpers/SystemInfoHelper.cs
gregkalapos: It's [this thing](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#the-downward-api) - I don't know anything about it either... More Kubernetes things will come, this PR is just about the container ID, but I thought I leave the comments there, since they will be probably very useful hints once we start working on reporting Kubernetes related things.

src/Elastic.Apm/Helpers/SystemInfoHelper.cs
gregkalapos: 👍  Yeah, will move it out to a field.

There are ways to boost the perf. of regex, eg. we could also make it compiled, but since this is called only once I won't do that, because that [would make the 1. (and only) run slower](https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regexoptions?view=netframework-4.8#fields)

So as you suggested, we only make this a field.

gregkalapos:
> It now passes all the tests on Windows.

And our CI is also back 🕺

src/Elastic.Apm/Helpers/SystemInfoHelper.cs
gregkalapos: I see; well that'd be a bigger refactoring. Also, this is stolen from Java, and they also do it this way. I think it's fine, I'd also prefer using this API than reimplementing it and do those things manually. Let's roll it this way for now 🚀 

cc @eyalkoren @felixbarny

eyalkoren: @gregkalapos I just ran into an issue with my unit tests failing in Windows because of using the FileSystem APIs. So, while I maintain the usage of those APIs for actual reading the file content (which actually requires real file system operations), I am removing them from the ID parsing part, which deals with stuff that only look like paths, but probably aren't...

gregkalapos: Thanks for the heads up! Are you working on docker windows support, or is it some other id?

eyalkoren: Since this is about container IDs, it is a Linux thing anyway. I changed that only to enable building our agent on Windows.

gregkalapos: Och I see... we have runtime checks and in our case this part only runs on Linux - both the prod code and also tests covering this.

PR #253, opened 2019-06-04
Fix stacktrace

gregkalapos:
We introduced 2 issues in elastic/apm-agent-dotnet#240 - fixing those. 

1: Never send null or empty `FileName` (since it's required)
2: ~Only send LineNo when the methodname was not changed manually (e.g. due to being async mehtod) - otherwise the LineNo would not match the method name~ This seems to work, so we send those, and will be further investigated in https://github.com/elastic/apm-agent-dotnet/issues/266, discussion: https://github.com/elastic/apm-agent-dotnet/pull/253#discussion_r291755759

+tests

### DIFF
--- a/src/Elastic.Apm/Helpers/StacktraceHelper.cs
+++ b/src/Elastic.Apm/Helpers/StacktraceHelper.cs
@@ -28,14 +28,15 @@ namespace Elastic.Apm.Helpers
 			try
 			{
 				retVal.AddRange(from item in frames
+					let fileName = item?.GetMethod()
+						?.ReflectedType?.FullName //see: https://github.com/elastic/apm-agent-dotnet/pull/240#discussion_r289619196
+					let functionName = GetRealMethodName(item?.GetMethod())
 					select new CapturedStackFrame
 					{
-						Function = GetRealMethodName(item?.GetMethod()),
-						FileName =
-							item?.GetMethod()
-								?.ReflectedType?.FullName, //see: https://github.com/elastic/apm-agent-dotnet/pull/240#discussion_r289619196
+						Function = functionName,
+						FileName = string.IsNullOrWhiteSpace(fileName)? "unknown" : fileName,
 						Module = item?.GetMethod()?.ReflectedType?.Assembly.FullName,
-						LineNo = item?.GetFileLineNumber() ?? 0
+						LineNo = item?.GetMethod().Name == functionName ? item?.GetFileLineNumber() ?? 0 : 0
 					});
 			}
 			catch (Exception e)
--- a/src/Elastic.Apm/Helpers/StacktraceHelper.cs
+++ b/src/Elastic.Apm/Helpers/StacktraceHelper.cs
@@ -29,12 +29,12 @@ namespace Elastic.Apm.Helpers
 			{
 				retVal.AddRange(from item in frames
 					let fileName = item?.GetMethod()
-						?.ReflectedType?.FullName //see: https://github.com/elastic/apm-agent-dotnet/pull/240#discussion_r289619196
+						?.DeclaringType?.FullName //see: https://github.com/elastic/apm-agent-dotnet/pull/240#discussion_r289619196
 					let functionName = GetRealMethodName(item?.GetMethod())
 					select new CapturedStackFrame
 					{
 						Function = functionName,
-						FileName = string.IsNullOrWhiteSpace(fileName)? "unknown" : fileName,
+						FileName = string.IsNullOrWhiteSpace(fileName)? "N/A" : fileName,
 						Module = item?.GetMethod()?.ReflectedType?.Assembly.FullName,
 						LineNo = item?.GetMethod().Name == functionName ? item?.GetFileLineNumber() ?? 0 : 0
 					});
--- a/src/Elastic.Apm/Helpers/StacktraceHelper.cs
+++ b/src/Elastic.Apm/Helpers/StacktraceHelper.cs
@@ -36,7 +36,7 @@ namespace Elastic.Apm.Helpers
 						Function = functionName,
 						FileName = string.IsNullOrWhiteSpace(fileName)? "N/A" : fileName,
 						Module = item?.GetMethod()?.ReflectedType?.Assembly.FullName,
-						LineNo = item?.GetMethod().Name == functionName ? item?.GetFileLineNumber() ?? 0 : 0
+						LineNo = item?.GetFileLineNumber() ?? 0
 					});
 			}
 			catch (Exception e)
--- a/test/Elastic.Apm.Tests/StackTraceTests.cs
+++ b/test/Elastic.Apm.Tests/StackTraceTests.cs
@@ -42,6 +42,8 @@ namespace Elastic.Apm.Tests
 
 		/// <summary>
 		/// Makes sure that the name of the async method is captured correctly
+		/// Also asserts that the line number is 0 in this case - that is because the fixed method name
+		/// does not match the line number in the state machine, so the agent sends LineNo 0.
 		/// </summary>
 		[Fact]
 		public async Task AsyncCallStackTest()
@@ -60,7 +62,8 @@ namespace Elastic.Apm.Tests
 
 			payloadSender.Errors.Should().NotBeEmpty();
 			(payloadSender.Errors.First() as Error).Should().NotBeNull();
-			(payloadSender.Errors.First() as Error)?.Exception.Stacktrace.Should().Contain(m => m.Function == nameof(ClassWithAsync.TestMethodAsync));
+			(payloadSender.Errors.First() as Error)?.Exception.Stacktrace.Should()
+				.Contain(m => m.Function == nameof(ClassWithAsync.TestMethodAsync) && m.LineNo == 0);
 		}
 
 		/// <summary>
@@ -118,6 +121,39 @@ namespace Elastic.Apm.Tests
 				.Contain(m => m.FileName == typeof(DerivedTestClass).FullName
 					&& m.Function == nameof(DerivedTestClass.TestMethod)
 					&& m.Module == typeof(DerivedTestClass).Assembly.FullName);
+		}
+
+		/// <summary>
+		/// Makes sure that the filename is never null or empty in the call stack, since it's a required field.
+		/// </summary>
+		[Fact]
+		public void StackTraceWithLambda()
+		{
+			Action action = () => { TestMethod(); };
+
+			var payloadSender = new MockPayloadSender();
+			var agent = new ApmAgent(new TestAgentComponents(payloadSender: payloadSender));
+
+			Assert.Throws<Exception>(() => { agent.Tracer.CaptureTransaction("TestTransaction", "Test", () => { action(); }); });
+
+			payloadSender.Errors.Should().NotBeEmpty();
+			(payloadSender.Errors.First() as Error).Should().NotBeNull();
+			(payloadSender.Errors.First() as Error)?.Exception.Stacktrace.Should().NotContain(frame => string.IsNullOrWhiteSpace(frame.FileName));
+		}
+
+		private void TestMethod() => InnerTestMethod(() => throw new Exception("TestException"));
+
+		private void InnerTestMethod(Action actionToRun)
+		{
+			try
+			{
+				actionToRun();
+			}
+			catch (Exception e)
+			{
+				Console.WriteLine(e);
+				throw;
+			}
 		}
 
 		private class ClassWithSyncMethods
--- a/test/Elastic.Apm.Tests/StackTraceTests.cs
+++ b/test/Elastic.Apm.Tests/StackTraceTests.cs
@@ -49,16 +49,17 @@ namespace Elastic.Apm.Tests
 		public async Task AsyncCallStackTest()
 		{
 			var payloadSender = new MockPayloadSender();
-			var agent = new ApmAgent(new TestAgentComponents(payloadSender: payloadSender));
-
-			await Assert.ThrowsAsync<Exception>(async () =>
+			using (var agent = new ApmAgent(new TestAgentComponents(payloadSender: payloadSender)))
 			{
-				await agent.Tracer.CaptureTransaction("TestTransaction", "Test", async () =>
+				await Assert.ThrowsAsync<Exception>(async () =>
 				{
-					var classWithAsync = new ClassWithAsync();
-					await classWithAsync.TestMethodAsync();
+					await agent.Tracer.CaptureTransaction("TestTransaction", "Test", async () =>
+					{
+						var classWithAsync = new ClassWithAsync();
+						await classWithAsync.TestMethodAsync();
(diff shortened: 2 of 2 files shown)


gregkalapos:
cc skynet2, jfyi, this is a follow up on your previous PR.

gregkalapos:
cc @skynet2, jfyi, this is a follow up on your previous PR.

test/Elastic.Apm.Tests/StackTraceTests.cs
```diff
@@ -42,6 +42,8 @@ public async Task HttpClientStackTrace()
 
 		/// <summary>
 		/// Makes sure that the name of the async method is captured correctly
+		/// Also asserts that the line number is 0 in this case - that is because the fixed method name
+		/// does not match the line number in the state machine, so the agent sends LineNo 0.
```
SergeyKleyman: Please see my comment for `LineNo = item?.GetMethod().Name == functionName ? item?.GetFileLineNumber() ?? 0 : 0`

src/Elastic.Apm/Helpers/StacktraceHelper.cs
```diff
@@ -28,14 +28,15 @@ internal static List<CapturedStackFrame> GenerateApmStackTrace(StackFrame[] fram
...
 					{
-						Function = GetRealMethodName(item?.GetMethod()),
-						FileName =
-							item?.GetMethod()
-								?.ReflectedType?.FullName, //see: https://github.com/elastic/apm-agent-dotnet/pull/240#discussion_r289619196
+						Function = functionName,
+						FileName = string.IsNullOrWhiteSpace(fileName)? "unknown" : fileName,
 						Module = item?.GetMethod()?.ReflectedType?.Assembly.FullName,
-						LineNo = item?.GetFileLineNumber() ?? 0
+						LineNo = item?.GetMethod().Name == functionName ? item?.GetFileLineNumber() ?? 0 : 0
```
SergeyKleyman: Doesn't compiler preserve original source code line numbers even after rewriting IL?

src/Elastic.Apm/Helpers/StacktraceHelper.cs
```diff
@@ -28,14 +28,15 @@ internal static List<CapturedStackFrame> GenerateApmStackTrace(StackFrame[] fram
...
+						?.ReflectedType?.FullName //see: https://github.com/elastic/apm-agent-dotnet/pull/240#discussion_r289619196
+					let functionName = GetRealMethodName(item?.GetMethod())
 					select new CapturedStackFrame
 					{
-						Function = GetRealMethodName(item?.GetMethod()),
-						FileName =
-							item?.GetMethod()
-								?.ReflectedType?.FullName, //see: https://github.com/elastic/apm-agent-dotnet/pull/240#discussion_r289619196
+						Function = functionName,
+						FileName = string.IsNullOrWhiteSpace(fileName)? "unknown" : fileName,
```
SergeyKleyman: Do we have example when method type name is not available? Can we use something more useful than "unknown"? Maybe we can fallback on some other property from the stack frame info.

skynet2: I think item.GetFileName() can work here

src/Elastic.Apm/Helpers/StacktraceHelper.cs
```diff
@@ -28,14 +28,15 @@ internal static List<CapturedStackFrame> GenerateApmStackTrace(StackFrame[] fram
 			try
 			{
 				retVal.AddRange(from item in frames
+					let fileName = item?.GetMethod()
+						?.ReflectedType?.FullName //see: https://github.com/elastic/apm-agent-dotnet/pull/240#discussion_r289619196
```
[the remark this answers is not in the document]
gregkalapos: I thought by using `DeclaringType` it's not possible to unambiguously what code is executed. 
From the stackoverflow example, with that we would show `Base in Method`, but that's not what we want... the code executed is acutally in `Derived in Method`. 

Or am I wrong? 🤔

src/Elastic.Apm/Helpers/StacktraceHelper.cs
gregkalapos: I don't know, but I'll research this...

src/Elastic.Apm/Helpers/StacktraceHelper.cs
gregkalapos: > Do we have example when method type name is not available?

Yes, run the `/home/index` in the ASP.NET Core sample application from master... the outgoing http request (where we capture the stacktrace) will be rejected, because the FileName remains `null`. It's a generated lambda. 

> I think item.GetFileName() can work here
Wouldn't that be very confusing? Also I think in that case `GetFileName` also returns `null`, but I haven't tested this, it's just a guess: if there is no `?.ReflectedType` maybe there is no filename either.

src/Elastic.Apm/Helpers/StacktraceHelper.cs
gregkalapos: Hahh, it does, but now the class name does not match - but we can also fix that... 

image

src/Elastic.Apm/Helpers/StacktraceHelper.cs
skynet2: Yes, GetFileName will not work for that 
image

gregkalapos: Thanks for trying @skynet2! I was about to do that.. appreciate the info.

skynet2: Do you think that it makes sense to include such data in stack trace? 
Without fileName, it makes no sense to have data in apm. As i can remember, previously, there was a check to check if fileName is not empty, and only then include it as a stackframe

gregkalapos: So.. to answer the question:

> Can we use something more useful than "unknown"? 

I don't think so... it's literally unknown. That's some runtime generated stuff, which has no type name and also no filename. Also, I think in msil it's possible to have global functions without having a class.

gregkalapos: > Do you think that it makes sense to include such data in stack trace?
> Without fileName, it makes no sense to have data in apm. As i can remember, previously, there was a check to check if fileName is not empty, and only then include it as a stackframe

Hmm, yeah, now that you say this, I remember that we dropped those things originally. I also think that this doesn't really add much value, on the other hand cutting out things from a stacktrace does not sound very good. What if you e.g. also have a memory dump and you want to search for similar callstack in that? If we skip frames we make that harder. But I'm not 100% sure.

skynet2: hm, I don't remember the code where I dropped it, so probably it was not me. Don`t have git log currently :) 
Working with a memory dump from prod, IMHO, that's not a common scenario

src/Elastic.Apm/Helpers/StacktraceHelper.cs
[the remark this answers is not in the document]
SergeyKleyman: Well, it depends on what you are going to do with class name. Let's say you have `Base` class with `SomeMethod()` method that throws an exception and `Derived` class that inherits from `Base`. I would argue seeing `Base` at the stack trace is more useful even if the instance for which `SomeMethod()` was called is `Derived`. It is because `Base` is where you would be able to see source code for `SomeMethod()` and `Derived` class won't have.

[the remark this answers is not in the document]
SergeyKleyman: Of course, if  `SomeMethod()` is `virtual` and `Derived` overrides it then we should see `Derived` and it will be as expected if we use `DeclaringType`.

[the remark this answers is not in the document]
gregkalapos: I get your example. 

But what if `Derived` overrides `SomeMethod`? In that case showing `Base` seems to be wrong to me, while in your example showing `Derived` isn't completely wrong - you can end up in the right place of the code simply looking for the method starting from `Derived` going through parent classes as long as you find the method.

Do we know what other tools do?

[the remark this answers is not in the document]
gregkalapos: > Of course, if SomeMethod() is virtual and Derived overrides it then we should see Derived and it will be as expected if we use DeclaringType.

If that's the case and there aren't any issues with that I'm ok with that.

src/Elastic.Apm/Helpers/StacktraceHelper.cs
SergeyKleyman: I just wanted to make sure we cannot find anything better, at least for the cases of which we know. If there isn't anything better then let's go with unknown. I would just make sure it cannot be confused with a real class name - maybe something like `"<unknown>"`?

gregkalapos: > hm, I don't remember the code where I dropped it, so probably it was not me. Don`t have git log currently :)

I think it was me, my original code skipped those. :) 

> Working with a memory dump from prod, IMHO, that's not a common scenario 
Sure, it's not common. My point is that if we skip frames then we don't really show "THE" stacktrace, we just show a modified stacktrace... it's not the real stacktrace, and I think this can be misleading.

gregkalapos: >  maybe something like "<unknown>"? 

Sounds good to me.

gregkalapos: > > maybe something like `"<unknown>"`?
> 
> Sounds good to me.

@skynet2 also ok with this?

skynet2: yep, let`s try that solution, thanks, @gregkalapos

SergeyKleyman: Nit: I thought about even better alternative - how about `"N/A"`? You guys choose :)

src/Elastic.Apm/Helpers/StacktraceHelper.cs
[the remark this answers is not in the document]
SergeyKleyman: Just to be on the safe side I also tested `DeclaringType` with `interface` and implementing `class` - `DeclaringType` returns implementing `class` so it works as expected.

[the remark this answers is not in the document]
gregkalapos: Ok, I change it to `DeclaringType` - I spent some time on this, and actually I think this does not matter and both return the same in this case.

When the `StackTrace` class builds the frames it builds `MethodBase` instances for each frame, where I think there is no real meaning of `ReflectedType` or `DeclaringType` - it's just a stackframe with a MethodTable pointer, it's not a type or an object instance that they have - and they seem to end up with a value that the `DeclaringType` stores in the samples above in both the `ReflectedType` and in the `DeclaringType` as well.

I added some unit tests pinning this behaviour (basically testing this `Base`,  `Derived` example), but the callstack we capture is the same no matter whether we have `DecleringType` or `ReflectedType`. Furthermore I tested it with our sample ASP.NET Core app and there were no `MethodBase` instances in the `StackFrame[] frames` array where `item?.GetMethod()?.ReflectedType?.FullName != item?.GetMethod()?.DeclaringType?.FullName` - they always seem to be the same.

I think this kind of makes sense, because in case of a stacktrace when a method name is resolved we start already from a Method Table pointer, while in case of all the discussions with sample code about `ReflectedType` vs. `DeclaringType` it starts from a type or an instance.

Some interesting code:
[`StackTrace.CaptureStackTrace`](https://github.com/dotnet/coreclr/blob/9773db1e7b1acb3ec75c9cc0e36bd62dcbacd6d5/src/System.Private.CoreLib/src/System/Diagnostics/StackTrace.CoreCLR.cs#L62)

[`StackFrameHelper.GetMethodBase`](https://github.com/dotnet/coreclr/blob/9773db1e7b1acb3ec75c9cc0e36bd62dcbacd6d5/src/System.Private.CoreLib/src/System/Diagnostics/StackFrameHelper.cs#L176)

[`GetTypicalMethodDefinitio`](https://github.com/dotnet/coreclr/blob/a64cb0a41a4ebeb4a61b8b0f1f2eeeff8dd539c9/src/vm/runtimehandles.cpp#L2268)

[`LoadTypicalMethodDefinition `](https://github.com/dotnet/coreclr/blob/3dd303f0004ed4771bc29167df30efda07e4cf7e/src/vm/method.cpp#L1732)

src/Elastic.Apm/Helpers/StacktraceHelper.cs
gregkalapos: Ok, let's go with `N/A`.

src/Elastic.Apm/Helpers/StacktraceHelper.cs
gregkalapos: @SergeyKleyman I think this makes sense to research more, it looks promising, but just to be on the safe side I'd rather not send the linenumber in case of async method for now. I think it'd be best to move on with this PR since it addresses a fairly important issue (all our events with stackframe are currently rejected by the server) and this change is just for being on the safe side and try to prevent sending incorrect info.

I opened https://github.com/elastic/apm-agent-dotnet/issues/266 to follow up on this.

gregkalapos:
@SergeyKleyman could you please take another look? Thanks!

src/Elastic.Apm/Helpers/StacktraceHelper.cs
SergeyKleyman: In the spirit of progress over perfection I would go with the opposite approach - sending line numbers for all the methods (async as well) and if we encounter an issue we will deal with it then. I think line number is very important information to just drop especially considering that we don't know of any case when it's incorrect.

I'm leaving it to you guys to decide.

gregkalapos: Ok, let's make that way.

Besides not being sure I just thought that it’s a little bit incorrect, since the class name we send does not match the method name, therefore the line number also does not match the class name (the `lineNo`  and the `function` seem to be correct, so that’s fine) - ok, I’ll change this and send line number, I don’t think sending it would be very bad, so let’s make it that way.

Also there is a followup issue, so we'll have another iteration on this question, let's send it always when we have it, and then go from there in #266.

test/Elastic.Apm.Tests/StackTraceTests.cs
gregkalapos: I changed this one to test the opposite, which matches the current agent behaviour (https://github.com/elastic/apm-agent-dotnet/pull/253#discussion_r291835766)

gregkalapos:
> Other than I would go the opposite way with the line numbers, it looks good.

Ok, adapted it and also adapted the corresponding test.

SergeyKleyman:
LGTM

PR #258, opened 2019-06-05
fix: transaction name; closes #257

### LINKED ISSUE #257: Transaction.Name is overridden
Transaction.Name is always overridden, so it`s impossible to change it.

Root case - https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs#L93

cc @SergeyKleyman

skynet2:


gregkalapos:
_We have CI troubles, ignore the failing builds, not related to this PR._

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
```diff
@@ -89,13 +91,16 @@ public async Task InvokeAsync(HttpContext context)
 			catch (Exception e) when (ExceptionFilter.Capture(e, transaction)) { }
 			finally
 			{
-				//fixup Transaction.Name - e.g. /user/profile/1 -> /user/profile/{id}
-				var routeData = (context.Features[typeof(IRoutingFeature)] as IRoutingFeature)?.RouteData;
-				if (routeData != null)
+				if (transactionName == transaction.Name)
```
SergeyKleyman: I would go with explicit flag in `Transaction` instance to signal that name was explicitly set.

sample/SampleAspNetCoreApp/Controllers/HomeController.cs
```diff
@@ -119,6 +119,13 @@ public IActionResult Sample(int id)
 			return Ok(id);
 		}
 
+		public IActionResult SampleWithCustomName()
```
SergeyKleyman: It is little bit confusing what “sample” means here - I would go with `TransactionWithCustomName`

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
skynet2: Hi @SergeyKleyman , 
any suggestions about implementation?

gregkalapos: We could use a full property for `Name` in `Transaction`, setting just the backing property in the constructor and then adding an additional bool flag that we switch when the setter of the full property is called.

On the other hand I don’t see why having an additional flag is better. The code code is not less complex (i'd say a little bit more complex) and that would introduce an additional field, which will be unused in most cases. I was fine with the proposed solution in this PR.

skynet2: I agree with @gregkalapos , adding a new field is too much for that functionality, IMHO.
Moreover, we have that problem only in one specific place ( issue was introduced in order to bypass some limitation (parsing routing parameters) from net.core), so I think it should be fixed locally.

SergeyKleyman: The reason I proposed to have a flag is to address the following issue. If the value to which transaction name is set via API call is “<verb> <URL path>” then it will still be “fixed”. It will make description for this feature quite contrived - “... setting transaction name via public API enforces the use of the set name unless it is <HTTP verb> <URL path>” in which case the name automatically deduced from routing will *still* be used”

Regarding it being only relevant for .NET Core - it’s correct for now but in the future we will implement similar mechanism for full .NET Framework and then we will have to choose whether to go with consistency and implement (IMHO contrived) behavior there as well, or break consistency or to change .NET Core implementation and break backward compatibility.

Essentially what I’m saying is that any feature should start with clear spec and then be implemented according to it not the other way around by making the spec harder to understand to save a couple lines of code.

sample/SampleAspNetCoreApp/Controllers/HomeController.cs
skynet2: Done

skynet2:
HI @gregkalapos and @SergeyKleyman ,
I`ve created one more pull request for fixing that issue, but with custom name flag. 
Please check https://github.com/elastic/apm-agent-dotnet/pull/259 and decide what works better.

Thanks,
Stas

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
gregkalapos: I got the scenario you describe @SergeyKleyman, and it's definitely a more consistent approach. Since @skynet2 already implemented that we can take that approach.

I'm not sure we solve a real issue with that, and I still think that we kind of make this more complicated than it needs to be. But let's make progress on this and go with that approach!

SergeyKleyman: I agree that we don't need to be perfectionists especially if it's significantly increases implementation complexity. But in this case the difference in implementation is a couple of lines for the cleaner specification.

gregkalapos:
@skynet2 this should be closed in favor of #259 - that seems to be the agreed solution.

PR #259, opened 2019-06-06
Fix transaction name v2

skynet2:


SergeyKleyman:
It looks good.

gregkalapos:
@skynet2 one small thing before we merge this: It'd be nice to also cover the case when the `Transaction.Name` is set to `{context.Request.Method} {context.Request.Path}` with a test. Same test your already have, just with a different `Transaction.Name`.

Would you mind pushing a commit with that? If you don't have time, I can also do it before we merge.

Thanks again for the PR!

skynet2:
hi @gregkalapos ,

If you can do that now please go ahead, i`m not at my pc currently

gregkalapos:
> hi @gregkalapos ,
> 
> If you can do that now please go ahead, i`m not at my pc currently

Sure; on it.

gregkalapos:
Added the test, the coverage is ok.

I also talked to devs from other agents and most of the agents have this problem. There would be other ways to solve this:
- In Java e.g. they start the transaction with an empty name and if it's empty at the end of the request then they calculate the default name, otherwise they use what the user provides.  
- In Node.js there are 2 fields, `defaultname` and `name`, `name` is exposed to the public API, so the agent knows if the user provided a custom name or not.
- Ruby does not support this feature.

I think this solution is good, it solves the problem and none of the other solutions are less complicated or significantly better. So this PR is ready to merge. I leave it open a little bit in case there is further feedback, otherwise will be merged soon.

PR #262, opened 2019-06-06
fix: fix release stages

kuisathaverat:
remove undefines stash objects, this will fix the release stages.

image


kuisathaverat:
In Jenkins you can see the header on grey, this means aborted, and it is aborted because we do not answer to the release input, it is because I've forced to execute all stages to check the PR

PR #277, opened 2019-06-13
Add net461 as a separate target for Elastic.Apm

gregkalapos:
#248 seems to break our Full Framework code, mainly because it adds `System.Diagnostics.PerformanceCounter` as a dependency to `Elastic.Apm`. We use the `PerformanceCounter` type from this package, but it turns out that this type is in `System.dll` in case of `net461`. The problem itself isn't specific to this class, I suspect we'd have this later with other things (so as a quick and sloppy summary: using `netstandard2.0` on `net461` is a pain, `net461` should be a separate target, and that's what this PR proposes).

[The recommendation from Microsoft is also](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/cross-platform-targeting#multi-targeting) to target `net461` and `netstandard2.0` separately. 

Current problems/questions: 
- if we simply add `<TargetFrameworks>netstandard2.0;net461</TargetFrameworks>` to `Elastic.Apm` builds will fail on Linux (CI) and macOS (my dev machine). Therefore I added `<PropertyGroup Condition=" '$(OS)' == 'Windows_NT' ">` checks to only build `net461` on Windows. I'm not 100% sure this is a good idea, because in this case the generated NuGet packages contain an additional `net461` dll on Windows, but those are missing when we generate the packages on a non Windows OSs. I think this is very dangerous, because we may end up with a release where the `net461` `Elastic.Apm` dll is missing. Open for ideas on how to solve this, otherwise we can go with this and think about it later.
- Do we want to include this in the beta release? This changes things on Full Framework, so that's a risk, on the other hand, this is probably the way to go, so if we spend time to test it, I don't see any reason not to do the beta release including this change.

@SergeyKleyman feel free to push your fix to the `RuntimeInformation.FrameworkDescription` issue.

### DIFF
--- a/src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
+++ b/src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
@@ -39,7 +39,7 @@ namespace Elastic.Apm.AspNetFullFramework
 		private HttpApplication _httpApp;
 
 		private static Version AspNetVersion => typeof(HttpRuntime).Assembly.GetName().Version;
-		private static string ClrDescription => RuntimeInformation.FrameworkDescription;
+		private static string ClrDescription => "aaa";  // RuntimeInformation.FrameworkDescription;
 		private static Version IisVersion => HttpRuntime.IISVersion;
 
 		private static void SetServiceInformation(Service service)
--- a/src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
+++ b/src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
@@ -1,11 +1,11 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Collections.Specialized;
-using System.Runtime.InteropServices;
 using System.Web;
 using Elastic.Apm.Api;
 using Elastic.Apm.DiagnosticSource;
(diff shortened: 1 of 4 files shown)


src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
```diff
@@ -39,7 +39,7 @@ static ElasticApmModule()
 		private HttpApplication _httpApp;
 
 		private static Version AspNetVersion => typeof(HttpRuntime).Assembly.GetName().Version;
-		private static string ClrDescription => RuntimeInformation.FrameworkDescription;
+		private static string ClrDescription => "aaa";  // RuntimeInformation.FrameworkDescription;
```
gregkalapos: Fix is already on @SergeyKleyman's machine.

v1v:
As I want to validate whether the GH notification checks are fine, I'll say `jenkins run the tests`

gregkalapos:
Merging this. 

❗️ Very important ❗️ 
After this we should only release packages built on Windows! Building on non-Windows won't generate the `net461` package.

gregkalapos:
Merging this. 

❗️ Very important ❗️ 
After this we should only release packages built on Windows! Building on non-Windows won't generate the `net461` package.

Issue #234, opened 2019-05-22
Rename UseElasticApm extension methods


gregkalapos:
From elastic/apm-agent-dotnet#230 

Currently we have 2 `UseElasticApm` methods and this is very confusing. There seems to be an agreement to rename those to better reflect their purposes.

This [comment describes](https://github.com/elastic/apm-agent-dotnet/issues/230#issuecomment-493554507) what those do.

Suggestions (from @SergeyKleyman 's [comment](https://github.com/elastic/apm-agent-dotnet/issues/230#issuecomment-494722757))
- `Elastic.Apm.AspNetCore.UseElasticApm` -> `Elastic.Apm.AspNetCore.UseElasticApmBase`
- `Elastic.Apm.All.UseElasticApm` -> `Elastic.Apm.All.UseAllElasticApm `

gregkalapos:
By looking at this I think the 2. one is a good name, but the first one (`Elastic.Apm.AspNetCore.UseElasticApmBase`) isn't really telling the truth (which was i guess my suggestion 😅 ) - That actually only registers the ASP.NET Core middleware. 

How about:
- `Elastic.Apm.AspNetCore.UseElasticApm` -> `Elastic.Apm.AspNetCore.UseElasticAspNetCoreMiddleware`
- `Elastic.Apm.All.UseElasticApm` -> `Elastic.Apm.All.UseAllElasticApm`

@SergeyKleyman?

gregkalapos:
Maybe not using the word `Middleware` in the method name would be better, because that is an implementation detail, and it can be changed (e.g. all we do with the middleware, also can be done with DiagnosticSource), and also we in fact register 1 diagnostic source listener for error capturing. 

So alternative:
- `Elastic.Apm.AspNetCore.UseElasticApm` -> `Elastic.Apm.AspNetCore.UseElasticAspNetCore` (or `Elastic.Apm.AspNetCore.UseElasticAspNetCoreMonitoring`
- `Elastic.Apm.All.UseElasticApm` -> `Elastic.Apm.All.UseAllElasticApm`

SergeyKleyman:
I would go with the following to stress that it's about ASP.NET Core APM and the difference is the set of features enabled (Basic vs All):

- `Elastic.Apm.AspNetCore.UseElasticApm` -> `Elastic.Apm.AspNetCore.UseBasicElasticAspNetCoreApm`
[...]

PR #296, opened 2019-06-18
Add override for AspNetCore UseElasticApm that enables HttpDiagnosticsSubscriber by default

lafriks:
It does not add additional dependencies so I don't see a reason not to enable it by default

### DIFF
--- a/src/Elastic.Apm.AspNetCore/ApmMiddlewareExtension.cs
+++ b/src/Elastic.Apm.AspNetCore/ApmMiddlewareExtension.cs
@@ -16,6 +16,22 @@ namespace Elastic.Apm.AspNetCore
 	public static class ApmMiddlewareExtension
 	{
 		/// <summary>
+		/// Adds the Elastic APM Middleware to the ASP.NET Core pipeline and enables <see cref="HttpDiagnosticsSubscriber" />.
+		/// </summary>
+		/// <returns>The elastic apm.</returns>
+		/// <param name="builder">Builder.</param>
+		/// <param name="configuration">
+		/// You can optionally pass the IConfiguration of your application to the Elastic APM Agent. By
+		/// doing this the agent will read agent related configurations through this IConfiguration instance.
+		/// If no <see cref="IConfiguration" /> is passed to the agent then it will read configs from environment variables.
+		/// </param>
+		public static IApplicationBuilder UseElasticApm(
+			this IApplicationBuilder builder,
+			IConfiguration configuration = null
+		) =>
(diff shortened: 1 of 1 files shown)


gregkalapos:
Thanks for the contrib @lafriks 

> It does not add additional dependencies so I don't see a reason not to enable it by default

The basic idea was that in `Elastic.Apm.AspNetCore` we only enable ASP.NET Core tracing and in the `Elastic.Apm.All` we enable everything - so unless you use `.All` you enable things manually.

My only fear with this change is that from this point writing `app.UseElasticApm(Configuration);` would mean that you enable outgoing HTTP tracing, but don't enable let's say database tracing. So the agent will track some things (`HttpClient`), but not others (`DbContext`) which I think can be very confusing and also hard to explain in the docs. The current rule is easy: `.All` enables everything, otherwise you enable things manually.

Another option would be to use a different method name like `UseElasticApmWithHttpTracing` (or something like that).

I'm kind of undecided on this, happy to hear other opinions.

Also, before merging we'll need some automated tests making sure the new method really does what it promises to do.

gregkalapos:
Some other thoughts:

I think the default case is that a user wants to turn on everything and monitor everything in an application. We can't offer this from the `Elastic.Apm.AspNetCore` project, because that'd require us to pull in additional dependencies which for some use-cases (like very small docker images with minimal dependencies) can be a problem.

Therefore we offer the `Elastic.Apm.All` package, which enables everything (`HttpClient`, EFCore, etc). That's our default way to enable the agent. Other packages are only recommended to use directly in case you want to minimize your dependencies.

On this note we realize that naming the `UseElasticApm` method on `Elastic.Apm.All` and in `Elastic.Apm.AspNetCore` the same is probably not the right decision. We address that in https://github.com/elastic/apm-agent-dotnet/issues/234 - we plan to have better names to make clear that 1 only enables ASP.NET Core traces, while the other enables everything.

Would having better names as discussed in #234 make the situation more clear? Instead of adding another overload as suggested here.

lafriks:
Renaming would definitely make it better as currently when you have used Elastic.Apm.All for one project and now you try to enable apm for other that does not use EF it took me some time to figure out why http client requests are not logged :)

lafriks:
I will close this PR in favor to be addressed in #234 but it would be great to have two methods, one with nothing enabled and other with everything that can be enabled without adding additional dependencies

gregkalapos:
Sounds good, nevertheless thanks for bringing this up @lafriks - another good reason to make a change in this area.

PR #309, opened 2019-06-20
Refactor unit tests to use a custom WebApplicationFactory

NeoXtreem:
Later package versions (not yet upgraded to) have stricter routing in the unit tests, so for `AspNetFullFrameworkSampleApp` (renamed from `SampleAspNetFullFrameworkApp` for better consistency), the unit tests have been refactored to use `CustomWebApplicationFactory` which allows a custom (fake) `Startup` to be used for configuring test startup.

This config is encapsulated in `StartupConfigService` which is instantiated by the unit tests (using `TestHelper` static methods).  The fake `Startup` classes derive from the real ones overriding only what needs to be configured for the unit tests.  The custom factory calls on its `WithWebHostBuilder` the `ConfigureTestServices` method to add `StartupConfigService` to the DI container so that this can be retrieved by the fake `Startup` for special configuration in the `ConfigureAgent` override method.  `UseSolutionRelativeContentRoot` is also called here to specify the root path for the views.

The above methodology follows [this article](https://dzone.com/articles/using-a-custom-startup-class-with-aspnet-core-inte) which is based on Microsoft documentation on [the Startup class](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/startup?view=aspnetcore-2.2) and [integration tests in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-2.2).

Overall, this refactor results in much cleaner and leaner unit test code without boiler-plate bootstrapping the test startup configuration.  For example, `TestHelper` (renamed from `Helper`) is now much leaner and simpler.

It should be noted that unit test `SubscribeAndUnsubscribeAspNetCoreDiagnosticListener` is still semi-disabled, but it succeeds when run indepedently.  A `TODO` note has been placed in the unit test to this effect.

These changes were originally from a feature branch which has since been deleted and PR #226 (with some historical discussion) closed.


NeoXtreem:
@gregkalapos Could you please look at this PR as soon as possible so that the unit test projects are cleaner and therefore easier to work with?  I merged the Full Framework test code into my fork recently, but this constant merging and conflict-resolving to keep up with the base branch is becoming time-consuming, and issues with the test code such as #315 will be more preventable or easier to resolve if this PR is approved.

NeoXtreem:
@gregkalapos Could you please look at this PR as soon as possible so that the unit test projects are cleaner and therefore easier to work with? I merged the Full Framework test code into my fork recently, but this constant merging and conflict-resolving to keep up with the base branch is becoming time-consuming, and issues with the test code such as #315 will be more preventable or easier to resolve if this PR is approved.

src/Elastic.Apm.AspNetCore/Elastic.Apm.AspNetCore.csproj
```diff
@@ -2,16 +2,13 @@
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
     <AssemblyName>Elastic.Apm.AspNetCore</AssemblyName>
-    <AssemblyVersion>1.0.0</AssemblyVersion>
```
gregkalapos: We should not have these changes - I think most of this is not up-to-date. I'd suggest just leave the `.csproj` files as they are - we have the current version on `master` and we'll increase those once we release.

src/Elastic.Apm.AspNetFullFramework/ApplicationConfigurationReader.cs
```diff
@@ -7,9 +9,11 @@
 
 namespace Elastic.Apm.AspNetFullFramework
 {
-	internal class FullFrameworkConfigReader : EnvironmentConfigurationReader
+	internal class ApplicationConfigurationReader : EnvironmentConfigurationReader
```
gregkalapos: I also don't really see why we should rename this.

src/Elastic.Apm.AspNetFullFramework/ApplicationConfigurationReader.cs
```diff
@@ -22,8 +26,10 @@ private string DiscoverFullFrameworkServiceName()
 
 			void AppendIfNotNull(string nameToAppend)
 			{
-				const string separator = "_";
-				if (nameToAppend != null) retVal.AppendSeparatedIfNotEmpty(separator, nameToAppend);
+				if (nameToAppend != null)
+				{
```
gregkalapos: Please no braces for 1 liners.

src/Elastic.Apm.AspNetFullFramework/ApplicationConfigurationReader.cs
```diff
@@ -33,7 +39,7 @@ private string FindAppPoolName()
 			{
 				return Environment.GetEnvironmentVariable("APP_POOL_ID");
 			}
-			catch (Exception ex)
+			catch (SecurityException ex)
```
gregkalapos: Why is this necessary? I think the catch block is a very generic error handler - even if the current `Environment.GetEnvironmentVariable` only throws `SecurityException` I don't see why it's bad to have a general error handler for every future case.

src/Elastic.Apm.AspNetFullFramework/ApplicationConfigurationReader.cs
```diff
@@ -46,11 +52,17 @@ private string FindSiteName()
 			{
 				return HostingEnvironment.SiteName;
 			}
-			catch (Exception ex)
+			catch (SecurityException ex)
```
gregkalapos: Same as above

src/Elastic.Apm.AspNetFullFramework/Elastic.Apm.AspNetFullFramework.csproj
```diff
@@ -1,62 +1,29 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
gregkalapos: Not sure we want to change this to the new format - not pushing back on this one, but isn't this unrelated?

src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
```diff
@@ -57,22 +58,14 @@ private static void SetServiceInformation(Service service, string aspNetVersion)
...
-		public void Dispose()
-		{
-			if (_httpApp != null)
-			{
-				_httpApp.BeginRequest -= OnBeginRequest;
-				_httpApp.EndRequest -= OnEndRequest;
-				_httpApp = null;
-			}
-		}
+		public void Dispose() => _context = null;
```
gregkalapos: Shouldn't we unsubscribe from the `BeginRequest` and `EndRequest` events?

src/Elastic.Apm.EntityFrameworkCore/Elastic.Apm.EntityFrameworkCore.csproj
```diff
@@ -1,17 +1,14 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
```
gregkalapos: Same as to other `.csproj` files - please don't change these.

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
```diff
@@ -44,27 +43,18 @@ protected internal static bool TryParseLogLevel(string value, out LogLevel? leve
...
-			return true;
-		}
+		protected bool ParseCaptureHeaders(ConfigurationKeyValue kv) => kv == null || string.IsNullOrEmpty(kv.Value) || !bool.TryParse(kv.Value, out var value) || value;
 
 		protected LogLevel ParseLogLevel(ConfigurationKeyValue kv)
 		{
 			if (TryParseLogLevel(kv?.Value, out var level)) return level.Value;
 
 			if (kv?.Value == null)
+			{
```
gregkalapos: Please remove braces.

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
```diff
@@ -114,11 +104,7 @@ bool TryParseUri(string u, out Uri uri)
 
 		protected double ParseMetricsInterval(ConfigurationKeyValue kv)
 		{
-			string value;
-			if (kv == null || string.IsNullOrWhiteSpace(kv.Value))
-				value = ConfigConsts.DefaultValues.MetricsInterval;
-			else
-				value = kv.Value;
+			var value = kv == null || string.IsNullOrWhiteSpace(kv.Value) ? ConfigConsts.DefaultValues.MetricsInterval : kv.Value;
```
gregkalapos: Was this a suggestion based on `.editorconfig`? To me the original one was way more readable, but this is just a nit.

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
```diff
@@ -156,7 +141,7 @@ protected double ParseMetricsInterval(ConfigurationKeyValue kv)
 			return valueInMilliseconds;
 		}
 
-		private bool TryParseTimeInterval(string valueAsString, out double valueInMilliseconds)
+		private static bool TryParseTimeInterval(string valueAsString, out double valueInMilliseconds)
```
gregkalapos: I see you sometimes adding static (I think mostly because of our `.editorconfig` prefers those) but here you remove it - why? Can't follow and these are so many unrelated changes.

src/Elastic.Apm/Elastic.Apm.csproj
```diff
@@ -8,30 +8,28 @@
...
     <PackageTags>apm, monitoring, elastic, elasticapm, analytics, sdk</PackageTags>
     <LangVersion>latest</LangVersion>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
   </PropertyGroup>
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="System.Diagnostics.PerformanceCounter" Version="4.5.0" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
-    <!-- TODO: Are we ok with this reference/version? -->
```
gregkalapos: Why do you remove the todo?

src/Elastic.Apm/Elastic.Apm.csproj
```diff
@@ -8,30 +8,28 @@
   <PropertyGroup>
     <RootNamespace>Elastic.Apm</RootNamespace>
     <AssemblyName>Elastic.Apm</AssemblyName>
-    <AssemblyVersion>1.0.0</AssemblyVersion>
```
gregkalapos: Same as to other `.csproj` files - please don't change these.

src/Elastic.Apm/Elastic.Apm.csproj
```diff
@@ -8,30 +8,28 @@
...
+    <VersionPrefix>1.0.1</VersionPrefix>
+    <VersionSuffix>beta1</VersionSuffix>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\..\build\elasticapm.snk</AssemblyOriginatorKeyFile>
     <DelaySign>false</DelaySign>
-    <PackageVersion>1.0.0-beta1</PackageVersion>
     <Authors>Elastic and contributors</Authors>
     <PackageId>Elastic.Apm</PackageId>
-    <copyright>2019 Elasticsearch BV</copyright>
+    <Copyright>2019 Elasticsearch BV</Copyright>
```
gregkalapos: 👍 This is ok, nice catch.

src/Elastic.Apm/Elastic.Apm.csproj
```diff
@@ -8,30 +8,28 @@
...
     <PackageLicenseUrl>https://github.com/elastic/apm-agent-dotnet/blob/master/LICENSE</PackageLicenseUrl>
     <PackageProjectUrl>https://github.com/elastic/apm-agent-dotnet</PackageProjectUrl>
     <PackageIconUrl>https://raw.githubusercontent.com/elastic/apm-agent-dotnet/master/build/nuget-icon.png</PackageIconUrl>
     <Description>Elastic APM .NET Agent base package. This package provides core functionality for transmitting of all Elastic APM types and is a dependent package for all other Elastic APM package. Additionally this package contains the public Agent API that allows you to manually capture transactions and spans. Please install the platform specific package for the best experience. See: https://github.com/elastic/apm-agent-dotnet/tree/master/docs</Description>
     <PackageTags>apm, monitoring, elastic, elasticapm, analytics, sdk</PackageTags>
     <LangVersion>latest</LangVersion>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
```
gregkalapos: We don't plan to have this, at least I don't know about it - also that's not related to the subject of the PR either.

src/Elastic.Apm.All/Elastic.Apm.All.csproj
```diff
@@ -1,17 +1,14 @@
-<Project Sdk="Microsoft.NET.Sdk">
```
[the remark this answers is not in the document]
NeoXtreem: It's not out of date - I've kept it updated from the base branch. If you're referring to the change in use of version properties, this is correct (good info [here](https://andrewlock.net/version-vs-versionsuffix-vs-packageversion-what-do-they-all-mean/)). `AssemblyVersion` defaults to `Version` without any prerelease label. `FileVersion` defaults to `AssemblyVersion`. `InformationalVersion` defaults to `Version` with the prerelease label. The end result is the same, but this is the proper way to do it if you don't require specifically different values for the removed version properties other than the defaults (which isn't the case in this case).

src/Elastic.Apm.AspNetCore/ApmMiddlewareExtension.cs
```diff
@@ -67,8 +56,11 @@ internal static void UpdateServiceInformation(Service service)
 			string version;
 			var versionQuery = AppDomain.CurrentDomain.GetAssemblies().Where(n => n.GetName().Name == "Microsoft.AspNetCore");
 			var assemblies = versionQuery as Assembly[] ?? versionQuery.ToArray();
+
 			if (assemblies.Any())
+			{
```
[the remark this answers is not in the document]
NeoXtreem: OK, that rule (if a repo has it) normally applies where there is no `else` statement that has a statement body. But if that's the requirement, I've removed it.

src/Elastic.Apm.AspNetCore/ApplicationConfigurationReader.cs
```diff
@@ -4,15 +4,16 @@
 using Elastic.Apm.Logging;
 using Microsoft.Extensions.Configuration;
 
-namespace Elastic.Apm.AspNetCore.Config
```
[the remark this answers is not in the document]
NeoXtreem: `Elastic.Apm.AspNetCore` and `Elastic.Apm.AspNetFullFramework` are corresponding projects, one for .NET Core, one for Full Framework. Therefore, any corresponding classes should follow a similar pattern.

- `MicrosoftExtensionsConfig` extends `AbstractConfigurationReader` and realises `IConfigurationReader`.
- `FullFrameworkConfigReader` extends `EnvironmentConfigurationReader`.

These are therefore corresponding classes between the two projects. It made sense to put both in the same `Config` subfolder in their respective projects, or to put both at root level. Since there is only one class in each project, and the class names already state they are config-related, it made sense to follow the same pattern of `FullFrameworkConfigReader` and put `MicrosoftExtensionsConfig` at root level. If you want both in a `Config` folder, then that's doable, but then `FullFrameworkConfigReader` would have to be moved to that namespace instead in order to keep alignment.

**Advantage:** It's less confusing for the reader, and they can see the correspondence between the two projects.

src/Elastic.Apm.AspNetCore/ApplicationConfigurationReader.cs
```diff
@@ -4,15 +4,16 @@
...
+namespace Elastic.Apm.AspNetCore
 {
 	/// <summary>
 	/// An agent-config provider based on Microsoft.Extensions.Configuration.IConfiguration.
-	/// It uses environment variables as fallback
+	/// It uses environment variables as fallback.
 	/// </summary>
-	internal class MicrosoftExtensionsConfig : AbstractConfigurationReader, IConfigurationReader
+	internal class ApplicationConfigurationReader : AbstractConfigurationReader, IConfigurationReader
```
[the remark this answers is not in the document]
NeoXtreem: `Elastic.Apm.AspNetCore` and `Elastic.Apm.AspNetFullFramework` are corresponding projects, one for .NET Core, one for Full Framework. Therefore, any corresponding classes should follow a similar pattern.

- `MicrosoftExtensionsConfig` extends `AbstractConfigurationReader` and realises `IConfigurationReader`.
- `FullFrameworkConfigReader` extends `EnvironmentConfigurationReader`.

Conventionally, derived classes include all or part of the name of the base class. Most classes in Elastic APM Agent follow this pattern. `FullFrameworkConfigReader` _almost_ follows that pattern but abbreviates `Configuration`. `MicrosoftExtensionsConfig` abbreviates `Configuration` _and_ drops `Reader`. This is incorrect because the class is a reader (a config reader specifically) and is not config. That is also implied by its base class. The class _extends_ the base class.

The fact that it uses `IConfiguration` reader, which simply happens to be in the `Microsoft.Extensions` namespace, should not affect the class name. This is implementation. The class name should reflect what the class _does_, not how it is _implemented_. It is highly unusual to name any class based on the namespace of other classes or interfaces it inherits. It is more usual to name it based on the inherited class or interface name itself.

**Advantage:** The reader can understand what the class does rather than seeing some obscure namespace embedded into the class name.

In any case, I've reverted the change and will put it in a separate PR.

src/Elastic.Apm.AspNetCore/Elastic.Apm.AspNetCore.csproj
NeoXtreem: It's not out of date - I've kept it updated from the base branch. If you're referring to the change in use of version properties, this is correct (good info [here](https://andrewlock.net/version-vs-versionsuffix-vs-packageversion-what-do-they-all-mean/)). `AssemblyVersion` defaults to `Version` without any prerelease label. `FileVersion` defaults to `AssemblyVersion`. `InformationalVersion` defaults to `Version` with the prerelease label. The end result is the same, but this is the proper way to do it if you don't require specifically different values for the removed version properties other than the defaults (which isn't the case in this case).

However, I'll ensure the version is aligned with _master_ on the base branch in any case.

gregkalapos:
@NeoXtreem I really appreciate your work, but is it for you completely out of discussion that you try things in smaller PRs? It would make our life much easier, and the chance that your PRs get merged quickly would be way higher.

Look at other community PRs, maybe those show what I mean.

src/Elastic.Apm.AspNetFullFramework/ApplicationConfigurationReader.cs
NeoXtreem: `Elastic.Apm.AspNetCore` and `Elastic.Apm.AspNetFullFramework` are corresponding projects, one for .NET Core, one for Full Framework. Therefore, any corresponding classes should follow a similar pattern.

- `MicrosoftExtensionsConfig` extends `AbstractConfigurationReader` and realises `IConfigurationReader`.
- `FullFrameworkConfigReader` extends `EnvironmentConfigurationReader`.

Since `MicrosoftExtensionsConfig` was renamed more appropriately as explained in [my reply](https://github.com/elastic/apm-agent-dotnet/pull/309/files#r300779132) to your other comment, `FullFrameworkConfigReader` should follow a similar pattern. Including `FullFramework` in the name makes little sense as it is already in the Full Framework project and therefore in the `AspNetFullFramework` namespace. Avoid repetition in namespaces and class names ensures DRY is followed. It is therefore renamed the same as `MicrosoftExtensionsConfig`. Both are `ApplicationConfigurationReader` as this indicates what they do in their respective projects. They'll never conflict because they're in independent projects with different namespaces.

**Advantage:** The reader can understand what the class does and will be able to see the correspondence between the .NET Core and Full Framework projects much more easily.

In any case, I've reverted the change and will put it in a separate PR.

src/Elastic.Apm.AspNetFullFramework/ApplicationConfigurationReader.cs
NeoXtreem: Catch-all is very bad practice and generally avoided and recommended against almost ubiquitously, and for very good reason. If you are unaware of the exceptions that can be expected, then this is a bad situation to be in because you are implemented generic handling of unexpected exceptions. You are treating all possible unexpected exceptions as a security exception would be treated. This can result in silent errors, or bad object states.

Unit tests should cover all possible exceptions, so if any future code is added that could generate an exception, this should be caught in unit tests and then handled accordingly. Adding a catch-all here is potentially dangerous, bad practice, and in relation to your reply, breaks YAGNI.

src/Elastic.Apm.AspNetFullFramework/ApplicationConfigurationReader.cs
NeoXtreem: Same reply as above.

src/Elastic.Apm.AspNetFullFramework/Elastic.Apm.AspNetFullFramework.csproj
NeoXtreem: It made it much easier for me to resolve the reference issues I was having as the new format means that Visual Studio and NuGet takes care of everything default instead of having to explicitly state everything in the old format. It also leans up the code and there is no disadvantage to this.

src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
NeoXtreem: No. You're using the `HttpApplication` initialization pipeline to bind to them (via `Init`). `Dispose` is outside of the scope `HttpApplication`, so unsubscribing from them in this `Dispose` method is dangerous as it can cause them to be called after being unsubscribed from. This caused crashes on our side. Allow the framework to take care of them being unsubscribed. You might find this [SO question](https://stackoverflow.com/q/2781346/197591) useful.

src/Elastic.Apm.EntityFrameworkCore/Elastic.Apm.EntityFrameworkCore.csproj
NeoXtreem: If you're referring to the change in use of version properties, this is correct (good info [here](https://andrewlock.net/version-vs-versionsuffix-vs-packageversion-what-do-they-all-mean/)). `AssemblyVersion` defaults to `Version` without any prerelease label. `FileVersion` defaults to `AssemblyVersion`. `InformationalVersion` defaults to `Version` with the prerelease label. The end result is the same, but this is the proper way to do it if you don't require specifically different values for the removed version properties other than the defaults (which isn't the case in this case).

However, I'll ensure the version is aligned with _master_ on the base branch in any case.

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
NeoXtreem: It's a VS or ReSharper suggestion. It's better because you're only assigning `value` once in the code. It's more robust. The operands of the ternary operator could go on separate lines, but I don't see how this would make it more readable for such short code.

src/Elastic.Apm/Elastic.Apm.csproj
NeoXtreem: I investigated the references in-depth in relation to System.Diagnostics.DiagnosticSource` and upgraded that and also `Newtonsoft.Json` when I determined it was both better and safe to do so. In conclusion, I deemed the new reference versions to be OK, and therefore the TODO no longer necessary.

src/Elastic.Apm/Elastic.Apm.csproj
NeoXtreem: If you're referring to the change in use of version properties, this is correct (good info [here](https://andrewlock.net/version-vs-versionsuffix-vs-packageversion-what-do-they-all-mean/)). `AssemblyVersion` defaults to `Version` without any prerelease label. `FileVersion` defaults to `AssemblyVersion`. `InformationalVersion` defaults to `Version` with the prerelease label. The end result is the same, but this is the proper way to do it if you don't require specifically different values for the removed version properties other than the defaults (which isn't the case in this case).

However, I'll ensure the version is aligned with _master_ on the base branch in any case.

src/Elastic.Apm/Elastic.Apm.csproj
NeoXtreem: Sure, this is for our purposes as we need to generate the package for our own NuGet repo. I'll remove it from the PR.

NeoXtreem:
> @NeoXtreem I really appreciate your work, but is it for you completely out of discussion that you try things in smaller PRs? It would make our life much easier, and the chance that your PRs get merged quickly would be way higher.
> 
> Look at other community PRs, maybe those show what I mean.

I'm splitting the config reader changes to another PR as this is definitely outside the scope of unit tests. The unit tests refactor however is a big change in itself that would be far more difficult to split up. However, the unit tests are still in place and test the same code. It's more how they're setup that has changed, and this is low risk. The functional code of course hasn't changed much.

PR #311, opened 2019-06-21
Fix ASP.NET version that agent reports on Full Framework [v2]

SergeyKleyman:
Fixes elastic/apm-agent-dotnet#287

PR #314, opened 2019-06-21
Implement service.runtime.{name and version}

### LINKED ISSUE #199: Agent should send service.runtime.{name and version}
`service` object in intake API includes `runtime` (with `name` and `version`) - 
https://github.com/elastic/apm-server/blob/master/docs/spec/service.json
.NET agent currently doesn't capture (and thus doesn't send) that data.

SergeyKleyman:
Closes elastic/apm-agent-dotnet#199


SergeyKleyman:
This PR builds on top of some changes in #311 so it should be merged after #311.

gregkalapos:
What do we aim to report here? I think if we would like to report the version and the name of the runtime then it should be the Common Language Runtime (CLR) version and name.

So in case of .NET Core it should be "CoreCLR" and the corresponding version and in case of .NET Framework it'd be something like "Clr" plus the version.

Currently the PR reports those as ".NET Core" and ".NET Framework". Are we sure this is what we want? 

So what I'd say:
- `Framework` should be the high level entry framework (like we already do - ASP.NET / Core) - alternatively this could be changed to report the .NET / .NET Core Framework name/version, but I think that'd be less useful.
- `Runtime` would be the CLR (CoreCLR vs. CLR, maybe later Mono, especially with .NET 5).

With this approach we don't report the .NET Framework version, which is I think fine.

Would this make sense?

SergeyKleyman:
My main motivation to use `.NET Framework` and `.NET Core` prefixes was the fact that  `RuntimeInformation.FrameworkDescription` reports it in this way so I assumed it would be natural for .NET developers to see it in the same way.

gregkalapos:
> My main motivation to use `.NET Framework` and `.NET Core` prefixes was the fact that `RuntimeInformation.FrameworkDescription` reports it in this way so I assumed it would be natural for .NET developers to see it in the same way.

Yeah, but that's the framework and not the runtime. We report this as `runtime` which sounds strange. I'd assume that's the CLR.

gregkalapos:
Some other thoughts on reporting runtime (aka CLR) vs. reporting framework (aka .NET/Core Framework) in the `runtime` field:

So… for .NET Core I think this is at the moment the same. 

But for .NET Framework it’s not the same… the Framework version is like 4.6.x and 4.7.x, but the runtime itself, the CLR, that is still 4.0.x… there are 2 main runtime versions for Full Framework, which are the CLR 2 and CLR 4. And reporting the .NET Framework version as runtime feels wrong to me. If we call it `runtime` then it’s the CLR.

Also for future (although my stronger argument is the previous one) - in .NET 5 MS wants to make the runtime interchangeable… you can use either CoreClr or Mono. I think that again would be a nice use case for reporting this as `runtime`, but that is not the .NET Core framework name/version, but the runtime.

SergeyKleyman:
It's`runtime in the sense that it reports version of actual runtime on top of which the code is currently running (not SDK that used to build or target platform).

gregkalapos:
> It's`runtime in the sense that it reports version of actual runtime on top of which the code is currently running (not SDK that used to build or target platform).

Not in case of .NET Framework (see my previous comment).

SergeyKleyman:
If I understand you correctly you are saying that CLR == runtime with which I don't agree. Runtime is whatever people need to install to run their applications. In case of .NET Framework it's the whole framework and not just CLR.

SergeyKleyman:
For example if the difference between .NET Framework 4.7.1 and 4.7.2 is in some BCL classes and no changes in CLR the version of runtime should be 4.7.2.

SergeyKleyman:
For example https://blogs.msdn.microsoft.com/karinm/2008/11/10/whats-the-difference-between-clr-and-net-framework/ both .NET Framework 2.0 and 3.0 had CLR version 2.0. I would argue that reporting runtime as `.NET Framework 3.0` is more useful than as `CLR 2.0` as it allows to distinguish between .NET Framework 2.0 and 3.0

gregkalapos:
> For example https://blogs.msdn.microsoft.com/karinm/2008/11/10/whats-the-difference-between-clr-and-net-framework/ both .NET Framework 2.0 and 3.0 had CLR version 2.0. I would argue that reporting runtime as `.NET Framework 3.0` is more useful than as `CLR 2.0` as it allows to distinguish between .NET Framework 2.0 and 3.0

I also think that this information is more useful, but then putting it into the `runtime` field is I think very misleading. 

Altenative idea:

- we change `framework`, and we put the .NET/Core Framework version there
- we put the CLR version into the `runtime`.

src/Elastic.Apm/Api/Service.cs
```diff
@@ -1,6 +1,11 @@
-﻿using System.Reflection;
+﻿using System;
+using System.Diagnostics.SymbolStore;
```
gregkalapos: nit: this is unused.

src/Elastic.Apm/Api/Service.cs
```diff
@@ -14,22 +19,122 @@ public class Service
...
 				{
 					Name = Consts.AgentName,
 					Version = typeof(Agent).Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion
-				}
+				},
+				Runtime = GetRuntime(logger)
 			};
+		}
+
+		private static Runtime GetRuntime(IApmLogger logger)
```
gregkalapos: Not so important, but this has quite some internal logic, but it's in the `Api` namespace. We have e.g. the `Helpers` folder, I feel that it'd be a better place for this code. If you don't wanna move it, feel free to leave it here, this is just my 2 cents.

SergeyKleyman: I agree it's better be in some internal helper - I moved it there.

src/Elastic.Apm/Api/Service.cs
SergeyKleyman: Done

SergeyKleyman: > I think this is ok... we have an ongoing discussion about .NET version vs. CLR version as runtime, but let's not block this PR because of that. We can reiterate on that.
> 
> I'd suggest not closing #199 and continue the discussion there, plus as agreed this will be also a topic on our next .NET sync.
> 
> Thanks @SergeyKleyman.

I prefer to close #199 as done since this PR does implement what is required in #199. I'll open a follow up to revisit per our discussion here. WDTY?

gregkalapos: 👍

SergeyKleyman: I've opened #317 to follow up.

PR #328, opened 2019-07-01
Metrics docs

gregkalapos:
- Mainly stolen from Java (and of course adapted)
- Documents the `MetricsInterval` setting
- And documents all metrics that the agent currently sends

### DIFF
--- a/docs/metrics.asciidoc
+++ b/docs/metrics.asciidoc
@@ -1,0 +1,102 @@
+ifdef::env-github[]
+NOTE: For the best reading experience,
+please view this documentation at https://www.elastic.co/guide/en/apm/agent/java[elastic.co]
+endif::[]
+
+[[metrics]]
+== Metrics
+
+The .NET agent tracks certain system and application metrics.
+Some of them have built-in visualizations and some can only be visualized with custom Kibana dashboards.
+
+These metrics will be sent regularly to the APM Server and from there to Elasticsearch.
+You can adjust the interval with the setting <<config-metrics-interval>>.
+
+The metrics will be stored in the `apm-*` index and have the `processor.event` property set to `metric`.
+
+"Platform: all" means that the metric is available on every platform where .NET Core is supported.
+
+[float]
+[[metrics-system]]
+=== System metrics
+
+As of APM version 6.6, these metrics will be visualized in the APM UI.
+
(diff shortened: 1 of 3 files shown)


docs/metrics.asciidoc
```diff
@@ -0,0 +1,102 @@
+ifdef::env-github[]
```
bmorelli25: This file needs to be included in `index.asciidoc` for it to build:
```js
include::./metrics.asciidoc[Metrics]
```

gregkalapos: Thanks, added - trying to have the same order as other agents, let me know if I should move it.

bmorelli25: That's good for now. Thanks!

docs/metrics.asciidoc
```diff
@@ -0,0 +1,102 @@
...
+NOTE: For the best reading experience,
+please view this documentation at https://www.elastic.co/guide/en/apm/agent/java[elastic.co]
```
SergeyKleyman: `https://www.elastic.co/guide/en/apm/agent/java` should probably be `https://www.elastic.co/guide/en/apm/agent/dotnet`, right? :)

docs/metrics.asciidoc
```diff
@@ -0,0 +1,102 @@
...
+endif::[]
+
+[[metrics]]
+== Metrics
+
+The .NET agent tracks certain system and application metrics.
+Some of them have built-in visualizations and some can only be visualized with custom Kibana dashboards.
+
+These metrics will be sent regularly to the APM Server and from there to Elasticsearch.
+You can adjust the interval with the setting <<config-metrics-interval>>.
```
SergeyKleyman: `config-metrics-interval` link is not in this doc but in `configuration.asciidoc`

docs/metrics.asciidoc
```diff
@@ -0,0 +1,102 @@
...
+*`system.memory.actual.free`*::
++
+--
+type: long
+
+format: bytes
+
+Platform: Windows and Linux only.
+
+Actual free memory in bytes.
```
SergeyKleyman: There is already `format: bytes` above and this is the only memory related metric description that reiterates that the units are bytes. I would just leave `Actual free memory`  for consistency.

PR #331, opened 2019-07-01
Fix Error timestamp after Span end bug

### LINKED ISSUE #330: Error event for exception thrown during a span has timestamp which is later than the span's end
I encountered this bug when `Elastic.Apm.AspNetFullFramework.Tests.ErrorsTests.CustomSpanThrowsTest` failed sporadically (BTW it happened only when tests were executed inside Windows docker container - I'm still not sure why...). I've added a few things to the log to trace what is going on (I removed parts that are not relevant to this issue):
```
2019-07-01 07:09:26.179 | New Span instance created: Span{..., Start time: 2019-07-01 04:09:26.179856 +00:00 (as timestamp: 1561954166179000)

2019-07-01 07:09:26.195 | New Error instance created: Error{..., Start time: 2019-07-01 04:09:26.195481 +00:00 (as timestamp: 1561954166195000)

2019-07-01 07:09:26.195 | Ended Span{..., Start time: 2019-07-01 04:09:26.179856 +00:00 (as timestamp: 1561954166179000), End time: 2019-07-01 04:09:26.195481 +00:00 (as timestamp: 1561954166195000), Duration: 15.6252ms, DurationUsingTimestamps: 16ms
```

We can see that the root cause is rounding error - we use rounded time for timestamp (span started at `2019-07-01 04:09:26.179856` but timestamp for span start is `1561954166179000` - `856` microseconds were discarded) but when we calculate duration we use original (not rounded) time points thus creating discrepancy - duration from span's start to error is 16ms while span's duration 15.6252ms which is impossible since exception was thrown before span ended.

SergeyKleyman:
Fixes #330.
The fix takes the approach of using timestamps (that we report) to calculate duration as well.


src/Elastic.Apm/Model/Span.cs
```diff
@@ -31,15 +29,16 @@ internal class Span : ISpan
 			IApmLogger logger
 		)
 		{
-			_start = DateTimeOffset.UtcNow;
+			Timestamp = TimeUtils.TimestampNow();
+			Id = RandomGenerator.GenerateRandomBytesAsString(new byte[8]);
+			_logger = logger?.Scoped($"{nameof(Span)}.{Id}");
```
gregkalapos: Is it a good idea to include the span id here? That's also logged as part of `Span.ToString()`, which I think we use everywhere:

```
[2019-07-03 02:33:43][Debug] - {Span.14aa6ce329bddc53} End() called on already ended Span{Id: 14aa6ce329bddc53, TransactionId: a857e56e2d6c1d78, ParentId: a857e56e2d6c1d78, TraceId: 8e063cef1f687d400c50db874609cdce, Name: PRAGMA foreign_keys=ON;, Type: db, IsSampled: True}. Start time: 2019-07-03 14:33:43.9529890 UTC (as timestamp: 1562164423952989), Duration: 3.0991ms
```
I think with this it'll be harder to filter only span related logs.

src/Elastic.Apm/Model/Span.cs
```diff
@@ -135,14 +141,25 @@ internal Span StartSpanInternal(string name, string type, string subType = null,
...
 			return retVal;
 		}
 
 		public void End()
 		{
-			_logger.Debug()?.Log("Ending {SpanDetails}", ToString());
-			if (!Duration.HasValue) Duration = (DateTimeOffset.UtcNow - _start).TotalMilliseconds;
+			if (Duration.HasValue)
+			{
+				_logger.Debug()?.Log("End() called on already ended {Span}. Start time: {Time} (as timestamp: {Timestamp}), Duration: {Duration}ms",
```
gregkalapos: I think this log sounds misleading... it sounds like you called `End()` already once, which is not the case. This happens when you manually set the duration of the span. If I remember correctly EF Core gives us durations, so we use those instead of calculating ourself. I'd say something like "End() called and duration was already passed to the span manually - no duration calculation. Start time: {Time} (as timestamp: {Timestamp}), Duration: {Duration}ms.:

src/Elastic.Apm/Model/Span.cs
```diff
@@ -135,14 +141,25 @@ internal Span StartSpanInternal(string name, string type, string subType = null,
...
+					this, TimeUtils.FormatTimestampForLog(Timestamp), Timestamp, Duration);
+			}
+			else
+			{
+				var endTimestamp = TimeUtils.TimestampNow();
+				Duration = TimeUtils.DurationBetweenTimestamps(Timestamp, endTimestamp);
+				_logger.Trace()?.Log("Ended {Span}. Start time: {Time} (as timestamp: {Timestamp}), End time: {Time} (as timestamp: {Timestamp}), Duration: {Duration}ms",
```
gregkalapos: Any reason this is `Trace` and above is `Debug`?

src/Elastic.Apm/Model/Transaction.cs
```diff
@@ -165,11 +168,21 @@ public override string ToString() => new ToStringBuilder(nameof(Transaction))
 
 		public void End()
 		{
-			if (!Duration.HasValue) Duration = (DateTimeOffset.UtcNow - _start).TotalMilliseconds;
+			if (Duration.HasValue)
+			{
+				_logger.Debug()?.Log("End() called on already ended {Transaction}. Start time: {Time} (as timestamp: {Timestamp}), Duration: {Duration}",
```
gregkalapos: Same as in `Span.End` - we should consider changing the text.

SergeyKleyman:
> Thanks @SergeyKleyman.
> 
> Hmm.. very strange. So basically previously we did `_start = DateTimeOffset.UtcNow;` in the `.ctor` and ` Duration = (DateTimeOffset.UtcNow - _start).TotalMilliseconds` in `End()`.
> 
> Now we do `DateTime.UtcNow` which we convert to long with `(long)((dateTimeToConvert - UnixEpochDateTime).TotalMilliseconds * 1000);` both in `.ctor` and in `End()` and calculate the difference at the end by using the 2 `long` values.
> 
> Honestly to me it's still a mystery why this makes a difference, but if you see that this fixes the issues, then it's fine by me.
> 
> I also added some minor comments. Maybe the wording of the logs in `End()` both in `Span` and `Transaction` in case `if (Duration.HasValue)` is something you should consider.

Please take a look at https://github.com/elastic/apm-agent-dotnet/issues/330 I've tried to explain the root cause for the bug there.

src/Elastic.Apm/Model/Span.cs
SergeyKleyman: The reason I thought it might useful to have Span ID in the category/logger name is this way we don't need to remember ti include `Span.ToString()` every time we print something from Span's context. 

> I think with this it'll be harder to filter only span related logs.

Maybe just a little bit harder
```
grep -e "{Span\.\o:h*}"
```
instead of
```
grep "{Span}"
```

You are - at the moment we don't print that much from Span's context and everything we print includes `Span.ToString()` - I just wanted to future proof it when for when add more logs to it.
If you think at the moment adding ID to category/logger does more harm than good I'll remove it.

gregkalapos: I just I was surprised by the logs, and it took me some seconds to realize that it's the span id - and realized this by seeing it again in the same lane later printed by `Span.ToString`.

src/Elastic.Apm/Model/Span.cs
SergeyKleyman: I see... I think we should not allow `End` and `Duration` to get out sync - it will be too error-prone. How about making setter for Duration private and adding another overload for `End` that has duration parameter?

src/Elastic.Apm/Model/Span.cs
SergeyKleyman: That was under assumption that `Duration` can be set only by `End` (which i think should be the case) and thus calling `End` more than once is very unusual and we might want to know about it.

src/Elastic.Apm/Model/Span.cs
gregkalapos: Fine by me. This is the example where it makes sense to get the duration from the library and not calculating it ourselves: https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm.EntityFrameworkCore/EfCoreDiagnosticListener.cs#L51

src/Elastic.Apm/Helpers/TimeExtensions.cs
```diff
@@ -0,0 +1,25 @@
+using System;
+using System.Globalization;
+
+namespace Elastic.Apm.Helpers
+{
+	public static class TimeExtensions
```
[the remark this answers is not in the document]
SergeyKleyman: 10x - changed to internal

src/Elastic.Apm/Model/Span.cs
gregkalapos: None of this is public (it's on on `ISpan`), so no breaking change here.

SergeyKleyman: I agree that it's a valid case when duration is determined outside - but OTOH if duration is already know then it means that the span has definitely ended, right? So it makes sense to have API that guides you to do it in one call to minimize chance of setting `Duration` but forgetting to call `End`. Do you agree?

SergeyKleyman:
@gregkalapos Does the explanation of the root cause at #330 make sense?

src/Elastic.Apm/Model/Span.cs
gregkalapos: Sure, sure. I'm fine with that change.

gregkalapos:
> @gregkalapos Does the explanation of the root cause at #330 make sense?

Not 100%. I understand the explanation but I don't see how this maps to code. How did you print that `timestamp`? Where does that come from?

SergeyKleyman:
Logs in #330 came from the code before the fix (plus additional data in the log but no changes to the calculations related to the bug) so timestamp came from
```
Timestamp => _start.ToUnixTimeMilliseconds() * 1000;
```
Let me try to explain in a different way - `Timestamp` calculation throws microseconds away (truncates) but before the fix `Duration` calculation
```
Duration = (DateTimeOffset.UtcNow - _start).TotalMilliseconds
```
does NOT throw microseconds so assuming the code before the fix and:
Start time: 2019-07-01 04:09:26.179856 becomes timestamp: 1561954166179000
End time: 2019-07-01 04:09:26.195481 becomes timestamp: 1561954166195000
Exception caught very shortly before Span's end so it has the same timestamp as span's end - 1561954166195000. The difference between span's start timestamp and error's timestamp is 16ms.
OTOH `Duration` is calculated without discarding microseconds and the result is 15.6252ms < 16ms which is a problem. 
Mock APM server sees only span's start timestamp, error timestamp and span's duration and according to this data span's end (= span's start timestamp + span's duration) is before error's timestamp.
Does it make sense?

gregkalapos:
> Logs in #330 came from the code before the fix (plus additional data in the log but no changes to the calculations related to the bug) so timestamp came from
> 
> ```
> Timestamp => _start.ToUnixTimeMilliseconds() * 1000;
> ```
> 
> Let me try to explain in a different way - `Timestamp` calculation throws microseconds away (truncates) but before the fix `Duration` calculation
> 
> ```
> Duration = (DateTimeOffset.UtcNow - _start).TotalMilliseconds
> ```
> 
> does NOT throw microseconds so assuming the code before the fix and:
> Start time: 2019-07-01 04:09:26.179856 becomes timestamp: 1561954166179000
> End time: 2019-07-01 04:09:26.195481 becomes timestamp: 1561954166195000
> Exception caught very shortly before Span's end so it has the same timestamp as span's end - 1561954166195000. The difference between span's start timestamp and error's timestamp is 16ms.
> OTOH `Duration` is calculated without discarding microseconds and the result is 15.6252ms < 16ms which is a problem.
> Mock APM server sees only span's start timestamp, error timestamp and span's duration and according to this data span's end (= span's start timestamp + span's duration) is before error's timestamp.
> Does it make sense?

Ach ok, so now we truncate everything, also for duration calculation - I see. 

Is it good though? 🤔

src/Elastic.Apm/Helpers/TimeExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+			switch (dateTimeKind)
+			{
+				case DateTimeKind.Utc: return "UTC";
+				case DateTimeKind.Local: return "Local";
+				case DateTimeKind.Unspecified: return "Unspecified";
+				default: return $"UNRECOGNIZED {dateTimeKind} ({(int)dateTimeKind} as int)";
+			}
+		}
+
+		internal static string FormatForLog(this DateTimeOffset dateTimeOffset) =>
```
gregkalapos: This is unused.

src/Elastic.Apm/Model/Span.cs
SergeyKleyman: It will require making `Duration` setter private (it was public before). Are you okay with it?

gregkalapos: It's on an `internal` type, so it's not a breaking change - seems good to me.

SergeyKleyman: Unfortunately it was `Duration` setter was `public` on `IExecutionSegment`

src/Elastic.Apm/Helpers/TimeExtensions.cs
SergeyKleyman: > Is it good though? 🤔

No, now we don't truncate microseconds neither for timestamp nor for durations :)

SergeyKleyman: > This is unused.

10x - Fixed.

SergeyKleyman:
> But still with `(long)((dateTimeToConvert - UnixEpochDateTime).TotalMilliseconds * 1000)` we have the same, so to be totally honest I still don't see why this fixes it.

The new way to calculate timestamp
```
(long)((dateTimeToConvert - UnixEpochDateTime).TotalMilliseconds * 1000)
```
is not the same as the old way
```
_start.ToUnixTimeMilliseconds()
```
Because `TotalMilliseconds` returns milliseconds as `double` and it include microseconds and `* 1000` moves microseconds to integer part of the floating point. So in the new way to calculate microseconds are not discarded. PR includes tests to verify that it's the case.

src/Elastic.Apm/Model/Span.cs
gregkalapos: 😬Haven't looked that far up the inheritance chain.. yeah, it was publich :S  In this case I'm not that sure that this was a good idea - sorry for not seeing earlier.

gregkalapos:
> Because TotalMilliseconds returns milliseconds as double and it include microseconds 

Ok, I haven't realized this difference, now I got it. Thx!

gregkalapos:
Added label `breakingchange` because we removed `IExecutionSegment.Duration` setter. This label will be useful when we create release notes - we can list those to users and explain the alternatives, since these things can cause failing builds once they update.

src/Elastic.Apm/Model/Span.cs
SergeyKleyman: Do you think anybody actually used that API? If you think we should revert that change I will do it. I just wanted to make API guide the users into less error prone flow but I can leave with the old API as well.

gregkalapos: > Do you think anybody actually used that API?

I don't know, but they could and this breaks it, so I'd treat this as a breaking change. If someone used this, they won't be able to compile once they update.

> If you think we should revert

I'm not sure. I agree with your API change. On the other hand, the next version is planned to be `1.0.0`, after the current `1.0.0-beta`, so no major version change and we should avoid breaking changes in such scenarios. I'd personally avoid breaking change now and open an issue to make the change on a major version bump.

But I'm also fine if we keep it as it is now. I leave this up to you.

SergeyKleyman: > I'm not sure. I agree with your API change. On the other hand, the next version is planned to be `1.0.0`, after the current `1.0.0-beta`, so no major version change and we should avoid breaking changes in such scenarios. I'd personally avoid breaking change now and open an issue to make the change on a major version bump.

I agree - let's keep breaking changes for when it's really necessary. I'll create a PR shortly to revert and an issue to revisit this for the next major release.

SergeyKleyman: #349 - PR to revert.
#350 - issue to revisit this for the next major release when we can slip in breaking changes.

SergeyKleyman: BTW should we remove "breaking change" label from this issue?

gregkalapos: 👍 

I also removed the braking change label.

PR #334, opened 2019-07-02
Fix TrimToLength and move it to SerializationUtils

SergeyKleyman:
- Fixed `TrimToLength` not using `maxLength` parameter
- Moved `TrimToLength` and `TrimToPropertyMaxLength` from `StringExtensions` to `SerializationUtils` since both are very specific to serialization so it would make more sense to place it there
- Added tests

### DIFF
--- a/src/Elastic.Apm/Report/Serialization/SerializationUtils.cs
+++ b/src/Elastic.Apm/Report/Serialization/SerializationUtils.cs
@@ -1,0 +1,20 @@
+using Elastic.Apm.Helpers;
+
+namespace Elastic.Apm.Report.Serialization
+{
+	internal static class SerializationUtils
+	{
+		internal static string TrimToLength(string input, int maxLength)
+		{
+			input.ThrowIfArgumentNull(nameof(input));
+
+			if (input.Length <= maxLength) return input;
+
+			if (maxLength <= 5) return input.Substring(0, maxLength);
+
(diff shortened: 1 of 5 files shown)


src/Elastic.Apm/Report/Serialization/SerializationUtils.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+	internal static class SerializationUtils
+	{
+		internal static string TrimToLength(string input, int maxLength)
+		{
+			input.ThrowIfArgumentNull(nameof(input));
+
+			if (input.Length <= maxLength) return input;
+
+			if (maxLength <= 5) return input.Substring(0, maxLength);
```
gregkalapos: Why is this a special case? We wanna avoid "..." if we end up with very short strings?

SergeyKleyman: I thought that it doesn't make much sense not leaving anything from the original string but I'm open to suggestions. Although it's only used by serialization and `Consts.PropertyMaxLength` is much higher than 5.

gregkalapos: Yeah, I think it's fine to leave it there. I agree that showing `"st..."` (for `string`)  would not be very helpful :)

PR #349, opened 2019-07-04
Revert breaking change introduced in PR #331

SergeyKleyman:
Revert breaking change introduced in PR elastic/apm-agent-dotnet#331 (see https://github.com/elastic/apm-agent-dotnet/pull/331#discussion_r300169858)

elastic/apm-agent-dotnet#350 - issue to revisit this for the next major release when we can slip in breaking changes.


src/Elastic.Apm/Model/Span.cs
```diff
@@ -145,33 +146,33 @@ internal Span StartSpanInternal(string name, string type, string subType = null,
...
-				else
-				{
-					var endTimestamp = TimeUtils.TimestampNow();
-					Duration = TimeUtils.DurationBetweenTimestamps(Timestamp, endTimestamp);
-					_logger.Trace()?.Log("Ended {Span}. Start time: {Time} (as timestamp: {Timestamp})," +
-						" End time: {Time} (as timestamp: {Timestamp}), Duration: {Duration}ms",
-						this, TimeUtils.FormatTimestampForLog(Timestamp), Timestamp,
-						TimeUtils.FormatTimestampForLog(endTimestamp), endTimestamp, Duration);
-				}
+				Assertion.IfEnabled?.That(! _dbgIsEnded, $"If a span doesn't have Duration set it means that the span did not end yet." +
```
gregkalapos: Does this really do what it supposed to do? 

If you call `.End()` twice the 2. time the call will end up in `if (duration.HasValue)` (so this only runs once)  and silently skip over `if (IsSampled && isFirstEndCall)` - so the span won't be recorded the 2. time, which is how is definitely how it should be. 

 I just don't see why the big gymnastics before that is necessary - I thought the point is that it'll throw if you call it 2 times, but that does not seem to be the case.

src/Elastic.Apm/Model/Span.cs
```diff
@@ -17,6 +17,7 @@ internal class Span : ISpan
 		private readonly Transaction _enclosingTransaction;
 		private readonly IApmLogger _logger;
 		private readonly IPayloadSender _payloadSender;
+		private bool _dbgIsEnded;
```
gregkalapos: Sure about the `dbg` prefix? This variable influences the behaviour of the `End()` method.

src/Elastic.Apm/Model/Span.cs
```diff
@@ -145,33 +146,33 @@ internal Span StartSpanInternal(string name, string type, string subType = null,
...
+				Duration = TimeUtils.DurationBetweenTimestamps(Timestamp, endTimestamp);
+				_logger.Trace()?.Log("Ended {Span}. Start time: {Time} (as timestamp: {Timestamp})," +
+					" End time: {Time} (as timestamp: {Timestamp}), Duration: {Duration}ms",
+					this, TimeUtils.FormatTimestampForLog(Timestamp), Timestamp,
+					TimeUtils.FormatTimestampForLog(endTimestamp), endTimestamp, Duration);
 			}
 
-			if (IsSampled) _payloadSender.QueueSpan(this);
+			var isFirstEndCall = ! _dbgIsEnded;
+			if (IsSampled && isFirstEndCall)
```
gregkalapos: nit: I think according the `.editorconfig` we should invert this (not error though). (although I personally find this version more readable)

src/Elastic.Apm/Model/Span.cs
gregkalapos: Also I don't really understand the text - if I get this then it'd be the text of the exception, but then it'd say that you called `Span.End()` but "If a span doesn't have Duration set it means that the span did not end yet" - I can't follow, probably I'm missing something.

src/Elastic.Apm/Model/Transaction.cs
```diff
@@ -166,32 +167,28 @@ public override string ToString() => new ToStringBuilder(nameof(Transaction))
...
-				else
-				{
-					var endTimestamp = TimeUtils.TimestampNow();
-					Duration = TimeUtils.DurationBetweenTimestamps(Timestamp, endTimestamp);
-					_logger.Trace()?.Log("Ended {Transaction}. Start time: {Time} (as timestamp: {Timestamp})," +
-						" End time: {Time} (as timestamp: {Timestamp}), Duration: {Duration}ms",
-						this, TimeUtils.FormatTimestampForLog(Timestamp), Timestamp,
-						TimeUtils.FormatTimestampForLog(endTimestamp), endTimestamp, Duration);
-				}
+				Assertion.IfEnabled?.That(! _dbgIsEnded, $"If a transaction doesn't have Duration set it means that the transaction did not end yet." +
```
gregkalapos: Same questions as above :)

test/Elastic.Apm.Tests/SerializationTests.cs
```diff
@@ -25,8 +25,7 @@ public void StringTruncationTest()
 		{
 			var str = new string('a', 1200);
 
-			var transaction = new Transaction(new TestAgentComponents(), str, "test") { Result = "fail" };
-			transaction.End(/* duration: */ 1);
```
gregkalapos: Did we not have a `.End()` call here previously? Maybe it's still worth calling `.End()` before we serialize it, just to make sure that the duration is not changed by the `.End()` call.

src/Elastic.Apm/Model/Span.cs
SergeyKleyman: `Assertion.IfEnabled?.That(...)` is a replacement for `System.Diagnostics.Contracts.Contract.Assert` for agent's internal use - for us to catch bugs in agent's code. I didn't want to use `Contract.Assert` because application might have it disabled or configured to do some strange things (pop GUI dialog boxes, etc.) so I thought we might want a separate runtime assertion facility that we can control independently of whatever monitored application uses.

This particular assertion goal is to verify an invariant that should hold by the end of a call to `End()` which is that `Duration` should have value which means that if `Duration` does not have value then `End()` was not called before which means `_isEnded` should be `false`.
I can rephrase the message as inversion of the what it's now - "By the end of call to `End()` `Duration` should have value" - it's just that it won't be clear what it has to do with the context where this assertion is located which is `else` branch of `if (Duration.HasValue)` meaning we know that `Duration` doesn't have a value at this point.

The goal of adding this assertion was definitely not to throw if `End()` is called more than once since this was not `End()`'s behavior so far.  The goal of adding this assertion is to verify invariant of the relation between `_isEnded` and `Duration.HasValue` and thus to protect against bugs that can introduced by changes to this code and break that invariant.

Regarding throwing if `End()` is called more than once
- Do we want to do it? For example calling `IDisposable.Dispose()` more than once is allowed and it's a no-op.
- It would be a breaking change, wouldn't it?

src/Elastic.Apm/Model/Transaction.cs
SergeyKleyman: Please see my explanation above.

test/Elastic.Apm.Tests/SerializationTests.cs
SergeyKleyman: Fixed

src/Elastic.Apm/Model/Span.cs
SergeyKleyman: This code had a bug (`_dbgIsEnded` was set to `true` only when `IsSampled` is `true`) so I fixed it and now there is no suggestion from Rider to invert `if`. There was a related bug in `Transaction.End()` (`End()` queued transaction on each call instead of only the 1st one). I fixed both issues and added a test (`Elastic.Apm.Tests.ApiTests.ApiTests.CallingEndMoreThanOnceIsNoop`) to verify.

BTW is it indeed settings in `.editorconfig` that prompt Rider's suggestion to invert that `if` or maybe it's just default Rider's inspections settings?

src/Elastic.Apm/Model/Span.cs
SergeyKleyman: Yes, you're right - I've removed `dbg` prefix.

src/Elastic.Apm/Helpers/Assertion.cs
```diff
@@ -0,0 +1,19 @@
+namespace Elastic.Apm.Helpers
+{
+	internal class Assertion
```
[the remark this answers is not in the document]
SergeyKleyman: Yes - Fixed

src/Elastic.Apm/Helpers/Assertion.cs
```diff
@@ -0,0 +1,19 @@
...
+
+		internal struct AssertIfEnabled
+		{
+			internal void That(bool condition, string message)
+			{
+				if (!condition) throw new AssertionFailedException(message);
+			}
+		}
+
+		internal static bool IsEnabled { get; set; } = true;
```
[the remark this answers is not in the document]
SergeyKleyman: Fixed

src/Elastic.Apm/Helpers/Assertion.cs
```diff
@@ -0,0 +1,19 @@
+namespace Elastic.Apm.Helpers
+{
+	internal class Assertion
+	{
+		private static readonly AssertIfEnabled _singletonAssertIfEnabled = new AssertIfEnabled();
```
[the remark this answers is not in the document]
SergeyKleyman: Fixed

test/Elastic.Apm.Tests/SerializationTests.cs
SergeyKleyman: I've added another test (since this one is called `StringTruncationTest` so it is not related to `Duration`) to verify that `Duration` if already has value then is not changed by `End()`  - `Elastic.Apm.Tests.ApiTests.ApiTests.CallToEndDoesNotChangeAlreadySetDuration`

src/Elastic.Apm/Model/Span.cs
gregkalapos: Ok, now that you write this it makes sense, but by looking at the code at least to me it looked that your intention is different. 

Overall I think it's fine. If I may add some minor suggestions:
- Can we maybe add the first part of your comment (or something similar) to the `Assertion` class? That'd shed some light on why we do that stuff et all. 

- > This particular assertion goal is to verify an invariant that should hold by the end of a call to End() which is that Duration should have value which means that if Duration does not have value then End() was not called before which means _isEnded should be false.

This totally makes sense, but by looking at the code I did not think that this is the goal:

```
Assertion.IfEnabled?.That(!_isEnded, $"If a span doesn't have Duration set it means that the span did not end yet." +
					$" this: {this}; {nameof(Duration)}: {Duration}, {nameof(_isEnded)}: {_isEnded}");
```

Mainly because you assert on `_isEnded` and the string talks about `Duration`. I think your suggestion with the other message and a short comment would be more clear - or keep the message, but add a more complete comment - like the one you added here.

> Regarding throwing if End()
> Do we want to do it?  

No, even if we do it, I'd not do it in this PR.

> It would be a breaking change, wouldn't it?

Yes, it'd be and I think we both prefer to avoid this breaking change at this point, so from my side we are ok with the current code.

SergeyKleyman: Done - I've added comment to `Assertion` class and included an explanation in the message. I prefer message to comment because the message will be available in logs and thus it won't be necessary to go to source code to understand it.

gregkalapos: > and included an explanation in the message.

Yeah, even better 👍  Great, thanks!

PR #363, opened 2019-07-06
Refactor unit tests to use a custom WebApplicationFactory

NeoXtreem:
Later package versions (not yet upgraded to) have stricter routing in the unit tests, so for `AspNetFullFrameworkSampleApp` (renamed from `SampleAspNetFullFrameworkApp` for better consistency), the unit tests have been refactored to use `CustomWebApplicationFactory` which allows a custom (fake) `Startup` to be used for configuring test startup.

This config is encapsulated in `StartupConfigService` which is instantiated by the unit tests (using `TestHelper` static methods).  The fake `Startup` classes derive from the real ones overriding only what needs to be configured for the unit tests.  The custom factory calls on its `WithWebHostBuilder` the `ConfigureTestServices` method to add `StartupConfigService` to the DI container so that this can be retrieved by the fake `Startup` for special configuration in the `ConfigureAgent` override method.  `UseSolutionRelativeContentRoot` is also called here to specify the root path for the views.

The above methodology follows [this article](https://dzone.com/articles/using-a-custom-startup-class-with-aspnet-core-inte) which is based on Microsoft documentation on [the Startup class](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/startup?view=aspnetcore-2.2) and [integration tests in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-2.2).

Overall, this refactor results in much cleaner and leaner unit test code without boiler-plate bootstrapping the test startup configuration.  For example, `TestHelper` (renamed from `Helper`) is now much leaner and simpler.

It should be noted that unit test `SubscribeAndUnsubscribeAspNetCoreDiagnosticListener` is still semi-disabled, but it succeeds when run indepedently.  A `TODO` note has been placed in the unit test to this effect.

These changes were originally in PRs #226 and #309 (with some historical discussion), but these have been closed in order to split the changes into separate PRs.


NeoXtreem:
@gregkalapos @SergeyKleyman Now that the application is non-beta, could this PR be looked at as I have just merged from the base master branch, and I can changes that are really unnecessary and wouldn't have to be made if the `CustomWebApplicationFactory` methodology described in the description of this PR (as per Microsoft documentation) is followed.

I can see boiler plate changes having to be made in `TransactionNameTests` and `AspNetCoreMiddlewareTests` that I have significantly simplified removing the need for members (`_agent`, `_factory`, `_payloadSender`) that should all be disposed and be used with the dispose pattern (which I have implemented locally within the unit tests).

NeoXtreem:
@gregkalapos @SergeyKleyman Any chance this could be looked at as I'm seeing new tests and test classes being added again (such as `SanitizeFieldNamesTests`) with lots of boiler plate code that my changes prevent? It's making forks difficult to manage where additional tests using the proper methods are being added.

gregkalapos:
@NeoXtreem 

I'd be open to discuss `WebApplicationFactory` (we are in general open to discuss any suggestion or improvement).

Could you please just move out the renaming of the sample project from this PR? - or open a new one, or just rebase and force push, whichever is the best for you. But please only have `WebApplicationFactory` related changes here.

Sorry to push back on this in its current form, but please also take the reviewer's perspective into account - ~12K lines are changed in 126 file, it'd be hard to review the whole thing and discuss `WebApplicationFactory`. 

We tried to summerize some thing that help us to review [here](https://github.com/elastic/apm-agent-dotnet/blob/master/CONTRIBUTING.md#general-advice).

Thanks, and I hope we can help each other with that approach.

NeoXtreem:
@gregkalapos I've reverted the rename of the sample app - many of the ~12k line changes were from this it seems, and that still shows as the case even though the files have been renamed back (a shortcoming of the diff in GitHub perhaps). But the PR now focuses on the implementation of `WebApplicationFactory`. This is a large refactor of the unit tests, so it is difficult to make this a much smaller change in terms of lines of code and files, but it shouldn't take someone that long (relatively) once they get going and understand the nature of the changes. I believe it is certainly a good change, and worth a reviewer's time (whether it takes an hour or a day) to review, as it certainly took me more than a few days to do this in the first place as well as the additional work in keeping the branch up to date with the base branch.

gregkalapos:
Closing for now - this diverged a lot from master. 

Happy to restart the discussion, but please consider the points above and our [CONTRIBUTING.md](https://github.com/elastic/apm-agent-dotnet/blob/master/CONTRIBUTING.md).

Issue #358, opened 2019-07-05
Support for chained exceptions


SergeyKleyman:
Derived from https://github.com/elastic/apm/issues/40

russcam:
Implemented in https://github.com/elastic/apm-agent-dotnet/pull/1277

PR #367, opened 2019-07-06
Capture chained (inner) exceptions

NeoXtreem:
Addresses issue #358.  Kibana doesn't have a specific property for this according to the [error schema](https://www.elastic.co/guide/en/apm/server/master/error-api.html), so the inner exception details have been added to the `CapturedException.Message` property.


SergeyKleyman:
Thank you for the PR.
There is [a change to the `error` schema](https://github.com/elastic/apm/issues/40#issuecomment-508619451) proposed at #358's parent issue.
I think we should wait for the change to be finalized and then implement .NET agent's part according the new schema.

NeoXtreem:
> Thank you for the PR.
> There is [a change to the `error` schema](https://github.com/elastic/apm/issues/40#issuecomment-508619451) proposed at #358's parent issue.
> I think we should wait for the change to be finalized and then implement .NET agent's part according the new schema.

@SergeyKleyman OK, thanks for letting me know. I'll do it this way for now just for my company's purposes (as we need to see this information), but will change it according to the new error schema once it supports inner exception information, and do it under this PR. Just ping me when it's ready if you can, and I'll happily do it - doubt it'll take long. I'll add you and @gregkalapos as contributors to my fork anyway.

gregkalapos:
https://github.com/elastic/apm-agent-dotnet/pull/1277 implements this.

Issue #243, opened 2019-05-29
Rename Elastic.Apm.All project


SergeyKleyman:
Now that we have code specific for .NET Core (`Elastic.Apm.AspNetCore` and `Elastic.Apm.EntityFrameworkCore`) and .NET Framework (`Elastic.Apm.AspNetFullFramework`) `Elastic.Apm.All` project still references only .NET Core projects. We should consider if we want the same kind `.All` project to include all features applicable to .NET Framework.

It seems that continuing with just one `.All` project and include both .NET Core and .NET Framework would not be very useful because most customers' projects target one of those platforms. So maybe we should have split the project tree at the very top:

| Current name | New  name | Notes |
| --------------- | ------------ | ------- |
| `Elastic.Apm.AspNetCore` | `Elastic.Apm.NetCore.Asp` | |
| `Elastic.Apm.EntityFrameworkCore` | `Elastic.Apm.NetCore.EntityFramework` | |
| `Elastic.Apm.All` | `Elastic.Apm.NetCore.All` | |
| `Elastic.Apm.AspNetFullFramework` | `Elastic.Apm.NetFramework.Asp` | |
| | `Elastic.Apm.NetFramework.EntityFramework` | New project (TBD) |
| | `Elastic.Apm.NetFramework.All` | New project. References `Elastic.Apm.NetFramework.Asp` and `Elastic.Apm.NetFramework.EntityFramework`) |

gregkalapos:
I'm afraid, the proposed scheme with `NetCore` and `NetFramework` won't really work - or it'd be misleading, for example the EF Core package (proposed `Elastic.Apm.NetCore.EntityFramework`) also works with EF Core on Full Framework, so that package is not specific to a .NET Core and I expect having similar situation with other libraries later. You can also run ASP.NET Core on top of Full Framework (see [here](https://blog.codeinside.eu/2018/03/31/running-aspnet-core-2-under-the-full-dotnet-framework/)) - MS plans to drop this long term, but it works with the current version.

Also this would be a big big breaking change.

On the other hand I also think that we should address the issue with having useful `All` packages for common scenarios, and adding ASP.NET (Classic) support breaks our current naming. 

How about keeping all names as they are, except changing the current `Elastic.Apm.All` to something that better reflects its purpose and also add something similar for the ASP.NET Classic scenario?

SergeyKleyman:
[...]

PR #371, opened 2019-07-08
Renaming the Elastic.Apm.All pacakge and the UseElasticApm method

gregkalapos:
Implements the decisions from #243  and #234. 

- `Elastic.Apm.All` package becomes  `Elastic.Apm.NetCoreAll`. Main reason: we want to make clear that this package is intended for ASP.NET Core and .NET Core use cases - not for .NET Framework use cases - more in elastic/apm-agent-dotnet#243 
- The `UseElasticApm()` method in the `Elastic.Apm.NetCoreAll` package is renamed to `UseAllElasticApm()`. Main reason: Previously we had 2 `UseElasticApm()` methods in 2 different packages with 2 different namespaces. They do different things (described in the   `docs/setup.asciidoc`), but this wasn't obvious and having those 2 methods with the same name was very confusing to users - more in #234. 

- Updated the docs

This is a breaking change, probably doing this before the beta would have been better, but I think it's still ok to do this before GA.


gregkalapos:
Integration tests and [`opbeans-dotnet`](https://github.com/elastic/opbeans-dotnet/blob/master/opbeans-dotnet/opbeans-dotnet.csproj#L11) need to be also adapted. Talking to @v1v, since he did a lot of work in that area. 

`opbeans-dotnet` will be especially fun to adapt, since it relies on the published NuGet packages, but now `.UseElasticApm()` also needs to be updated, which has not been published yet. 🐔 🥚

gregkalapos:
Merging now, IT on `master` will fail, once the 2 referenced issues are merged IT should be green again.

PR #372, opened 2019-07-12
ci: test windows immutable workers

kuisathaverat:
This PR is to test the new windows immutable workers.


Jenkinsfile
```diff
@@ -6,7 +6,7 @@ pipeline {
   agent any
   environment {
     REPO = 'apm-agent-dotnet'
-    BASE_DIR = "src/github.com/elastic/${env.REPO}"
+    BASE_DIR = "src"
```
kuisathaverat: we've reached the max path length 248 characters because of that NuGet 5.1.0 fails when makes the restore.

ElasticApmAgent.sln
```diff
@@ -1,4 +1,4 @@
-﻿Microsoft Visual Studio Solution File, Format Version 12.00
+Microsoft Visual Studio Solution File, Format Version 12.00
```
kuisathaverat: revert this change

PR #375, opened 2019-07-17
Implement ServerUrls configuration option limitation

### LINKED ISSUE #144: Support for multiple APM servers in serverUrls config
From https://github.com/elastic/apm-agent-dotnet/pull/136#discussion_r264608577 

Maybe relevant: https://github.com/elastic/apm-agent-nodejs/issues/447

Java agent [server-urls](https://www.elastic.co/guide/en/apm/agent/java/current/config-reporter.html#config-server-urls) configuration setting supports multiple URLs which are used to implement fail-over behavior as follows
1. Randomly shuffle the list at the start. Shuffled list is the one to be used in the rest of the steps.
2. Set the first URL in the list as the current one.
3. Send data using the current URL.
4. If it fails set the next URL in the list as the current one and go to step 3.

See https://github.com/elastic/apm-agent-java/blob/master/apm-agent-core/src/main/java/co/elastic/apm/agent/report/IntakeV2ReportingEventHandler.java#L356 for details.

SergeyKleyman:
Closes elastic/apm-agent-dotnet#376

- Document ServerUrls configuration option limitation
- Log warning if user specified more than URL is in the configuration

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -172,9 +172,9 @@ Use this setting in case the APM Server requires a token (e.g. APM Server in Ela
 | `http://localhost:8200` | List
 |============
 
-The URLs for your APM Servers. The URLs must be fully qualified, including protocol (`http` or `https`) and port. To add multiple servers, separate them with a comma (`,`).
+The URLs for your APM Servers. The URLs must be fully qualified, including protocol (`http` or `https`) and port.
 
-NOTE: Providing multiple URLs only works with APM Server v6.5+.
+NOTE: Providing multiple URLs is not supported by the agent yet. If multiple URLs are provided only the first one will be used.
 
 [float]
 [[config-log-level]]
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -169,10 +169,10 @@ Use this setting in case the APM Server requires a token (e.g. APM Server in Ela
 [options="header"]
 |============
 | Default                 | Type
-| `http://localhost:8200` | List
+| `http://localhost:8200` | String
 |============
 
-The URLs for your APM Servers. The URLs must be fully qualified, including protocol (`http` or `https`) and port.
+The URL for your APM Server. The URL must be fully qualified, including protocol (`http` or `https`) and port.
 
 NOTE: Providing multiple URLs is not supported by the agent yet. If multiple URLs are provided only the first one will be used.
 
--- a/src/Elastic.Apm.AspNetCore/Config/MicrosoftExtensionsConfig.cs
+++ b/src/Elastic.Apm.AspNetCore/Config/MicrosoftExtensionsConfig.cs
@@ -12,7 +12,7 @@ namespace Elastic.Apm.AspNetCore.Config
 	/// </summary>
 	internal class MicrosoftExtensionsConfig : AbstractConfigurationReader, IConfigurationReader
 	{
-		internal const string Origin = "Configuration Provider";
+		internal const string Origin = "Microsoft.Extensions.Configuration";
 
 		internal static class Keys
 		{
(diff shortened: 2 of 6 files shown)


docs/configuration.asciidoc
```diff
@@ -172,9 +172,9 @@ Use this setting in case the APM Server requires a token (e.g. APM Server in Ela
 | `http://localhost:8200` | List
 |============
 
-The URLs for your APM Servers. The URLs must be fully qualified, including protocol (`http` or `https`) and port. To add multiple servers, separate them with a comma (`,`).
```
gregkalapos: I'd keep the `To add multiple servers, separate them with a comma (`,`).` part - I think that adds value, even if the feature is not implemented.

SergeyKleyman: I'm not sure I follow your reasoning - you would like to let users know how to do something we don't support yet?

gregkalapos: Yes, otherwise it's not clear how to set multiple values. 

It's like we say:
- The setting enables you to add multiple servers.
- You add multiple values by separating by comma
- Currently only the 1. one is used.

So reading this short part makes it clear.

Otherwise we say:
- The setting enables you to add multiple values
- We use only the 1. one.
- (we don't even tell you how to set multiple values) 

But then why we say `The URLs for your APM Servers`?

SergeyKleyman: I see. So wouldn't fixing all other mentions of "URLs" in plural be a better approach?

gregkalapos: I'd be very open in the docs about this and not try to hide it. I personally would prefer the other one, because without the plural we still leave questions open (like why is the name of the setting itself in plural and how to set multiple values - even if only the 1. is used). But I'm also fine with this version. Maybe @bmorelli25 has an idea on this - I'd usually be super open in the docs and try to cover everything and just list the limitations.

SergeyKleyman: I agree that we should cover the limitations (which this PR already does). I just don't see the point of describing how to use a feature that is not implemented yet.

gregkalapos: >  I just don't see the point of describing how to use a feature that is not implemented yet.

To be fair, it's a partly implemented feature - we indeed parse the list, it's just that the `PayloadSenderV2` only uses the 1. item - that causes the trouble here.

I think it's just not good to diverge in the docs from the real implementation in order to hide something that we should not be ashamed of.

This setting is indeed a `List`, not a `String` - I don't see any problem with fully describing the feature and say that it only partly works in the current version instead of having docs that diverge from the real implementation (stating Type `String` when in reality it's a `List`) and leave questions open (why is the setting plural and how to set multiple values?).

From the users point of view I think an approach like this would be best (pseudo docs):
- The setting is `ServerUrls`, it's type is `List`
- Set multiple values by separating by comma
- The current version only uses the 1. one, feature will be implemented to use multiple in a later version.

docs/configuration.asciidoc
```diff
@@ -169,12 +169,12 @@ Use this setting in case the APM Server requires a token (e.g. APM Server in Ela
 [options="header"]
 |============
 | Default                 | Type
-| `http://localhost:8200` | List
+| `http://localhost:8200` | String
```
gregkalapos: Added a comment on this one: https://github.com/elastic/apm-agent-dotnet/pull/375/files#r304312327

docs/configuration.asciidoc
SergeyKleyman: It doesn't matter to the end user if feature implemented 0% or 99% - in the end it either works or it doesn't. Setting this configuration option to a list doesn't work so there's no point in wasting user's time on reading how to configure more than one URL just to follow it up with a note saying that it's not yet supported.

gregkalapos: Ok, I give up on this.

Issue #378, opened 2019-07-17
Fix flaky test - Elastic.Apm.Tests.MetricsTests.MetricsWithRealAgent


SergeyKleyman:
For example it failed when testing changes from [PR #375](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-dotnet%2Fapm-agent-dotnet-mbp/detail/PR-375/3/pipeline/#step-160-log-531) and [PR #380](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-dotnet%2Fapm-agent-dotnet-mbp/detail/PR-380/2/pipeline/143/#step-185-log-490) even though those PRs did not change anything related to the test.

Currently the test instructs agent to collect metrics every second and wait 2.2 seconds for the agent to collect metrics at least once. It seems that agent doesn't get an opportunity to collect metrics even during 2.2 seconds. We can either investigate why it happens (Does the timer even starts during 2.2 seconds? If it does then why it cannot complete in time?) or we can just increase wait time (let's say to 10 instead of 2.2 seconds) and see if it helps.

gregkalapos:
I opened #385 to address this.

It's very strange. I `XunitOutputLogger`, probably the logs will help. 

> or we can just increase wait time 
I think it's also worth doing it - did also that in the PR.

gregkalapos:
I think this is ok to close since #385 is merged. I'd just close this and if the test fails again we can reopen this. I think it's better than leaving it open and "wait" for failing again.

Anything against that?

SergeyKleyman:
I agree - we can close it.

PR #385, opened 2019-07-18
MetricsWithRealAgent - add XunitOutputLogger and increase timeout

gregkalapos:
Addressing #378

- Increased timeout
- Added `XunitOutputLogger` with debug logs which hopefully will tell us how far the metrics collection went in case it fails again


SergeyKleyman:
1. `Elastic.Apm.Metrics.MetricsCollector.CollectAllMetrics` doesn't have any logging statements at the start so it won't be clear when the timer was actually triggered.
2. If we want to understand which provider takes longer than expected I would also add logging statements before and after calling each provider.

SergeyKleyman:
> Added more logging.

It looks good.

PR #389, opened 2019-07-18
Change Level property of IApmLogger to IsEnabled method

vhatsura:
The following idea of refactoring `IApmLogger` interface was introduced in https://github.com/elastic/apm-agent-dotnet/pull/249#discussion_r300036181 discussion.


test/Elastic.Apm.Tests/ConfigTests.cs
```diff
@@ -162,31 +162,35 @@ public void SetDebugLogLevelTest()
 		{
 			var agent = new ApmAgent(new TestAgentComponents("Debug"));
 			agent.ConfigurationReader.LogLevel.Should().Be(LogLevel.Debug);
-			agent.Logger.Level.Should().Be(LogLevel.Debug);
+			agent.Logger.IsEnabled(LogLevel.Debug).Should().BeTrue();
```
gregkalapos: Would something like this method make sense and calling it from the respective `SetXDebugLogLevelTest`?

```
private static void TestAllLevelsWithEnabledLevel(IApmAgent agent, LogLevel enabledLevel)
{
	agent.Logger.IsEnabled(enabledLevel).Should().BeTrue();
	foreach (LogLevel enumValue in Enum.GetValues(typeof(LogLevel)))
	{
		if(enabledLevel <= enumValue)
			agent.Logger.IsEnabled(enumValue).Should().BeTrue();
		else
			agent.Logger.IsEnabled(enumValue).Should().BeFalse();
	}
}
```

And in this case the Debug test would just call `TestAllLevelsWithEnabledLevel(agent, LogLevel.Debug);` - and some for the other tests.

vhatsura: The valid point, will refactor unit tests

vhatsura:
@gregkalapos, thanks for quick review. I totally agree with you that changes in public interface should be interpreted as breaking changes, however, I think it's eligible to intriduce this changes in GA release.
As we look into usage of `IApmLogger` interface we see:
1. custom implementation of `IApmLogger` cannot be injected for Asp.Net Core applications (no suitable extension method which can accept `IApmLogger` impelementation)
2. custom implementation of `IApmLogger` cannot be injeted for Asp.Net Full Framework applications in case of using `ElasticApmModule` class.

This breaking changes has impact only for users who use agent public api thorugh `Agent.Setup` method. For such users we can suggest to make changes as it did for `ConsoleLogger`: `return Level <= level;` (a few lines in release notes).

It's only my thoughts. The final decision is yours.

PR #390, opened 2019-07-19
Add Elastic APM for SmartSql

RocherKong:
Add Support For SmartSql.


src/Elastic.Apm.SmartSql/Elastic.Apm.SmartSql.csproj
```diff
@@ -0,0 +1,36 @@
...
+    <Description>Elastic APM for SmartSql. This package contains auto instrumentation for SmartSql. See: https://github.com/elastic/apm-agent-dotnet/tree/master/docs</Description>
+    <PackageTags>apm, monitoring, elastic, elasticapm, analytics, entiryframeworkcore, efcore, SmartSql</PackageTags>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
+  </PropertyGroup>
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="SmartSql" Version="4.0.3" PrivateAssets="All" />
```
gregkalapos: It's worth targeting the lowest possible version. If there is an older one that has all the types we need in this project, we should use that one - I'm not familiar with SmartSql, so maybe this is the oldest we can use here. This would mean Elastic APM would support it from `4.0.3`. Feel free to decide if this is something worth following up on.

src/Elastic.Apm.SmartSql/SmartSqlDiagnosticListener.cs
```diff
@@ -0,0 +1,92 @@
...
+		public void OnError(Exception error) { }
+
+		public void OnNext(KeyValuePair<string, object> kv)
+		{
+			switch (kv.Key)
+			{
+				case string k when k == SmartSqlDiagnosticListenerExtensions.SMART_SQL_BEFORE_COMMAND_EXECUTER_EXECUTE:
+					if (kv.Value is CommandExecuterExecuteBeforeEventData executerExecuteBeforeEventData)
+					{
+						var newsSpan = Agent.TransactionContainer.Transactions.Value.StartSpanInternal(executerExecuteBeforeEventData.Operation,
```
gregkalapos: The current transaction can be null. This is when there is no transaction started. In the sample application this never happens, because for each incoming HTTP request our ASP.NET Core auto instrumentation starts a transaction. 

I think this one can help you - it should be similar: https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm.EntityFrameworkCore/EfCoreDiagnosticListener.cs#L31 - spot `&& Agent.TransactionContainer.Transactions.Value != null`

src/Elastic.Apm.SmartSql/SmartSqlDiagnosticListener.cs
```diff
@@ -0,0 +1,92 @@
...
+							ApiConstants.TypeDb);
+						newsSpan.Tags.Add("statement",executerExecuteBeforeEventData.ExecutionContext.Request.RealSql);
```
gregkalapos: `Tags` is preserved for users of the public API - so when users manually trace things. We don't use it for auto instrumentation. I see below you already fill the db related fields - as we do EF Core instrumentation here: https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm.EntityFrameworkCore/EfCoreDiagnosticListener.cs#L45.

What's the intention here with adding the same to `Tags`?

Btw. I think the confusion on the meaning of `Tags` is our fault - already saw multiple people using it for these kinds of thing - I'll make sure we document this better!

src/Elastic.Apm.SmartSql/SmartSqlDiagnosticListener.cs
```diff
@@ -0,0 +1,92 @@
...
+					{
+						if (_spans.TryRemove(commandExecutedEventData.OperationId,out var span))
+						{
+							span.Context.Db = new Database
+							{
+								Statement = commandExecutedEventData.ExecutionContext.Request.RealSql,
+								Instance = commandExecutedEventData.ExecutionContext.Request.DataSourceChoice.ToString(),
+								Type = Database.TypeSql
+							};
+							span.Tags.Add("fromcache", commandExecutedEventData.ExecutionContext.Result.FromCache.ToString());
```
gregkalapos: Same with tags as above.

src/Elastic.Apm.SmartSql/SmartSqlDiagnosticListener.cs
```diff
@@ -0,0 +1,92 @@
...
+							var dbConnection = commandExecutedEventData.ExecutionContext.DbSession.Connection;
+							if (dbConnection == null)
+							{
+								return;
```
gregkalapos: In this case the span won't be sent (`End()` is not called) - is this a big enough problem to drop the span? I see you only set `Tags` here - any reason not just skipping that part?

src/Elastic.Apm.SmartSql/Elastic.Apm.SmartSql.csproj
RocherKong: Yes.SmartSql version `4.0.3` is the lowest possible version for the types needed in the project. 
Smart Sql is an ORM similar to mybatis in Java.

src/Elastic.Apm.SmartSql/SmartSqlDiagnosticListener.cs
RocherKong: ok,I will add null check in next pr.

src/Elastic.Apm.SmartSql/SmartSqlDiagnosticListener.cs
RocherKong: this is my fault. there is no reason to drop the span.

src/Elastic.Apm.SmartSql/SmartSqlDiagnosticListener.cs
RocherKong: ok

src/Elastic.Apm.SmartSql/SmartSqlDiagnosticListener.cs
RocherKong: Thanks @gregkalapos !
Thank you for your comments on the submission code.
I will fix a series of questions raised in the comments in the near future.
1.add the minimal test module 
2.remove the custom tag content.
3.add null check for transaction and make end() called.
This project is great. I plan to put it into production.

RocherKong:
For cover Elastic.Apm.SmartSql with automated tests. I Add a test file under /test which name is DiagnosticListenerSmartSqlTests Referred to the Efcore example(/tests/DiagnosticListenerTests.cs), but it doesn't seem to work and it's not clear where the failure occurred.Can you help me deal with this problem?

gregkalapos:
> For cover Elastic.Apm.SmartSql with automated tests. I Add a test file under /test which name is DiagnosticListenerSmartSqlTests Referred to the Efcore example(/tests/DiagnosticListenerTests.cs), but it doesn't seem to work and it's not clear where the failure occurred.Can you help me deal with this problem?

When I send the request to the test app, it throws an error:

```
 Request matched multiple actions resulting in ambiguity. Matching actions: SampleAspNetCoreApp.Controllers.HomeController.Index (SampleAspNetCoreApp)
```

It seems that you also start up the `SampleAspNetCoreApp` in the test. I'm trying to find the code where this happens. Will update here once I know more.

gregkalapos:
> For cover Elastic.Apm.SmartSql with automated tests. I Add a test file under /test which name is DiagnosticListenerSmartSqlTests Referred to the Efcore example(/tests/DiagnosticListenerTests.cs), but it doesn't seem to work and it's not clear where the failure occurred.Can you help me deal with this problem?

When I send the request to the test app, it throws an error:

```
 Request matched multiple actions resulting in ambiguity. Matching actions: SampleAspNetCoreApp.Controllers.HomeController.Index (SampleAspNetCoreApp)
```

It seems that you also start up the `SampleAspNetCoreApp` in the test. I'm trying to find the code where this happens. Will update here once I know more.

gregkalapos:
> For cover Elastic.Apm.SmartSql with automated tests. I Add a test file under /test which name is DiagnosticListenerSmartSqlTests Referred to the Efcore example(/tests/DiagnosticListenerTests.cs), but it doesn't seem to work and it's not clear where the failure occurred.Can you help me deal with this problem?

When I send the request to the test app, it throws an error:

```
 Request matched multiple actions resulting in ambiguity. Matching actions: SampleAspNetCoreApp.Controllers.HomeController.Index (SampleAspNetCoreApp)
```

It seems that you also start up the `SampleAspNetCoreApp` in the test. I'm trying to find the code where this happens. Will update here once I know more.

gregkalapos:
> For cover Elastic.Apm.SmartSql with automated tests. I Add a test file under /test which name is DiagnosticListenerSmartSqlTests Referred to the Efcore example(/tests/DiagnosticListenerTests.cs), but it doesn't seem to work and it's not clear where the failure occurred.Can you help me deal with this problem?

When I send the request to the test app, it throws an error:

```
 Request matched multiple actions resulting in ambiguity. Matching actions: SampleAspNetCoreApp.Controllers.HomeController.Index (SampleAspNetCoreApp)
```

It seems that you also start up the `SampleAspNetCoreApp` in the test. I'm trying to find the code where this happens. Will update here once I know more.

gregkalapos:
> For cover Elastic.Apm.SmartSql with automated tests. I Add a test file under /test which name is DiagnosticListenerSmartSqlTests Referred to the Efcore example(/tests/DiagnosticListenerTests.cs), but it doesn't seem to work and it's not clear where the failure occurred.Can you help me deal with this problem?

When I send the request to the test app, it throws an error:

```
 Request matched multiple actions resulting in ambiguity. Matching actions: SampleAspNetCoreApp.Controllers.HomeController.Index (SampleAspNetCoreApp)
```

It seems that you also start up the `SampleAspNetCoreApp` in the test. I'm trying to find the code where this happens. Will update here once I know more.

RocherKong:
Create New Test Project For Elastic.Apm.SmartSql in _/test_ Which Name is **Elastic.Apm.SmartSql.Tests**. And Replace SqlServer with SQLite as Default DB Storage.

RocherKong:
Hello, can you check whether the submission meets the requirements of merging this time？

RocherKong:
Any other problem? about merge .

gregkalapos:
Hi @RocherKong,

Sorry for the delay. 

We discussed this in the team, and we are not totally sure that the best place for this code would be in the main agent repository.

I personally when I looked at it first I thought this is not an issue, but we started discussing it, and we probably won’t have the bandwidth to maintain it and we also don’t really know when and how to release it. 

Therefore the suggestion would be that I help you to basically create the same code in an external repository.

First of all, I’m super sorry about this, I know you worked on this a lot, and I personally would be happy to see these things in the main repo. On the other hand it’s true that we probably are not the best people to control the release of this and maintain support for it. 

I’d be happy to work with you on moving this into a different repository, and from that point you can control the release of it.

The way to do this would be to take a dependency on the [`Elastic.Apm`](https://www.nuget.org/packages/Elastic.Apm/) package and write code against the [Public Agent API](https://www.elastic.co/guide/en/apm/agent/dotnet/current/public-api.html).

So… I think the heart of this PR is the `SmartSqlDiagnosticListener.cs`. Currently you use `Agent.TransactionContainer.Transactions.Value.StartSpanInternal`, which is internal, so that won’t be visible. 

The first step would be to switch to `Agent.TransactionContainer.Transactions.Value.StartSpan` - that is a public method which returns an `ISpan`. I think you mainly use things that are alsp available on `ISpan` anyway… so first we should see if there is anything missing. If there is we’ll try to make those available for you.

How does this sound? I’m very sorry about the inconvenience.

RocherKong:
Hi @gregkalapos ,
I am very sorry for this result.

1. SmartSql, like EntityFrame Core, is an ORM based on. net standard platform. The difference is that one is a LINQ system and the other is xml-centric. It is also the two ORM factions in all languages today. Each one has a fairly wide user base and application scenarios.

2. The Agent for SmartSql is Easy to manage which doesn't need much modification, and Package for Diagnostic is quite stable. It also has been approved and released in [SkyApm](https://github.com/apache/skywalking/blob/48ffd0923941e61df778a8fa6588f7a87c05728f/oap-server/server-core/src/test/resources/component-libraries.yml#L258). 
Management in your repository should be the best destination. I can't find any other repository to place this package. I also have time to participate in the maintenance and update of this package (Release is actually as simple as other packages).

3. Of course I will to contribute to this community and help apm-agent-dotnet go further.

4. I don't quite understand what you said about the lack of bandwidth.

5. Thank you for joining me in discussing the issues arising from this submission. If we can't meet your requirements for this merger, I can only say that we have done our best. Thank you again!

RocherKong:
ok，I got it. I‘ll move it to another project.

Issue #245, opened 2019-05-30
An automatically created span should be a child span of the current one


SergeyKleyman:
After we introduced support for child spans we didn't adapt code that automatically creates spans. Currently we automatically create spans for HTTP client and Entity Framework Core client. An automatically created span is always generated as a top level span while the expected behavior is for it to be a child span of the current one (in the same async/thread context).

See calls to `StartSpanInternal` in [EfCoreDiagnosticListener.cs](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm.EntityFrameworkCore/EfCoreDiagnosticListener.cs#L34)  and [HttpDiagnosticListenerImplBase.cs](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerImplBase.cs#L131)

SergeyKleyman:
Implemented in #391

Issue #180, opened 2019-04-11
Agent.Setup is never called in non ASP.NET Core scenarios


gregkalapos:
E.g. when only the public API is used `Agent.Setup()` is never called.

Until this method is not called, only a few things are initialized. E.g. the logger won't read the environment variables and the log level is ignored.

We should decide if
- the agent should work with 0 config, so without explicitly calling `Agent.Setup()` we initialize it during the first usage (whatever 'first usage' means) - I personally would prefer this one.
- or we document this and ask people to call it 😞

gregkalapos:
I take this. cc @SergeyKleyman

SergeyKleyman:
I did some refactoring in `AgentComponents` as part of work on #308 and #245 so it might be easier to wait with this one to avoid complicated merges.

gregkalapos:
Ok, I can wait.

Do you maybe have a branch to take a look at it? I'm already thinking about this stuff and I'd be interested.

SergeyKleyman:
NP - I'll make draft PR today.

SergeyKleyman:
https://github.com/elastic/apm-agent-dotnet/pull/391

gregkalapos:
This works now. [Here](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm/Agent.cs#L64) we instantiate a new `AgentComponent` in case the current one is `null`. And [that defaults](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm/AgentComponents.cs#L28) to an `EnvironmentConfigurationReader`. 

Tried it, it works.

PR #391, opened 2019-07-19
Track the current span: add public API, fix automatically created spans

### LINKED ISSUE #308: Add API method/property to get current span
Currently there is no way to get current span only current transaction

Add option to get current span like it is possible to get `CurrentTransaction` a new getter should probably be added `CurrentSpan`. 

**Additional context**
Something like it is possible in Java API - https://www.elastic.co/guide/en/apm/agent/java/master/public-api.html#api-current-span

This could probably be requirement to fix elastic/apm-agent-dotnet#245

SergeyKleyman:
Closes #308 and #245

- [x] Added public API to get the current span (#308)
- [x] Make an automatically created span a child span of the current one (#245)
- [x] Updated public API documentation


src/Elastic.Apm/Helpers/IntExtensions.cs
```diff
@@ -15,5 +16,11 @@ public static void Repeat(this int repeatCount, Action<int> action)
 			for (var i = 0; i < repeatCount; ++i)
 				action(i);
 		}
+
+		public static async Task Repeat(this int repeatCount, Func<int, Task> action)
```
gregkalapos: I think every usage from this class is only in test projects. Would it maybe better to move it to `Elastic.Apm.Tests`? 

...if I overlooked some prod usage, then nevermind.

test/Elastic.Apm.AspNetCore.Tests/TransactionNameTests.cs
```diff
@@ -23,7 +24,10 @@ public class TransactionNameTests : IClassFixture<WebApplicationFactory<Startup>
 		public TransactionNameTests(WebApplicationFactory<Startup> factory)
 		{
 			_factory = factory;
-			_agent = new ApmAgent(new TestAgentComponents(payloadSender: _payloadSender));
+
+			AgentSingletonUtils.EnsureInstanceCreated();
```
gregkalapos: Is this needed? I think there is no relationship between what this does to the static instance and the `_agent = new ApmAgent` we create in the next line. - but fine by me, in https://github.com/elastic/apm-agent-dotnet/issues/180 we can come back to this.

src/Elastic.Apm.AspNetCore/DiagnosticListener/AspNetCoreDiagnosticListener.cs
```diff
@@ -25,7 +24,7 @@ public void OnNext(KeyValuePair<string, object> kv)
 
 			var exception = kv.Value.GetType().GetTypeInfo().GetDeclaredProperty("exception").GetValue(kv.Value) as Exception;
 
-			var transaction = Agent.TransactionContainer.Transactions?.Value;
+			var transaction = _agent.Tracer.CurrentTransaction;
```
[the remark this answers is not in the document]
SergeyKleyman: 10x

test/Elastic.Apm.AspNetCore.Tests/TransactionNameTests.cs
SergeyKleyman: I've added a comment to clarify.
I'm not sure #180 is related to this - here it's all about `CurrentExecutionSegmentsContainer`.

src/Elastic.Apm/Helpers/IntExtensions.cs
SergeyKleyman: Yes, `IntExtensions` is used only in tests - I've moved it to `Elastic.Apm.Tests`.

sample/SampleAspNetCoreApp/Controllers/HomeController.cs
```diff
@@ -43,17 +45,33 @@ public async Task<IActionResult> Index()
 		}
 
 		[HttpPost]
-		public async Task<IActionResult> AddSampleData([FromForm] string enteredName)
+		public Task<IActionResult> AddSampleData(IFormCollection formFields)
```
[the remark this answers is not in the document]
SergeyKleyman: It's a good idea - I've added the comment.

test/Elastic.Apm.AspNetCore.Tests/TransactionNameTests.cs
gregkalapos: Oh, now I see... 👍

sample/SampleAspNetCoreApp/Controllers/HomeController.cs
```diff
@@ -23,25 +23,41 @@ public class HomeController : Controller
 
 		public HomeController(SampleDataContext sampleDataContext) => _sampleDataContext = sampleDataContext;
 
-		public async Task<IActionResult> Index()
+		private bool GetCaptureControllerActionAsSpanFromQueryString()
```
gregkalapos: Why don't we just use `AddSampleData` for this test? I thought that's done for testing that.

gregkalapos: The question applies to the `Index` method...

Och.. yeah.. that's HTTP post, and this is easier to test, right? 

But then why do we use `SafeCaptureSpan<IActionResult>` in `AddSampleData`?

SergeyKleyman: I've added `SafeCaptureSpan<IActionResult>` to `AddSampleData` at the point when I mistakenly thought that there are no automatic tests for Entity Framework Core so I used `AddSampleData` to verify manually. After you told me that there are automatic Entity Framework Core tests I thought it would be a good idea to add test for the code in this PR affecting Entity Framework Core auto captured spans.

gregkalapos: I see, yeah.

But that means the `AddSampleData` is not really utilized in this PR, right? ...with the comment I think it's fine to leave it as it is. I just thought we should use the existing method for testing this, but given it's a POST now I also think just adding another one is easier.

SergeyKleyman: We can try to use `AddSampleData` for tests as well and yes it will take more effort to simulate form submission but I'm more concerned about DB growing after each test run.

gregkalapos: > We can try to use AddSampleData for tests as well
I think we are good for now. I don't see it needed within this PR.

Issue #50, opened 2019-01-04
Bug: service name detection fails when the app is called from unmanaged code (e.g. IIS)


gregkalapos:
If the user does not give us a service name we automatically use the name of the entry assembly:

```
            get
            {
                if(service == null)
                {
                    service = new Service
                    {
                        Name = Assembly.GetEntryAssembly()?.GetName()?.Name,
                        Agent = new Model.Payload.Agent
                        {
                            Name = Consts.AgentName,
                            Version = Consts.AgentVersion
                        }
                    };
                }

                return service;
            }
```

This bugs, because `Assembly.GetEntryAssembly()` can return null as the [docu](
https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.getentryassembly?view=netframework-4.7.2) says:

> The GetEntryAssembly method can return null when a managed assembly has been loaded from an unmanaged application. 

This happens e.g. on Full Framework when the app runs in IIS.

gregkalapos:
First idea: We could just walk up the stack and look for the first non elastic frame. This prototype code uses the assembly name, once we sign our assemblies we should compare the public key tokens, that imo would be totally reliable.

```
if(String.IsNullOrEmpty(service.Name))
{
    var stackFrames = new StackTrace().GetFrames();

    foreach (var frame in stackFrames)
    {
        var currentAssembly = frame.GetMethod()?.DeclaringType.Assembly.GetName().Name;
        if (!String.IsNullOrEmpty(currentAssembly) && !currentAssembly.StartsWith("Elastic.Apm"))
        {
            service.Name = currentAssembly;
            break;
        }
    }

    service.Name = "unknown";
}

```

Comments:

- The Service instance is only calculated once, so this would only happen once, which is acceptable from the performance perspective.
- I know that with some application architectures we may not hit the correct assembly. I'd say it's impossible to select the right one, so the behaviour is that if users don't set the service name we set it to something which is more meaningful than "unknown" or something like that, but of course it won't be always perfect. On the other hand, users are free to set it manually if they are not happy with our heuristic.

rushtu22:
Hello, 

My application name is BL.Mines.API so Assembly.GetEntryAssembly()?.GetName()?.Name get this name, but APM throwing validation error because of dots. So please add ServiceName to the configuration file. 

Error is data validation error: Problem validating JSON document against schema: I[#] S[#] doesn't validate with "transaction#"
[...]

PR #393, opened 2019-07-19
Add ability to specify service version

### LINKED ISSUE #198: Add support  for Service Version configuration
Today I can set some data about the service, like service name and service URLs, but I can't set the service version.

Solution:
Add a field Version to Service class, get this data from IConfigurationReader, and use this data when create the json.

vhatsura:
to close #198

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -125,8 +125,8 @@ of the deployed revision, e.g. the output of git rev-parse HEAD.
 
 [options="header"]
 |============
-| Default                          | Type
-| Version of the entry assembly    | string 
+| Default                                        | Type
+| Informational version of the entry assembly    | string 
 |============
 
 [float]
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -110,6 +110,26 @@ NOTE: The service name must conform to this regular expression: ^[a-zA-Z0-9 _-]+
 |============
 
 [float]
+[[config-service-version]]
+==== `ServiceVersion`
+
+This is used to know exact version of your service.
+
+NOTE: The service version must conform to this regular expression: ^[0-9]+\.[0-9]+(?:\.[0-9]+)?(?:\.[0-9]+)?$. In less regexy terms: Your service version must contain only numbers (from 2 to 4), which are separated by `.` (dot).
+
+[options="header"]
+|============
+| Environment variable name      | IConfiguration key 
+| `ELASTIC_APM_SERVICE_VERSION` | `ElasticApm:ServiceVersion`
+|============
+
+[options="header"]
+|============
+| Default                          | Type
+| Version of the entry assembly    | Version 
+|============
+
+[float]
 [[config-transaction-sample-rate]]
 ==== `TransactionSampleRate`
 
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -113,9 +113,9 @@ NOTE: The service name must conform to this regular expression: ^[a-zA-Z0-9 _-]+
 [[config-service-version]]
 ==== `ServiceVersion`
 
-This is used to know exact version of your service.
-
-NOTE: The service version must conform to this regular expression: ^[0-9]+\.[0-9]+(?:\.[0-9]+)?(?:\.[0-9]+)?$. In less regexy terms: Your service version must contain only numbers (from 2 to 4), which are separated by `.` (dot).
+A version string for the currently deployed version of the service. If you don’t
+version your deployments, the recommended value for this field is the commit identifier
+of the deployed revision, e.g. the output of git rev-parse HEAD.
 
 [options="header"]
 |============
@@ -126,7 +126,7 @@ NOTE: The service version must conform to this regular expression: ^[0-9]+\.[0-9
 [options="header"]
 |============
 | Default                          | Type
-| Version of the entry assembly    | Version 
+| Version of the entry assembly    | string 
 |============
 
 [float]
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -126,7 +126,7 @@ of the deployed revision, e.g. the output of git rev-parse HEAD.
 [options="header"]
 |============
 | Default                                        | Type
-| Informational version of the entry assembly    | string 
+| Informational version of the entry assembly    | String 
 |============
 
 [float]
--- a/src/Elastic.Apm/Api/Service.cs
+++ b/src/Elastic.Apm/Api/Service.cs
@@ -19,6 +19,7 @@ namespace Elastic.Apm.Api
 		[JsonConverter(typeof(TrimmedStringJsonConverter))]
 		public string Name { get; set; }
 
+		[JsonConverter(typeof(TrimmedStringJsonConverter))]
 		public string Version { get; set; }
 
 		public Runtime Runtime { get; set; }
--- a/src/Elastic.Apm/Api/Service.cs
+++ b/src/Elastic.Apm/Api/Service.cs
@@ -1,4 +1,5 @@
-﻿using System.Reflection;
+﻿using System;
+using System.Reflection;
 using Elastic.Apm.Config;
 using Elastic.Apm.Helpers;
 using Elastic.Apm.Logging;
@@ -18,11 +19,13 @@ namespace Elastic.Apm.Api
 		[JsonConverter(typeof(TrimmedStringJsonConverter))]
 		public string Name { get; set; }
 
+		public Version Version { get; set; }
+
 		public Runtime Runtime { get; set; }
 
 		public override string ToString() => new ToStringBuilder(nameof(Service))
 		{
-			{ "Name", Name }, { "Agent", Agent }, { "Framework", Framework }, { "Language", Language },
+			{ "Name", Name }, {"Version", Version}, { "Agent", Agent }, { "Framework", Framework }, { "Language", Language },
 		}.ToString();
 
 		internal static Service GetDefaultService(IConfigurationReader configurationReader, IApmLogger loggerArg)
@@ -31,6 +34,7 @@ namespace Elastic.Apm.Api
 			return new Service
 			{
 				Name = configurationReader.ServiceName,
+				Version = configurationReader.ServiceVersion,
 				Agent = new AgentC
 				{
 					Name = Consts.AgentName,
--- a/src/Elastic.Apm/Api/Service.cs
+++ b/src/Elastic.Apm/Api/Service.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using System.Reflection;
+﻿using System.Reflection;
 using Elastic.Apm.Config;
 using Elastic.Apm.Helpers;
 using Elastic.Apm.Logging;
--- a/src/Elastic.Apm/Api/Service.cs
+++ b/src/Elastic.Apm/Api/Service.cs
@@ -19,7 +19,7 @@ namespace Elastic.Apm.Api
 		[JsonConverter(typeof(TrimmedStringJsonConverter))]
 		public string Name { get; set; }
 
-		public Version Version { get; set; }
+		public string Version { get; set; }
 
 		public Runtime Runtime { get; set; }
 
--- a/src/Elastic.Apm/Config/AbstractConfigurationReader.cs
+++ b/src/Elastic.Apm/Config/AbstractConfigurationReader.cs
@@ -364,10 +364,7 @@ namespace Elastic.Apm.Config
 		{
 			var versionInConfig = kv.Value;
 
-			if (!string.IsNullOrEmpty(versionInConfig))
-			{
-				return versionInConfig;
-			}
+			if (!string.IsNullOrEmpty(versionInConfig)) return versionInConfig;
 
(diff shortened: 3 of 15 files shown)


docs/configuration.asciidoc
```diff
@@ -108,6 +108,26 @@ NOTE: The service name must conform to this regular expression: ^[a-zA-Z0-9 _-]+
 | Name of the entry assembly       | String 
 |============
 
+[float]
+[[config-service-version]]
+==== `ServiceVersion`
+
+This is used to know exact version of your service.
+
+NOTE: The service version must conform to this regular expression: ^[0-9]+\.[0-9]+(?:\.[0-9]+)?(?:\.[0-9]+)?$. In less regexy terms: Your service version must contain only numbers (from 2 to 4), which are separated by `.` (dot).
```
SergeyKleyman: Could you please clarify where these constrains are coming from? I don't see any constrains neither in [other agents documentation](https://docs.google.com/spreadsheets/d/1JJjZotapacA3FkHc2sv_0wiChILi3uKnkwLTjtBmxwU/) nor in [APM server API specification](https://github.com/elastic/apm-server/blob/6.5/docs/spec/service.json#L79).

docs/configuration.asciidoc
```diff
@@ -108,6 +108,26 @@ NOTE: The service name must conform to this regular expression: ^[a-zA-Z0-9 _-]+
 | Name of the entry assembly       | String 
 |============
 
+[float]
+[[config-service-version]]
+==== `ServiceVersion`
+
+This is used to know exact version of your service.
```
SergeyKleyman: Since this configuration option is common to all Elastic APM agents I recommend copying documentation from some other agent - for example from [Java agent](https://www.elastic.co/guide/en/apm/agent/java/current/config-core.html#config-service-version).

docs/configuration.asciidoc
```diff
@@ -108,6 +108,26 @@ NOTE: The service name must conform to this regular expression: ^[a-zA-Z0-9 _-]+
...
+[options="header"]
+|============
+| Environment variable name      | IConfiguration key 
+| `ELASTIC_APM_SERVICE_VERSION` | `ElasticApm:ServiceVersion`
+|============
+
+[options="header"]
+|============
+| Default                          | Type
+| Version of the entry assembly    | Version
```
SergeyKleyman: All [other Elastic APM agents](https://docs.google.com/spreadsheets/d/1JJjZotapacA3FkHc2sv_0wiChILi3uKnkwLTjtBmxwU/) don't have default value for this configuration option and it's not mandatory [APM Server API specification](https://github.com/elastic/apm-server/blob/6.5/docs/spec/service.json#L79) so I recommend to align with other agents' approach.

docs/configuration.asciidoc
vhatsura: Ouch. I didn't look into documentation. My fault, will fix it.

docs/configuration.asciidoc
vhatsura: Will fix it

vhatsura: will fix it

docs/configuration.asciidoc
vhatsura: Ouch. I didn't look into documentation. My fault, will fix it.

docs/configuration.asciidoc
vhatsura: I believe we need to discuss it in details. From user perspective I want to use service version to determine exact version of my application, which has some issues. And with this we don't any problems in case of Elastic.Apm. However, we need to look into set up process of service version. In my own experience, we always set proper version for assemblies during CI process (lately, [`GitVersion`](https://github.com/GitTools/GitVersion) heps us with this). In case of we don't have default value, a different approach should be defined to set service version into configuration file during build or environment variable during deploy process. But this reason, I think it's nice to have default value for service version which is fetched from executing assembly.
@SergeyKleyman, what do you think? From one side, it's different behaviour from various agents, fro other one - is good for end users.

gregkalapos: Yeah, this is tricky - I'd have implemented it the same way @vhatsura as you did. But usually we default to what other agents do. I think for .NET devs it'd be better to have the version of the entry assembly as default though. I asked other agent devs what they think - will report back.

One technical note: if the process is hosted by a native application (e.g. IIS) then `Assembly.GetEntryAssembly()` returns null, so this default won't work for apps directly hosted in IIS - I guess typically ASP.NET Classic, or ASP.NET Core in process - but I think that's fine.

vhatsura: @gregkalapos, in case of Assembly.GetEntryAssembly() returns null, fallback to current stack trace will be involved (implemented as for service name).

gregkalapos: Yes, I see it, but that may have some drawbacks, see the end of this comment: https://github.com/elastic/apm-agent-dotnet/issues/50#issuecomment-455134362 

In case we do implement it with trying to have a default maybe in case of the service version we should only use it when `Assembly.GetEntryAssembly()` is not null. The service name is mandatory, so we needed to set it to something, but we know it's not always correct, in that case it can be corrected by setting it manually. Probably for an optional field skipping it would be better if we know that we just "guess" the version.

vhatsura: Yeap, make sense. Will stop on `Assembly.GetEntryAssembly`

SergeyKleyman: `service.name` is [mandatory for 6.5](https://github.com/elastic/apm-server/blob/6.5/docs/spec/service.json#L55) but [optional for 7.0](https://github.com/elastic/apm-server/blob/7.0/docs/spec/service.json#L52) so if we want to make it optional we will have to increase [APM Server version in compatibility matrix](https://www.elastic.co/guide/en/apm/get-started/7.3/agent-server-compatibility.html#dotnet-compatibility).
OTOH `service.version` is [optional at least from 6.5](https://github.com/elastic/apm-server/blob/6.5/docs/spec/service.json#L81)

gregkalapos: I asked devs from other agents and we tend to agree that it's fine to try to report a default version this way - most of them planning to have something similar. After dc3f68d we are ok here.

xitaocrazy:
I think that it is the same that I have opened in #198. Unfortunately, I had no time in lasts months and couldn't work on it, but I think that when this item be done,the #198 should be closed too.

gregkalapos:
> I think that it is the same that I have opened in #198. Unfortunately, I had no time in lasts months and couldn't work on it, but I think that when this item be done,the #198 should be closed too.

Right, it solved #198. Will close once this is merged 👍

vhatsura:
@gregkalapos, @SergeyKleyman. Do you have a chance to look into changes?
I changed behavior to retrieve default service version. Previously it was `Version` from `AssemblyName`. Now it's [`InformationalVersion` from `AssemblyInformationalAttribute`](https://docs.microsoft.com/en-us/dotnet/core/tools/csproj#assemblyinfo-properties)

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
```diff
@@ -345,6 +351,38 @@ protected string ParseServiceName(ConfigurationKeyValue kv)
...
+			var discoveredVersion = DiscoverServiceVersion();
+			if (discoveredVersion != null)
+			{
+				Logger?.Info()
+					?.Log("The agent was started without a service version. The automatically discovered service version is {ServiceVersion}",
+						discoveredVersion);
+				return discoveredVersion;
+			}
+
+			return string.Empty;
```
gregkalapos: 2 small things to this:
- Could we also add an info log before we return here? Otherwise in the logs in case the `if` block above is not entered we'll just print the `...version will be automatically discovered.` text, but won't send any version, which could be confusing.
- Any reason we return `string.Empty` here? Other methods return `null`. The general concept (`null` vs `Empty`) can be of course discussed, but I think mixing them (so having some methods returning `null`, some `Empty`) is I think the least favorable option.

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
```diff
@@ -405,15 +443,15 @@ internal static bool IsMsOrElastic(byte[] array)
 
 			for (var i = 0; i < 8; i++)
 			{
-				if (array[i] != elasticToken[i])
+				if (isElasticApm && array[i] != elasticToken[i])
```
gregkalapos: Are these changes with `and`-ing the array item with the `bool` variable meant to be an optimization?

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
```diff
@@ -345,6 +351,38 @@ protected string ParseServiceName(ConfigurationKeyValue kv)
...
+
+			return null;
+		}
+
+		protected string ParseServiceVersion(ConfigurationKeyValue kv)
+		{
+			var versionInConfig = kv.Value;
+
+			if (!string.IsNullOrEmpty(versionInConfig))
+			{
```
gregkalapos: codestlye: `{` `}` could be removed.

vhatsura: fixed

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
vhatsura: it was a kind of. After deeper analysis with benchmarks it doesn't make sense: differences within the limit of error

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
vhatsura: fixed

docs/configuration.asciidoc
```diff
@@ -109,6 +109,26 @@ NOTE: The service name must conform to this regular expression: ^[a-zA-Z0-9 _-]+
...
+[options="header"]
+|============
+| Environment variable name      | IConfiguration key 
+| `ELASTIC_APM_SERVICE_VERSION` | `ElasticApm:ServiceVersion`
+|============
+
+[options="header"]
+|============
+| Default                                        | Type
+| Informational version of the entry assembly    | string
```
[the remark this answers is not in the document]

src/Elastic.Apm/Api/Service.cs
```diff
@@ -1,4 +1,5 @@
-﻿using System.Reflection;
+﻿using System;
```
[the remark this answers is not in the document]
vhatsura: fixed

vhatsura:
@gregkalapos, thanks for review. I fixed all your comments

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
gregkalapos: I see. Yeah, I agree with your decision.

Also huge 👍 for measuring it!

gregkalapos:
We had 2 failing tests, the main issue is that in the context of XUnit tests the service detection does not work - it's one of those cases when the EntryAssembly is one of the known Microsoft assemblies and we skip reporting that as the service - which makes sense, so I'd suggest we adapt the tests accordingly.

- `Elastic.Apm.Tests.HttpDiagnosticListenerTests.OnNextWithStartAndStopTwice`: It's just because of the new log suggested [here](https://github.com/elastic/apm-agent-dotnet/pull/393#discussion_r315270198). I adapted the test, so this one is fixed.
- ` Elastic.Apm.AspNetCore.Tests.AspNetCoreMiddlewareTests.HomeSimplePageTransactionTest`: this one is also because of [that](https://github.com/elastic/apm-agent-dotnet/pull/393#discussion_r315270198) suggestion, this time because previously we returned `String.Empty`, now we return `null`, so `_agent.Service.Version.Should().NotBeNull();` fails. I think that assert was a little bit unfortunate anyway, because it passed with `String.Empty`, and probably that's not what we wanted to have there. Given that we have other tests covering this (e.g. `ReadServiceVersionViaEnvironmentVariable` and some more) I'd say let's just not assert on service version in this test.

gregkalapos:
We have some CI troubles - this PR is ready to merge, I just wait for a green CI - current CI failures are infrastructure related, the PR itself should be ok.

vhatsura:
> We had 2 failing tests, the main issue is that in the context of XUnit tests the service detection does not work - it's one of those cases when the EntryAssembly is one of the known Microsoft assemblies and we skip reporting that as the service - which makes sense, so I'd suggest we adapt the tests accordingly.
> 
> * `Elastic.Apm.Tests.HttpDiagnosticListenerTests.OnNextWithStartAndStopTwice`: It's just because of the new log suggested [here](https://github.com/elastic/apm-agent-dotnet/pull/393#discussion_r315270198). I adapted the test, so this one is fixed.
> * ` Elastic.Apm.AspNetCore.Tests.AspNetCoreMiddlewareTests.HomeSimplePageTransactionTest`: this one is also because of [that](https://github.com/elastic/apm-agent-dotnet/pull/393#discussion_r315270198) suggestion, this time because previously we returned `String.Empty`, now we return `null`, so `_agent.Service.Version.Should().NotBeNull();` fails. I think that assert was a little bit unfortunate anyway, because it passed with `String.Empty`, and probably that's not what we wanted to have there. Given that we have other tests covering this (e.g. `ReadServiceVersionViaEnvironmentVariable` and some more) I'd say let's just not assert on service version in this test.

totally forgot about unit tests. Thanks for updating them

PR #394, opened 2019-07-19
thread-safety for increment of started and dropped span count

### LINKED ISSUE #184: Started/dropped spans counters should be thread-safe
Currently `started` span counter is a regular (i.e., non-atomic) integer and it's incremented without any thread-safety control (lock, etc.) which might cause a race condition if spans for the same transaction are created on more than one thread.
Currently `dropped` count is not incremented since we still don't support `transaction_max_spans` configuration option elastic/apm-agent-dotnet#178

vhatsura:
to close #184

### DIFF
--- a/src/Elastic.Apm/Model/Span.cs
+++ b/src/Elastic.Apm/Model/Span.cs
@@ -48,7 +48,7 @@ namespace Elastic.Apm.Model
 
 			if (IsSampled)
 				// Started spans should be counted only for sampled transactions
-				enclosingTransaction.SpanCount.Started++;
+				enclosingTransaction.SpanCount.IncrementStarted();
 
 			_logger.Trace()
 				?.Log("New Span instance created: {Span}. Start time: {Time} (as timestamp: {Timestamp})",
--- a/src/Elastic.Apm/Model/SpanCount.cs
+++ b/src/Elastic.Apm/Model/SpanCount.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Threading;
 using Elastic.Apm.Api;
 using Elastic.Apm.Helpers;
 
@@ -6,13 +7,15 @@ namespace Elastic.Apm.Model
 {
 	internal class SpanCount
 	{
-		public int Started { get; set; }
-		public int Dropped { get; set; }
+		private int _started;
+		private int _dropped;
+		public int Started => _started;
+		public int Dropped => _dropped;
 
-		public override string ToString() => new ToStringBuilder(nameof(SpanCount))
-		{
-			{ "Started", Started },
-			{ "Dropped", Dropped },
-		}.ToString();
+		public void IncrementStarted() => Interlocked.Increment(ref _started);
+
+		public void IncrementDropped() => Interlocked.Increment(ref _dropped);
+
+		public override string ToString() => new ToStringBuilder(nameof(SpanCount)) { { "Started", Started }, { "Dropped", Dropped }, }.ToString();
 	}
 }


SergeyKleyman:
> LGTM too.
> 
> Also measured the perf. difference since we have `Interlocked.Increment` here - no measurable perf diff with 1000 spans on a single thread.

It's all about contention - the more contention there is between cores accessing these counters the bigger performance degradation there will be. 
Of course there is also a potential of [false sharing](http://joeduffyblog.com/2009/10/19/false-sharing-is-no-fun/) - I don't know what are the chances of compacting GC placing counters from different transactions on the same cache line but if we want to be on the safe side we can add padding as explained [here](https://www.codeproject.com/Articles/51553/Concurrency-Hazards-False-Sharing).

vhatsura:
> > LGTM too.
> > Also measured the perf. difference since we have `Interlocked.Increment` here - no measurable perf diff with 1000 spans on a single thread.
> 
> It's all about contention - the more contention there is between cores accessing these counters the bigger performance degradation there will be.
> Of course there is also a potential of [false sharing](http://joeduffyblog.com/2009/10/19/false-sharing-is-no-fun/) - I don't know what are the chances of compacting GC placing counters from different transactions on the same cache line but if we want to be on the safe side we can add padding as explained [here](https://www.codeproject.com/Articles/51553/Concurrency-Hazards-False-Sharing).

All of Interlocked’s methods generate a [full fence](http://www.albahari.com/threading/part4.aspx#_Interlocked), which prevents any kind of instruction reordering or caching around that fence. By this reason, I believe we can forget about cache lines.

SergeyKleyman:
> All of Interlocked’s methods generate a [full fence](http://www.albahari.com/threading/part4.aspx#_Interlocked), which prevents any kind of instruction reordering or caching around that fence. By this reason, I believe we can forget about cache lines.

All what you described above is about correctness - and what you did does work correctly. False sharing is not about correctness - it's about performance degradation. I recommend you to read about it.

gregkalapos:
😂 All I wanted to say is that there is no measurable perf diff. on a single thread - did not measured with multiple threads, since that bugged anyway, so there was nothing to make worse for that. So in sum: 👍 👍 

Let's merge this! 🚀 🚀

gregkalapos:
_assigning myself just to not overlook this during planning meetings_

Thanks for the work @vhatsura.

SergeyKleyman:
> 😂 All I wanted to say is that there is no measurable perf diff. on a single thread - did not measured with multiple threads, since that bugged anyway, so there was nothing to make worse for that. So in sum: 👍 👍

It's not entirely correct - the use case that worked incorrectly is transaction creating spans on multiple threads simultaneously. The most wide spread use case with transaction running only one thread simultaneously (possibly switching threads because of `async`/`await`) worked correctly but now there's potential that unrelated transactions running on different CPU cores will contend over the same CPU cache line and cause degradation in performance in all of the contending transactions.

SergeyKleyman:
@vhatsura 
> ... or caching around that fence

In case you are interested in the topic - it's not correct that memory fences prohibit use of cache. In modern CPU architecture CPU core always accesses main memory via all the levels of cache (L1, L2, etc.) To make sure CPU cores see coherent picture of the main memory even though each core has its own cache (at some level(s)) there is [cache coherence](https://en.wikipedia.org/wiki/Cache_coherence) between cores. OTOH cache coherence is the reason false sharing causes performance degradation - each core will synchronize with all others on any change to the cache line even though each core reads/writes to a different part of the cache line.

Modern CPU architecture is not a simple thing and unfortunately it's not abstracted away from the programmers especially if you want to reason about performance.

vhatsura:
@SergeyKleyman , now I got it what you said. Thanks for detailed clarification

PR #397, opened 2019-07-19
add user agent header to payload sender

### LINKED ISSUE #377: Standardize user-agent header
See https://github.com/elastic/apm/issues/88

vhatsura:
to close #377 

Implemented according to https://github.com/elastic/apm/issues/88#issuecomment-491192841

### DIFF
--- a/src/Elastic.Apm/Report/PayloadSenderV2.cs
+++ b/src/Elastic.Apm/Report/PayloadSenderV2.cs
@@ -56,7 +56,7 @@ namespace Elastic.Apm.Report
 
 			_httpClient = new HttpClient(handler ?? new HttpClientHandler()) { BaseAddress = serverUrlBase };
 			_httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue($"elasticapm-{Consts.AgentName}", _service.Agent.Version));
-			_httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue("http-lib",
+			_httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue("system.net.http",
 				typeof(HttpClient).Assembly.GetName().Version.ToString()));
 
 			if (configurationReader.SecretToken != null)
--- a/src/Elastic.Apm/Report/PayloadSenderV2.cs
+++ b/src/Elastic.Apm/Report/PayloadSenderV2.cs
@@ -55,7 +55,7 @@ namespace Elastic.Apm.Report
 			servicePoint.ConnectionLimit = 20;
 
 			_httpClient = new HttpClient(handler ?? new HttpClientHandler()) { BaseAddress = serverUrlBase };
-			_httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue(new ProductHeaderValue("elasticapm-dotnet", _service.Agent.Version)));
+			_httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue(new ProductHeaderValue($"elasticapm-{Consts.AgentName}", _service.Agent.Version)));
 
 			if (configurationReader.SecretToken != null)
 			{
--- a/src/Elastic.Apm/Report/PayloadSenderV2.cs
+++ b/src/Elastic.Apm/Report/PayloadSenderV2.cs
@@ -55,7 +55,9 @@ namespace Elastic.Apm.Report
 			servicePoint.ConnectionLimit = 20;
 
 			_httpClient = new HttpClient(handler ?? new HttpClientHandler()) { BaseAddress = serverUrlBase };
-			_httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue(new ProductHeaderValue($"elasticapm-{Consts.AgentName}", _service.Agent.Version)));
+			_httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue($"elasticapm-{Consts.AgentName}", _service.Agent.Version));
+			_httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue("http-lib",
+				typeof(HttpClient).Assembly.GetName().Version.ToString()));
 
 			if (configurationReader.SecretToken != null)
 			{
--- a/src/Elastic.Apm/Report/PayloadSenderV2.cs
+++ b/src/Elastic.Apm/Report/PayloadSenderV2.cs
@@ -55,6 +55,7 @@ namespace Elastic.Apm.Report
 			servicePoint.ConnectionLimit = 20;
 
 			_httpClient = new HttpClient(handler ?? new HttpClientHandler()) { BaseAddress = serverUrlBase };
+			_httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue(new ProductHeaderValue("elasticapm-dotnet", _service.Agent.Version)));
 
 			if (configurationReader.SecretToken != null)
 			{
--- a/src/Elastic.Apm/Report/PayloadSenderV2.cs
+++ b/src/Elastic.Apm/Report/PayloadSenderV2.cs
@@ -7,6 +7,7 @@ using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Reflection;
 using System.Text;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Threading.Tasks.Dataflow;
@@ -42,7 +43,9 @@ namespace Elastic.Apm.Report
 		private readonly SingleThreadTaskScheduler _singleThreadTaskScheduler = new SingleThreadTaskScheduler(CancellationToken.None);
 		private readonly Metadata _metadata;
 
-		public PayloadSenderV2(IApmLogger logger, IConfigurationReader configurationReader, Service service, Api.System system, HttpMessageHandler handler = null)
+		public PayloadSenderV2(IApmLogger logger, IConfigurationReader configurationReader, Service service, Api.System system,
+			HttpMessageHandler handler = null
+		)
 		{
 			_service = service;
 			_system = system;
@@ -56,9 +59,12 @@ namespace Elastic.Apm.Report
 			servicePoint.ConnectionLimit = 20;
 
 			_httpClient = new HttpClient(handler ?? new HttpClientHandler()) { BaseAddress = serverUrlBase };
-			_httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue($"elasticapm-{Consts.AgentName}", _service.Agent.Version));
-			_httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue("system.net.http",
-				typeof(HttpClient).Assembly.GetCustomAttribute<AssemblyFileVersionAttribute>().Version));
+			_httpClient.DefaultRequestHeaders.UserAgent.Add(
+				new ProductInfoHeaderValue($"elasticapm-{Consts.AgentName}", AdaptUserAgentValue(_service.Agent.Version)));
+			_httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue("System.Net.Http",
+				AdaptUserAgentValue(typeof(HttpClient).Assembly.GetCustomAttribute<AssemblyFileVersionAttribute>().Version)));
+			_httpClient.DefaultRequestHeaders.UserAgent.Add(
+				new ProductInfoHeaderValue(AdaptUserAgentValue(_service.Runtime.Name), AdaptUserAgentValue(_service.Runtime.Version)));
 
 			if (configurationReader.SecretToken != null)
 			{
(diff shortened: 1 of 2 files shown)


src/Elastic.Apm/Report/PayloadSenderV2.cs
```diff
@@ -55,6 +55,7 @@ public PayloadSenderV2(IApmLogger logger, IConfigurationReader configurationRead
 			servicePoint.ConnectionLimit = 20;
 
 			_httpClient = new HttpClient(handler ?? new HttpClientHandler()) { BaseAddress = serverUrlBase };
+			_httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue(new ProductHeaderValue("elasticapm-dotnet", _service.Agent.Version)));
```
SergeyKleyman: Shouldn't we add `http-lib/version` and `language/version` parts while we are at it? How much more effort would you estimate it is?
I think for 
- `http-lib/version` we can use `System.Net.Http.HttpClient`/<System.Net.Http.HttpClient assembly version?>
- `language/version` we can use [`Elastic.Apm.Api.Runtime.Name`](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm/Api/Service.cs#L81)/[`Elastic.Apm.Api.Runtime.Version`](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm/Api/Service.cs#L84)

What do you think?

SergeyKleyman: I recommend to use `$"elasticapm-{Elastic.Apm.Consts.AgentName}"` instead of `"elasticapm-dotnet"`

test/Elastic.Apm.Tests/BasicAgentTests.cs
```diff
@@ -87,6 +87,33 @@ public async Task PayloadSentWithBearerToken()
...
+			var payloadSender = new PayloadSenderV2(logger, new TestAgentConfigurationReader(logger),
+				service, new Api.System(), handler);
+
+			using (var agent = new ApmAgent(new TestAgentComponents(payloadSender: payloadSender)))
+			{
+				agent.PayloadSender.QueueTransaction(new Transaction(agent, "TestName", "TestType", new TestAgentConfigurationReader(logger)));
+			}
+
+			await isRequestFinished.Task;
+			userAgentHeader.Should().ContainSingle(x => x.Product.Name == "elasticapm-dotnet" && x.Product.Version == service.Agent.Version);
```
SergeyKleyman: I recommend to enforce the order and verify it in the test. Based on [RFC 2616](https://tools.ietf.org/html/rfc2616#section-14.43)
> By convention, the product tokens are listed in order of their significance for identifying the application.

Based on https://github.com/elastic/apm/issues/88#issuecomment-491192841 `elasticapm-$language/$agent_version` should be the first.

src/Elastic.Apm/Report/PayloadSenderV2.cs
vhatsura: Ops, I didn't see this constant. Will change it

test/Elastic.Apm.Tests/BasicAgentTests.cs
vhatsura: fixed

src/Elastic.Apm/Report/PayloadSenderV2.cs
vhatsura: fixed

vhatsura: I didn't add `http-lib/version` and `language/version` due to there aren't implemented in all agents. However, it isn't a problem. Will do it,

vhatsura: @SergeyKleyman, I need a quick consult with you about `language/version`. `ProductInfoHeaderValue` has a validation of each character. This validation can be found at [`HttpRuleParser`](https://github.com/dotnet/corefx/blob/e64cac6dcacf996f98f0b3f75fb7ad0c12f588f7/src/System.Net.Http/src/System/Net/Http/HttpRuleParser.cs#L63).

`Elastic.Apm.Api.Runtime.Name` can be `.NET Core`, `.NET Framework` or `N/A` which contains illegal symbols (space and `/`). Resolution of this can be done by 3 ways:

1. analyze each character
```csharp
string AdaptProductName(string productName)
{
    var x = new HashSet<char> { ' ', '(', ')', '<', '>', '@', ',', ';', ':', '\\', '"', '/', '[', ']', '?', '=', '{', '}', (char)127 /*DEL*/ };

    var chars = productName.ToCharArray();

    for (var i = 0; i < chars.Length; i++)
    {
        if (x.Contains(chars[i]))
	    chars[i] = '-';
    }
    return new string(chars);
}
```
 
2. use `Regex.Replace`
3. use `Elastic.Apm.Api.Runtime.Name.Replace(' ', '-').Replace('/', '-')` 

What is the preferable way?

SergeyKleyman: I would just set `User-Agent` header to a raw string instead of using `ProductInfoHeaderValue`, for example as suggested [here](https://stackoverflow.com/questions/43215527/c-sharp-httpclient-adding-user-agent-header-shows-up-as-several-different-head).

SergeyKleyman: I thought a little bit more about `http-lib/version` part (for which we can use `System.Net.Http.HttpClient`/<System.Net.Http.HttpClient assembly version?>) and it seems to me it will be redundant for .NET agent since we use HTTP client provided by the runtime and not a 3rd party library so if we know .NET runtime name/version (which will be in `language`/`version`) that will be enough. What do you think?

vhatsura: > I would just set `User-Agent` header to a raw string instead of using `ProductInfoHeaderValue`, for example as suggested [here](https://stackoverflow.com/questions/43215527/c-sharp-httpclient-adding-user-agent-header-shows-up-as-several-different-head).

It can be the case, however, we still need to check all strings for spaces and `/`, because [`space` symbol is used as separator between sub-products](https://tools.ietf.org/html/rfc2616#section-3.8) and `/` symbol is used to separate product name with product version. I don't find mention of other symbols which are included in `HttpRuleParser`, but I'm afraid it's reasonable. Will investiagate it deeper

vhatsura: > I thought a little bit more about `http-lib/version` part (for which we can use `System.Net.Http.HttpClient`/<System.Net.Http.HttpClient assembly version?>) and it seems to me it will be redundant for .NET agent since we use HTTP client provided by the runtime and not a 3rd party library so if we know .NET runtime name/version (which will be in `language`/`version`) that will be enough. What do you think?

As I understand, `System.Net.Http` can be references as a [NuGet package](https://www.nuget.org/packages/System.Net.Http/) in application host and by this reason `.NET runtime name/version` cannot guaranteed correct map.

SergeyKleyman: I see. In that case let’s include it.

SergeyKleyman: > > I would just set `User-Agent` header to a raw string instead of using `ProductInfoHeaderValue`, for example as suggested [here](https://stackoverflow.com/questions/43215527/c-sharp-httpclient-adding-user-agent-header-shows-up-as-several-different-head).
> 
> It can be the case, however, we still need to check all strings for spaces and `/`, because [`space` symbol is used as separator between sub-products](https://tools.ietf.org/html/rfc2616#section-3.8) and `/` symbol is used to separate product name with product version. I don't find mention of other symbols which are included in `HttpRuleParser`, but I'm afraid it's reasonable. Will investiagate it deeper

I see your point - there’re advantages to having this header easily machine parseable. I would go with your suggestion to replace invalid characters. I would replace spaces with underscores and slashes with dashes.

src/Elastic.Apm/Report/PayloadSenderV2.cs
```diff
@@ -55,6 +55,9 @@ public PayloadSenderV2(IApmLogger logger, IConfigurationReader configurationRead
 			servicePoint.ConnectionLimit = 20;
 
 			_httpClient = new HttpClient(handler ?? new HttpClientHandler()) { BaseAddress = serverUrlBase };
+			_httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue($"elasticapm-{Consts.AgentName}", _service.Agent.Version));
+			_httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue("http-lib",
```
SergeyKleyman: Judging by the following node.js example

> elastic-apm-node/2.6.0 elastic-apm-http-client/7.1.1 node/12.2.0

I don't think the intent was to have literally `http-lib` - I think `http-lib` is a placeholder for the actual HTTP client library name that agent uses. In .NET agent case it's `System.Net.Http`, WDTY?

test/Elastic.Apm.Tests/BasicAgentTests.cs
```diff
@@ -87,6 +87,38 @@ public async Task PayloadSentWithBearerToken()
...
+
+			HttpHeaderValueCollection<ProductInfoHeaderValue> userAgentHeader = null;
+			var handler = new MockHttpMessageHandler((r, c) =>
+			{
+				userAgentHeader = r.Headers.UserAgent;
+				isRequestFinished.SetResult(null);
+				return Task.FromResult(new HttpResponseMessage(HttpStatusCode.OK));
+			});
+
+			var logger = ConsoleLogger.Instance;
```
SergeyKleyman: A small nitpick - we have `NoopLogger` that we use in tests that don't care about what SUT writes to logs.

src/Elastic.Apm/Report/PayloadSenderV2.cs
```diff
@@ -55,6 +55,9 @@ public PayloadSenderV2(IApmLogger logger, IConfigurationReader configurationRead
...
+				typeof(HttpClient).Assembly.GetName().Version.ToString()));
```
SergeyKleyman: On my machine (.NET Core Runtime 2.1.11)  
- `typeof(HttpClient).Assembly.GetName().Version.ToString()` returns `4.2.1.0`
- `typeof(HttpClient).Assembly.GetCustomAttribute<AssemblyFileVersionAttribute>().Version` returns `4.6.27617.02`
- `typeof(HttpClient).Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion` returns `4.6.27617.02 @BuiltBy: dlab14-DDVSOWINAGE051 @Branch: release/2.1-MSRC @SrcCode: https://github.com/dotnet/corefx/tree/c6cf790234e063b855fcdb50f3fb1b3cf
ac73275`

Which makes me wonder what would be more useful to see  `4.2.1.0` or `4.6.27617.02`? WDTY?

vhatsura: So, I checked it with .Net Core 2.2 and 3.0 and got the following results:

```
.Net Core 3.0
4.2.1.0
4.700.19.30308
3.0.0-preview6.19303.8+d47cae744ddfb625db8e391cecb261e4c3d7bb1c

.Net Core 2.2
4.2.1.0
4.6.27618.01
4.6.27618.01 @BuiltBy: dlab14-DDVSOWINAGE054 @Branch: release/2.2-MSRC @SrcCode: https://github.com/dotnet/corefx/tree/9b5053f09fac4a2d615619197924072b9571da78
```

So, I agreed with you that assembly file version is more suitable for our use case. Will change it

src/Elastic.Apm/Report/PayloadSenderV2.cs
```diff
@@ -55,6 +56,9 @@ public PayloadSenderV2(IApmLogger logger, IConfigurationReader configurationRead
 			servicePoint.ConnectionLimit = 20;
 
 			_httpClient = new HttpClient(handler ?? new HttpClientHandler()) { BaseAddress = serverUrlBase };
+			_httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue($"elasticapm-{Consts.AgentName}", _service.Agent.Version));
+			_httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue("system.net.http",
```
SergeyKleyman: I would leave the original case i.e., `System.Net.Http` - it looks more .NET-y to me :)

src/Elastic.Apm/Report/PayloadSenderV2.cs
```diff
@@ -55,6 +56,9 @@ public PayloadSenderV2(IApmLogger logger, IConfigurationReader configurationRead
...
+				typeof(HttpClient).Assembly.GetCustomAttribute<AssemblyFileVersionAttribute>().Version));
```
SergeyKleyman: I think you forgot to append `language`/`version` (`Elastic.Apm.Api.Runtime.Name`/`Elastic.Apm.Api.Runtime.Version`) with the adaptation as you proposed at https://github.com/elastic/apm-agent-dotnet/pull/397#discussion_r305664084

test/Elastic.Apm.Tests/BasicAgentTests.cs
```diff
@@ -87,6 +87,38 @@ public async Task PayloadSentWithBearerToken()
...
+				agent.PayloadSender.QueueTransaction(new Transaction(agent, "TestName", "TestType", new TestAgentConfigurationReader(logger)));
+			}
+
+			await isRequestFinished.Task;
+			userAgentHeader.Should()
+				.NotBeEmpty()
+				.And.HaveCount(2)
+				.And.ContainInOrder(
+					new ProductInfoHeaderValue($"elasticapm-{Consts.AgentName}", service.Agent.Version),
+					new ProductInfoHeaderValue("http-lib", typeof(HttpClient).Assembly.GetName().Version.ToString()));
```
SergeyKleyman: You forgot to update the test.

vhatsura: updated

src/Elastic.Apm/Report/PayloadSenderV2.cs
vhatsura: changed

src/Elastic.Apm/Report/PayloadSenderV2.cs
vhatsura: I left it to last and decided to replace all invalid characters by `"[ /()<>@,:;={}?\\[\\]\"\\\\]"` regex because it doesn't look complicated but guaranteed valid header

vhatsura:
@SergeyKleyman, can you review the PR, again?

SergeyKleyman:
> @SergeyKleyman, can you review the PR, again?

I'm sorry for the delay - I'll review it today.

Issue #208, opened 2019-04-25
Exceptions in Agent's code should not escape to monitored application code 


SergeyKleyman:
I think it's worth considering a mechanism that prevents an exception in Agent's code from propagating to monitored application code and causing havoc.

We do need to distinguish between exception caused by some runtime condition from which agent can recover vs. bugs in agent's code that we should treat as critical events and take fail-fast approach. For example as mentioned in this case - https://github.com/elastic/apm-agent-dotnet/pull/374#discussion_r304429633

SergeyKleyman:
We need to make sure that all the agent's methods that can be called from application code are wrapped in `try`/`catch` that don't allow exceptions to escape.  One of the approaches is to implement static analysis tool (for example using Roslyn) that can be invoked as a unit test and fail the build if not all the required methods are wrapped as expected.

Please note that we still need to find a way to identify the methods that need to be guarded. All the public methods should be included but we should probably also include explicit implementations of interfaces from namespaces outside agent's and public interfaces from agent's namespaces.

PR #402, opened 2019-07-21
Capture request body aspnet core

### LINKED ISSUE #213: Capture request body in ASP.NET Core
Trigger by [this](https://github.com/elastic/apm-agent-dotnet/issues/117#issuecomment-486407571) comment.

In elastic/apm-agent-dotnet#134 we added the `Request.Body` field, but that is only utilized by API tests, so the only way to use it is to manually set the request body.

Other agents have a setting for this. E.g. see [here](https://www.elastic.co/guide/en/apm/agent/nodejs/current/configuration.html#capture-body).

We should add this config setting and extend our ASP.NET Core instrumentation to behave accordingly.

katzdan:
Closes #213

initial feature commit - 
Added config values + behavior in the Apm middle-ware class.
Errors body extraction still pending.

@SergeyKleyman , @gregkalapos 
As there are differences between the node.js and java docs - Iv'e kept the implementation to whatever is common :
 - the accepted values in 'captrueBody' config value can be 'off', 'all', 'errors' or 'transactions'
 - the request body, when extracted, is truncated after 2kb.
please let me know if anything else is required for this feature.

I added tests to verify this works along side a test that creates a 'post' http call ad in order for that to work properly I added a new controller that accepts post requests to the 'SampleAspNetCoreApp' project. There are other controllers that have post actions but these redirect the UI to another page and that would have made the tests more complicated.

There is one part of the implementation I don't know how to do - when the 'captureBody' is set to 'errors' than the apm module should retrieve the body from the request but there is no HttpContext in the error class or anywhere in the classes that handle this scenario. I'll need some guidance with this.


katzdan:
FIY - all the tests related to new feature run successfully in my machine (there is one that fails - register user test which expects a web service that's not running ). I'm not sure why there are conflicts. I have the latest version of master locally, branched of course, but still - up to date. There shouldn't be any conflicts.

gregkalapos:
@katzdan thanks for the PR! 

To address your question:
> There is one part of the implementation I don't know how to do - when the 'captureBody' is set to 'errors' than the apm module should retrieve the body from the request but there is no HttpContext in the error class or anywhere in the classes that handle this scenario. I'll need some guidance with this.

Maybe you already figured this out, but just in case: 
The `Error` uses the same [`Context`](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm/Model/Error.cs#L32) instance as the current transaction. Long term it'd be probably better to not share the same instance, but let's ignore it now - that's not directly related top this PR anyway. So, this `Context` instance has a [`Request`](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm/Api/Context.cs#L18) property and if you look at the [`Request` type](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm/Api/Request.cs#L17) you'll se that it's the same one that you already use.

There are 2 ways when we create errors in ASP.NET Core. 

- when ASP.NET Core fires an `Microsoft.AspNetCore.Diagnostics.UnhandledException` or `..HandledException` event, we handle this [here](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm.AspNetCore/DiagnosticListener/AspNetCoreDiagnosticListener.cs#L23) This happens when there is an error handler page registered.

As you can see in this screenshot you have access to the `HttpContext` there, similar to the exception you just need to read it with reflection:

image

- the other case is when our middleware catches the exception [here](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs#L92) This happens when there is not error handler page registered in the pipeline.

Now one thing you can already do is that before we catch the error you simply set the `transaction.Context.Request.Body` - in both cases you have the `HttpContext`, I guess this won't bee too much code.

A nicer way would be to give you access to the error before it's sent out, but that'd require more refactoring, and I think it'd better to do that in a separate PR. So probably the option described above would be more feasible. 

Hope this helps.

SergeyKleyman:
I recommend to test use cases with
- [Chunked transfer encoding](https://en.wikipedia.org/wiki/Chunked_transfer_encoding)
- HTTP compression - it's mostly used for responses but [it seems it can be used for requests as well](https://serverfault.com/questions/56700/is-it-possible-to-enable-http-compression-for-requests)
- Both Chunked transfer encoding and HTTP compression

By testing I don't imply we have to capture request body in these use cases (it's worth syncing with other agents) - we should have tests verifying whatever is decided agent should do in these use cases.

gregkalapos:
>  I'm not sure why there are conflicts. I have the latest version of master locally, branched of course, but still - up to date. There shouldn't be any conflicts.

I'm not sure this is the case. I think you branch does not have #374. If you compare e.g `MicrosoftExtensionsConfig.cs ` on [your branch](https://github.com/katzdan/apm-agent-dotnet/blob/capture-request-body-aspnet-core/src/Elastic.Apm.AspNetCore/Config/MicrosoftExtensionsConfig.cs) and on [master](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm.AspNetCore/Config/MicrosoftExtensionsConfig.cs) you'll see it's different - those are mostly just auto code cleanups, so things got moved, but no logical changes. (so my cleanup causes you trouble 👼 ) - also if I look at the history I see you branch misses some commits. Try merging `origin/master` to your branch again.

katzdan:
Thanks @SergeyKleyman , @gregkalapos  , I'll implement according to your guidelines and report back here.

sample/SampleAspNetCoreApp/Controllers/ValuesController.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System.Collections.Generic;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+
+namespace SampleAspNetCoreApp.Controllers
+{
+	/**
```
gregkalapos: Just a small nit: we use [Xml doc](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/) format here in the repo, like [here](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm/Api/ITracer.cs#L8) - In Rider or Visual Studio just hit `/` 3 times and the IDE will generate it for you.

sample/SampleAspNetCoreApp/Controllers/ValuesController.cs
```diff
@@ -0,0 +1,40 @@
...
+
+		//	return new[] { "value1", "value2", lastValue };
+		//}
+
+		//// GET api/values/5
+		//[HttpGet("{id}")]
+		//public ActionResult<string> Get(int id) => "value" + id;
+
+		[HttpPost("{id}")]
+		public ActionResult<string> Post([FromForm]string id) => "value=" + id;
```
gregkalapos: Since this is basically a 1 liner, I think it's fine to add it to the `HomeController` - just add a comment to it that describes the use case - basically what you have above.

sample/SampleAspNetCoreApp/Controllers/ValuesController.cs
```diff
@@ -0,0 +1,40 @@
...
+		//	}
+
+		//	return new[] { "value1", "value2", lastValue };
+		//}
+
+		//// GET api/values/5
+		//[HttpGet("{id}")]
+		//public ActionResult<string> Get(int id) => "value" + id;
+
+		[HttpPost("{id}")]
```
gregkalapos: Are you sure the `"{id}"` part is needed? I think if it's there then it must be something like `api/Values/1` - so without `/{id}` URLs won't matc., but the id value comes from `[FromForm]string id`, so the `/1` will be ignored. If you remove it then it'll match to `api/Values`, isn't that the intention? The id comes from the body, not from the url.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
```diff
@@ -265,5 +286,7 @@ private static string GetHttpVersion(string protocolString)
 					return protocolString.Replace("HTTP/", string.Empty);
 			}
 		}
+
+		private bool hasBody(HttpRequest request) => request.Body != null;
```
gregkalapos: Another a small nit: we use capital method names, like `HasBody`. - not that we don't like Java :)

test/Elastic.Apm.Tests/ConfigTest.cs
```diff
@@ -127,6 +130,17 @@ public void SetCaptureHeadersTest()
 			config.CaptureHeaders.Should().Be(false);
 		}
 
+		[Fact]
+		public void SetCaptureBodyTest()
+		{
+			//Possible values : "off", "all", "errors", "transactions"
+			foreach (var value in SupportedValues.CaptureBodySupportedValues){
```
gregkalapos: Nice how you cover all cases here 👍

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
```diff
@@ -322,5 +323,25 @@ internal static bool IsMsOrElastic(byte[] array)
 
 			return true;
 		}
+
+		protected string ParseCaptureBody(ConfigurationKeyValue kv)
+		{
+			var captureBodyInConfig = kv.Value;
+			if (string.IsNullOrEmpty(captureBodyInConfig))
+			{
+				return ConfigConsts.SupportedValues.CaptureBodyOff;
```
gregkalapos: nit: why not `return DefaultValues.captureBody` here, as below?

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
```diff
@@ -144,6 +146,24 @@ private void FillSampledTransactionContextRequest(HttpContext context, Transacti
...
+			var body = Consts.BodyRedacted; // According to the documentation - the default value of 'body' is '[Redacted]'
+			//If 'CaptureBody' is not 'off' and 
+			if (hasBody(context.Request) &&
+				(_configurationReader.CaptureBody.Equals(ConfigConsts.SupportedValues.CaptureBodyAll) ||
+				_configurationReader.CaptureBody.Equals(ConfigConsts.SupportedValues.CaptureBodyTransactions))
+				)
+			{
+				using (var reader = new StreamReader(context.Request.Body))
+				{
+					body = reader.ReadToEnd();
```
gregkalapos: I see some exceptions here when I run the sample app:

```
fail: Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware[1]
      An unhandled exception has occurred while executing the request.
System.IO.IOException: Unexpected end of Stream, the content may have already been read by another component. 
   at Microsoft.AspNetCore.WebUtilities.MultipartReaderStream.ReadAsync(Byte[] buffer, Int32 offset, Int32 count, CancellationToken cancellationToken)
   at Microsoft.AspNetCore.WebUtilities.StreamHelperExtensions.DrainAsync(Stream stream, ArrayPool`1 bytePool, Nullable`1 limit, CancellationToken cancellationToken)
   at Microsoft.AspNetCore.WebUtilities.MultipartReader.ReadNextSectionAsync(CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Http.Features.FormFeature.InnerReadFormAsync(CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Mvc.ModelBinding.FormValueProviderFactory.AddValueProviderAsync(ValueProviderFactoryContext context)
   at Microsoft.AspNetCore.Mvc.ModelBinding.CompositeValueProvider.CreateAsync(ActionContext actionContext, IList`1 factories)
   at Microsoft.AspNetCore.Mvc.ModelBinding.CompositeValueProvider.CreateAsync(ControllerContext controllerContext)
   at Microsoft.AspNetCore.Mvc.Internal.ControllerBinderDelegateProvider.<>c__DisplayClass0_0.<<CreateBinderDelegate>g__Bind|0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker.InvokeInnerFilterAsync()
   at Microsoft.AspNetCore.Mvc.Internal.ResourceInvoker.InvokeNextResourceFilter()
   at Microsoft.AspNetCore.Mvc.Internal.ResourceInvoker.Rethrow(ResourceExecutedContext context)
   at Microsoft.AspNetCore.Mvc.Internal.ResourceInvoker.Next(State& next, Scope& scope, Object& state, Boolean& isCompleted)
   at Microsoft.AspNetCore.Mvc.Internal.ResourceInvoker.InvokeFilterPipelineAsync()
   at Microsoft.AspNetCore.Mvc.Internal.ResourceInvoker.InvokeAsync()
   at Microsoft.AspNetCore.Builder.RouterMiddleware.Invoke(HttpContext httpContext)
   at Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
   at Microsoft.AspNetCore.StaticFiles.StaticFileMiddleware.Invoke(HttpContext context)
   at Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware.Invoke(HttpContext context)

```

I googled this and it seems to be a common thing. [This one](https://stackoverflow.com/questions/49867343/unexpected-end-of-stream-the-content-may-have-already-been-read-by-another-comp) suggest disabling model binding, but that does not help in our case - we can't tell this to all our users.

I'm trying to help here and come up with something. I'd say this is definitely something where we should come up with a solution for.

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
katzdan: In order not have to deal with a null/empty value passed in the config file or env variables. Let me know if you prefer to change that. I'd feel better to align the code to the preference of the devs with more experience in this project.

test/Elastic.Apm.Tests/ConfigTest.cs
katzdan: Thanks.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
katzdan: Java conventions die hard... I'm back to capital casing.

sample/SampleAspNetCoreApp/Controllers/ValuesController.cs
katzdan: I removed it. You will see it in the next commit/push.

sample/SampleAspNetCoreApp/Controllers/ValuesController.cs
katzdan: OK. Done. Values controller has been removed and I added the post method to the Home controller with a custom route in order not to affect any of the functionality.

sample/SampleAspNetCoreApp/Controllers/ValuesController.cs
katzdan: OK. got it.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
katzdan: Looking into this as well.

katzdan:
> I recommend to test use cases with
> 
> * [Chunked transfer encoding](https://en.wikipedia.org/wiki/Chunked_transfer_encoding)
> * HTTP compression - it's mostly used for responses but [it seems it can be used for requests as well](https://serverfault.com/questions/56700/is-it-possible-to-enable-http-compression-for-requests)
> * Both Chunked transfer encoding and HTTP compression

@SergeyKleyman  - I started to go over the resources you mentioned and will comment with some recommendations.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
katzdan: @gregkalapos - I wasn't able to reproduce this problem (please provide instructions on who to do it). However - the problem of an "already read" request/response stream is a well known one. There are several approaches to resolve this , with the one most appropriate for this problem being 
1. Resetting the body stream reader to position 0 when it's supported.
2. Reading the body and then writing it back to the memory to prevent subsequent attempts to the read the body from failing. 
I'm implementing and testing it now.

There is another other solution which I think are not suitable for this case-
- EnableRewind() for all requests in the pipeline but that would mean we are changing the behavior of the web app on the server. 

refs : 
https://gunnarpeipman.com/aspnet/aspnet-core-request-body/
http://www.palador.com/2017/05/24/logging-the-body-of-http-request-and-response-in-asp-net-core/
https://devblogs.microsoft.com/aspnet/re-reading-asp-net-core-request-bodies-with-enablebuffering/

sample/SampleAspNetCoreApp/Controllers/ValuesController.cs
katzdan: Code committed.

sample/SampleAspNetCoreApp/Controllers/ValuesController.cs
katzdan: Code commited.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
katzdan: Code committed. Please check on your end if possible or provide instructions for recreating this problem.

katzdan:
@gregkalapos  - I added the second case of error handling in ApmMiddleware.cs.
@SergeyKleyman  - Please comment on which content-types should be supported.
aside from that - I'll handle the file conflicts later today (@gregkalapos - I tried what you proposed but it didn't resolve the situation, I'll have to try other solutions).

SergeyKleyman:
> @SergeyKleyman - Please comment on which content-types should be supported.

I recommend to follow Java agent approach - https://www.elastic.co/guide/en/apm/agent/java/current/config-http.html#config-capture-body-content-types

SergeyKleyman:
@katzdan If you arrive to conclusion that supporting Chunked transfer encoding and/or HTTP compression is a lot of additional effort we can declare those use cases as not supported for now and see if there is any demand for them from the customers.

katzdan:
@gregkalapos - indeed, I think that supporting text content is enough for now. Supporting these cases would increase the amount of code here considerably. I think it would make sense to align with the Java agent and add support for non text content types in the future based on customers' input.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
gregkalapos: How do you test? I simply sent an HTTP POST to `http://localhost:5000/api/Home/Post` with Postman. If I remember correctly, HTTP GET requests from a browser did not repro the problem for me either (I should have said this earlier.)

To the current version: I still see some problems, and this time it happens always:
```
Application started. Press Ctrl+C to shut down.
fail: Elastic.Apm[0]
      {ApmMiddleware} General Error reading request body , Error msg : Specified method is not supported. , stack trace :    at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpRequestStream.set_Position(Int64 value)
         at Elastic.Apm.AspNetCore.Extensions.HttpRequestExtensions.ExtractRequestBody(HttpRequest request, IApmLogger logger) in /Users/gergelykalapos/Repos/apm-agent-dotnet/src/Elastic.Apm.AspNetCore/Extensions/RequestExtentions.cs:line 28 
fail: Elastic.Apm[0]
      {ApmMiddleware} General Error reading request body , Error msg : Specified method is not supported. , stack trace :    at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpRequestStream.set_Position(Int64 value)
         at Elastic.Apm.AspNetCore.Extensions.HttpRequestExtensions.ExtractRequestBody(HttpRequest request, IApmLogger logger) in /Users/gergelykalapos/Repos/apm-agent-dotnet/src/Elastic.Apm.AspNetCore/Extensions/RequestExtentions.cs:line 28 
^CApplication is shutting down...
```
It's very strange that [on this post](https://gunnarpeipman.com/aspnet/aspnet-core-request-body/) no one mentions this problem.

Doesn't this happen on your machine?

This GitHub issue is the same: https://github.com/aspnet/KestrelHttpServer/issues/1548 

And the solution for that is to call `request.EnableRewind();` - that's currently not available, for that we need a reference to the `Microsoft.AspNetCore.Http` package. But it's still very strange that on the 1. blogpost people seem to suggest that as an alternative, and no one says that the original version without `request.EnableRewind();`  does not seem to work. It defiantly fails on my machine and [the referenced GitHub issue](https://github.com/aspnet/KestrelHttpServer/issues/1548 ) seems to verify that without `request.EnableRewind();` it should not work.

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
gregkalapos: I'm not sure I can follow you.

So, this is the part I talk about:

```
if (string.IsNullOrEmpty(captureBodyInConfig))
{
   return ConfigConsts.SupportedValues.CaptureBodyOff;
}
```

All I suggest is to replace `ConfigConsts.SupportedValues.CaptureBodyOff` with `DefaultValues.captureBody`, because in case the entered value is `null` or an empty string we want to return the default value - which is indeed the same as `ConfigConsts.SupportedValues.CaptureBodyOff`, it just communicates the intent better.

So in this sense... we do deal with null/empty values: we simply return the default in that case.

gregkalapos:
> @SergeyKleyman - indeed, I think that supporting text content is enough for now. Supporting these cases would increase the amount of code here considerably. I think it would make sense to align with the Java agent and add support for non text content types in the future based on customers' input.

Yeah, I'd also take this approach.

>  (@gregkalapos - I tried what you proposed but it didn't resolve the situation, I'll have to try other solutions).

Regarding the conflicted changed: It seems you resolved those 👍

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
katzdan: @gregkalapos - I tested using the SampleAspNetCoreApp and with test methods (
TestJsonBodyRetrievalOnRequestFailureInAspNetCore, FailingPostRequestWithoutConfiguredExceptionPage, HomeSimplePagePostTransactionTest, all in AspNetCoreMiddlewareTests.cs).
I'll try it your way. The test methods for this worked but since I merged master into my working branch - these tests fail. It looks like the initialization of the agent in the test method (with the captureBody setting being set to all/errors/transactions) is being ignored and then debugging I can see that the loaded value in the website being used in the test is "off". I'm still trying to understand why this is happening.

src/Elastic.Apm.AspNetCore/Extensions/RequestExtentions.cs
```diff
@@ -0,0 +1,56 @@
...
+					//Write the body into the body in case it will be read by other components
+					//after this one
+					var bytesToWrite = Encoding.UTF8.GetBytes(body);
+					injectedRequestStream.Write(bytesToWrite, 0, bytesToWrite.Length);
+					injectedRequestStream.Seek(0, SeekOrigin.Begin);
+					request.Body = injectedRequestStream;
+
+				}
+			} catch (IOException ioException){
+				logger.Error()?.Log("IO Error reading request body , Error msg : {exception} , stack trace : {stacktrace} " , ioException.Message , ioException.StackTrace);
```
gregkalapos: jfyi: we have an overload that logs the exception for you:

```
logger.Error()?.LogException(ioException, "IO Error reading request body");
```

The result is same that you did here manually.

Same below.

src/Elastic.Apm.AspNetCore/Helpers/AgentConfigHelper.cs
```diff
@@ -0,0 +1,36 @@
+using System;
```
gregkalapos: Nit: some unused `using`s here.

src/Elastic.Apm.AspNetCore/Helpers/AgentConfigHelper.cs
```diff
@@ -0,0 +1,36 @@
...
+		/// <returns></returns>
+		public static bool ShouldExtractRequestBodyOnError() =>
+			(Agent.Config.CaptureBody.Equals(ConfigConsts.SupportedValues.CaptureBodyAll) ||
+				Agent.Config.CaptureBody.Equals(ConfigConsts.SupportedValues.CaptureBodyErrors));
+
+		/// <summary>
+		/// Returns weather we should collect the request body in transaction logging (accoridng to the Apm configuration)
+		/// </summary>
+		/// <returns></returns>
+		public static bool ShouldExtractRequestBodyOnTransactios() =>
```
gregkalapos: nit: small typo: `ShouldExtractRequestBodyOnTransactios` -> `ShouldExtractRequestBodyOnTransactions`

src/Elastic.Apm.AspNetCore/Helpers/AgentConfigHelper.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Elastic.Apm.AspNetCore.Helpers
+{
+	/// <summary>
+	/// Simplifies the way to determine if we should collect request field's body during trx, errors etc.
+	/// </summary>
+    public static class AgentConfigHelper
+    {
+		/// <summary>
+		/// Returns weather we should collect the request body in case of error (accoridng to the Apm configuration)
```
gregkalapos: nit: small typo: `accoridng` -> `according `

src/Elastic.Apm.AspNetCore/Helpers/AgentConfigHelper.cs
```diff
@@ -0,0 +1,36 @@
...
+		/// </summary>
+		/// <returns></returns>
+		public static bool ShouldExtractRequestBodyOnError() =>
+			(Agent.Config.CaptureBody.Equals(ConfigConsts.SupportedValues.CaptureBodyAll) ||
+				Agent.Config.CaptureBody.Equals(ConfigConsts.SupportedValues.CaptureBodyErrors));
+
+		/// <summary>
+		/// Returns weather we should collect the request body in transaction logging (accoridng to the Apm configuration)
```
gregkalapos: nit: `accoridng` -> `according `

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
katzdan: OK. Will do.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
```diff
@@ -88,7 +92,13 @@ public async Task InvokeAsync(HttpContext context)
 			{
 				await _next(context);
 			}
-			catch (Exception e) when (ExceptionFilter.Capture(e, transaction)) { }
+			catch (Exception e) {
```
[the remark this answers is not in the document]
katzdan: I was asking myself the exact question about the scope of the body capturing code. I wanted to keep it in the `Elastic.Apm.AspNetCore` project and I certainly don't feel comfortable changing things in my first issue so it's good to hear that new Capture method can be in the `Elastic.Apm.AspNetCore` project. 
The only thing that the new filter is supposed to do is collect the request body and then execute the current filter (from `Elastic.Apm`) right?

src/Elastic.Apm.AspNetCore/Extensions/RequestExtentions.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.IO;
+using System.Text;
+using Elastic.Apm.Logging;
+using Elastic.Apm.Model;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Internal;
+
+namespace Elastic.Apm.AspNetCore.Extensions
+{
+    public static class HttpRequestExtensions
```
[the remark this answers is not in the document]
katzdan: is there a specific linting tool use in this project?

src/Elastic.Apm.AspNetCore/Helpers/AgentConfigHelper.cs
katzdan: I just found out that for some reason my VS2019 installation didn't have a spell checker installed. Going over the files to verify there are no more typos.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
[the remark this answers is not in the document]
gregkalapos: > The only thing that the new filter is supposed to do is collect the request body and then execute the current filter (from Elastic.Apm) right?

Yes, exactly. Something like this (this was just me quickly writing this up):

```
...
try
{
	await _next(context);
}
catch (Exception e) when (CaptureExceptionWithHttpRequest(transaction, context.Request, e)) { }
finally
{
...
```

```
bool CaptureExceptionWithHttpRequest(Transaction transaction, HttpRequest request, Exception e)
{
	transaction.CollectRequestInfo(request, _logger);
	return  ExceptionFilter.Capture(e, transaction);
}
```

Maybe CaptureExceptionWithHttpRequest on a separate `static class` similar to `ExceptionFilter.Capture`.

src/Elastic.Apm.AspNetCore/Extensions/RequestExtentions.cs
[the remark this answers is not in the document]
gregkalapos: Most of us working on this project use JetBrains Rider, here is how the file looks in it (spot the yellow things)

image

I suspect you use Visual Studio, is that right? VS also shows some of the issues:

image

As you can see it shows the unused `using`s, plus I turned on showing whitespace characters, that's how I saw that spaces and tabs are mixed (as you can see... there are seom `. . . . ` lines and some `-> ->`.  Regarding the missing new lines: those are unfortunately not shown in VS, but if you reformat, VS takes care of those - same for the tabs vs. spaces issue.

I think the only thing I saw and VS did not find and fix is the `string body` instead of `var body`.

So.. the way this works is that we have an [`.editorconfig`](https://github.com/elastic/apm-agent-dotnet/blob/master/.editorconfig) file, and the editor picks that up. Everything is defined there.

src/Elastic.Apm.AspNetCore/Helpers/AgentConfigHelper.cs
gregkalapos: 👍  No worries, this happens to me even with spell checker 😄

src/Elastic.Apm.AspNetCore/Extensions/RequestExtentions.cs
```diff
@@ -0,0 +1,56 @@
...
+					var bytesToWrite = Encoding.UTF8.GetBytes(body);
+					injectedRequestStream.Write(bytesToWrite, 0, bytesToWrite.Length);
+					injectedRequestStream.Seek(0, SeekOrigin.Begin);
+					request.Body = injectedRequestStream;
+
+				}
+			} catch (IOException ioException){
+				logger.Error()?.Log("IO Error reading request body , Error msg : {exception} , stack trace : {stacktrace} " , ioException.Message , ioException.StackTrace);
+			} catch (Exception e) {
+				//TODO : remove this after havning finalized all cases and testing all scenarios
```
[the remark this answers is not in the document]
katzdan: If there is no exception handler in a wrapper context (calling function/class etc) then obviously we need to keep the general exception handler block. I'll remove the comment.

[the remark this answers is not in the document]
gregkalapos: 👍 Sounds good! We plan to have a better solution for this long term (https://github.com/elastic/apm-agent-dotnet/issues/208) but currently every exception caused by the agent will bubble up to the user code unless we manually try-catch.

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
katzdan: Done. code is pushed.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
katzdan: I removed the request.body.position=0 statement after re-creating the problem.
Still working on the 'Unexpected end of Stream' problem (Which I haven't been able to recreate yet).

src/Elastic.Apm.AspNetCore/Extensions/RequestExtentions.cs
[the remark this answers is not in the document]
katzdan: I verified all files are cleaned-up before my last commit from a few minutes ago.

src/Elastic.Apm.AspNetCore/Helpers/AgentConfigHelper.cs
katzdan: Fixed.

src/Elastic.Apm.AspNetCore/Helpers/AgentConfigHelper.cs
katzdan: Fixed.

src/Elastic.Apm.AspNetCore/Helpers/AgentConfigHelper.cs
katzdan: Fixed.

src/Elastic.Apm.AspNetCore/Helpers/AgentConfigHelper.cs
katzdan: This file and all the other files I edited are now cleaned-up.

src/Elastic.Apm.AspNetCore/Extensions/RequestExtentions.cs
katzdan: I updated the code to use `logger.Error()?.LogException`.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO;
```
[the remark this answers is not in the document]
katzdan: I did a file clean-up in my last commit.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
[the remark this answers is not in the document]
katzdan: Done.

katzdan:
@gregkalapos , @SergeyKleyman - I'm done with the implementation, with 3 issues:

1. I'm still unable to re-create the case where reading the request body fails with the error `Unexpected end of Stream, the content may have already been read by another component.`. 
From what I read - **'EnableRewind' is available for .net standard 2.0 apps** via the Microsoft.AspNetCore.Http.Extensions nuget package which is supposed to resolve this problem. I just wanted your guys' approval before adding a new package to the project.

2. When the user configures 'CaptureBody' to all/errors/transactions but there is no 'CaptureBodyContentTypes' property in the appsettings.json file - the Apm agent will report it in the logs  ` ("Capture_Body is on but no content types are configured. Request bodies will not be captured.")` and the agent will not extract the request body in any scenario because the list of allowed content-types is empty (and the agent matches the request content-type with the entries in the list). In this case - I'm asking if it makes more sense to use the default content types instead of an empty one. Your thoughts?

3. The checks for this commit are failing due to some code checkout error. How can I trigger the checks again (without committing)? All tests run fine on my machine.

katzdan:
@SergeyKleyman , @gregkalapos , 
Have you see my latest comment?

SergeyKleyman:
@katzdan Hi Dan. We have a few urgent things to finish for the latest release and we will get back to this feature ASAP.

gregkalapos:
Hey @katzdan, sorry for the delay! 

1. I can’t repro that problem either anymore.. I checked out your latest code and I don’t see that problem. Nevertheless I think it’d be nice to use `EnableRewind` - it’s fine to reference `Microsoft.AspNetCore.Http.Extensions` 👍. Please make sure to reference Version `2.0.0`. 

2. In such situations we usually look at what other Elastic APM Agents do, and if it is not something totally unnatural for .NET, then we tend to just adapt it. The [Java agent seem to do exactly what you suggest](https://www.elastic.co/guide/en/apm/agent/java/current/config-http.html#config-capture-body-content-types) - it has a default, and that is used when no `CaptureBodyContentTypes` is configured - the only thing I don’t see in the default list is `multipart/form-data`. I’m happy to hear thoughts on this: shouldn’t we also add that?

3. I think there are 2 reasons for not building the PR - 1) there are some conflicts and that’s why Jenkins did not build it. 2) our CI only automatically builds PRs from either devs from the Elastic org, or PRs that one of us approved. This is a security mechanism. I lost track on this.. was this built by the CI previously? I think it wasn’t. Could you please resolve the conflicts? I’ll then take care of running it in CI once that’s done.

I think we are almost done here. :)

katzdan:
@gregkalapos , @SergeyKleyman  - 
1. I'll add the HttpExtensions lib (2.0), no problem.
2. In regard to multipart/form-data - it may contain binary data (uploaded files in a HTML form) so I would think it makes sense not to include it in this PR , stay in line with the Java implementation ad review the option to include it in the future.
3. I'll merge the master branch from the Elastic .net Apm repo into my work branch , commit in order to resolve the conflicts and then we'll see if the ci builds still fail.

gregkalapos:
> @gregkalapos , @SergeyKleyman -
> 
> 1. I'll add the HttpExtensions lib (2.0), no problem.
> 2. In regard to multipart/form-data - it may contain binary data (uploaded files in a HTML form) so I would think it makes sense not to include it in this PR , stay in line with the Java implementation ad review the option to include it in the future.
> 3. I'll merge the master branch from the Elastic .net Apm repo into my work branch , commit in order to resolve the conflicts and then we'll see if the ci builds still fail.

All sounds good @katzdan 👍

katzdan:
@gregkalapos , @SergeyKleyman  - 
I added the call to 'EnableRewind' (it turns out that I needed to add a reference to 'Microsoft.AspNetCore.Http' 2.0.0 and not to the HttpExtensions lib). The tests pass and a manual check also passed.  I resolved the conflicts yesterday so I guess you guys can continue with the merge (unless there anything additional you need from me).

gregkalapos:
> @gregkalapos , @SergeyKleyman -
> I added the call to 'EnableRewind' (it turns out that I needed to add a reference to 'Microsoft.AspNetCore.Http' 2.0.0 and not to the HttpExtensions lib). The tests pass and a manual check also passed. I resolved the conflicts yesterday so I guess you guys can continue with the merge (unless there anything additional you need from me).

Thanks, great; will take a look at it soon!

skynet2:
hi all,
quick question, net core give us the ability to use middlewares for that purpose, maybe it will be easier to use it? for example here is the code which I wrote to have APM logging for both request + response https://gist.github.com/skynet2/b0c9b96f0c3a75175962d6831b403339 

Thanks,
Stas

gregkalapos:
> hi all,
> quick question, net core give us the ability to use middlewares for that purpose, maybe it will be easier to use it? for example here is the code which I wrote to have APM logging for both request + response https://gist.github.com/skynet2/b0c9b96f0c3a75175962d6831b403339
> 
> Thanks,
> Stas

Hi @skynet2,

yes, that's exactly what's happening here. In this PR the reading of the request body happens in the `ApmMiddleware` [here](https://github.com/elastic/apm-agent-dotnet/pull/402/files#diff-5b459610a75d3fb704b839753632e7ceR155).

So I think we have the same approach here as in your code. Or do I miss something?

src/Elastic.Apm/Config/ConfigConsts.cs
```diff
@@ -1,12 +1,22 @@
 ﻿using System;
+using System.Collections.Generic;
 
 namespace Elastic.Apm.Config
 {
 	public static class ConfigConsts
 	{
 		public static class Constraints
 		{
+			public const string LogLevel = "ELASTIC_APM_LOG_LEVEL";
```
gregkalapos: I think none of the new consts added here from `Constraints` is used. These new values seem to be the same as values from `EnvVarNames` - could we maybe remove these?

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
```diff
@@ -428,5 +429,46 @@ private enum TimeSuffix
 			Ms,
 			S
 		}
+
```
gregkalapos: nit: there are 2 new lines here.

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
```diff
@@ -428,5 +429,46 @@ private enum TimeSuffix
...
+
+		protected string ParseCaptureBody(ConfigurationKeyValue kv)
+		{
+			var captureBodyInConfig = kv.Value;
+			if (string.IsNullOrEmpty(captureBodyInConfig))
+			{
```
gregkalapos: codestyle: the `{` `}` can be removed here.

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
```diff
@@ -428,5 +429,46 @@ private enum TimeSuffix
...
+			//CaptureBodyContentTypes and CaptureBody are linked. Verify that in case CaptureBody is ON - then CaptureBodyContentTypes is not empty
+			if (string.IsNullOrEmpty(captureBodyContentTypesInConfig) && captureBody != SupportedValues.CaptureBodyOff)
+			{
+				Logger?.Error()?.Log("Capture_Body is on but no content types are configured. Request bodies will not be captured.");
+				return new List<string>();
+			}
+
+			var captureBodyContentTypes = new List<string>();
+			if (captureBodyContentTypesInConfig != null)
+			{
```
gregkalapos: Same as above: `{` and `}` can be removed for `if` bodies with a single line. This is defined in our `.editorconfig`

image

src/Elastic.Apm.AspNetCore/Helpers/ExceptionFilter.cs
```diff
@@ -0,0 +1,28 @@
...
+	/// Advantage: it avoid stack unwinding.
+	/// </summary>
+	internal static class ExceptionFilter
+	{
+		internal static bool Capture(Exception e, ITransaction transaction, HttpContext httpContext, IConfigurationReader configurationReader, IApmLogger logger)
+		{
+			transaction.CaptureException(e);
+
+			if (httpContext != null && configurationReader.ShouldExtractRequestBodyOnError())
+			{
```
gregkalapos: codestyle: `{` `}` can be removed.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
```diff
@@ -144,6 +146,16 @@ private void FillSampledTransactionContextRequest(HttpContext context, Transacti
 					requestHeaders.Add(header.Key, header.Value.ToString());
 			}
 
+			var body = Consts.BodyRedacted; // According to the documentation - the default value of 'body' is '[Redacted]'
+			if (!string.IsNullOrEmpty(context?.Request?.ContentType))
+			{
+				var contentType = new ContentType(context.Request.ContentType);
+				if (_configurationReader.ShouldExtractRequestBodyOnTransactions() && _configurationReader.CaptureBodyContentTypes.ContainsLike(contentType.MediaType))
+				{
```
gregkalapos: codestyle: `{` `}` can be removed.

katzdan:
@gregkalapos - I'll make the changes today or tomorrow.

src/Elastic.Apm/Config/ConfigConsts.cs
katzdan: Done.

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
katzdan: Done.

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
katzdan: Done.

src/Elastic.Apm.AspNetCore/Helpers/ExceptionFilter.cs
katzdan: Done.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
katzdan: Done.

katzdan:
@gregkalapos  - I pushed the files updated in the code clean up. You can merge this into the Master branch unless there are additional changes to be implemented.

gregkalapos:
Merged! 🎉 

Thanks for your work and your patience on this @katzdan. Welcome on the contributors list! 🚀

brezaie:
Hi pals, 
Adding the following configuration to the `appsettings.json`:

`"ElasticApm": {
    "ServerUrls": "***",
    "ServiceName": "SomeService",
    "CaptureHeaders": true,
    "CaptureBody": "all",
    "LogLevel": "Info",
    "StackTraceLimit": -1
  }`

I received the following log message with does not show what the body request and response are:

image

gregkalapos:
Hi @brezaie 


I see you don’t configure `CaptureBodyContentTypes` ([here](https://github.com/elastic/apm-agent-dotnet/pull/402/files#diff-ab2b52889f8a07e7c07735b0e9ae9dbeR13) is an example for that). I suspect that this is some content type that is not included in the default. Is it maybe `multipart/form-data`? 


Also please note that this PR is only about capturing the `request` body - we don’t capture the response body.

brezaie:
Hi @gregkalapos 

Thanks for your response. It did work now.

Do you have any plans for adding the `response` body?

gregkalapos:
> Hi @gregkalapos
> 
> Thanks for your response. It did work now.
> 
> Do you have any plans for adding the `response` body?

Unfortunately we have no such plans currently. `response` body is not part of [the API](https://github.com/elastic/apm-server/blob/master/docs/spec/context.json#L15) that the APM server expects from us. We have a [field for it in `request`](https://github.com/elastic/apm-server/blob/master/docs/spec/request.json#L7), bot not in `response`.

brezaie:
Dear @gregkalapos 

Thanks for your reply.

When adding the following configuration, surprisingly, I receive a `null` for the input of my `POST` method
`
"CaptureBodyContentTypes": "application/x-www-form-urlencoded*, text/*, application/json*, application/xml*",
`

Here is my `POST` method and the corresponding filter and configuration:
```
[HttpPost("send")]
public JsonResult Send([FromBody] BaseReportFilter<SendMessageFilter> filter)
{
	//Do something
	return new JsonResult("");
}


public class BaseReportFilter<T>
{
    public T ReportFilter { get; set; }
}

public class SendMessageFilter{
	public string SenderApplicationCode { get; set; }
	public string MediaType { get; set; }
	public string Body { get; set; }
	public List<string> Recipients { get; set; }
}

//appsettings.json:
"ElasticApm": {
    "ServerUrls": "http://localhost:8200",
    "ServiceName": "ServiceName",
    "CaptureHeaders": true,
    "CaptureBody": "all",
    "CaptureBodyContentTypes": "application/x-www-form-urlencoded*, text/*, application/json*, application/xml*",
    "LogLevel": "Info",
    "StackTraceLimit": -1
  }
``` 

Also, you can see my API call from Postman here:
image

gregkalapos:
> Dear @gregkalapos
> 
> Thanks for your reply.
> 
> When adding the following configuration, surprisingly, I receive a `null` for the input of my `POST` method
> `"CaptureBodyContentTypes": "application/x-www-form-urlencoded*, text/*, application/json*, application/xml*",`
> 
> Here is my `POST` method and the corresponding filter and configuration:
> 
> ```
> [HttpPost("send")]
> public JsonResult Send([FromBody] BaseReportFilter<SendMessageFilter> filter)
> {
> 	//Do something
> 	return new JsonResult("");
> }
> 
> 
> public class BaseReportFilter<T>
> {
>     public T ReportFilter { get; set; }
> }
> 
> public class SendMessageFilter{
> 	public string SenderApplicationCode { get; set; }
> 	public string MediaType { get; set; }
> 	public string Body { get; set; }
> 	public List<string> Recipients { get; set; }
> }
> 
> //appsettings.json:
> "ElasticApm": {
>     "ServerUrls": "http://localhost:8200",
>     "ServiceName": "ServiceName",
>     "CaptureHeaders": true,
>     "CaptureBody": "all",
>     "CaptureBodyContentTypes": "application/x-www-form-urlencoded*, text/*, application/json*, application/xml*",
>     "LogLevel": "Info",
>     "StackTraceLimit": -1
>   }
> ```
> 
> Also, you can see my API call from Postman here:
> image

Thanks for reporting this @brezaie. This is not good  - I also managed to reproduce this. I opened a separate bug for this: https://github.com/elastic/apm-agent-dotnet/issues/460. If there is anything to add to it, feel free to comment there.

PR #403, opened 2019-07-22
Add documentation about stack trace configs

gregkalapos:
- Following up on #374 - adding docs
- Text is completely copied [from Java](https://github.com/elastic/apm-agent-java/blob/master/docs/configuration.asciidoc)
- Noticed that in `MicrosoftExtensionsConfig.cs` the name was not plural - also fixed that

Adding @Titch990 for docs reviewer. Thanks for jumping in! :)


gregkalapos:
> Line 274 "durations equal or longer than . . . " -> "durations equal **to** or longer than . . . ". That's a cut-and-paste of the same error in the Java. Otherwise, LGTM.

Thank you! Fixed - I'll also open a PR for the Java agent ( _This sometimes happens when I'm stealing their docs_ 😈 )

Issue #407, opened 2019-07-26
Should trim strings read from environment


rickardp:
**Describe the bug**
It is quite common for strings read from environment to contain whitespaces. A typical example is when they are populated from Kubernetes secrets.

**To Reproduce**
1. Append newline to end of ELASTIC_APM_SERVER_URLS
2. Start APM
4. Error: Failed parsing server URL from environment: ELASTIC_APM_SERVER_URLS, value: ...

**Expected behavior**
While possible to work around, it is an annoyance that could easily be mitigated by trimming whitespace. Expected behavior is then that above string should be parsed successfully.

gregkalapos:
Yeah, let's fix this!

Already ongoing on https://github.com/elastic/apm-agent-dotnet/pull/408. 

Thanks @rickardp!

gregkalapos:
Merged #408 which fixes this - thanks @rickardp.

PR #408, opened 2019-07-26
Fix for issue #407: Trim environment variables

rickardp:
Trim whitespace when reading environment variables.


rickardp:
That makes sense. I will look into it right now!

rickardp:
I took your test almost as it were, the only thing I added was newline characters at the end, since this is typically how it will look as per the original issue.

PR #409, opened 2019-07-26
fix/remove-invalid-char

kuisathaverat:
Remove special character at the beginning of the file that breaks the NuGet restore command


kuisathaverat:
finally, it is not needed

PR #413, opened 2019-07-28
add environment property to service class

### LINKED ISSUE #320: Send service.environment (configured by user)
See the [spec](https://github.com/elastic/apm-server/blob/master/docs/spec/service.json#L61)

We need it for remote configuration (https://github.com/elastic/apm-agent-dotnet/issues/190) - for details search `service.environment` at https://github.com/elastic/apm/issues/76
There also seem to be [other use cases](https://discuss.elastic.co/t/are-env-labels-supported-in-elastic-apm-agent/187556/3).

vhatsura:
closes #320 

Changes:
1. Add `environment` property to service class
2. Make `logger` in `MicrosoftExtensionsConfig` as required
3. Add `environmentName` parameter of `string` type to `MicrosoftExtensionsConfig`
4. Add dependency to `Microsoft.AspNetCore.Hosting.Abstractions`

In case of ASP.Net Core application `environment` is read from `appsettings` configuration with fallback to environment variable. If configuration and environment variable don't have value, parameter is read from `IHostingEnvironment`, that can be an issue due to in .Net Core 3.0 [`IHostingEnvironment` is obsolete](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.hosting.ihostingenvironment?view=aspnetcore-3.0) and [`IHostEnvironment` should be used instead](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/generic-host?view=aspnetcore-3.0#ihostenvironment).
However, to official support of .Net Core 3.0 in elastic apm library, all dependencies in `Elastic.Apm.AspNetCore` should be updated to 3.0, which means `IHostingEnvironment` can be used right now.

What should be done after review of draft pull request?
1. Update documentation
2. Write unit tests

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -151,7 +151,7 @@ You must use the query bar to filter for a specific environment in versions prio
 [options="header"]
 |============
 | Default                                        | Type
-| EnvironmentName from IHostingEnvironment    | String 
+| https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.hosting.ihostingenvironment.environmentname?view=aspnetcore-2.2#Microsoft_AspNetCore_Hosting_IHostingEnvironment_EnvironmentName[EnvironmentName from IHostingEnvironment] | String 
 |============
 
 [float]
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -130,6 +130,31 @@ of the deployed revision, e.g. the output of git rev-parse HEAD.
 |============
 
 [float]
+[[config-environment]]
+==== `Environment`
+
+The name of the environment this service is deployed in, e.g. "production" or "staging".
+
+Environments allow you to easily filter data on a global level in the APM UI.
+It's important to be consistent when naming environments across agents.
+See {kibana-ref}/filters.html#environment-selector[environment selector] in the Kibana UI for more information.
+
+NOTE: This feature is fully supported in the APM UI in Kibana versions >= 7.2.
+You must use the query bar to filter for a specific environment in versions prior to 7.2.
+
+[options="header"]
+|============
+| Environment variable name      | IConfiguration key 
+| `ELASTIC_APM_ENVIRONMENT`      | `ElasticApm:Environment`
+|============
+
+[options="header"]
+|============
+| Default                                        | Type
+| EnvironmentName from IHostingEnvironment    | String 
+|============
+
+[float]
 [[config-transaction-sample-rate]]
 ==== `TransactionSampleRate`
 
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -150,9 +150,11 @@ You must use the query bar to filter for a specific environment in versions prio
 
 [options="header"]
 |============
-| Default                                        | Type
-| https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.hosting.ihostingenvironment.environmentname?view=aspnetcore-2.2#Microsoft_AspNetCore_Hosting_IHostingEnvironment_EnvironmentName[EnvironmentName from IHostingEnvironment] | String 
+| Default        | Type
+| See note below | String 
 |============
+
+NOTE: On ASP.NET Core application the agent uses https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.hosting.ihostingenvironment.environmentname?view=aspnetcore-2.2#Microsoft_AspNetCore_Hosting_IHostingEnvironment_EnvironmentName[EnvironmentName from IHostingEnvironment] as default environment name.
 
 [float]
 [[config-transaction-sample-rate]]
--- a/src/Elastic.Apm/Api/Service.cs
+++ b/src/Elastic.Apm/Api/Service.cs
@@ -23,9 +23,11 @@ namespace Elastic.Apm.Api
 
 		public Runtime Runtime { get; set; }
 
+		public string Environment { get; set; }
+
 		public override string ToString() => new ToStringBuilder(nameof(Service))
 		{
-			{ "Name", Name }, {"Version", Version}, { "Agent", Agent }, { "Framework", Framework }, { "Language", Language },
+			{ "Name", Name }, {"Version", Version}, { "Agent", Agent }, { "Framework", Framework }, { "Language", Language }, {"Environment", Environment}
 		}.ToString();
 
 		internal static Service GetDefaultService(IConfigurationReader configurationReader, IApmLogger loggerArg)
@@ -40,7 +42,8 @@ namespace Elastic.Apm.Api
 					Name = Consts.AgentName,
 					Version = typeof(Agent).Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion
 				},
-				Runtime = PlatformDetection.GetServiceRuntime(logger)
+				Runtime = PlatformDetection.GetServiceRuntime(logger),
+				Environment = configurationReader.Environment
 			};
 		}
 
--- a/src/Elastic.Apm/Api/Service.cs
+++ b/src/Elastic.Apm/Api/Service.cs
@@ -23,6 +23,7 @@ namespace Elastic.Apm.Api
 
 		public Runtime Runtime { get; set; }
 
+		[JsonConverter(typeof(TrimmedStringJsonConverter))]
 		public string Environment { get; set; }
 
 		public override string ToString() => new ToStringBuilder(nameof(Service))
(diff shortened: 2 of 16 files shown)


vhatsura:
@gregkalapos, @SergeyKleyman. I will glad to hear your comments to go forward.

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
```diff
@@ -345,6 +345,13 @@ protected string ParseServiceName(ConfigurationKeyValue kv)
 			return ConfigConsts.DefaultValues.UnknownServiceName;
 		}
 
+		protected string ParseEnvironment(ConfigurationKeyValue kv)
+		{
+			if (kv == null || string.IsNullOrEmpty(kv.Value)) return null;
```
SergeyKleyman: This means that it will impossible to configure empty string as environment name - is this indeed what we want to do?

test/Elastic.Apm.AspNetCore.Tests/MicrosoftExtensionsConfigTests.cs
```diff
@@ -30,7 +30,7 @@ public class MicrosoftExtensionsConfigTests
 		public void ReadValidConfigsFromAppSettingsJson()
 		{
 			var config = new MicrosoftExtensionsConfig(GetConfig($"TestConfigs{Path.DirectorySeparatorChar}appsettings_valid.json"),
-				new TestLogger());
+				new NoopLogger(), "test");
```
SergeyKleyman: I would add an `internal` constructor overload to `MicrosoftExtensionsConfig` that doesn't require `environmentName` parameter - it will be easier to maintain.

test/Elastic.Apm.AspNetCore.Tests/MicrosoftExtensionsConfigTests.cs
```diff
@@ -49,10 +49,12 @@ public void ReadValidConfigsFromAppSettingsJson()
 		public void ReadInvalidLogLevelConfigFromAppsettingsJson()
 		{
 			var logger = new TestLogger();
-			var config = new MicrosoftExtensionsConfig(GetConfig($"TestConfigs{Path.DirectorySeparatorChar}appsettings_invalid.json"), logger);
+			var config = new MicrosoftExtensionsConfig(GetConfig($"TestConfigs{Path.DirectorySeparatorChar}appsettings_invalid.json"), logger,
```
SergeyKleyman: Please see my comment above regarding adding an `internal` constructor overload to `MicrosoftExtensionsConfig`.

test/Elastic.Apm.AspNetCore.Tests/MicrosoftExtensionsConfigTests.cs
```diff
@@ -74,11 +76,13 @@ public void ReadInvalidLogLevelConfigFromAppsettingsJson()
 		public void ReadInvalidServerUrlsConfigFromAppsettingsJson()
 		{
 			var logger = new TestLogger();
-			var config = new MicrosoftExtensionsConfig(GetConfig($"TestConfigs{Path.DirectorySeparatorChar}appsettings_invalid.json"), logger);
+			var config = new MicrosoftExtensionsConfig(GetConfig($"TestConfigs{Path.DirectorySeparatorChar}appsettings_invalid.json"), logger,
+				"test");
```
SergeyKleyman: Please see my comment above regarding adding an `internal` constructor overload to `MicrosoftExtensionsConfig`.

test/Elastic.Apm.AspNetCore.Tests/MicrosoftExtensionsConfigTests.cs
```diff
@@ -109,7 +113,7 @@ public void ReadConfingsFromEnvVarsViaIConfig()
 				.AddEnvironmentVariables()
 				.Build();
 
-			var config = new MicrosoftExtensionsConfig(configBuilder, new TestLogger());
+			var config = new MicrosoftExtensionsConfig(configBuilder, new NoopLogger(), "test");
```
SergeyKleyman: Please see my comment above regarding adding an `internal` constructor overload to `MicrosoftExtensionsConfig`.

test/Elastic.Apm.AspNetCore.Tests/MicrosoftExtensionsConfigTests.cs
```diff
@@ -126,7 +130,8 @@ public void ReadConfingsFromEnvVarsViaIConfig()
 		public void LoggerNotNull()
 		{
 			var testLogger = new TestLogger();
-			var config = new MicrosoftExtensionsConfig(GetConfig($"TestConfigs{Path.DirectorySeparatorChar}appsettings_invalid.json"), testLogger);
+			var config = new MicrosoftExtensionsConfig(GetConfig($"TestConfigs{Path.DirectorySeparatorChar}appsettings_invalid.json"), testLogger,
+				"test");
```
SergeyKleyman: Please see my comment above regarding adding an `internal` constructor overload to `MicrosoftExtensionsConfig`.

test/Elastic.Apm.AspNetCore.Tests/MicrosoftExtensionsConfigTests.cs
```diff
@@ -152,7 +157,7 @@ public MicrosoftExtensionsConfigIntegrationTests(WebApplicationFactory<Startup>
 			var capturedPayload = new MockPayloadSender();
 
 			var config = new MicrosoftExtensionsConfig(
-				MicrosoftExtensionsConfigTests.GetConfig($"TestConfigs{Path.DirectorySeparatorChar}appsettings_invalid.json"), _logger);
+				MicrosoftExtensionsConfigTests.GetConfig($"TestConfigs{Path.DirectorySeparatorChar}appsettings_invalid.json"), _logger, "test");
```
SergeyKleyman: Please see my comment above regarding adding an `internal` constructor overload to `MicrosoftExtensionsConfig`.

src/Elastic.Apm/Api/Service.cs
```diff
@@ -20,9 +20,11 @@ public class Service
 
 		public Runtime Runtime { get; set; }
 
+		public string Environment { get; set; }
```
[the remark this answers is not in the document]
vhatsura: Hm, I believe we can have two options:

1. Analyzer which will analyze conformity according to the spec. However, it's expensive and only needed in this repo.
2. Unit test which will check conformity according to the spec through reflection. However, the main point which we want to cover is maxLength. In such case we can check only presence of `JsonConverter` attribute with proper converter. However, as I see, `maxLength` is set not for all string properties in the spec.
I believe the 2nd option is more realistic. The only once question is do we need parse spec or we can specify certain classes which will be tested. WDYT?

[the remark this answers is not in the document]
SergeyKleyman: I remembered that I already created a dedicated issue for this - https://github.com/elastic/apm-agent-dotnet/pull/413
So we can continue our discussion there without mixing it with this PR.

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
vhatsura: All other values check to null or empty. This is the reason why I also use such validation. Also, I don't see the point to use empty string as environment.

test/Elastic.Apm.AspNetCore.Tests/MicrosoftExtensionsConfigTests.cs
vhatsura: From one point of view, it would be easier to maintain, however, from another point of view it need to be as required parameter. In case of adding internal constructor overload it will be possible to call constructor w/o environment name inside asp.net core configuration. By this reason, I prefer left only one constructor

docs/configuration.asciidoc
```diff
@@ -129,6 +129,31 @@ of the deployed revision, e.g. the output of git rev-parse HEAD.
...
+[options="header"]
+|============
+| Environment variable name      | IConfiguration key 
+| `ELASTIC_APM_ENVIRONMENT`      | `ElasticApm:Environment`
+|============
+
+[options="header"]
+|============
+| Default                                        | Type
+| https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.hosting.ihostingenvironment.environmentname?view=aspnetcore-2.2#Microsoft_AspNetCore_Hosting_IHostingEnvironment_EnvironmentName[EnvironmentName from IHostingEnvironment] | String
```
SergeyKleyman: Thank for the docs :)

It might worth adding that this default is only for ASP.NET Core applications - I don't it will fit well in that table cell so I would put in the table cell a phrase like "See note below" and then in the note below the table clarification that `EnvironmentName` from `IHostingEnvironment` is the default for ASP.NET Core applications.

PR #414, opened 2019-07-28
make service.name as trimmed string with limitation of 1024 characters

### LINKED ISSUE #412: service.name should be limited to 1024 characters
According to elastic apm server specification, service name should be limited to 1024 characters
https://github.com/elastic/apm-server/blob/7b716ad540a68d0f95060340651d4ec5dbfdc017/docs/spec/service.json#L55

vhatsura:
closes #412

### DIFF
--- a/test/Elastic.Apm.Tests/SerializationTests.cs
+++ b/test/Elastic.Apm.Tests/SerializationTests.cs
@@ -134,7 +134,7 @@ namespace Elastic.Apm.Tests
 		/// Makes sure the name is truncated.
 		/// </summary>
 		[Fact]
-		public void ServiceVersionLengthTest()
+		public void ServiceNameLengthTest()
 		{
 			var logger = new NoopLogger();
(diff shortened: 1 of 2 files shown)


test/Elastic.Apm.Tests/SerializationTests.cs
```diff
@@ -129,6 +129,26 @@ public void DbStatementLengthTest()
 			Assert.Equal(str, deserializedDb.Statement);
 		}
 
+		/// <summary>
+		///	Creates a service instance with a name that is longer than <see cref="Consts.PropertyMaxLength"/>.
+		/// Makes sure the name is truncated.
+		/// </summary>
+		[Fact]
+		public void ServiceVersionLengthTest()
```
gregkalapos: Shouldn't this be `ServiceNameLengthTest` instead of `ServiceVersionLengthTest`?

gregkalapos: Hope it's ok that I renamed it 👼  

That way I can merge it now - before we release.

PR #415, opened 2019-07-29
feat: add IIS test support

kuisathaverat:
add IIS test support 

related to https://github.com/elastic/apm-agent-dotnet/pull/319

### DIFF
--- a/.ci/docker/iis/Dockerfile
+++ b/.ci/docker/iis/Dockerfile
@@ -29,7 +29,4 @@ RUN Invoke-WebRequest -UseBasicParsing "https://download.microsoft.com/download/
 
 ENV PATH="C:\\Windows\\system32;C:\\Windows;C:\\Windows\\System32\\Wbem;C:\\Windows\\System32\\WindowsPowerShell\\v1.0;C:\\Windows\\System32\\OpenSSH;C:\\Users\ContainerAdministrator\\AppData\\Local\\Microsoft\\WindowsApps;c:\\tools;c:\\tools\\dotnet;;C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\BuildTools\\MSBuild\\15.0\\Bin"
 
-#docker run --name iis -w C:\inetpub\wwwroot\ iis
-#docker cp . iis:C:\inetpub\wwwroot
-#docker exec iis .ci\windows\test-iis.bat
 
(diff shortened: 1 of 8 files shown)


.ci/docker/iis/Dockerfile
```diff
@@ -0,0 +1,35 @@
+FROM mcr.microsoft.com/dotnet/framework/aspnet
```
kuisathaverat: this Docker image is not needed for tests, but it is an environment where the test can run, we invest time to make it and I think it worth to be keep

.ci/windows/msbuild.bat
```diff
@@ -6,7 +6,8 @@
 ::vswhere -latest -requires Microsoft.Component.MSBuild -find MSBuild\**\Bin\MSBuild.exe`
 ::vswhere -latest -requires Microsoft.Component.MSBuild -property installationPath
```
kuisathaverat: cosmetic changes

.ci/docker/iis/Dockerfile
v1v: Sounds good, what about a README.md to explain this particular Dockerfile is not used but it could be, and if so, what commands are required to use it locally?

.ci/docker/iis/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+                                                "--add", "Microsoft.Net.Core.Component.SDK", \
+                                                "--add", "Microsoft.VisualStudio.Workload.MSBuildTools", \
+                                                "--add", "Microsoft.VisualStudio.Component.WebDeploy ", \
+                                                "--add", "Microsoft.VisualStudio.Workload.WebBuildTools", \
+                                                "--add", "Microsoft.VisualStudio.Component.NuGet.BuildTools", \
+                                                "--quiet", "--norestart", "--nocache" -NoNewWindow -Wait;
+
+ENV PATH="C:\\Windows\\system32;C:\\Windows;C:\\Windows\\System32\\Wbem;C:\\Windows\\System32\\WindowsPowerShell\\v1.0;C:\\Windows\\System32\\OpenSSH;C:\\Users\ContainerAdministrator\\AppData\\Local\\Microsoft\\WindowsApps;c:\\tools;c:\\tools\\dotnet;;C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\BuildTools\\MSBuild\\15.0\\Bin"
+
+#docker run --name iis -w C:\inetpub\wwwroot\ iis
```
v1v: What's the reason for those commented comments?

kuisathaverat: they are the commands to launch it locally

PR #425, opened 2019-07-31
Surface .ci folder as project in solution

Mpdreamz:
Adds a ci project that automatically includes the .ci folder's contents so these files show up in solution explorer.


SergeyKleyman:
What is end goal of this change?

Mpdreamz:
To make sure all project related files show up in the editor (vs/rider/etc).

SergeyKleyman:
What about `build` and `docs` folders?

SergeyKleyman:
Also `.ci` folder contain a few files but only `Jenkinsfile` file is included even though it's not even in `.ci` folder.

We will have to maintain this project file when infra team makes changes to `.ci` folder. What are the advantages of having all these files in a project that justifies this effort?

Mpdreamz:
The goal is to have all build related files editable in the same editor under a single project. 

Changes in the `.ci` folder (new/removed files/directories/) will show up so if the infra teams does changes they show up.

I would do the same for `docs` since these are meant to be editted manually.

The `build` folder only contains binaries at the moment, these might find a better home in the future or might warrant inclusion once we start to introduce shared `msbuild` properties files that are meant to be editted manually.

Mpdreamz:
> Also `.ci` folder contain a few files but only `Jenkinsfile` file is included even though it's not even in `.ci` folder.

To be clear this file is included **in addition** to the contents of the folder which are picked up automatically in the  new sdk csproj

SergeyKleyman:
I still have concerns about using C# project for this purpose. Do you do this in Elasticsearch .NET client solution? Are there any potential issues with this approach?

Mpdreamz:
Can you elaborate on your concerns?

SergeyKleyman:
My main concern is using C# project in a way that I personally didn't encounter before, namely to hold a list of random files without producing any kind of artifact - executable or library. My suggestion is to estimate potential risk involved in doing that and weigh it against potential gains. One way to estimate the risk is to see if some other long running projects did something like that and what was their experience.

Mpdreamz:
Gains:

Using one editor to edit the files in a way that contributors who don't use a .NET IDE (rider/vs/omnisharp/code) have to care about. With the benefit that those that do .NET development can see/search/edit these files without having to open a separate file editor to peruse these files.

Cons:

Will emit `bin` `obj` folders if opened in a .NET IDE, these are ignored though so won't show up for anyone not editing through a .NET IDE. 


Alternative approach is to keep adding these files as solution items which won't update automatically on addition and removal.

Not quite the same since our build files are .NET (fsharp) that compiles and is run with `dotnet` cli tool but we do the same where we include the related files from the root (appveyor/azure-devops/build.bat/build.sh) in the project much like the jenkins file is included in this PR.

SergeyKleyman:
It sounds that you don't foresee any negative consequences based on the experience.

What about the `docs` directory - would you prefer to place the files in `docs` directory into a separate project?

PR #426, opened 2019-07-31
Move Elastic.Apm.AspNetFullFramework over to new sdk style project

Mpdreamz:
This will allow all artifacts to use the same mechanics to describe and publish the packages and we can look into abstracting common bits to shared msbuild files.


Mpdreamz:
@gregkalapos @SergeyKleyman 

I updated this PR so that its easy to build the solution from non windows OS's again.

Everything is an SDK style project except for the sample app for fullframework ASP.NET. 

Through some trickery this project exists but is basically empty on linux. 

This alleviates the need to this in the build scripts too (cc @kuisathaverat)

https://github.com/elastic/apm-agent-dotnet/blob/master/.ci/linux/build.sh#L8-L10

A simple `dotnet build` suffices.

Mpdreamz:
Tests on windows run fine but the script returns `1` not sure why though

gregkalapos:
> Tests on windows run fine but the script returns `1` not sure why though

I think we need to update msbuild on the Windows CI machine:

```
 error : Version 2.2.401 of the .NET Core SDK requires at least version 16.0.0 of MSBuild. The current available version of MSBuild is 15.9.21.664. Change the .NET Core SDK specified in global.json to an older version that requires the MSBuild version currently available.
```

Directory.Build.props
```diff
@@ -0,0 +1,7 @@
+<Project>
+  <!-- Root Directory Build Properties -->
+  <PropertyGroup>
+    <LangVersion>8.0</LangVersion>
```
gregkalapos: This caused some trouble to me on macOS with the latest Rider (2019.2.2)

image

It [seems](https://developercommunity.visualstudio.com/content/problem/478557/error-cs1617-invalid-option-800-for-langversion.html) that mono is a little bit behind with C# 8.0 support - According to [this](https://developercommunity.visualstudio.com/content/problem/469096/selecting-c-8-breaks-solution-preferences-screen-i.html) VS for Mac already has a version supporting it - I try to figure this out, maybe it's just a mono update. 

Anyway, I think we don't use anything `8.0` related yet, right? We could use `7.3`  or `latest`, can't we? I'll report back if I could make it work on macOS.

src/Elastic.Apm.AspNetFullFramework/Properties/AssemblyInfo.cs
```diff
@@ -1,35 +0,0 @@
-using System.Reflection;
```
gregkalapos: ❤️  

I won't miss this file :)

gregkalapos:
Also, just some organizational things:
This covers https://github.com/elastic/apm-agent-dotnet/pull/47 and https://github.com/elastic/apm-agent-dotnet/pull/471, which is totally fine by me. If this gets merged those 2 can be also closed.

Directory.Build.props
Mpdreamz: Aye the `8.0` is a copypasta issue :)

Mpdreamz: Updated back to `latest`

kuisathaverat:
>I think we need to update msbuild on the Windows CI machine:

This could be a pain, we have to make it on the worker better than in the build, Is it need? if so I'll open an issue to request the change to Infra.

Mpdreamz:
Working on dropping down instead, using 2.1.505 everywhere instead (the one that vs2017  msbuild understands). When we move to 3.0 the image will need to be updated.

Directory.Build.props
gregkalapos: Awesome! A VS for Mac update (which probably triggered a Mono update) solved it anyway for me. But yeah, let's roll with `latest` for now.

test/Elastic.Apm.AspNetFullFramework.Tests/AppSettingsTests.cs
```diff
@@ -2,10 +2,9 @@
 using System.Collections.Generic;
 using System.Configuration;
 using Elastic.Apm.Config;
-using FluentAssertions;
 using Xunit;
 
-namespace Elastic.Apm.AspNetFullFramework.Tests
+namespace Elastic.Apm.AspNetFullFramework.Tests.FullFramework
```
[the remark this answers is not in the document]
Mpdreamz: Leftover from trying to isolate before going a different route. Will revert

test/Elastic.Apm.AspNetFullFramework.Tests/Elastic.Apm.AspNetFullFramework.Tests.csproj
```diff
@@ -9,19 +9,30 @@
     <!-- Always generate debug symbols this allows fluent symbols exception messages to include variable names -->
     <DebugSymbols>True</DebugSymbols>
   </PropertyGroup>
-  <ItemGroup>
+  <ItemGroup Condition="'$(OS)' == 'WINDOWS_NT'">
+    <PackageReference Include="Microsoft.AspNetCore.App" />
```
SergeyKleyman: Could you please clarify why do we need `Microsoft.AspNetCore.App` dependency?

Mpdreamz: https://github.com/dotnet/sdk/issues/2253 

This is too squash compiler warnings since this test project does not end up pulling these in.

.ci/linux/tools.sh
```diff
@@ -8,4 +8,4 @@ set -euxo pipefail
 
 # Install .Net SDK
 curl -O https://dot.net/v1/dotnet-install.sh
-/bin/bash ./dotnet-install.sh --install-dir ${HOME}/dotnet -Channel LTS
+/bin/bash ./dotnet-install.sh --install-dir ${HOME}/dotnet -version 2.2.401
```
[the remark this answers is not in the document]
Mpdreamz: No, the current solution can not use `2.2` since the windows CI loads VS 2017 build tools and calls msbuild explictly.

src/Elastic.Apm.AspNetFullFramework/Elastic.Apm.AspNetFullFramework.csproj
```diff
@@ -1,64 +1,31 @@
...
+    <RootNamespace>Elastic.Apm.AspNetFullFramework</RootNamespace>
+    <AssemblyVersion>1.0.0</AssemblyVersion>
+    <InformationalVersion>1.0.0-beta1</InformationalVersion>
+    <FileVersion>1.0.0</FileVersion>
     <SignAssembly>true</SignAssembly>
-  </PropertyGroup>
-  <PropertyGroup>
     <AssemblyOriginatorKeyFile>..\..\build\elasticapm.snk</AssemblyOriginatorKeyFile>
+    <DelaySign>false</DelaySign>
+    <PackageVersion>1.0.0-beta1</PackageVersion>
```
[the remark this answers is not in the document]
Mpdreamz: Bumped anyway 👍 We can now put most metadata in a shared `props` file 😄

PR #427, opened 2019-08-02
Auto code cleanup

gregkalapos:
During reviews I saw an increasing number of code style updates - which is good, we should maintain the coding style as defined in .editorconfig. In order to avoid these in PRs later and making reviews easier I ran auto cleanup on the whole solution.

No functional change here - only applied auto clenaup.

After [having a release this week](https://github.com/elastic/apm-agent-dotnet/releases/tag/1.0.0) probably this is a good time to do this.

Will try to merge it at a point where there are not much PRs open to minimize conflicts.


gregkalapos:
Opened #535 - this one is already out-of-date.

Issue #428, opened 2019-08-02
Null pointer in PayloadSenderV2


skynet2:
Having that error on remote server, but can not reproduce on local
```
Aug  3 01:22:24 server1 integration[4834]: [2019-08-02 11:22:24][Debug] - {Tracer} Starting Transaction{Id: 8525f9f45ad1798f, TraceId: 1b20305c5f76358cb3424438c407522f, ParentId: null, Name: POST /api/xx/deposit, Type: request, IsSampled: True}
```
Then there is a bunch of spans from entity framework 

And last lines are : 
```
Aug  3 01:22:26 server1 integration[4834]: [2019-08-02 11:22:26][Debug] - {PayloadSenderV2} Transaction added to the queue, Transaction{Id: 8525f9f45ad1798f, TraceId: 1b20305c5f76358cb3424438c407522f, ParentId: null, Name: POST Xx/Deposit, Type: request, IsSampled: True}
Aug  3 01:22:26 server1 integration[4834]: [2019-08-02 11:22:26][Debug] - {Transaction} Ending Transaction{Id: 8525f9f45ad1798f, TraceId: 1b20305c5f76358cb3424438c407522f, ParentId: null, Name: POST Xx/Deposit, Type: request, IsSampled: True}
Aug  3 01:22:26 server1 integration[4834]: [2019-08-02 11:22:26][Warning] - {PayloadSenderV2} Failed sending events. Following events were not transferred successfully to the server:
Aug  3 01:22:26 server1 integration[4834]: Span{Id: 8bb96752463c20a8, TransactionId: 8525f9f45ad1798f, ParentId: 8525f9f45ad1798f, TraceId: 1b20305c5f76358cb3424438c407522f, Name: SELECT x.id, x.extra, x.game_id, x."order", x.partner_aggregator_id, x.product_type, x.status, "x.Game".id, "x.Game".device_type, "x.Game".external_id, "x.Game".extra, "x.Game".has_demo, "x.Game".image_url, "x.Game".provider_id, "x.Game".recalled_at, "x.Game".released_at, "x.Game".status, "x.Game".token
Aug  3 01:22:26 server1 integration[4834]: FROM common.partner_games_new AS x
Aug  3 01:22:26 server1 integration[4834]: INNER JOIN common.partner_aggregators AS "x.PartnerAggregator" ON x.partner_aggregator_id = "x.PartnerAggregator".id
[...]

PR #429, opened 2019-08-03
Handle labels with `null`

gregkalapos:
Fixing #428

Idea: 
- In case a label is `null` we serialize it as `null` (string), so `Agent.Tracer.CurrentTransaction.Labels["foo"] = null;` is valid and we send `null` (as string) to the server.
- Keys as null are not valid, and that's because adding `null` as key to a `Dictionary` throws an exception and we just follow this.

### DIFF
--- a/src/Elastic.Apm/Report/Serialization/LabelsJsonConverter.cs
+++ b/src/Elastic.Apm/Report/Serialization/LabelsJsonConverter.cs
@@ -12,7 +12,11 @@ namespace Elastic.Apm.Report.Serialization
 			foreach (var keyValue in labels)
 			{
 				writer.WritePropertyName(SerializationUtils.TrimToPropertyMaxLength(keyValue.Key));
-				writer.WriteValue(string.IsNullOrEmpty(keyValue.Value) ? "null" : SerializationUtils.TrimToPropertyMaxLength(keyValue.Value));
+
+				if (keyValue.Value != null)
+					writer.WriteValue(SerializationUtils.TrimToPropertyMaxLength(keyValue.Value));
+				else
+					writer.WriteNull();
 			}
 			writer.WriteEndObject();
 		}
--- a/src/Elastic.Apm/Report/Serialization/LabelsJsonConverter.cs
+++ b/src/Elastic.Apm/Report/Serialization/LabelsJsonConverter.cs
@@ -12,7 +12,7 @@ namespace Elastic.Apm.Report.Serialization
 			foreach (var keyValue in labels)
 			{
 				writer.WritePropertyName(SerializationUtils.TrimToPropertyMaxLength(keyValue.Key));
-				writer.WriteValue(SerializationUtils.TrimToPropertyMaxLength(keyValue.Value));
+				writer.WriteValue(string.IsNullOrEmpty(keyValue.Value) ? "null" : SerializationUtils.TrimToPropertyMaxLength(keyValue.Value));
 			}
 			writer.WriteEndObject();
 		}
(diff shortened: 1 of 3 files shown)


src/Elastic.Apm/Report/Serialization/LabelsJsonConverter.cs
```diff
@@ -12,7 +12,7 @@ public override void WriteJson(JsonWriter writer, Dictionary<string, string> lab
 			foreach (var keyValue in labels)
 			{
 				writer.WritePropertyName(SerializationUtils.TrimToPropertyMaxLength(keyValue.Key));
-				writer.WriteValue(SerializationUtils.TrimToPropertyMaxLength(keyValue.Value));
+				writer.WriteValue(string.IsNullOrEmpty(keyValue.Value) ? "null" : SerializationUtils.TrimToPropertyMaxLength(keyValue.Value));
```
SergeyKleyman: JSON has `null` so I think we should use `Newtonsoft.Json.JsonWriter.WriteNull` if `keyValue.Value` is `null`.

BTW the only interesting case is when `keyValue.Value` is `null` - there's no reason to treat empty string any differently than any other non-null string so there is no need for `string.IsNullOrEmpty` in this case.

test/Elastic.Apm.Tests/SerializationTests.cs
```diff
@@ -16,6 +16,16 @@ namespace Elastic.Apm.Tests
 	/// </summary>
 	public class SerializationTests
 	{
+		// ReSharper disable once MemberCanBePrivate.Global
+		public static TheoryData SerializationUtilsTrimToPropertyMaxLengthVariantsToTest => new TheoryData<string, string>
```
gregkalapos: This is just due to auto cleanup 🤷‍♂️ - #427 addresses this...

test/Elastic.Apm.Tests/SerializationTests.cs
```diff
@@ -115,6 +125,22 @@ public void LabelsTruncationSpanContext()
...
+		public void LabelWithNullValueShouldBeCaptured()
+		{
+			var context = new SpanContext();
+			context.Labels["foo"] = null;
+
+			var json = SerializePayloadItem(context);
+			var deserializedContext = JsonConvert.DeserializeObject(json) as JObject;
+
+			Assert.NotNull(deserializedContext);
+			deserializedContext["foo"].Should().BeNull();
```
SergeyKleyman: This test passes even though `Elastic.Apm.Report.Serialization.LabelsJsonConverter.WriteJson` is not correct because the assert above is not correct neither, The correct assert is `deserializedContext["tags"]["foo"].Value<string>().Should().BeNull();`

SergeyKleyman:
As I mentioned at https://github.com/elastic/apm-agent-dotnet/issues/428#issuecomment-517884444:
> I think it's worth adding a code comment for `Labels` drawing user's attention to the fact that null value is completely okay but null key will throw ArgumentNullException.

src/Elastic.Apm/Report/Serialization/LabelsJsonConverter.cs
gregkalapos: Done

test/Elastic.Apm.Tests/SerializationTests.cs
gregkalapos: Fixed

gregkalapos:
> As I mentioned at [#428 (comment)](https://github.com/elastic/apm-agent-dotnet/issues/428#issuecomment-517884444):
> 
> > I think it's worth adding a code comment for `Labels` and public API documentation drawing user's attention to the fact that `null` value is completely okay but `null` key will throw ArgumentNullException.

Done

Issue #443, opened 2019-08-20
Nullref in HttpDiagnosticListenerImplBase.ProcessExceptionEvent


gregkalapos:
From https://discuss.elastic.co/t/apm-bug-causes-null-pointer-for-web-requests-in-net-4-6-1/195923

SergeyKleyman:
It might be also worth improving logging while we at it. At the moment it would have been not very useful for this bug:
```
[2019-08-20 11:56:43][Trace] - {HttpDiagnosticListenerImplBase} Processing exception event... Request URL: https://xxxx/api/georef/v1/regions/details?orqId=Cs_6a91a9c4-e40c-4457-89f2-06570b3f9805
Application startup exception: System.Net.WebException: The request was aborted: The request was canceled. ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Elastic.Apm.DiagnosticListeners.HttpDiagnosticListenerImplBase`2.ProcessExceptionEvent(Object eventValue, Uri requestUrl)
   at Elastic.Apm.DiagnosticListeners.HttpDiagnosticListenerImplBase`2.OnNext(KeyValuePair`2 kv)
   at System.Diagnostics.DiagnosticListener.Write(String name, Object value)
   at System.Diagnostics.HttpHandlerDiagnosticListener.RaiseResponseEvent(HttpWebRequest request, HttpStatusCode statusCode, WebHeaderCollection headers)
   at System.Diagnostics.HttpHandlerDiagnosticListener.HttpWebRequestArrayList.RemoveAt(Int32 index)
   at System.Net.Connection.ReadStartNextRequest(WebRequest currentRequest, ConnectionReturnResult& returnResult)
   at System.Net.ConnectStream.CallDone(ConnectionReturnResult returnResult)
   at System.Net.HttpWebResponse..ctor(Uri responseUri, KnownHttpVerb verb, CoreResponseData coreData, String mediaType, Boolean usesProxySemantics, DecompressionMethods decompressionMethod, Boolean isWebSocketResponse, String connectionGroupName)
   at System.Net.HttpWebRequest.SetResponse(CoreResponseData coreResponseData)
   --- End of inner exception stack trace ---
   at System.Net.HttpWebRequest.GetResponse()
   at ServiceStack.ServiceClientBase.Send[TResponse](String httpMethod, String relativeOrAbsoluteUrl, Object request)
```
[...]

PR #450, opened 2019-08-23
Fix nullref in HttpDiagnosticListenerFullFrameworkImpl

gregkalapos:
Proposed fix for #443. 

General thoughts:
- This problem is in `Elastic.Apm` and it causes a NullRef exception, so if possible we should address it fairly soon.
- Therefore, I'd suggest to only focus on fixing this issue here - happy to open follow ups. 

Idea:
I made `ProcessExceptionEvent` in `HttpDiagnosticListenerImplBase` `virtual` and I have a specialized implantation of that method in `HttpDiagnosticListenerFullFrameworkImpl` - that takes care of the case that currently causes the nullref.

By seeing the code, I think having a common base for the 2 cases (.NET Core and Full Framework) is maybe not the best way to go. ApplicationInsights for example does not have that common base class either ([Full Framework](https://github.com/microsoft/ApplicationInsights-dotnet-server/blob/3cc166e3c9ec9d3f4ce77f48d3ab8c6bd6218808/Src/DependencyCollector/Shared/Implementation/HttpDesktopDiagnosticSourceListener.cs), [.NET Core](https://github.com/microsoft/ApplicationInsights-dotnet-server/blob/055bcaaec7249cf91ca3e1e59e8bcc08393e10e7/Src/DependencyCollector/Shared/HttpCoreDiagnosticSourceListener.cs)). I think we have unnecessary complexity in our code by adding the common base with little benefit. I think the cleanest way would be to have 2 separate implementations. Nevertheless to address the main issue (which is in my opinion the nullref in the current release) I'd propose to merge this short fix and maybe address other things later.

Also added a test to cover the specific case that caused the nullref.

### DIFF
--- a/test/Elastic.Apm.Tests.MockApmServer/AssertValidExtensions.cs
+++ b/test/Elastic.Apm.Tests.MockApmServer/AssertValidExtensions.cs
@@ -217,7 +217,16 @@ namespace Elastic.Apm.Tests.MockApmServer
 
 			thisObj.Code?.AssertValid();
 			thisObj.StackTrace?.AssertValid();
-			thisObj.Type.AssertValid();
+
+			if (string.IsNullOrEmpty(thisObj.Type))
+				thisObj.Message.AssertValid();
+			else
+				thisObj.Message?.AssertValid();
+
+			if (string.IsNullOrEmpty(thisObj.Message))
+				thisObj.Type.AssertValid();
+			else
+				thisObj.Type?.AssertValid();
 		}
 
 		internal static void AssertValid(this Database thisObj)
--- a/sample/AspNetFullFrameworkSampleApp/Controllers/HomeController.cs
+++ b/sample/AspNetFullFrameworkSampleApp/Controllers/HomeController.cs
@@ -23,6 +23,8 @@ namespace AspNetFullFrameworkSampleApp.Controllers
 
 		internal const string CustomChildSpanThrowsPageRelativePath = HomePageRelativePath + "/" + nameof(CustomChildSpanThrows);
 
+		internal const string ForbidHttpResponsePageRelativePath = HomePageRelativePath + "/" + nameof(ForbidHttpResponse);
+
 		internal const string CustomSpanThrowsInternalMethodName = nameof(CustomSpanThrowsInternal);
 		internal const string CustomSpanThrowsPageRelativePath = HomePageRelativePath + "/" + nameof(CustomSpanThrows);
 
@@ -63,6 +65,15 @@ namespace AspNetFullFrameworkSampleApp.Controllers
 					CaptureControllerActionAsSpanQueryStringKey);
 			}
 			return bool.Parse(captureControllerActionAsSpanValues[0]);
+		}
+
+		public async Task<ActionResult> ForbidHttpResponse()
+		{
+			//see https://github.com/elastic/apm-agent-dotnet/issues/443
+			var httpClient = new HttpClient();
+			var res = await httpClient.GetAsync("https://httpstat.us/403");
+			var retVal = new ContentResult { Content = res.IsSuccessStatusCode.ToString() };
+			return retVal;
 		}
 
 		public Task<ActionResult> Contact()
--- a/src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerFullFrameworkImpl.cs
+++ b/src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerFullFrameworkImpl.cs
@@ -37,13 +37,13 @@ namespace Elastic.Apm.DiagnosticListeners
 		/// </summary>
 		protected override void ProcessExceptionEvent(object eventValue, Uri requestUrl)
 		{
-			_logger.Trace()?.Log("Processing stop.ex event... Request URL: {RequestUrl}", requestUrl);
+			Logger.Trace()?.Log("Processing stop.ex event... Request URL: {RequestUrl}", requestUrl);
 
 			var requestObject = eventValue.GetType().GetTypeInfo().GetDeclaredProperty(EventRequestPropertyName)?.GetValue(eventValue);
 
 			if (!(requestObject is HttpWebRequest request))
 			{
-				_logger.Trace()
+				Logger.Trace()
 					?.Log("Actual type of object ({EventRequestPropertyActualType}) in event's {EventRequestPropertyName} property " +
 						"doesn't match the expected type ({EventRequestPropertyExpectedType}) - exiting",
 						requestObject.GetType().FullName, EventRequestPropertyName, typeof(HttpWebRequest).FullName);
@@ -51,7 +51,7 @@ namespace Elastic.Apm.DiagnosticListeners
 			}
 			if (!ProcessingRequests.TryRemove(request, out var span))
 			{
-				_logger.Warning()
+				Logger.Warning()
 					?.Log("Failed capturing request (failed to remove from ProcessingRequests) - " +
 						"This Span will be skipped in case it wasn't captured before. " +
 						"Request: method: {HttpMethod}, URL: {RequestUrl}", RequestGetMethod(request), requestUrl);
@@ -74,7 +74,7 @@ namespace Elastic.Apm.DiagnosticListeners
 					}
 					else
 					{
-						_logger.Trace()
+						Logger.Trace()
 							?.Log("Actual type of object ({EventStatusCodePropertyActualType}) in event's {EventStatusCodePropertyName} property " +
 								"doesn't match the expected type ({EventStatusCodeePropertyExpectedType})",
 								statusCodeObject.GetType().FullName, "StatusCode", typeof(HttpStatusCode).FullName);
--- a/src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerFullFrameworkImpl.cs
+++ b/src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerFullFrameworkImpl.cs
@@ -1,5 +1,7 @@
 using System;
 using System.Net;
+using System.Reflection;
+using Elastic.Apm.Logging;
 
 namespace Elastic.Apm.DiagnosticListeners
 {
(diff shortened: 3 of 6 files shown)


test/Elastic.Apm.Tests.MockApmServer/AssertValidExtensions.cs
```diff
@@ -217,7 +217,16 @@ internal static void AssertValid(this CapturedException thisObj)
 
 			thisObj.Code?.AssertValid();
 			thisObj.StackTrace?.AssertValid();
-			thisObj.Type.AssertValid();
```
gregkalapos: `Message` or `Type` is required. If `Message` is present `Type` can be null. 

https://github.com/elastic/apm-server/blob/master/docs/spec/errors/error.json#L98

SergeyKleyman:
I think it's worth discussing if `"System.Net.Http.Desktop.HttpRequestOut.Ex.Stop"` should be treated as _Exception_ event or as _Stop_ event (similar to `"...Stop"`) with different payload? I think `"...Ex.Stop"` meaning is much closer to `"...Stop"` than to .NET Core's `"System.Net.Http.Exception"`. If I follow implementation correctly [`"...Stop"`](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/HttpHandlerDiagnosticListener.cs#L540) and [`"...Ex.Stop"`](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/HttpHandlerDiagnosticListener.cs#L564) are generated under very similar conditions and the reasons for two payload kinds for _Stop_ event are just technical.

gregkalapos:
> I think it's worth discussing if `"System.Net.Http.Desktop.HttpRequestOut.Ex.Stop"` should be treated as _Exception_ event or as _Stop_ event (similar to `"...Stop"`) with different payload? I think `"...Ex.Stop"` meaning is much closer to `"...Stop"` than to .NET Core's `"System.Net.Http.Exception"`. If I follow implementation correctly [`"...Stop"`](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/HttpHandlerDiagnosticListener.cs#L540) and [`"...Ex.Stop"`](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/HttpHandlerDiagnosticListener.cs#L564) are generated under very similar conditions and the reasons for two payload kinds for _Stop_ event are just technical.

In my opinion `"System.Net.Http.Desktop.HttpRequestOut.Ex.Stop"` should be treated as `"System.Net.Http.Desktop.HttpRequestOut.Ex.Stop"` and we should not trying to find something identical for .NET Core, so in my opinion the cleanest way to solve this would be something similar to what Application Insights is doing: not trying to share code between the two. 

Do you see this question as a blocker to this PR? In my opinion refactoring the PR to handle it as `Stop` won't make a big difference. 

I agree with this part:
> "...Stop" and "...Ex.Stop" are generated under very similar conditions 

On the other hand, this is just the current implementation and that can change. 

So, overall I think both handling it as "Stop" or "Exception" makes the code harder to read and ultimately the simplest and cleanest way would be to not trying to map it. Unless you think it's a blocker I'd keep it this way and maybe think about removing the common base later. What do you think?

gregkalapos:
To your question added in your edit:

>  Also AFAIK .NET Core HTTP calls with 403 responses do not result in Error events so it would make sense to align .NET Framework and .NET Core behavior in the respect, don't you agree?

Yes, that's true, on the other hand this was already the case before (since in Full Framework the code that generated the Errors never worked) and I see this more like a "nice to have". I'd focus in this PR on the main problem which is I think the nullref. So I'd fix that and I'm happy to open follow ups, but preferably I'd focus on just fixing the nullref here.

SergeyKleyman:
> Do you see this question as a blocker to this PR? In my opinion refactoring the PR to handle it as `Stop` won't make a big difference.

I do because after this change 403 HTTP calls generate error events for .NET Framework but not for .NET Core. That is the main reason I think keeping common base is very useful because it makes us think about potential differences in behavior between .NET Framework and .NET Core and whether each of those differences is valid or not.

> since in Full Framework the code that generated the Errors never worked

It's true but it's not what I'm saying above. You are talking here about HTTP calls that throw exceptions (for example failed to resolve host name to IP, etc.) not HTTP calls with responses > 300. Just because there's one aspect where .NET Framework and .NET Core are not aligned (hopefully just before we find a solution for that as well) should we give up on keeping them in sync altogether?

> On the other hand, this is just the current implementation and that can change.

Yes, it's possible but we have to assume some contract for this API otherwise how can we use it? Unfortunately there's no clear documentation specifying the contract so looking at the implementation is the only way to find information.

> So, overall I think both handling it as "Stop" or "Exception" makes the code harder to read

I'm not sure what you mean by "both" could you clarify that please?
As I mentioned in my comment in the issue (https://github.com/elastic/apm-agent-dotnet/issues/443#issuecomment-523141935) I could not find any _Exception_ event for .NET Framework so `"...Stop"` and `"...Ex.Stop"` are both _Stop_ events.

I think the code becomes more complex exactly because of trying to treat `"...Ex.Stop"` as _Exception_ event which it's not.
Later I will push a suggestion for change to your PR that treats both `"...Stop"` and `"...Ex.Stop"` as `Stop` events and we can discuss it.

SergeyKleyman:
So it seems we agree that both `"...Stop"` and `"...Ex.Stop"` should be treated as _Stop_  events so are there any reasons not do it in this PR?

test/Elastic.Apm.Tests.MockApmServer/AssertValidExtensions.cs
```diff
@@ -217,7 +217,16 @@ internal static void AssertValid(this CapturedException thisObj)
...
+
+			if (string.IsNullOrEmpty(thisObj.Type))
```
SergeyKleyman: Is there a reason for special treatment for empty string? it seems intake API spec treats it as valid.

test/Elastic.Apm.Tests.MockApmServer/AssertValidExtensions.cs
```diff
@@ -217,7 +217,16 @@ internal static void AssertValid(this CapturedException thisObj)
...
+				thisObj.Message.AssertValid();
+			else
+				thisObj.Message?.AssertValid();
+
+			if (string.IsNullOrEmpty(thisObj.Message))
```
SergeyKleyman: Is there a reason for special treatment for empty string? it seems intake API spec treats it as valid.

gregkalapos:
> > Do you see this question as a blocker to this PR? In my opinion refactoring the PR to handle it as `Stop` won't make a big difference.
> 
> I do because after this change 403 HTTP calls generate error events for .NET Framework but not for .NET Core. That is the main reason I think keeping common base is very useful because it makes us think about potential differences in behavior between .NET Framework and .NET Core and whether each of those differences is valid or not.

Better to have this difference than crashing the user's app ;) Trying to keep the 2 in sync is a nice effort, but given that on Full Framework we never get Exception objects except [here](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/HttpHandlerDiagnosticListener.cs#L84) - which we currently don't handle, I think this won't work - those 2 send different payloads. So I think we won't be able to keep those 100% in sync anyway. It's nice in theory, but won't work in practice.

> > since in Full Framework the code that generated the Errors never worked
> 
> It's true but it's not what I'm saying above. You are talking here about HTTP calls that throw exceptions (for example failed to resolve host name to IP, etc.) not HTTP calls with responses > 300. Just because there's one aspect where .NET Framework and .NET Core are not aligned (hopefully just before we find a solution for that as well) should we give up on keeping them in sync altogether?

Yes, because I think the differences are too big, the payloads are different, the keys are different, and the meaning of the keys are different. Probably that's why AppInsights treats those differently. 

> > On the other hand, this is just the current implementation and that can change.
> 
> Yes, it's possible but we have to assume some contract for this API otherwise how can we use it? Unfortunately there's no clear documentation specifying the contract so looking at the implementation is the only way to find information.

Very good point! Yes, the contract is I'd say the key - those are different for Full Framework and .NET Core - I think it's ok to do assumptions based on the keys, but nothing more - I'd not make assumptions on the fact that 2 methods are called close to each other in a method.

> > So, overall I think both handling it as "Stop" or "Exception" makes the code harder to read
> 
> I'm not sure what you mean by "both" could you clarify that please?
> As I mentioned in my comment in the issue ([#443 (comment)](https://github.com/elastic/apm-agent-dotnet/issues/443#issuecomment-523141935)) I could not find any _Exception_ event for .NET Framework so `"...Stop"` and `"...Ex.Stop"` are both _Stop_ events.

By "both" I mean that in my opinion it does not matter if the `OnNext` calls with `"System.Net.Http.Desktop.HttpRequestOut.Ex.Stop"` end up in `ProcessStopEvent` or in `ProcessStopEvent` - this mapping is in my opinion is unfortunate.

> I think the code becomes more complex exactly because of trying to treat `"...Ex.Stop"` as _Exception_ event which it's not.
> Later I will push a suggestion for change to your PR that treats both `"...Stop"` and `"...Ex.Stop"` as _Stop_ events and we can discuss it.

No, the code becomes complex, because we try to share. Look at our code and then the Application Insights code.

gregkalapos:
> So it seems we agree that both `"...Stop"` and `"...Ex.Stop"` should be treated as _Stop_ events so are there any reasons not do it in this PR?

No, I say it does not matter. My main reason is that we just waste time on it. But ok, let's make it to `Stop`. I'll push it.

SergeyKleyman:
By keeping in sync Full Framework and .NET Core I don't mean we have to have common base - we can do it without common base. I think the behavior visible to user should be aligned - either we generate error events for HTTP calls with response status > 300 for both of them or neither one.

gregkalapos:
> By keeping in sync Full Framework and .NET Core I don't mean we have to have common base - we can do it without common base. I think the behavior visible to user should be aligned - either we generate error events for HTTP calls with response status > 300 for both of them or neither one.

Ok, I'll remove that part from the PR.

So:
- I'll change handling this case in stop and not in the exception part
- And I'll remove sending an Error.

Happy?

SergeyKleyman:
I think introducing a change that generates error events on HTTP calls with response status > 300 and only for Full Framework is a blocker.

The way I thought to go about is
1. Instead of `HttpDiagnosticListenerImplBase` assuming that each event has just one name (which is not the case anymore) `HttpDiagnosticListenerImplBase` can declare an `enum` for `Start`, `Stop`, `Exception` and `Unrecognized` and each derived can implement a parse method to map `Key` string to enum.
2. I think `HttpDiagnosticListenerImplBase` doesn't actually need `TResponse` type parameter because it's only used for extracting status code and it can refactored to derived class to extract status code directly from `eventValue`.

Those are just initial thoughts - I didn't try them out yet so you would like to use them please take that into account.

gregkalapos:
> I think introducing a change that generates error events on HTTP calls with response status > 300 and only for Full Framework is a blocker.
> 
> The way I thought to go about is
> 
> 1. Instead of `HttpDiagnosticListenerImplBase` assuming that each event has just one name (which is not the case anymore) `HttpDiagnosticListenerImplBase` can declare an `enum` for `Start`, `Stop`, `Exception` and `Unrecognized` and each derived can implement a parse method to map `Key` string to enum.
> 2. I think `HttpDiagnosticListenerImplBase` doesn't actually need `TResponse` type parameter because it's only used for extracting status code and it can refactored to derived classes to extract status code directly from `eventValue`.
> 
> Those are just initial thoughts - I didn't try them out yet so if you would like to use them please take that into account.

Ok, so... that would be more than what I'd do quickly on the weekend. Let's come back to this on Monday.

SergeyKleyman:
@gregkalapos I've pushed changes with my suggestion above and I refactored `HttpDiagnosticListenerTests` to simulate evens for both Full .NET Framework and .NET Core - the tests that don't simulate events but invoke `HttpClient` are still .NET Core only but we can go back to it later and consider if we still want `Elastic.Apm.Tests` to target both .NET Core and Full .NET Framework.
Now that we have tests making sure that the behavior for .NET Core and Full .NET Framework is aligned we can consider if we still want implementation to have common base without risk of unintended diverging.

test/Elastic.Apm.Tests.MockApmServer/AssertValidExtensions.cs
gregkalapos: This is related to this one: https://github.com/elastic/apm-agent-dotnet/pull/450/files#r317177339 

I'm trying here to make sure either `Type` or `Message` is set - either one of those or both. Both can't be empty.

test/Elastic.Apm.Tests.MockApmServer/AssertValidExtensions.cs
gregkalapos: Same as above.

gregkalapos:
As discussed Let's go with the smaller change and release soon.

> Therefore, I'd suggest to only focus on fixing this issue here - happy to open follow ups.

#455 is already open for the rest. 🚀

PR #451, opened 2019-08-25
Fix flaky DefaultStackTraceLimitAndSpanFramesMinDuration_ShortSpan 

gregkalapos:
I see this test failing sometimes. I think it's because in the CI machine which is sometimes slower than expected the span takes more than 5ms.

### DIFF
--- a/test/Elastic.Apm.Tests/StackTraceTests.cs
+++ b/test/Elastic.Apm.Tests/StackTraceTests.cs
@@ -281,7 +281,11 @@ namespace Elastic.Apm.Tests
 			}
 
 			payloadSender.FirstSpan.Should().NotBeNull();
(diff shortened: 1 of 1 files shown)


SergeyKleyman:
After this change will there be at least one test that verifies that stack trace is not captured for spans shorter than `SpanFramesMinDuration`? If not maybe it's worth adding a test that sets `SpanFramesMinDuration` to very high value to ensure no-op span is shorter?

gregkalapos:
> After this change will there be at least one test that verifies that stack trace is not captured for spans shorter than `SpanFramesMinDuration`? If not maybe it's worth adding a test that sets `SpanFramesMinDuration` to very high value to ensure no-op span is shorter?

I pushed [one](https://github.com/elastic/apm-agent-dotnet/pull/451/files#diff-8c7ba9914ef04d48944496cb34487ae1R299).

PR #455, opened 2019-08-26
Treat Full Framework HTTP DiagnosticListener event "...Ex.Stop" event as "...Stop"

SergeyKleyman:
Fixes elastic/apm-agent-dotnet#443 without generating error events for Full Framework HTTP call with status code > 300

- Treat "...Ex.Stop" event as "...Stop"
- Add test for Full Framework InitializationFailed event
- Add NoExceptionEscapesFromOnNext test case
- Refactor HttpDiagnosticListenerTests to simulate events for both Full .NET Framework and .NET Core


russcam:
@gregkalapos I think this can be closed, since #450 is merged?

gregkalapos:
> @gregkalapos I think this can be closed, since #450 is merged?

Agree, let's close this.

PR #465, opened 2019-09-02
Add support for Server certificate verification

russcam:
This commit adds support for server certificate verification. A configuration value can be supplied

1. through environment variables with key `"ELASTIC_APM_VERIFY_SERVER_CERT"` and a value that can be parsed to a boolean
2. through configuration with key `"ElasticApm:VerifyServerCert"` and a value than can be parsed to a boolean

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -394,6 +394,26 @@ Both the agents and the APM server have to be configured with the same secret to
 Use this setting in case the APM Server requires a token (e.g. APM Server in Elastic Cloud).
 
 [float]
+[[config-verify-server-cert]]
+==== `VerifyServerCert`
+
+[options="header"]
+|============
+| Environment variable name        | IConfiguration or Web.config key
+| `ELASTIC_APM_VERIFY_SERVER_CERT` | `ElasticApm:VerifyServerCert`
+|============
+
+[options="header"]
+|============
+| Default                 | Type
+| `true`                  | Boolean
+|============
+
+By default, the agent verifies the SSL certificate if you use an HTTPS connection to the APM server.
+
+Verification can be disabled by changing this setting to false.
+
+[float]
 [[config-flush-interval]]
 ==== `FlushInterval` (added[1.1])
 
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -395,7 +395,7 @@ Use this setting in case the APM Server requires a token (e.g. APM Server in Ela
 
 [float]
 [[config-verify-server-cert]]
-==== `VerifyServerCert`
+==== `VerifyServerCert` (added[1.3])
 
 [options="header"]
 |============
--- a/src/Elastic.Apm/Config/AbstractConfigurationReader.cs
+++ b/src/Elastic.Apm/Config/AbstractConfigurationReader.cs
@@ -117,6 +117,13 @@ namespace Elastic.Apm.Config
 			return kv.Value;
 		}
 
+		protected bool ParseVerifyServerCert(ConfigurationKeyValue kv)
+		{
+			if (kv == null || string.IsNullOrEmpty(kv.Value)) return DefaultValues.VerifyServerCert;
+			// ReSharper disable once SimplifyConditionalTernaryExpression
+			return bool.TryParse(kv.Value, out var value) ? value : DefaultValues.VerifyServerCert;
+		}
+
 		protected bool ParseCaptureHeaders(ConfigurationKeyValue kv) => ParseBoolOption(kv, DefaultValues.CaptureHeaders, "CaptureHeaders");
 
 		protected LogLevel ParseLogLevel(ConfigurationKeyValue kv)
--- a/src/Elastic.Apm/BackendComm/BackendCommUtils.cs
+++ b/src/Elastic.Apm/BackendComm/BackendCommUtils.cs
@@ -3,7 +3,9 @@ using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
+using System.Net.Security;
 using System.Reflection;
+using System.Security.Cryptography.X509Certificates;
 using System.Text;
 using System.Text.RegularExpressions;
 using Elastic.Apm.Api;
@@ -105,6 +107,18 @@ namespace Elastic.Apm.BackendComm
 				servicePoint.ConnectionLimit = 20;
 			});
 
+		private static HttpClientHandler CreateHttpClientHandler(bool verifyServerCert, IApmLogger logger)
+		{
+			bool ServerCertificateCustomValidationCallback(HttpRequestMessage message, X509Certificate2 certificate, X509Chain chain, SslPolicyErrors policyError)
+			{
+				if (policyError == SslPolicyErrors.None) return true;
+
+				logger.Trace()?.Log("Certificate validation failed. Policy error {PolicyError}", policyError);
+				return !verifyServerCert;
+			}
+
+			return new HttpClientHandler { ServerCertificateCustomValidationCallback = ServerCertificateCustomValidationCallback };
+		}
 
 		internal static HttpClient BuildHttpClient(IApmLogger loggerArg, IConfigSnapshot config, Service service, string dbgCallerDesc
 			, HttpMessageHandler httpMessageHandler = null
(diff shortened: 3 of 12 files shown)


russcam:
There aren't tests for this in this PR. Testing could be supported

1. through `HttpListener` by creating a self-signed cert for a given HTTPS prefix and adding to the certificate store as part of `LocalServer` instantiation
2. through `MockApmServer`

src/Elastic.Apm/Report/PayloadSenderV2.cs
```diff
@@ -91,6 +92,13 @@ internal class PayloadSenderV2 : IPayloadSender, IDisposable
 			string AdaptUserAgentValue(string value) => Regex.Replace(value, "[ /()<>@,:;={}?\\[\\]\"\\\\]", "_");
 		}
 
+		private static HttpClientHandler CreateHttpClientHandler(IConfigurationReader configurationReader) =>
+			new HttpClientHandler
+			{
+				ServerCertificateCustomValidationCallback = (message, certificate, chain, policyError) =>
+					policyError == SslPolicyErrors.None || !configurationReader.VerifyServerCert
```
SergeyKleyman: Call `configurationReader.VerifyServerCert` cause re-parsing of the option - I'm working on fixing this performance issue in another PR but for this PR I think it would be better to just cache the value for example as one of the lambda captures    .

SergeyKleyman: For supportability purposes I think it's worth to add logging (`DEBUG` level for example) if there is indeed an issue with the certificate that we are overriding and what is that issue (I assume it's in `policyError` parameter, right?)

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
```diff
@@ -52,6 +52,13 @@ protected internal static bool TryParseLogLevel(string value, out LogLevel level
 			}
 		}
 
+		protected bool ParseVerifyServerCert(ConfigurationKeyValue kv)
+		{
+			if (kv == null || string.IsNullOrEmpty(kv.Value)) return true;
```
SergeyKleyman: I would add a named constant for the default value - https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm/Config/ConfigConsts.cs#L13

src/Elastic.Apm/Report/PayloadSenderV2.cs
russcam: ++

I thought there may be some desire to make configuration changes reloadable, based on https://github.com/elastic/apm-agent-dotnet/blob/9efee030ac820e66767c71d5e61948d6c188c38f/src/Elastic.Apm.AspNetCore/Config/MicrosoftExtensionsConfig.cs#L47, but can pass the `bool` configuration value directly to `CreateHttpClientHandler()`

russcam: You'd want to log why certification validation fails, even when the user has specified to skip validation?

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
russcam: ++

src/Elastic.Apm/Report/PayloadSenderV2.cs
SergeyKleyman: I think it might be useful to know if the option has any effect and what can kind of validation issues does it override - we can lower the level to `TRACE` which is disabled by default.

SergeyKleyman: Yes, we will make it re-loadable but from performance point of view the right way to do it is by re-parsing configuration only on change and not on each read access. Do you agree? (BTW this what I plan to do in that other PR that I mentioned above - I'll add a note to myself to make sure we read option value from the latest configuration snapshot in this callback).

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
```diff
@@ -52,6 +52,13 @@ protected internal static bool TryParseLogLevel(string value, out LogLevel level
 			}
 		}
 
+		protected bool ParseVerifyServerCert(ConfigurationKeyValue kv)
+		{
+			if (kv == null || string.IsNullOrEmpty(kv.Value)) return DefaultValues.VerifyServerCert;
+
+			return !bool.TryParse(kv.Value, out var value) || value;
```
SergeyKleyman: That is why I don't like these "too smart for its own good" Rider refactoring suggestions - for example here it assumes that `DefaultValues.VerifyServerCert` is `true`. If you change `DefaultValues.VerifyServerCert` to `false` then `return !bool.TryParse(kv.Value, out var value) || value;` will be incorrect since it will return `true` when it cannot parse.

src/Elastic.Apm/Report/PayloadSenderV2.cs
russcam: It's going to override all verification issues 😄 Logging can be added if you think it's worthwhile, but I'm not sure it is in this case, because setting `VerifyServerCert` to `false` says the user doesn't care.

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
russcam: added a test for the default configuration value

src/Elastic.Apm/Report/PayloadSenderV2.cs
SergeyKleyman: > It's going to override all verification issues

That is why logging the actual issue being overridden (`policyError` value) is useful.

> if you think it's worthwhile

I think it can helpful supportability feature. For example we might want to temporarily set `VerifyServerCert` to `false` just to see in logs what errors prevent normal HTTPS communication (I don't know if exception thrown on connection failure contains that level of detail).

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
SergeyKleyman: Test is good although it doesn't get to `return !bool.TryParse(kv.Value, out var value) || value;` line - the test covers `string.IsNullOrEmpty(kv.Value)` is `true` branch.

wsz34:
Hello,

Do you hanve any information when this PR will be merged and released?

SergeyKleyman:
> Hello,
> 
> Do you have any information when this PR will be merged and released?

Hi,

The main roadblock for merging this PR is absence of automated tests. We do have mock APM Server for tests (`test/Elastic.Apm.Tests.MockApmServer` used by `test/Elastic.Apm.AspNetFullFramework.Tests`) but we don't have any tests for agent's communication with APM Server over HTTPS yet so there's no code to set up HTTPS for automated tests. We welcome community PRs if you would like to take a stab at it.

se:
Hi guys, any update on this?

russcam:
I've reset and rebased on top of master, applied changes on top of new component layout and force push updated this branch/PR.

gregkalapos:
> I've reset and rebased on top of master, applied changes on top of new component layout and force push updated this branch/PR.

After @russcam's cleanup I merged some things, that caused new conflicts  🤦‍♂  So I cleaned it again. 

Also pushed documentation for it.

docs/configuration.asciidoc
```diff
@@ -393,6 +393,26 @@ This string is used to ensure that only your agents can send data to your APM se
 Both the agents and the APM server have to be configured with the same secret token.
 Use this setting in case the APM Server requires a token (e.g. APM Server in Elastic Cloud).
 
+[float]
+[[config-verify-server-cert]]
+==== `VerifyServerCert`
```
```suggestion
==== `VerifyServerCert` (added[1.3])
```

gregkalapos: Oh, that is a very good point!

russcam:
> Regarding testing: installing the cert to the machine is not so easy for tests - maybe we could have 1 test that we limit to windows only and install a cert there.

MockApmServer could be amended to allow configuring a HTTPS endpoint with a certificate, without relying on certificate stores. For example,

```c#
	public class MockApmServer
	{
		// ...

		public MockApmServer(IApmLogger logger, string dbgCurrentTestName, bool https = false, string certFile = null, string certPassword = null)
		{
			InternalLogger = logger.Scoped(ThisClassName);
			_dbgCurrentTestName = dbgCurrentTestName;
			_https = https;
			_certFile = certFile ?? "tests.pfx";
			_certPassword = certPassword ?? string.Empty;
		}

                // ...

		private IWebHostBuilder CreateWebHostBuilder()
		{
			var builder = WebHost.CreateDefaultBuilder(new string[0])
				.ConfigureServices(services =>
				{
					services.AddMvc()
						.AddApplicationPart(typeof(MockApmServer).Assembly)
						.SetCompatibilityVersion(CompatibilityVersion.Version_2_1);

					// Add this instance of MockApmServer as injected dependency for controllers
					services.AddSingleton(this);
				})
				.UseStartup<Startup>();

			if (_https)
			{
				builder.UseUrls($"https://localhost:{_port}")
					.UseKestrel((builderContext, options) =>
					{
						options.Configure(builderContext.Configuration.GetSection("Kestrel"));
						options.ConfigureHttpsDefaults(listenOptions =>
						{
							listenOptions.SslProtocols = SslProtocols.Tls12;
							listenOptions.ServerCertificate = new X509Certificate2(_certFile, _certPassword);
						});
					});
			}
			else
			{
				builder.UseUrls($"http://localhost:{_port}");
			}

			return builder;
		}
        }
```

`MockApmServer` seems like it would be useful to easily instantiate for tests where a request needs to be made, e.g.

```c#
using (var server = new MockApmServer(https: true))
{
    // perform some transactions with agent

    // make some assertions on agent and/or server
}
```

gregkalapos:
Thanks @russcam. I'm merging this now, since I'd say this would be good to have in the next release. 

Opened https://github.com/elastic/apm-agent-dotnet/issues/722 to cover the test - and linked the comment above in that issue.

PR #472, opened 2019-09-04
Add ability to configure transaction max spans

### LINKED ISSUE #178: Add support for transaction_max_spans configuration option
`transaction_max_spans` configuration option (type: Integer) specifies maximum number of spans for a single transaction. Following spans are ignored and `transaction.span_count.dropped.total` is bumped.
Default value: 500

vhatsura:
closes #178 

Open questions:
* comparing started transaction with transaction max spans from configuration reader. The main question: if we increment `started` with `Interlocked.Increment` is it enough for simple comparing?

TBD:
* [x] tests
* [x] documentation


test/Elastic.Apm.Tests/Mocks/TestAgentConfigurationReader.cs
```diff
@@ -64,6 +67,12 @@ public class TestAgentConfigurationReader : AbstractConfigurationReader, IConfig
 		public string Environment => ParseEnvironment(Kv(ConfigConsts.EnvVarNames.Environment, _environment, Origin));
 		public string SecretToken => ParseSecretToken(Kv(ConfigConsts.EnvVarNames.SecretToken, _secretToken, Origin));
 		public bool CaptureHeaders => ParseCaptureHeaders(Kv(ConfigConsts.EnvVarNames.CaptureHeaders, _captureHeaders, Origin));
+
+		public double TransactionSampleRate =>
```
SergeyKleyman: You copied `TransactionSampleRate` by mistake - I've pushed a change reverting it.

SergeyKleyman:
> if we increment started with Interlocked.Increment is it enough for simple comparing?

Unfortunately not in this case since you use `Interlocked.Increment` only when updating `started` and `dropped` counters but not when reading them:
```
if (enclosingTransaction.SpanCount.Started >= _configurationReader.TransactionMaxSpans)
```
and even if you did use some atomic API when reading the counters ([`Volatile.Read`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.volatile.read?view=netframework-4.8#System_Threading_Volatile_Read_System_Int32__) would be the best fit IMHO) that would have been two atomic calls but with a possible race condition between them. I recommend to introduce a separate `_total` counter (`private` - not included in the event sent to APM Server). I've pushed a suggested change - please take a look.

SergeyKleyman:
After discussion with other agents teams it seems that reusing span's `IsSampled` flag for both not sampled and dropped spans can be problematic. For example for distributed tracing over HTTP works by injecting [`traceparent` header](https://www.w3.org/TR/trace-context/) into HTTP request on the client side and using it on server side to continue the trace. In addition to other fields `traceparent` has `sampled` field (in the past it was called `recorded` field that is why we still call it `recorded` in .NET agent's `DistributedTracingData` class). So this `sampled` field should be set to 1 for sampled transaction (even if the span is dropped) and 0 for not sampled transaction. So if we will re-use span's `IsSampled` for both not sampled and dropped spans it will create an issue for `traceparent`'s `sampled` field.

I think the best approach would be:
1) `Span`'s `IsSampled` should use `IsSampled` from `_enclosingTransaction` because those two should always be in sync.
2) Add `private` `_isDropped` flag to `Span` and use it together with `IsSampled` to decide whether span should be sent to APM server.

vhatsura:
@SergeyKleyman, I got your point, however have additional question:
> Add private _isDropped flag to Span and use it together with IsSampled to decide whether span should be sent to APM server.

As, I understand in Elastic APM, non-sampled transactions do get reported to the APM server. Do we need to have the same logic for spans? Because span can be a parent for transaction in another service or `_dropped` will have higher priority on `IsSampled`?

SergeyKleyman:
@vhatsura

> As, I understand in Elastic APM, non-sampled transactions do get reported to the APM server. Do we need to have the same logic for spans?

Currently Elastic APM requirement is for the agent not to send spans to APM Server for non-sampled transactions. It might change in the future (not that there are any plans at the moment but you never know :) but that is the requirement at the moment. BTW code that implements that is at https://github.com/elastic/apm-agent-dotnet/blob/61ee66e1988da987af267005a5b006b80252f6e7/src/Elastic.Apm/Model/Span.cs#L188


> Because span can be a parent for transaction in another service or _dropped will have higher priority on IsSampled?

You are right - thank you for reminding me. I discussed it with other agent teams but forgot to include in my message above. `traceparent` header also has `parent-id` field. Elastic APM requirement for `parent-id` field is described at https://github.com/elastic/apm/issues/16. To summarize -
`parent-id` field should be set to `span-id` only for spans that agent will send to APM Server, otherwise agent should set `traceparent`'s `parent-id` to `transaction-id`. It means that agent should set `traceparent`'s `parent-id` to `transaction-id` for dropped spans of sampled transactions and all spans on non-sampled transaction. So that means that that code that decides what to use `transaction-id` or `span-id`
(which is https://github.com/elastic/apm-agent-dotnet/blob/61ee66e1988da987af267005a5b006b80252f6e7/src/Elastic.Apm/Model/Span.cs#L99) should choose `transaction-id` if `_isDropped || IsSampled` is `true`. I would extract this condition to `Span`'s private property (`ShouldSendToApmServer` or something) and use it in the code locations mentioned above.

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
```diff
@@ -430,6 +430,32 @@ protected double ParseTransactionSampleRate(ConfigurationKeyValue kv)
...
+						ConfigConsts.DefaultValues.TransactionMaxSpans);
+				return ConfigConsts.DefaultValues.TransactionMaxSpans;
+			}
+
+			if (int.TryParse(kv.Value, NumberStyles.Integer, CultureInfo.InvariantCulture, out var result))
+			{
+				Logger?.Debug()
+					?.Log("Using provided transaction max spans `{ProvidedTransactionMaxSpans}' parsed as {ParsedTransactionMaxSpans}",
+						kv.Value, result);
+				return result;
```
vhatsura: @SergeyKleyman, what do you think about validation for negative and zero value? I didn't find any limitations in other agents. And what about disable limitation? Does we need to support such case?

SergeyKleyman: I found that the only agent that specifies the expected behavior if negative value is provided is [Node.js](https://www.elastic.co/guide/en/apm/agent/nodejs/current/configuration.html#transaction-max-spans)
> Setting to -1 means that spans will never be dropped.

I can see why it can useful. I assume setting the option to zero causes agent not to send any spans at all. I would recommend going with the same approach (and documenting it)
- positive number N - agent drops spans above N
- `0` - agent drops all the spans
- `-1` - agent doesn't drops any spans
- any other negative value - invalid and agent falls back on the default value

docs/configuration.asciidoc
```diff
@@ -180,6 +180,24 @@ It also expects that a period (`.`) is used to separate the integer and the frac
 | 1.0     | Double 
 |============
 
+[float]
+[[config-transaction-max-spans]]
+==== `TransactionMaxSpans` (performance)
+
+Limits the amount of spans that are recorded per transaction. This is helpful in cases where a transaction creates a very high amount of spans (e.g. thousands of SQL queries). Setting an upper limit will prevent overloading the agent and the APM server with too much work for such edge cases.
```
SergeyKleyman: As we discussed at (https://github.com/elastic/apm-agent-dotnet/pull/472#discussion_r324145863), I would recommend adding specification for agent's behavior for the special cases.

src/Elastic.Apm/Model/Span.cs
```diff
@@ -47,14 +48,21 @@ internal class Span : ISpan
...
 
 			ParentId = parentId;
 			TraceId = traceId;
 
 			if (IsSampled)
-				// Started spans should be counted only for sampled transactions
-				enclosingTransaction.SpanCount.IncrementStarted();
+			{
+				// Started and dropped spans should be counted only for sampled transactions
+				if (enclosingTransaction.SpanCount.IncrementTotal() >= _configurationReader.TransactionMaxSpans)
```
SergeyKleyman: I think it should be strictly greater i.e.,
`if (enclosingTransaction.SpanCount.IncrementTotal() > _configurationReader.TransactionMaxSpans)`

ElasticApmAgent.sln
```diff
@@ -51,6 +51,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Elastic.Apm.AspNetFullFrame
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Elastic.Apm.Tests.MockApmServer", "test\Elastic.Apm.Tests.MockApmServer\Elastic.Apm.Tests.MockApmServer.csproj", "{AFFB238B-10C0-4229-9527-184EFCF0F5AC}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Elastic.Apm.Tests.Data", "test\Elastic.Apm.Tests.Data\Elastic.Apm.Tests.Data.csproj", "{BE2156EB-6C2C-4D13-925D-7794713C03ED}"
```
SergeyKleyman: You can share `TransactionMaxSpansTestData` between `Elastic.Apm.Tests` and `Elastic.Apm.AspNetCore.Tests` by placing `TransactionMaxSpansTestData` class in `Elastic.Apm.Tests` (for example in new `Data` sub-directory under `Elastic.Apm.Tests` directory and `Elastic.Apm.Tests.Data` namespace). It will work because `Elastic.Apm.AspNetCore.Tests.csproj` references `Elastic.Apm.Tests.csproj`. I don't think it's worth creating a new project just for this. WDYT?

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
vhatsura: Implemented

docs/configuration.asciidoc
vhatsura: Documentation was updated

src/Elastic.Apm/Model/Span.cs
vhatsura: Changed

ElasticApmAgent.sln
vhatsura: Agree with you. I didn't notice reference to Elastic.Apm.Tests.csproj by those projects. Will fix it.

src/Elastic.Apm/Model/Span.cs
```diff
@@ -47,14 +48,22 @@ internal class Span : ISpan
...
 
 			ParentId = parentId;
 			TraceId = traceId;
 
 			if (IsSampled)
-				// Started spans should be counted only for sampled transactions
-				enclosingTransaction.SpanCount.IncrementStarted();
+			{
+				// Started and dropped spans should be counted only for sampled transactions
+				if (enclosingTransaction.SpanCount.IncrementTotal() > _configurationReader.TransactionMaxSpans
```
SergeyKleyman: Since you were concerned about multi-thread safety with this new feature you might consider adding a test with multiple threads but only if you think it's worth your time. In case you decide that it's worth the time there is a test helper utility `MultiThreadsTestUtils.TestOnThreads` you can use to do all the thread starting, stopping and result gathering. Below is one of its usages in already existing tests:
https://github.com/elastic/apm-agent-dotnet/blob/045a07bfc1d77962b98cca78779ea4be063851c7/test/Elastic.Apm.Tests/HelpersTests/AgentSpinLockTests.cs#L113

docs/configuration.asciidoc
```diff
@@ -217,6 +217,26 @@ It also expects that a period (`.`) is used to separate the integer and the frac
...
```
bmorelli25: Can you please update this section to this (including the line breaks)
```
Limits the amount of spans that are recorded per transaction.
This is helpful in cases where a transaction creates a very high amount of spans,
for example, thousands of SQL queries.
Setting an upper limit helps prevent overloading the Agent and APM server in these edge cases.
```

docs/configuration.asciidoc
```diff
@@ -217,6 +217,26 @@ It also expects that a period (`.`) is used to separate the integer and the frac
...
+
+NOTE: Setting it to `-1` means that spans will never be dropped. Setting it to `0` means that spans will never be collected.
```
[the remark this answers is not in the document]
bmorelli25: What about:
```
NOTE: A value of `0` means that spans will never be collected.
Setting `-1` means that spans will never be dropped.
The Agent will revert to the default value if the value is set below `-1`.
```

bmorelli25:
Thanks so much! Just a couple docs comments.

docs/configuration.asciidoc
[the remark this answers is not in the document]
vhatsura: updated

docs/configuration.asciidoc

src/Elastic.Apm/Model/Span.cs
vhatsura: @SergeyKleyman, I'll look into it in the weekend. If you it's block release or smth else, then you can merge it and tests will be in the next PR.

vhatsura: @SergeyKleyman, I added test for concurrent capturing spans. I don't use `MultiThreadsTestUtils.NumberOfThreadsForTest` to be sure concurrent effect can be achieved independently from hardware and test environment.

docs/configuration.asciidoc
```diff
@@ -217,6 +217,31 @@ It also expects that a period (`.`) is used to separate the integer and the frac
 | 1.0     | Double 
 |============
 
+[float]
+[[config-transaction-max-spans]]
+==== `TransactionMaxSpans` (performance)
```
vhatsura: @SergeyKleyman, it what version such option will be available? It would be nice to mention it here.

SergeyKleyman: It's hard to say for sure but most likely the next release will be for [the milestone currently marked as 7.5 (which means .NET agent release aligned with 7.5 Elastic stack release)](https://github.com/elastic/apm-agent-dotnet/milestone/5) and agent's version will be 1.2

src/Elastic.Apm/Model/Span.cs
SergeyKleyman: Thank you - it looks good.

SergeyKleyman:
It's interesting - your `LimitedAmountOfSpansShouldBeSent_WhenSpansAreCapturedConcurrently` test fails because `MockPayloadSender` is not thread safe :)

vhatsura:
@SergeyKleyman, what does you mean under fail? I cannot see in Jenkins any failed test. Also, non-tread safe `MockPayloadSender` affects only order in spans collection

SergeyKleyman:
Can you access this link https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-dotnet%2Fapm-agent-dotnet-mbp/detail/PR-472/42/pipeline/128 (it's Windows/.NET Core pipeline)? If you can please right click `Show complete log`, download the file and search for `LimitedAmountOfSpansShouldBeSent_WhenSpansAreCapturedConcurrently`
```
[2019-10-02T18:58:12.865Z] Failed   Elastic.Apm.Tests.TransactionSamplingTests.LimitedAmountOfSpansShouldBeSent_WhenSpansAreCapturedConcurrently
[2019-10-02T18:58:12.865Z] Error Message:
[2019-10-02T18:58:12.865Z]  Expected mockPayloadSender.Spans.Count to be 2, but found 1.
[2019-10-02T18:58:12.865Z] Stack Trace:
[2019-10-02T18:58:12.865Z]    at FluentAssertions.Execution.XUnit2TestFramework.Throw(String message) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Execution\XUnit2TestFramework.cs:line 32
[2019-10-02T18:58:12.865Z]    at FluentAssertions.Execution.AssertionScope.FailWith(Func`1 failReasonFunc) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Execution\AssertionScope.cs:line 181
[2019-10-02T18:58:12.865Z]    at FluentAssertions.Numeric.NumericAssertions`1.Be(T expected, String because, Object[] becauseArgs) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Numeric\NumericAssertions.cs:line 50
[2019-10-02T18:58:12.865Z]    at Elastic.Apm.Tests.TransactionSamplingTests.LimitedAmountOfSpansShouldBeSent_WhenSpansAreCapturedConcurrently() in C:\Users\jenkins\workspace\tnet_apm-agent-dotnet-mbp_PR-472\apm-agent-dotnet\test\Elastic.Apm.Tests\TransactionSamplingTests.cs:line 115
```
It also failed on [Windows/Full .NET Framework CI pipeline](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-dotnet%2Fapm-agent-dotnet-mbp/detail/PR-472/42/pipeline/115)
```
[2019-10-02T18:57:30.396Z] Failed   Elastic.Apm.Tests.TransactionSamplingTests.LimitedAmountOfSpansShouldBeSent_WhenSpansAreCapturedConcurrently
[2019-10-02T18:57:30.396Z] Error Message:
[2019-10-02T18:57:30.396Z]  Expected mockPayloadSender.Spans.Count to be 2, but found 1.
[2019-10-02T18:57:30.396Z] Stack Trace:
[2019-10-02T18:57:30.396Z]    at FluentAssertions.Execution.XUnit2TestFramework.Throw(String message) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Execution\XUnit2TestFramework.cs:line 32
[2019-10-02T18:57:30.396Z]    at FluentAssertions.Execution.AssertionScope.FailWith(Func`1 failReasonFunc) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Execution\AssertionScope.cs:line 181
[2019-10-02T18:57:30.396Z]    at FluentAssertions.Numeric.NumericAssertions`1.Be(T expected, String because, Object[] becauseArgs) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Numeric\NumericAssertions.cs:line 50
[2019-10-02T18:57:30.396Z]    at Elastic.Apm.Tests.TransactionSamplingTests.LimitedAmountOfSpansShouldBeSent_WhenSpansAreCapturedConcurrently() in C:\Users\jenkins\workspace\tnet_apm-agent-dotnet-mbp_PR-472\apm-agent-dotnet\test\Elastic.Apm.Tests\TransactionSamplingTests.cs:line 115
```

> Also, non-tread safe `MockPayloadSender` affects only order in spans collection
Unfortunately it's not the case as you can see above. In this case because `System.Collections.Generic.List` used by `MockPayloadSender` is not thread safe it lost of the spans added that is why `mockPayloadSender.Spans.Count` is only 1 instead expected 2.

SergeyKleyman:
The test passed for run 44 of this PR - https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-dotnet%2Fapm-agent-dotnet-mbp/detail/PR-472/44/pipeline but it's because list loosing one of the spans is the matter of timing between threads. On my machine it fails all the time.

SergeyKleyman:
I've re-run it again (run 45) - let's see how it goes...

SergeyKleyman:
For run [45 it failed again](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-dotnet%2Fapm-agent-dotnet-mbp/detail/PR-472/45/pipeline/174). [The log](https://apm-ci.elastic.co/blue/rest/organizations/jenkins/pipelines/apm-agent-dotnet/pipelines/apm-agent-dotnet-mbp/branches/PR-472/runs/45/nodes/174/steps/227/log/?start=0) shows:

```
[2019-10-02T19:56:16.488Z] Failed   Elastic.Apm.Tests.TransactionSamplingTests.LimitedAmountOfSpansShouldBeSent_WhenSpansAreCapturedConcurrently
[2019-10-02T19:56:16.488Z] Error Message:
[2019-10-02T19:56:16.488Z]  Expected mockPayloadSender.Spans.Count to be 2, but found 1.
[2019-10-02T19:56:16.488Z] Stack Trace:
[2019-10-02T19:56:16.488Z]    at FluentAssertions.Execution.XUnit2TestFramework.Throw(String message) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Execution\XUnit2TestFramework.cs:line 32
[2019-10-02T19:56:16.488Z]    at FluentAssertions.Execution.AssertionScope.FailWith(Func`1 failReasonFunc) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Execution\AssertionScope.cs:line 181
[2019-10-02T19:56:16.488Z]    at FluentAssertions.Numeric.NumericAssertions`1.Be(T expected, String because, Object[] becauseArgs) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Numeric\NumericAssertions.cs:line 50
[2019-10-02T19:56:16.488Z]    at Elastic.Apm.Tests.TransactionSamplingTests.LimitedAmountOfSpansShouldBeSent_WhenSpansAreCapturedConcurrently() in C:\Users\jenkins\workspace\tnet_apm-agent-dotnet-mbp_PR-472\apm-agent-dotnet\test\Elastic.Apm.Tests\TransactionSamplingTests.cs:line 115
```

vhatsura:
Just want to say that issue can be occurred during adding to list. I didn't think that it can be an issue in this test due to all runs on my machine was successful.
What if we use concurrent bag instead of  list?

SergeyKleyman:
`ConcurrentBag` might not be the right choice because there are tests that depend on the order of events ([docs say](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentbag-1) "unordered collection of objects").
I've just added a lock to serialize access to `MockPayloadSender`'s lists. All the tests pass. I'll push the fix in a few minutes.

vhatsura:
Also, I found out why I didn't see issue in Jenkins. I looked into job result: https://apm-ci.elastic.co/job/apm-agent-dotnet/job/apm-agent-dotnet-mbp/job/PR-472/45/testReport/ instead of blue ocean

vhatsura:
> `ConcurrentBag` might not be the right choice because there are tests that depend on the order of events ([docs say](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentbag-1) "unordered collection of objects").
> I've just added a lock to serialize access to `MockPayloadSender`'s lists. All the tests pass. I'll push the fix in a few minutes.

Didn't know.anout tests with ordering check. So, in this case lock is good solution. Thanks a lot for fixing my stuff.

SergeyKleyman:
NP. Multi-threading is tricky - that is why I always recommend to avoid it and share the least amount of data between threads as possible both for correctness and performance.

gregkalapos:
Hey, due to some unrelated thing I pushed a new release, which is `1.1.1` and that was generated from `master` - since this was already on master, this feature ended up being in that release. Probably it would have been better to generate the release from the `1.x` branch. Anyway, since this is already out I'll open a PR to change the version where this got added in the docs [in this line](https://github.com/elastic/apm-agent-dotnet/pull/472/files#diff-01c1a79a5a1e1b67003a1885365f2cf5R222).

Issue #214, opened 2019-04-26
Tests in Elastic.Apm.Tests should be executed on both .NET Core and full .NET Framework


SergeyKleyman:
Currently unit tests in Elastic.Apm.Tests are executed only on .NET Core even though code under test is used in production on full .NET Framework as well. So we should have unit tests for the code in Elastic.Apm to run on all the environments for which the code is used in production.

This issue is related to #200   but the difference is that this issue is about unit tests while elastic/apm-agent-dotnet#200 is about integration tests.

gregkalapos:
Cleaning up milestone 1.1 - this is not merged yet due to some discussion, but I leave it on 1.1, since most of the work is already done and no prod code is affected.

PR #480, opened 2019-09-09
Tests on full framework

gregkalapos:
Solves #214.

# Summary

We have ~500 tests in `Elastic.Apm.Tests` that test the core features of the agent, but only executed those on .NET Core previously. This PR makes `Elastic.Apm.Tests` into a multi target project, so it tests both on .NET Framework and .NET Core.

I also added `mono` to the known frameworks - mainly because in case you run the Full Framework tests on macOS with Rider, Rider starts up mono and runs all the tests - with this PR that is not an issue anymore.

## HttpClient tests

See https://github.com/elastic/apm-agent-dotnet/issues/516 

## What changed for CI: 

cc @kuisathaverat @v1v  Could one of you please review this part (specifically the `jenkinsfile` and `.ci/windows/testnet461.bat`) and maybe answer my questions below? - Thx!

### Previously: 
In the stage `Windows .NET Framework` under stage `Test` we actually did not ran the `Elastic.Apm.Tests` project on .NET Framework, we ran it on .NET Core - that's because that project only targeted previously .NET Core, so the `dotnet test` command automatically executed the tests on .NET Core. 

### Now:
Now the `Elastic.Apm.Tests` project is a multi target project, it targets both .NET Core and .NET Framework. If you run it with `dotnet run` then it defaults to executing on .NET Core, with `dotnet run -f net461` it runs on .NET Framework. I created a `.ci/windows/testnet461.bat` file, it runs the `Elastic.Apm.Tests` project with `-f net461`. In the `jenkinsfiel` now I use this file on the `Windows .NET Framework` stage. With this in the `Windows .NET Framework` we really run tests only on .NET Framework, and .NET Core tests run on the `Windows .NET Core` stage.

~# Questions:~ Update: all fixed.

~On the jenkins test reporting part I see this:~

~image~

~1. It says "All 39 tests". That can't be right... we have thousands of tests... Is this something we should take a look at? Why is this?~

~2. Code coverage report seems to be broken, but it broke before I started messing with the `jenkinsfile`. I don't know why, could someone please help out with that?~

### DIFF
--- a/test/Elastic.Apm.Tests/HelpersTests/AgentTimerTests.cs
+++ b/test/Elastic.Apm.Tests/HelpersTests/AgentTimerTests.cs
@@ -319,15 +319,13 @@ namespace Elastic.Apm.Tests.HelpersTests
 
 			public void VerifyTryAwaitCompletedSuccessfully(Task tryAwaitOrTimeoutTask, Task delayTask)
 			{
-				tryAwaitOrTimeoutTask.IsCompleted.Should().BeTrue();
-				tryAwaitOrTimeoutTask.IsFaulted.Should().BeFalse();
+				tryAwaitOrTimeoutTask.IsCompletedSuccessfully().Should().BeTrue();
 
 				UnpackTryAwaitOrTimeoutTaskResult(tryAwaitOrTimeoutTask, out var hasTaskToAwaitCompleted, out var taskToAwaitResult);
 				hasTaskToAwaitCompleted.Should().BeTrue();
 				taskToAwaitResult.Should().Be(_resultValue);
 
-				_taskToAwaitTcs.Task.IsCompleted.Should().BeTrue();
-				_taskToAwaitTcs.Task.IsFaulted.Should().BeFalse();
+				_taskToAwaitTcs.Task.IsCompletedSuccessfully().Should().BeTrue();
 				_taskToAwaitTcs.Task.Result.Should().Be(_resultValue);
 
 				VerifyFinalAgentTimerState(delayTask);
@@ -335,13 +333,11 @@ namespace Elastic.Apm.Tests.HelpersTests
 
 			public void VerifyAwaitCompletedSuccessfully(Task awaitOrTimeoutTask, Task delayTask)
 			{
-				awaitOrTimeoutTask.IsCompleted.Should().BeTrue();
-				awaitOrTimeoutTask.IsFaulted.Should().BeFalse();
+				awaitOrTimeoutTask.IsCompletedSuccessfully().Should().BeTrue();
 
 				if (!_isVoid) ((Task<TResult>)awaitOrTimeoutTask).Result.Should().Be(_resultValue);
 
-				_taskToAwaitTcs.Task.IsCompleted.Should().BeTrue();
-				_taskToAwaitTcs.Task.IsFaulted.Should().BeFalse();
+				_taskToAwaitTcs.Task.IsCompletedSuccessfully().Should().BeTrue();
 				_taskToAwaitTcs.Task.Result.Should().Be(_resultValue);
 
 				VerifyFinalAgentTimerState(delayTask);
@@ -349,8 +345,7 @@ namespace Elastic.Apm.Tests.HelpersTests
 
 			public void VerifyTryAwaitTimeout(Task tryAwaitOrTimeoutTask, Task delayTask)
 			{
-				tryAwaitOrTimeoutTask.IsCompleted.Should().BeTrue();
-				tryAwaitOrTimeoutTask.IsFaulted.Should().BeFalse();
+				tryAwaitOrTimeoutTask.IsCompletedSuccessfully().Should().BeTrue();
 
 				UnpackTryAwaitOrTimeoutTaskResult(tryAwaitOrTimeoutTask, out var hasTaskToAwaitCompleted, out var taskToAwaitResult);
 				hasTaskToAwaitCompleted.Should().BeFalse();
@@ -434,10 +429,7 @@ namespace Elastic.Apm.Tests.HelpersTests
 				if (wasDelayCancelled)
 					delayTask.IsCanceled.Should().BeTrue();
 				else
-				{
-					delayTask.IsCompleted.Should().BeTrue();
-					delayTask.IsFaulted.Should().BeFalse();
-				}
+					delayTask.IsCompletedSuccessfully().Should().BeTrue();
 			}
 		}
 	}
--- a/test/Elastic.Apm.Tests/HelpersTests/AgentTimerTests.cs
+++ b/test/Elastic.Apm.Tests/HelpersTests/AgentTimerTests.cs
@@ -319,13 +319,15 @@ namespace Elastic.Apm.Tests.HelpersTests
 
 			public void VerifyTryAwaitCompletedSuccessfully(Task tryAwaitOrTimeoutTask, Task delayTask)
 			{
-				tryAwaitOrTimeoutTask.IsCompletedSuccessfully.Should().BeTrue();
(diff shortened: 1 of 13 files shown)


test/Elastic.Apm.Tests/Mocks/TestLogger.cs
```diff
@@ -16,6 +16,9 @@ internal class TestLogger : ConsoleLogger
 
 		public TestLogger(LogLevel level, StringWriter writer) : base(level, writer, writer) => _writer = writer;
 
-		public List<string> Lines => _writer.GetStringBuilder().ToString().Split(Environment.NewLine, StringSplitOptions.RemoveEmptyEntries).ToList();
+		public List<string> Lines => _writer.GetStringBuilder()
+			.ToString()
+			.Split(Environment.NewLine.ToCharArray(), StringSplitOptions.RemoveEmptyEntries)
```
gregkalapos: `Environment.NewLine.ToCharArray()` was needed to build on `net461`, the parameter list used originally does not exist on `net461`.

test/Elastic.Apm.Tests/HelpersTests/AgentTimerTests.cs
```diff
@@ -319,33 +319,38 @@ public void CompleteTaskSuccessfully()
 
 			public void VerifyTryAwaitCompletedSuccessfully(Task tryAwaitOrTimeoutTask, Task delayTask)
 			{
-				tryAwaitOrTimeoutTask.IsCompletedSuccessfully.Should().BeTrue();
```
gregkalapos: Unfortunately there is no `Task.IsCompletedSuccessfully` on `net461`.

test/Elastic.Apm.Tests/HelpersTests/TimeUtilsTests.cs
```diff
@@ -23,8 +23,10 @@ public class TimeUtilsTests
 			{ 5 * 24 * 60 * 60 * 1_000_000L, new DateTime(1970, 1, 6, 0, 0, 0, DateTimeKind.Utc) },
 		};
 
+#if NETCOREAPP2_1
```
gregkalapos: There is no `DateTime.UnixEpoch` on `net461`, I think there is no alternative for that - I'd say the easiest is to just not compile this part to `net461`, so I solved this one with just `#if NETCOREAPP2_1`.

test/Elastic.Apm.Tests/Mocks/TestLogger.cs
```diff
@@ -17,6 +17,6 @@ internal class TestLogger : ConsoleLogger
 		public TestLogger(LogLevel level, StringWriter writer) : base(level, writer, writer) => _writer = writer;
 
 		public IReadOnlyList<string> Lines =>
-			_writer.GetStringBuilder().ToString().Split(Environment.NewLine, StringSplitOptions.RemoveEmptyEntries).ToList();
+			_writer.GetStringBuilder().ToString().Split(Environment.NewLine.ToCharArray(), StringSplitOptions.RemoveEmptyEntries).ToList();
```
gregkalapos: The `Split` method with the previous parameter list does not exist on `net461`, that's why I added the `.ToCharArray()` here.

kuisathaverat:
>It says "All 39 tests". That can't be right... we have thousands of tests... Is this something we should take a look at? Why is this?

Jenkins it taking any file matches the following pattern `"${BASE_DIR}/**/junit-*.xml` or `${BASE_DIR}/target/**/TEST-*.xml"` , any file named `TEST-*.xml` or `junit-*.xml` in any folder, Do this pattern match to the generated files?

SergeyKleyman:
As described at https://xunit.net/docs/why-no-netstandard the root cause for `"All 39 tests"`  might be target platform for not Windows OS. [I've changed it from `netstandard2.0` to `netcoreapp2.1`](https://github.com/elastic/apm-agent-dotnet/pull/517) and [it seems to be working again](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-dotnet%2Fapm-agent-dotnet-mbp/detail/TEMP_dual_target_Tests/3/tests).

test/Elastic.Apm.Tests/HelpersTests/AgentTimerTests.cs
SergeyKleyman: I would add `IsCompletedSuccessfully()` extension method - https://github.com/elastic/apm-agent-dotnet/pull/517/commits/8e3f8afc2171bffc97b4b583736b4459d3d998f7 and https://github.com/elastic/apm-agent-dotnet/pull/517/commits/0826ed36910b94d3fb2349717142812068eb49c9

test/Elastic.Apm.Tests/HelpersTests/TimeUtilsTests.cs
SergeyKleyman: I think `#if NETCOREAPP` might be better in case we change the target to netcoreapp2.2 or later.

test/Elastic.Apm.Tests/HttpDiagnosticListenerTests.cs
```diff
@@ -84,7 +84,7 @@ public void OnErrorLog()
 		/// Builds an HttpRequestMessage and calls HttpDiagnosticListener.OnNext directly with it.
 		/// Makes sure that the processingRequests dictionary captures the ongoing transaction.
 		/// </summary>
-		[Fact]
+		[NetCoreFact] //see: https://github.com/elastic/apm-agent-dotnet/issues/516
```
SergeyKleyman: Could you please clarify what are the advantages of using runtime condition (via `[NetCoreFact]`) to skip tests vs. using `#if NETCOREAPP` (https://github.com/elastic/apm-agent-dotnet/pull/517/commits/d2813dac06bb808479aefe02572c02a0c020039d)? The advantage of using `#if NETCOREAPP` is that .NET Core only tests won't show as skipped for .NET Framework - the tests won't show at all.

It's worth discussing when tests should be marked as skipped. I think we should use skipped to mean that some preconditions **can be** but are not satisfied. If the preconditions cannot be satisfied then tests should not be included at all. For example running Full Framework tests on Linux should not include them at all while running Full Framework tests on Windows without environment variable `ELASTIC_APM_TESTS_FULL_FRAMEWORK_ENABLED` set should include the tests and mark them as skipped. This way a developer knows that if there are skipped tests that means they don't have some components in their environment to support all the available tests.

gregkalapos:
Thanks for the responses. 

I think @SergeyKleyman's comment solves the CI issues - we should target `netcoreapp2.1`, that seems to fix it - thx! 🎉 

To other comments - those seem to be single comments and I can't answer them in-place. 

> I would add IsCompletedSuccessfully() extension 

Ok, fine by me - I'm usually not a big fan of having so many helper classes, but yeah, this is fine by me. 

> I think #if NETCOREAPP might be better  

Agreed 

To the `ifdef`s I respond directly.

I cherry-picked 3 commit from the other branch - those are already included in this branch. The `ifdef`s are TBD.

test/Elastic.Apm.Tests/HttpDiagnosticListenerTests.cs
gregkalapos: I usually try to avoid `#if`s when possible. One benefit is having the same binary both for .NET Core and Full Framework - the more `#if`s we have the bigger the difference is between those two binaries. Also I think it’s harder to read that code, since you have to keep in mind on which .NET Flavor you currently are. 

I’d prefer having runtime checks - I think it’s better to have skipped tests with a proper explanation (in this case the `NetCoreFact` has it) than having tests that disappear. Also to me this seems to be a simpler option.

So I don’t see having skipped tests as a big problem - I’d say if from some reason a test cannot be executed it should be skipped but not hidden.

Issue #209, opened 2019-04-25
Logger should never throw


SergeyKleyman:
Currently logger throws when number of placeholders in format string doesn't match number of arguments. In addition to fixing that particular issue we need to make sure that issues in logging related code doesn't crash production code.

One way to approach the implementation would be to wrap logger implementation, capture escaping exceptions and log them under a special category (something like `Logging subsystem`).

This issue is somewhat related to #208 although I think mechanisms in this issue and the one in #208 should be independent. That is mechanism in this issue should prevent exceptions propagating out of logging code to the rest of the agent's code.

SergeyKleyman:
We agreed for short term to leave the logging interface/implementation as it is and just implement minimal change to fix "logger throwing exceptions" issue. #337 is the follow up to implement a better solution.

gregkalapos:
Decision: let's do it with a quick fix.

PR #485, opened 2019-09-10
Logger exceptions

gregkalapos:
Solves #209 

In case a structured log does not contain the expected number of parameters we make sure the logger does not throw exceptions.

Additionally this PR also adds some additional info to the log, so this won't get unnoticed.

So e.g.:
`Log("This is a log with params: {param1} {param2}", param1Value)` 

Here is how we'd log with this (the test itself is just an example)

## Sample1

Code:
```
var i = 42;
Logger?.Info()?.Log("The agent was started without a service name. The service name will be automatically discovered - args: {myArg1} {myArg2}", i);
```

Corresponding log: 
```
info: Elastic.Apm[0]
      {MicrosoftExtensionsConfig} The agent was started without a service name. The service name will be automatically discovered - args: 42
      Above line is from an invalid structured log and may not be complete: number of arguments is not matching the number of placeholders, placeholders with missing values: myArg2
```

## Sample2

Code:

```
var argVal1 = 42;
var argVal2 = 42;
Logger?.Info()?.Log("The agent was started without a service name. The service name will be automatically discovered - args: {myArg1}", argVal1, argVal2);
```

Corresponding log:

```
info: Elastic.Apm[0]
      {MicrosoftExtensionsConfig} The agent was started without a service name. The service name will be automatically discovered - args: 42
      Above line is from an invalid structured log and may not be complete: number of placeholders in the log message does not match the number of parameters.
```

### DIFF
--- a/src/Elastic.Apm/Logging/LogValuesFormatter.cs
+++ b/src/Elastic.Apm/Logging/LogValuesFormatter.cs
@@ -1,7 +1,3 @@
-// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0.
-
-using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
@@ -16,13 +12,20 @@ namespace Elastic.Apm.Logging
 	/// </summary>
 	internal class LogValuesFormatter
 	{
-		private readonly string _scope;
 		private const string NullValue = "(null)";
 		private static readonly object[] EmptyArray = new object[0];
 		private static readonly char[] FormatDelimiters = { ',', ':' };
 		private readonly string _format;
+		private readonly bool _isArgumentNumberMismatch;
 
-		public LogValuesFormatter(string format, string scope = null)
+		/// <summary>
+		/// Holds the list of placeholders that do not have corresponding values in the structured log.
+		/// </summary>
+		private readonly List<string> _mismatchedValueNames = new List<string>();
+
+		private readonly string _scope;
+
+		public LogValuesFormatter(string format, int expectedArgs, string scope = null)
 		{
 			_scope = scope;
 			OriginalFormat = format;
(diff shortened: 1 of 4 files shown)


src/Elastic.Apm/Logging/LogValuesFormatter.cs
```diff
@@ -1,7 +1,3 @@
-// Copyright (c) .NET Foundation. All rights reserved.
```
gregkalapos: Not sure why this got removed... maybe auto format.

Do we need this Copyright header? I don't have strong opinion on this, I'm ok with adding it back.

src/Elastic.Apm/Logging/LogValuesFormatter.cs
```diff
@@ -16,13 +12,20 @@ namespace Elastic.Apm.Logging
...
 		private const string NullValue = "(null)";
 		private static readonly object[] EmptyArray = new object[0];
 		private static readonly char[] FormatDelimiters = { ',', ':' };
 		private readonly string _format;
+		private readonly bool _isArgumentNumberMismatch;
+
+		/// <summary>
+		/// Holds the list of placeholders that do not have corresponding values in the structured log.
+		/// </summary>
+		private readonly List<string> _mismatchedValueNames = new List<string>();
```
SergeyKleyman: IIUC `_isArgumentNumberMismatch` and `_mismatchedValueNames` are used only in the constructor so you can make them local variables. Also `_isArgumentNumberMismatch == true` == `mismatchedValueNames.Count > 0` so you can keep only `mismatchedValueNames`.

SergeyKleyman: A nit: `mismatchedValueNames` is somewhat confusing name because you refer to `{xyz}` as placeholders which I agree - those are placeholders. Maybe `placeholdersMissingArgs`?

src/Elastic.Apm/Logging/LogValuesFormatter.cs
```diff
@@ -46,19 +49,41 @@ public LogValuesFormatter(string format, string scope = null)
...
+				sb.AppendLine();
+				sb.AppendLine(
+					$"Above line is from an invalid structured log and may not be complete: number of arguments is not matching the number of placeholders, placeholders with missing values: {string.Join(", ", _mismatchedValueNames.ToArray())}");
+			}
+
+			if (ValueNames.Count != expectedArgs)
+			{
+				sb.AppendLine();
+				sb.AppendLine(
+					$"Above line is from an invalid structured log and may not be complete: number of placeholders in the log message does not match the number of parameters.");
```
SergeyKleyman: I would also appending 
1. How many arguments miss placeholders 
2. Arguments that miss placeholders. It can be very useful if the invalid structured log line came up for the first time while handling a customer support escalation and those arguments are very important for investigation. Otherwise we will have to release a build with the invalid structured log statement fixed.

src/Elastic.Apm/Logging/LogValuesFormatter.cs
```diff
@@ -46,19 +49,41 @@ public LogValuesFormatter(string format, string scope = null)
...
 
 					scanIndex = closeBraceIndex + 1;
 				}
 			}
 
+			if (_isArgumentNumberMismatch)
+			{
+				sb.AppendLine();
+				sb.AppendLine(
+					$"Above line is from an invalid structured log and may not be complete: number of arguments is not matching the number of placeholders, placeholders with missing values: {string.Join(", ", _mismatchedValueNames.ToArray())}");
```
SergeyKleyman: As I mentioned before, having it on the same line as the message itself will make life easier for those of us that use grep and such.

src/Elastic.Apm/Logging/LogValuesFormatter.cs
```diff
@@ -46,19 +49,41 @@ public LogValuesFormatter(string format, string scope = null)
...
+			}
+
+			if (ValueNames.Count != expectedArgs)
+			{
+				sb.AppendLine();
```
SergeyKleyman: The same comment regarding keeping everything on the same line.

src/Elastic.Apm/Logging/LogValuesFormatter.cs
Mpdreamz: This did originate from: https://github.com/aspnet/Extensions/blob/master/src/Logging/Logging.Abstractions/src/LogValuesFormatter.cs

Under Apache2 we are free to create derivatives but we have to retain this attribution. 

See also e.g: https://softwareengineering.stackexchange.com/questions/220068/file-with-apache-2-0-and-my-modifications

gregkalapos: Oh, I see, thank you, I'll add it back.

src/Elastic.Apm/Logging/LogValuesFormatter.cs
gregkalapos: Oh, right. In a previous version they were used outside, but not anymore - nice catch, changed it accordingly.

gregkalapos: Yes, done.

src/Elastic.Apm/Logging/LogValuesFormatter.cs
gregkalapos: Yeah, that makes sense. I implemented this.

src/Elastic.Apm/Logging/LogValuesFormatter.cs
gregkalapos: Fine by me - now it's in the same line.

src/Elastic.Apm/Logging/LogValuesFormatter.cs
gregkalapos: This is also done.

Issue #496, opened 2019-09-18
Automated tests for Central Configuration feature


SergeyKleyman:
PR (#497) for central configuration implementation doesn't include automated tests - this issue is to add automated tests for central configuration feature.

- [ ] Unit tests using `MockCentralConfigFetcher` to simulate dynamic changes in configuration and verify that relevant features of the Agent (at the moment only transaction sampling) reacts to the configuration changes as expected
- [ ] Tests that run on both Full .NET Framework and .NET Core (and use `MockApmServer`?)
   - **Note:** If we want to use `MockApmServer` we won't be able to place the tests into `Elastic.Apm.Tests` because dependecies go into the opposite direction -  `Elastic.Apm.Tests.MockApmServer` depends on `Elastic.Apm.Tests`.
      - One possible approach is to create a new test project (for example `Elastic.Apm.ComponentTests`) that will depend on `Elastic.Apm.Tests.MockApmServer` and include tests that use only via Agent public API and data Agent sends to APM Server.
An advantage of this approach is that we don't need to allow `Elastic.Apm.ComponentTests` access to implementation `internal`-s thus reducing the chance of deviating from real world use cases.
- [ ] Tests for `Elastic.Apm.AspNetFullFramework.Tests` to verify that implementation works correctly when running in context of classic ASP.NET application on top of IIS.
   - It would be nice to re-use the same logic of (scenario, expected outcome) used in the tests from the previous item.

stevejgordon:
Closing as this an old issue and we have no plans to implement specific further testing

PR #497, opened 2019-09-18
Central configuration implementation

### LINKED ISSUE #352: Implement remote (AKA central) configuration

SergeyKleyman:
**Note:** This PR is temporarily in draft mode - see https://github.com/elastic/apm-agent-dotnet/pull/497#issuecomment-533244308 for details.

Closes #352

It doesn't include automated tests - I've opened #496 to follow up.


docs/configuration.asciidoc
```diff
@@ -180,6 +180,24 @@ It also expects that a period (`.`) is used to separate the integer and the frac
 | 1.0     | Double 
 |============
 
+[float]
+[[config-central-config]]
+==== `CentralConfig` (added[1.1])
+
+If set to `true`, the agent makes periodic requests to the APM Server to fetch the latest configuration.
```
bmorelli25: We should link to the central docs as well for this feature.
https://www.elastic.co/guide/en/kibana/7.3/agent-configuration.html

```suggestion
If set to `true`, the agent makes periodic requests to the APM Server to fetch the latest {kibana-ref}/agent-configuration.html[APM Agent configuration].
```

SergeyKleyman: @bmorelli25 Thank you and done. I actually copied that line [from Java Agent docs](https://www.elastic.co/guide/en/apm/agent/java/current/config-core.html#config-central-config) :) and it doesn't have that link. Also neither do [Go](https://www.elastic.co/guide/en/apm/agent/go/current/configuration.html#config-central-config) and  [Node.js](https://www.elastic.co/guide/en/apm/agent/nodejs/current/configuration.html) (or is it *neither ... **or** ...*? ;)) Only [Python](https://www.elastic.co/guide/en/apm/agent/python/current/configuration.html) and [Ruby](https://www.elastic.co/guide/en/apm/agent/ruby/current/configuration.html) Agents docs have the link although in Ruby's case it's named as `APM Agent Configuration via Kibana`. So you might want to do something it... ;)

SergeyKleyman:
I'm still investigating why tests were hanging on CI's Linux environment. It doesn't happen in this PR after I introduced `NoopCentralConfigFetcher` to be used in the tests that don't test Central Configuration feature (which at the moment are all the tests since tests for Central Configuration feature were deferred to a future PR, see #496).
I'm going to use a temporary PR (#499) for experiments since the issue of hanging tests doesn't reproduce in my environment (which is Windows - it's not clear at the moment if it's a Linux vs Windows issue).
Let's consider this PR as a draft. I will remove draft status after I find the root cause. If the root cause is related to Central Configuration feature I'll add a fix.

docs/configuration.asciidoc
bmorelli25: Thank you! Yes, some of the Agents got their docs done before the central documentation was ready. I appreciate the research, I'll create an issue to update them.

And yes, _neither_ is correct 😄

SergeyKleyman:
After quite a few experiments I found quite a few issues (most are only relevant for test use cases so I'll submit fixes for them later as separate PRs) but the main root causes for tests hanging on Linux were:
1. `HttpClient`'s `SendAsync` under certain circumstances not respecting cancellation (even though it accepts `CancellationToken`). It seems that those problematic circumstances occur when there are multiple simultaneously running tests instantiate `CentralConfigFetcher` which tries to connect to APM Server using `HttpClient.SendAsync(... GET request ...)` although I didn't succeed in creating a minimal reproducing example (i.e., the using only `HttpClient` without any of Agent's code). I solved the issue by wrapping `HttpClient.SendAsync` into `AwaitOrTimeout` which is cancellable,
2. `CentralConfigFetcher`'s `Dispose` signalling cancellation to `HttpClient`'s `SendAsync`. [It turns out that `CancellationTokenSource.Cancel` invokes callbacks synchronously](https://stackoverflow.com/questions/31495411/a-call-to-cancellationtokensource-cancel-never-returns) which in our cases caused deadlock because the thread calling `CancellationTokenSource.Cancel` after that tries to `Join` with the thread on which `HttpClient.SendAsync` is running that the first thread tries to cancel. Following the advice at the abov  e link and wrapping `CancellationTokenSource.Cancel` into `Task.Run` solved the deadlock.

gregkalapos:
Are we sure we merge all this 5000+ lines without review?

PR #499, opened 2019-09-19
[TEMP PR] Investigate why Central configuration implementation #497 has tests hanging on Linux

SergeyKleyman:


SergeyKleyman:
Investigation results are summarized at https://github.com/elastic/apm-agent-dotnet/pull/497#issuecomment-534656593

PR #364, opened 2019-07-06
Add Full Framework Web.config support to configure Elastic variables

NeoXtreem:
Addresses issue #275.

PR #503, opened 2019-09-23
Add Full Framework Web.config support to configure Elastic variables

### LINKED ISSUE #275: Make server URL configurable in Web.config for ASP.NET Full Framework
As discussed [here](https://discuss.elastic.co/t/apm-net-agent-public-api-configuration/181649/5?u=neox), just as the server URL can be configured for ASP.NET Core using `IConfiguration` (by setting `ElasticApm.ServerUrls` in a settings JSON file), support to configure this in a similar manner (namely **Web.config**) should be implemented for ASP.NET Full Framework so that `ELASTIC_APM_SERVER_URLS` doesn't need to be set.

Without this, it is difficult to use Elastic APM agent on servers where multiple websites are being hosted besides our own as it requires us to ask the host to add the environment variable.

lafriks:
Fixes #275 

Replaces #364

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -40,7 +40,7 @@ in the `IConfiguration or Web.config key` column of the corresponding option's d
 NOTE: By simply calling `app.UseElasticApm()` without the overload, the agent will read configurations only from environment variables.
 
 [float]
-[[sample-config]]
+[[configuration-on-asp-net-core-sample-config]]
 ==== Sample configuration file
 
 Below is a sample `appsettings.json` configuration file for a typical ASP.NET Core application. There are two important takeaways:
@@ -101,7 +101,7 @@ You can find the key of each configuration option
 in the `IConfiguration or Web.config key` column of the corresponding option's description.
 
 [float]
-[[sample-config]]
+[[configuration-on-asp-net-sample-config]]
 ==== Sample configuration file
 
 Below is a sample `Web.config` configuration file for a ASP.NET application.
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -33,7 +33,9 @@ public class Startup
 }
 ----
 
-With this you can use any configuration source that you configured on the `IConfiguration` instance that you passed to the APM Agent. You can find the key of each configuration below in the IConfiguration key column.
+With this you can use any configuration source that you configured on the `IConfiguration` instance that you passed to the APM Agent.
+You can find the key of each configuration option
+in the `IConfiguration or Web.config key` column of the corresponding option's description.
 
 NOTE: By simply calling `app.UseElasticApm()` without the overload, the agent will read configurations only from environment variables.
 
@@ -44,7 +46,7 @@ NOTE: By simply calling `app.UseElasticApm()` without the overload, the agent wi
 Below is a sample `appsettings.json` configuration file for a typical ASP.NET Core application. There are two important takeaways:
 
 1. The part below `ElasticApm` is fetched by the agent if the corresponding `IConfiguration` is passed to the agent.
-2. With ASP.NET Core, you must set `LogLevel for the internal APM logger in the standard `Logging` section with the `ElasticApm` category name.
+2. With ASP.NET Core, you must set `LogLevel` for the internal APM logger in the standard `Logging` section with the `ElasticApm` category name.
 
 [source,js]
 ----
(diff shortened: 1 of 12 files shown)


lafriks:
I created new class `AbstractConfigurationWithEnvFallbackReader` and moved most of properties from `MicrosoftExtensionsConfig` to not to have duplicate same code twice

SergeyKleyman:
@bmorelli25 Please take a look at the documentation changes. 10x.

src/Elastic.Apm/Config/ConfigConsts.cs
```diff
@@ -50,6 +50,27 @@ public static class EnvVarNames
...
 
+		public static class KeyNames
+		{
+			public const string CaptureBody = "ElasticApm:CaptureBody";
+			public const string CaptureBodyContentTypes = "ElasticApm:CaptureBodyContentTypes";
+			public const string CaptureHeaders = "ElasticApm:CaptureHeaders";
+			public const string Environment = "ElasticApm:Environment";
+			public const string FlushInterval = "ElasticApm:FlushInterval";
+			public const string LogLevel = "ElasticApm:LogLevel";
+			public const string LogLevelSubKey = "LogLevel";
```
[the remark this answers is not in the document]
lafriks: done

lafriks:
@SergeyKleyman I moved `ReadEnvVarValue` to `AbstractConfigurationReader` to reuse it also in `EnvironmentConfigurationReader` class

lafriks:
Fixed test that now was conflicting with previous test and was correctly failing because of fixed bug

lafriks:
`IIS Tests` fail with `Could not load file or assembly 'System.Threading.Tasks.Dataflow, Version=4.6.3.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)` but that should not be related to this PR

SergeyKleyman:
> `IIS Tests` fail with `Could not load file or assembly 'System.Threading.Tasks.Dataflow, Version=4.6.3.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)` but that should not be related to this PR

There seems to be a problem with CI environment. Let me see what I can do to fix it.

lafriks:
Yes, they work for me in my environment :)

docs/configuration.asciidoc
```diff
@@ -33,18 +33,20 @@ public class Startup
...
-With this you can use any configuration source that you configured on the `IConfiguration` instance that you passed to the APM Agent. You can find the key of each configuration below in the IConfiguration key column.
+With this you can use any configuration source that you configured on the `IConfiguration` instance that you passed to the APM Agent.
+You can find the key of each configuration option
+in the `IConfiguration or Web.config key` column of the corresponding option's description.
 
 NOTE: By simply calling `app.UseElasticApm()` without the overload, the agent will read configurations only from environment variables.
 
 [float]
-[[sample-config]]
+[[configuration-on-asp-net-core-sample-config]]
```
bmorelli25: In the future, we want to avoid changing topic IDs. This breaks external links to the documentation. If a link _has_ to be changed, we want to add the old slug to a redirects file to catch any old links.

We should also take care to not have super long ids like this one. It makes it difficult to reference them.

SergeyKleyman: @bmorelli25 Do you think it would preferable to change the old link (this one) back to `sample-config` and [for the new one](https://github.com/elastic/apm-agent-dotnet/pull/503/files/5fadd946daff7f09456068a766103f0d026c4142#diff-01c1a79a5a1e1b67003a1885365f2cf5R104) use something like `asp-net-sample-config`?

docs/configuration.asciidoc
```diff
@@ -85,6 +87,41 @@ In this case, you can set the log level of the agent with `ElasticApm:LogLevel`,
...
+Web.config has precedence over environment variables which means that
+the agent first tries to find a configuration option value by its key in Web.config.
+Only if it's not present then the agent tries to look for it among environment variables
+and if it's not present there as well the agent falls back on the option's default value.
+
+You can find the key of each configuration option
+in the `IConfiguration or Web.config key` column of the corresponding option's description.
+
+[float]
+[[configuration-on-asp-net-sample-config]]
```
bmorelli25: Same story with this ID. There's no need to mention `configuration` and `config` in the ID. This really should be shortened.

docs/configuration.asciidoc
bmorelli25: Yes, I'm working on a PR right now

docs/configuration.asciidoc
SergeyKleyman: I see - please see my comment above.

docs/configuration.asciidoc
bmorelli25: https://github.com/elastic/apm-agent-dotnet/pull/511

PR #504, opened 2019-09-24
[TEMP PR] Investigate System.Threading.Tasks.Dataflow build errors

SergeyKleyman:
Investigate build errors related to System.Threading.Tasks.Dataflow occurring during build of "Add Full Framework Web.config support to configure Elastic variables" #503


SergeyKleyman:
See #506 for root cause explanation and the fix.

Issue #346, opened 2019-07-03
Document agent's support of ASP.NET Full Framework


SergeyKleyman:
1. Document how to instrument ASP.NET Full Framework application using ElasticApmModule
    1.1 Mention that by adding [`managedHandler` to modules `precondition`](https://docs.microsoft.com/en-us/iis/configuration/system.webserver/modules/add)  monitoring can be limited to dynamic pages (URLs handled by ASP.NET) excluding static content (images, etc.). By default `ElasticApmModule` creates transactions for all the requests passed to it and since the way `ElasticApmModule` is currently configured in `sample\AspNetFullFrameworkSampleApp\Web.config` (i.e., without `managedHandler` in `precondition`) transaction will be created for static content as well.
2. Update "Supported technologies" section

PR #507, opened 2019-09-24
PR for Document agent's support of ASP.NET Full Framework #346

SergeyKleyman:
Closes elastic/apm-agent-dotnet#346

1. Document how to instrument ASP.NET Full Framework application using ElasticApmModule
    1.1 Mention that by adding [`managedHandler` to modules `precondition`](https://docs.microsoft.com/en-us/iis/configuration/system.webserver/modules/add)  monitoring can be limited to dynamic pages (URLs handled by ASP.NET) excluding static content (images, etc.). By default `ElasticApmModule` creates transactions for all the requests passed to it and since the way `ElasticApmModule` is currently configured in `sample\AspNetFullFrameworkSampleApp\Web.config` (i.e., without `managedHandler` in `precondition`) transaction will be created for static content as well.
2. Update "Supported technologies" section.


docs/setup.asciidoc
```diff
@@ -31,4 +36,42 @@ With this you enable every agent component including ASP.NET Core tracing, monit
...
-In case you only want to use the <<public-api>>, you don't need to do any initialization, you can simply start using the API and the agent will send the data to the APM Server.
\ No newline at end of file
+In case you only want to use the <<public-api>>, you don't need to do any initialization, you can simply start using the API and the agent will send the data to the APM Server.
+
+
+[float]
+[[setup-asp-net]]
+=== ASP.NET
+
+For ASP.NET (Full .NET Framework), once you referenced the `Elastic.Apm.AspNetFullFramework` package,
```
bmorelli25:
```suggestion
For ASP.NET (Full .NET Framework), once you've referenced the `Elastic.Apm.AspNetFullFramework` package,
```

docs/setup.asciidoc
```diff
@@ -31,4 +36,42 @@ With this you enable every agent component including ASP.NET Core tracing, monit
...
+you can enable auto instrumentation by including `ElasticApmModule` IIS Module in application's `web.config`:
```
bmorelli25:
```suggestion
you can enable auto instrumentation by including the `ElasticApmModule` IIS Module in your application's `web.config`: 
```

docs/setup.asciidoc
```diff
@@ -31,4 +36,42 @@ With this you enable every agent component including ASP.NET Core tracing, monit
...
+<configuration>
+    <system.webServer>
+        <modules>
+            <add name="ElasticApmModule" type="Elastic.Apm.AspNetFullFramework.ElasticApmModule" />
+        </modules>
+    </system.webServer>
+</configuration>
+----
+
+By default the agent creates transactions for all HTTP requests including for the ones for static content
```
bmorelli25:
```suggestion
By default the agent creates transactions for all HTTP requests, including the ones for static content:
```

docs/setup.asciidoc
```diff
@@ -31,4 +36,42 @@ With this you enable every agent component including ASP.NET Core tracing, monit
...
+(.html pages, images, etc.). If you would like to create transactions only for HTTP requests only for dynamic content
```
bmorelli25: Only for HTTP requests with dynamic content?

```suggestion
.html pages, images, etc. If you would like to create transactions only for HTTP requests with dynamic content,
```

docs/setup.asciidoc
```diff
@@ -31,4 +36,42 @@ With this you enable every agent component including ASP.NET Core tracing, monit
...
+such as `.aspx` pages you can add `managedHandler` `preCondition`
```
bmorelli25:
```suggestion
such as `.aspx` pages, you can add `managedHandler` `preCondition`
```

docs/setup.asciidoc
```diff
@@ -31,4 +36,42 @@ With this you enable every agent component including ASP.NET Core tracing, monit
...
+(https://docs.microsoft.com/en-us/iis/configuration/system.webserver/modules/add[official documentation])
+as in the following example:
```
bmorelli25:
```suggestion
as shown in the following example:
```

SergeyKleyman: Thank you for all the corrections :)

docs/setup.asciidoc
```diff
@@ -31,4 +36,42 @@ With this you enable every agent component including ASP.NET Core tracing, monit
...
+
+For ASP.NET (Full .NET Framework), once you've referenced the `Elastic.Apm.AspNetFullFramework` package,
+you can enable auto instrumentation by including the `ElasticApmModule` IIS Module in your application's `web.config`: 
+[source,xml]
+----
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+    <system.webServer>
+        <modules>
+            <add name="ElasticApmModule" type="Elastic.Apm.AspNetFullFramework.ElasticApmModule" />
```
gregkalapos: I think the order of the modules matter in case there are more, right? Maybe it'd be worth pointing out in the text below that the `ElasticAPMModule` should be the 1. one, otherwise everything that is before the `ElasticAPMModule` won't be measured as part of the transaction.

docs/supported-technologies.asciidoc
```diff
@@ -47,6 +47,10 @@ We support automatic instrumentation for the following web frameworks.
 |2.x
 |1.0
 
+|ASP.NET (.NET Framework) on IIS
+|3.5 and higher (IIS 7.0 and higher)
```
gregkalapos: Is `3.5` here the .NET Framework version? If so, was that tested? 🤔  I think the new package would only work from `4.6.1` and higher, this is what we define in the `Elastic.Apm.AspNetFullFramework.csproj`: 
`<TargetFrameworkVersion>v4.6.1</TargetFrameworkVersion>`. I guess you can't add that to a `3.5` project. But maybe I'm wrong... just making sure `3.5` is ok here.

SergeyKleyman: 10x - Good catch. I've changed it to `4.6.1`.

docs/setup.asciidoc
SergeyKleyman: I agree that if placing module first in `web.config` would have had the same effect as placing ASP.NET Core middleware first then documenting it would be important. But unfortunately in this case in more complicated than that, What we are actually interested in is to be **first** to be called on `BeginRequest` event but to be **last** to be called on `EndRequest` event and as described [here for example](https://stackoverflow.com/questions/14973667/can-i-be-sure-that-httpmodules-are-executed-in-the-order-in-which-they-are-liste/14974151#14974151
) it seems to be impossible with just one module.

We can have two modules: `ElasticApmModuleBeginRequest` (goes the first in web.config) and `ElasticApmModuleEndRequest` (goes the last in web.config) but I think it's a complication that is not worth the benefit.

Even though it sounds counter-intuitive it might actually be better to place our module last in `web.config` so it's called last on `EndRequest` event. The reasoning is that if there's any heavy work done by modules when request comes in this work will be around authentication and authorization and those have their own evens that are handled after `BeginRequest` so our module will account for them even if it's called last on `BeginRequest` event. On the other hand heavy work that modules might do when response is sent out (HTML links rewriting, JavaScript injection) is done on `EndRequest` event so if we want to account for that work then our module should be the last to be called on `EndRequest` event which is possible only if it's the last in web.config.

But it gets even more complicated than that... Regarding the statement above:
> (HTML links rewriting, JavaScript injection is done on `EndRequest`

Unfortunately even this is not exactly how modules might do things - for example here are two completely different ways to implement a module that adds content to HTTP response body:
1. [Using `OnEndRequest` event](https://stackoverflow.com/questions/8678424/how-to-inject-html-before-closing-html-tag-in-a-httpmodule)
2. [Using only `BeginReques` event and `HttpApplication.Context.Response.Filter`](https://stackoverflow.com/questions/33391083/injecting-javascript-on-every-page-in-asp-net)

I'm not sure if placing our module after the one that uses `HttpApplication.Context.Response.Filter` will account for work done in custom response filter.

Seeing that the whole issue is quite complicated - do we want to recommend where to place our module relative to the other modules?

gregkalapos: I see... yes I agree, it's best to just leave the doc as it is and maybe think about this based on feedback. 👍

PR #508, opened 2019-09-24
Fix AspNetFullFrameworkSampleApp logging

SergeyKleyman:
**Note:** this PR is built on top of #507 and it should be reviewed only after #507 is merged.

1. Before this change `AspNetFullFrameworkSampleApp` redirected agent's logging to NLog only if `ELASTIC_APM_ASP_NET_FULL_FRAMEWORK_SAMPLE_APP_LOG_FILE` environment variable was set (it's set by `Elastic.Apm.AspNetFullFramework.Tests`). It is useful to have `AspNetFullFrameworkSampleApp` log to the rest of target other than file (Trace, etc.) even when `ELASTIC_APM_ASP_NET_FULL_FRAMEWORK_SAMPLE_APP_LOG_FILE` is not set. For example when `AspNetFullFrameworkSampleApp` is started manually from IDE and not automatically by `Elastic.Apm.AspNetFullFramework.Tests`.
2. Add `Elastic APM .NET AspNetFullFrameworkSampleApp>` prefix to `TraceTarget` output (goes to `System.Diagnostics.Trace`) so it's easier to filter `AspNetFullFrameworkSampleApp` log even the log statements producing multiple lines (exception stack traces, etc.)

### DIFF
--- a/sample/AspNetFullFrameworkSampleApp/App_Start/LoggingConfig.cs
+++ b/sample/AspNetFullFrameworkSampleApp/App_Start/LoggingConfig.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.IO;
 using Elastic.Apm.AspNetFullFramework;
 using NLog;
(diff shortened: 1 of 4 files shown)


sample/AspNetFullFrameworkSampleApp/App_Start/LoggingConfig.cs
```diff
@@ -53,5 +58,60 @@ public static void SetupLogging()
...
+
+			internal PrefixingTraceTarget(string prefix = "")
+			{
+				_prefix = prefix;
+				OptimizeBufferReuse = true;
+			}
+
+			protected override void Write(LogEventInfo logEvent)
+			{
+				var message = this.RenderLogEvent(this.Layout, logEvent);
```
gregkalapos: nit, codestyle: `this` should be removed according to `.editorconfig`.

Issue #491, opened 2019-09-16
Nullreference in AWS Lambda


gregkalapos:
Callstack:

```
[Error] Amazon.Lambda.AspNetCoreServer.APIGatewayProxyFunction: Unknown error responding to request: NullReferenceException:
System.NullReferenceException: Object reference not set to an instance of an object.
at Elastic.Apm.AspNetCore.ApmMiddleware.GetHttpVersion(String protocolString)
at Elastic.Apm.AspNetCore.ApmMiddleware.FillSampledTransactionContextRequest(HttpContext context, Transaction transaction)
at Elastic.Apm.AspNetCore.ApmMiddleware.InvokeAsync(HttpContext context)
at Amazon.Lambda.AspNetCoreServer.AbstractAspNetCoreFunction`2.ProcessRequest(ILambdaContext lambdaContext, Context context, InvokeFeatures features, Boolean rethrowUnhandledError)
```

Based on the stack I'd say this happens when we read the HTTP version.

Based on the callstack this happens [here](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs#L265) and i suspect we pass `null` into that method [here](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs#L164) to get the HTTP version. Maybe `context.Request.Protocol` is `null`.

Given that HTTP version is optional we can have a quick fix for this. It'd be also worth reproducing it and see why we can't get this info.

gregkalapos:
The nullref is fixed in #512  - I tested it and there is no crash anymore, the application works as expected. 

Unfortunately I discovered another problem during my testing - I opened #518. Closing this now, since the nullref problem is solved and the other problem is either not related to AWS and can also happen somewhere else, or it's something AWS related but that'd mean that offering AWS Lambda support will take more effort than expected.

PR #512, opened 2019-09-25
add try-catch block for FillSampledTransactionContextRequest method

### LINKED ISSUE #502: Exception in Elastic.Apm.AspNetCore.ApmMiddleware.FillSampledTransactionContextRequest
From https://discuss.elastic.co/t/net-core-exception/200485

Logs: 
```
{ "time": "2019-09-20 14:18:59.9471", "application_trace_id": "0HLPTVNSIHH2O:00000005", "user_id": "561b95df-3047-47b5-b82d-a4e88c892bd2", "level": "ERROR", "nested": { "message": "Connection id "0HLPTVNSIHH2O", Request id "0HLPTVNSIHH2O:00000005": An unhandled exception was thrown by the application.", "exception": "System.FormatException: The specified content type is invalid.\n at System.Net.Mime.ContentType.ParseValue()\n at System.Net.Mime.ContentType..ctor(String contentType)\n at Elastic.Apm.AspNetCore.ApmMiddleware.FillSampledTransactionContextRequest(HttpContext context, Transaction transaction)\n at Elastic.Apm.AspNetCore.ApmMiddleware.InvokeAsync(HttpContext context)\n at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol.ProcessRequests[TContext](IHttpApplication`1 application)" }, "ConnectionId": "0HLPTVNSIHH2O", "TraceIdentifier": "0HLPTVNSIHH2O:00000005", "EventId_Id": 13, "EventId_Name": "ApplicationError", "EventId": "ApplicationError" }
```

The call to `Elastic.Apm.AspNetCore.ApmMiddleware.FillSampledTransactionContextRequest` should be in a try block, which is currently not the case. #491 is another one in the same area.

vhatsura:
fixes #502 and #491

### DIFF
--- a/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
+++ b/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
@@ -44,54 +44,79 @@ namespace Elastic.Apm.AspNetCore
 
 		public async Task InvokeAsync(HttpContext context)
 		{
-			Transaction transaction;
-
-			var transactionName = $"{context.Request.Method} {context.Request.Path}";
-
-			if (context.Request.Headers.ContainsKey(TraceParent.TraceParentHeaderName))
-			{
-				var headerValue = context.Request.Headers[TraceParent.TraceParentHeaderName].ToString();
-
-				var distributedTracingData = TraceParent.TryExtractTraceparent(headerValue);
-
-				if (distributedTracingData != null)
-				{
-					_logger.Debug()
-						?.Log(
-							"Incoming request with {TraceParentHeaderName} header. DistributedTracingData: {DistributedTracingData}. Continuing trace.",
-							TraceParent.TraceParentHeaderName, distributedTracingData);
-
-					transaction = _tracer.StartTransactionInternal(
-						transactionName,
-						ApiConstants.TypeRequest,
-						distributedTracingData);
-				}
-				else
-				{
-					_logger.Debug()
-						?.Log(
-							"Incoming request with invalid {TraceParentHeaderName} header (received value: {TraceParentHeaderValue}). Starting trace with new trace id.",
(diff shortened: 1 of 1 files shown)


gregkalapos:
Just had a quick look - one thing that came to my mind:
Shouldn't we also move [this `try` block](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs#L89) to the 1. line of the [`public async Task InvokeAsync(HttpContext context)`](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs#L46) method? I'd definitely sleep better if'd have that :)

vhatsura:
@gregkalapos, thanks for review. I will look what I can to do it with that, however, moving try block to the 1st line cause another problem: in case of exception during transaction start we will skip calling the next middleware. I suggest to divide `InvokeAsync` method to three parts: the 1st one is start transaction with capturing request details, 2nd - call the next middleware, the last - stop transaction with capturing response details. In such case we will have 3 try-catch blocks for each part in `InvokeAsync` method. In such case I believe it will be more safe WDYT?

gregkalapos:
> @gregkalapos, thanks for review. I will look what I can to do it with that, however, moving try block to the 1st line cause another problem: in case of exception during transaction start we will skip calling the next middleware. I suggest to divide `InvokeAsync` method to three parts: the 1st one is start transaction with capturing request details, 2nd - call the next middleware, the last - stop transaction with capturing response details. In such case we will have 3 try-catch blocks for each part in `InvokeAsync` method. In such case I believe it will be more safe WDYT?

Yes, that sounds very good, I agree with that. Thanks!

PR #514, opened 2019-09-25
Fix for Central configuration doesn't work if environment name contains characters that have to be URL encoded #513

### LINKED ISSUE #513: Central configuration doesn't work if environment name contains characters that have to be URL encoded
For example if user configures:
```
"ElasticApm": {
    ...
    "Environment": "my env amp:(&) plus:(+) ang:(<>) eq:(=) qm:(?)"
    ...
}
```
Kibana does display environment name correctly and user can create configuration for the corresponding service and environment but that agent will not receive the created configuration,

The root cause is the fact that agent doesn't URL encode service and environment names when sending them to APM Server's get-central-config API (`config/v1/agents`) so in the example above agents sends:
```
http://.../config/v1/agents?service.name=...&service.environment=my%20env%20amp:(&)%20plus:(+)%20ang:(%3C%3E)%20eq:(=)%20qm:(?)
```
while the correct URL should be
```
http://.../config/v1/agents?service.name=...&service.environment=my+env+amp%3A%28%26%29+plus%3A%28%2B%29+ang%3A%28%3C%3E%29+eq%3A%28%3D%29+qm%3A%28%3F%29
```

SergeyKleyman:
**Note:** this PR is built on top of #509 and it should be reviewed only after #509 is merged.

Fixes #513
1. URL encode service name and environment.
2. In addition make sure combining server's base URL and API path handles various cases.

### DIFF
--- a/src/Elastic.Apm/BackendComm/BackendCommUtils.cs
+++ b/src/Elastic.Apm/BackendComm/BackendCommUtils.cs
@@ -1,10 +1,10 @@
 using System;
-using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Reflection;
+using System.Text;
 using System.Text.RegularExpressions;
 using Elastic.Apm.Api;
 using Elastic.Apm.Config;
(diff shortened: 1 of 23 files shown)


axw:
There seems to be a lot of unrelated .NET Full Framework bits mixed into this PR.

SergeyKleyman:
> There seems to be a lot of unrelated .NET Full Framework bits mixed into this PR.

@axw Thank you for bringing it to my attention - this PR is built on top of #509 and it should be reviewed only after #509 is merged. I'll add it as a note to top of this PR.

src/Elastic.Apm/BackendComm/BackendCommUtils.cs
```diff
@@ -21,6 +21,58 @@ internal class BackendCommUtils
 
 		private static readonly TimeSpan DnsTimeout = TimeSpan.FromMinutes(1);
 
+		internal static class ApmServerEndpoints
+		{
+			internal static Uri BuildIntakeV2EventsAbsoluteUrl(Uri baseUrl) =>
+				CombineAbsoluteAndRelativeUrls(baseUrl, "intake/v2/events");
+
+			internal static Uri BuildGetConfigAbsoluteUrl(Uri baseUrl, Service service)
```
gregkalapos: I guess this is the fix itself, maybe I'd add some comments here, something like:

```
/// <summary>
/// Builds the absolute URL that points to the server URL that gives the configuration to the agent with the specific service.environment
/// </summary>
/// <param name="baseUrl"></param>
/// <param name="service">Points to the service which will be URL encoded in the returned URL</param>
/// <returns></returns>
```

Just saying because it took me a while to figure out what's going on here.

src/Elastic.Apm/Report/PayloadSenderV2.cs
```diff
@@ -44,6 +44,8 @@ internal class PayloadSenderV2 : BackendCommComponentBase, IPayloadSender
 		{
 			_logger = logger?.Scoped(ThisClassName + (dbgName == null ? "" : $" (dbgName: `{dbgName}')"));
 
+			_intakeV2EventsAbsoluteUrl = BackendCommUtils.ApmServerEndpoints.BuildIntakeV2EventsAbsoluteUrl(config.ServerUrls.First());
```
gregkalapos: These changes are not related to central config, right? Is it because you think we'd have the same relative path problem with the events endpoint?

gregkalapos:
Merging this due to release - questions can be discussed later - overall I think this is ready-to-merge, no question is open that would be a blocker.

SergeyKleyman:
I opened https://github.com/elastic/apm-agent-dotnet/pull/521 to address review feedback.

PR #525, opened 2019-10-02
update documentation and sample for manually propagating distributed …

### LINKED ISSUE #476: Sample code in "Manually propagating distributed tracing context" should use CurrentSpan and not only CurrentTransaction
Currently [sample code in "Manually propagating distributed tracing context" ](https://www.elastic.co/guide/en/apm/agent/dotnet/current/public-api.html#manually-propagating-distributed-tracing-context)
```
string outgoingDistributedTracingData = Agent.Tracer.CurrentTransaction
                     .OutgoingDistributedTracingData.SerializeToString();
```

while it should be
```
string outgoingDistributedTracingData =
    (Agent.Tracer.CurrentSpan?.OutgoingDistributedTracingData 
        ?? Agent.Tracer.CurrentTransaction?.OutgoingDistributedTracingData)?.SerializeToString();
```

We should also update `sample/ApiSamples/Program.cs` https://github.com/elastic/apm-agent-dotnet/blob/6fb3ca96d723d099f747fd46b513b59b8459b449/sample/ApiSamples/Program.cs#L158

vhatsura:
…tracing

closes #476 

@SergeyKleyman, I'm curious why sample need to be updated if in sample transaction is started manually?

### DIFF
--- a/docs/public-api.asciidoc
+++ b/docs/public-api.asciidoc
@@ -135,8 +135,9 @@ you can manually propagate distributed tracing context from the caller to the ca
 First you serialize distributed tracing context on the caller side:
 [source,csharp]
 ----
-string outgoingDistributedTracingData = Agent.Tracer.CurrentTransaction
-                     .OutgoingDistributedTracingData.SerializeToString();
+string outgoingDistributedTracingData =
+    (Agent.Tracer.CurrentSpan?.OutgoingDistributedTracingData 
+        ?? Agent.Tracer.CurrentTransaction?.OutgoingDistributedTracingData)?.SerializeToString();
 ----
 Then you transfer the resulted string to the callee side
 and you continue the trace by passing deserialized distributed tracing context to any of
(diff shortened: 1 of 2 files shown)


SergeyKleyman:
The problem with using `CurrentTransaction.OutgoingDistributedTracingData` instead of `CurrentSpan?.OutgoingDistributedTracingData` is that `parent-id` in the returned `DistributedTracingData` [will be `transaction-id`](https://github.com/elastic/apm-agent-dotnet/blob/045a07bfc1d77962b98cca78779ea4be063851c7/src/Elastic.Apm/Model/Transaction.cs#L137) while [it should be `span-id`](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm/Model/Span.cs#L95) (assuming that span is not dropped or non-sampled). I think you already encountered that area of the code while working on dropped spans.

vhatsura:

PR #526, opened 2019-10-02
fix parent span restore after ending span

### LINKED ISSUE #475: Span.End() doesn't restore current span if transaction is not sampled
Current span is restored
https://github.com/elastic/apm-agent-dotnet/blob/61ee66e1988da987af267005a5b006b80252f6e7/src/Elastic.Apm/Model/Span.cs#L204

only when span's `IsSampled` is `true`

https://github.com/elastic/apm-agent-dotnet/blob/61ee66e1988da987af267005a5b006b80252f6e7/src/Elastic.Apm/Model/Span.cs#L188

vhatsura:
closes #475


vhatsura:

test/Elastic.Apm.Tests/SpanTests.cs
```diff
@@ -0,0 +1,32 @@
...
+		[Fact]
+		public void End_ShouldRestoreParentSpan_WhenTransactionIsNotSampled()
+		{
+			// Arrange
+			var currentExecutionSegmentsContainer = new NoopCurrentExecutionSegmentsContainer();
+			var payloadSender = new NoopPayloadSender();
+			var logger = new NoopLogger();
+			var configurationReader = new MockConfigSnapshot();
+
+			var transaction = new Transaction(logger, "transaction", "type", new Sampler(1.0), null, payloadSender, configurationReader,
```
SergeyKleyman: As I mentioned at https://github.com/elastic/apm-agent-dotnet/pull/533#pullrequestreview-306361626, we should minimize use of agent's implementation internals in tests and use agent's public API instead.

SergeyKleyman: I assume it should be `new Sampler(0)` (or [`config: new MockConfigSnapshot(transactionSampleRate: "0")`](https://github.com/elastic/apm-agent-dotnet/blob/480cb4fdcd5b2034b62ed6b0fb9d5da6026ff3e7/test/Elastic.Apm.Tests/SpanTests.cs#L64) if we use only public API) because otherwise the test passes even without the fix to production code.

test/Elastic.Apm.Tests/SpanTests.cs
```diff
@@ -0,0 +1,30 @@
...
+				var transaction = agent.Tracer.StartTransaction("transaction", "type");
+
+				var parentSpan = transaction.StartSpan("parent", "type");
+
+				// Act
+				parentSpan.CaptureSpan("span", "type", span => { });
+
+				// Assert
+				payloadSender.Spans.Count.Should().Be(0);
+				agent.TracerInternal.CurrentExecutionSegmentsContainer.CurrentSpan.Should().Be(parentSpan);
```
SergeyKleyman:
```suggestion
				agent.Tracer.CurrentSpan.Should().Be(parentSpan);
```

I think `agent.Tracer.CurrentSpan` is preferable to `agent.TracerInternal.CurrentExecutionSegmentsContainer` since the former is public API.

Issue #510, opened 2019-09-25
Enforce code style in CI


axw:
I propose that we introduce a step in CI that uses [dotnet-format](https://github.com/dotnet/format#how-to-use) to enforce the code style defined in `.editorconfig`, by running `dotnet format --check --dry-run`

gregkalapos:
cc @elastic/dotnet

gregkalapos:
Done in #532.

PR #532, opened 2019-10-03
Enforce code style with dotnet format 

gregkalapos:
Addressing #510.

## What's new:
With `dotnet format --dry-run --check` we make sure all our C# code is formatted the way it should be formatted according to our `.editorconfig` file. If that is not the case then `dotnet format` returns with non-zero exit code.

I added this step only to the Linux build - I think that's enough, no need to run it on multiple OSs. Let me know if there is a better place for it.

## Next step:

~This build currently fails because we have code style violations - those are addressed in #535 (opened a separate PR, otherwise it'd be hard to review this), so that'll be merged first. #536 shows the status after #535 is merged into that branch:~ Update: code cleanup done, this PR branch should be also ok. 

Screen Shot 2019-10-04 at 12 52 40

### DIFF
--- a/.ci/linux/codestyle.sh
+++ b/.ci/linux/codestyle.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+#
+# This script installs the required tools to be used to make sure there are no code violations in the source code.
+#
+set -euxo pipefail
+
+# Install
+dotnet tool install -g dotnet-format
+
+# Check
+dotnet format --dry-run --check
--- a/.ci/linux/codestyle.sh
+++ b/.ci/linux/codestyle.sh
@@ -5,7 +5,7 @@
 set -euxo pipefail
 
 # Install
-dotnet tool install -g dotnet-format
+dotnet tool install -g dotnet-format --version 3.1.37601
 
 # Check
 dotnet format --dry-run --check
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -48,10 +48,13 @@ pipeline {
               environment {
                 MSBUILDDEBUGPATH = "${env.WORKSPACE}"
               }
+              /**
+              Make sure there are no code style violation in the repo.
+              */
               stages{
                 stage('CodeStyleCheck') {
                   steps {
-                    withGithubNotify(context: 'CodeStyle', tab: 'tests') {
+                    withGithubNotify(context: 'CodeStyle check', tab: 'tests') {
                       deleteDir()
                       unstash 'source'
                       dir("${BASE_DIR}"){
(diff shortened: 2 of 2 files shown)


Jenkinsfile
```diff
@@ -50,7 +50,23 @@ pipeline {
               environment {
                 MSBUILDDEBUGPATH = "${env.WORKSPACE}"
               }
+              /**
+              Make sure there are no code style violation in the repo.
+              */
               stages{
+                stage('CodeStyleCheck') {
+                  steps {
+                    withGithubNotify(context: 'CodeStyle check', tab: 'tests') {
```
kuisathaverat: the tab should be the default one, simply remove the `tab` field, this does not generate a xunit file with tests results

kuisathaverat:
```suggestion
                    withGithubNotify(context: 'CodeStyle check') {
```

gregkalapos: Thanks, done!

.ci/linux/codestyle.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+#
+# This script installs the required tools to be used to make sure there are no code violations in the source code.
+#
+set -euxo pipefail
+
+# Install
+dotnet tool install -g dotnet-format
```
kuisathaverat:
```suggestion
dotnet tool install -g dotnet-format --version 3.1.37601
```

kuisathaverat:
>It would be good for .ci to start moving to local .NET tools when this repository moves over to .NET Core 3.0 so that the build stays repeatable. cc @kuisathaverat

probably we have to have two build one for .NET 2.x and another for .NET 3.x to ensure compatibility

.ci/linux/codestyle.sh
gregkalapos: Just to make sure: with this we pin the version of `dotnet-format` (which is accidentally also `3.x`, but that is unrelated to the .NET Core 3.0 comment from above. I'm ok with this, just want to make sure...

gregkalapos:
> > It would be good for .ci to start moving to local .NET tools when this repository moves over to .NET Core 3.0 so that the build stays repeatable. cc @kuisathaverat
> 
> probably we'd have to have two builds, one for .NET 2.x, and another for .NET 3.x to ensure compatibility

For compatibility actually we'll also need some 2.x and 3.x projects that use the agent. Now I'm not totally sure, but I think if we have 2.x and 3.x projects that test the agent then 1 build is enough, because that'd pull the corresponding .NET Core version anyway. I opened https://github.com/elastic/apm-agent-dotnet/issues/546

PR #533, opened 2019-10-03
use transaction id as parent id for error when span non sampled or dr…

### LINKED ISSUE #477: Error's parent_id should be transaction_id for non sampled and dropped spans
To align with https://github.com/elastic/apm/issues/16 agent should not include `Span.Id` in other events for spans that are not going to be sent out (non sampled and dropped spans). At the moment it seems that we do comply with the above requirement except for Error event's `parent_id` which is set to `Span.Id` even for non sampled and dropped spans. 

**Implementation notes**
Code in question is https://github.com/elastic/apm-agent-dotnet/blob/61ee66e1988da987af267005a5b006b80252f6e7/src/Elastic.Apm/Model/Span.cs#L207-L217 and https://github.com/elastic/apm-agent-dotnet/blob/61ee66e1988da987af267005a5b006b80252f6e7/src/Elastic.Apm/Model/Span.cs#L244-L254

By the time we get to fixing this bug we might already have changes discussed at https://github.com/elastic/apm-agent-dotnet/pull/472#issuecomment-528399483 already implemented so we should use `ShouldBeSentToApmServer` and instead of just passing `parentId` to `ExecutionSegmentCommon.CaptureException`/`ExecutionSegmentCommon.CaptureError` we should pass something like `parentId ?? (ShouldBeSentToApmServer ? null : _enclosingTransaction.Id)`

vhatsura:
…opped

closes #477

### DIFF
--- a/src/Elastic.Apm/Model/Span.cs
+++ b/src/Elastic.Apm/Model/Span.cs
@@ -226,7 +226,7 @@ namespace Elastic.Apm.Model
 				_enclosingTransaction,
 				culprit,
 				isHandled,
-				parentId
+				parentId ?? (ShouldBeSentToApmServer ? null : _enclosingTransaction.Id)
 			);
 
 		public void CaptureSpan(string name, string type, Action<ISpan> capturedAction, string subType = null, string action = null)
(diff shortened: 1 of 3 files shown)


vhatsura:

vhatsura:
> It looks good with one comment - we should minimize use of agent's implementation internals in tests and use agent's public API instead. I've adapted the tests in [480cb4f](https://github.com/elastic/apm-agent-dotnet/commit/480cb4fdcd5b2034b62ed6b0fb9d5da6026ff3e7) - please take a look and merge if you find the changes acceptable.

I rewrote tests.

SergeyKleyman:
@vhatsura Thank you very much for your contribution.

PR #535, opened 2019-10-04
Fix all code style violations

gregkalapos:
#532 adds `dotnet format` checks to make sure all PRs respect our `.editorconfig`. Currently #532 fails, because we already have lots of violations in our code base. This PR fixes those.

All done by auto code cleanup - there is no manual code change here. I think this does not need too much manual review, it'd be very hard to do anyway.


gregkalapos:
This auto cleanup became out of date (due to #551). Closing and doing the cleanup in #536.

Issue #441, opened 2019-08-13
Don't use static Agent API on APM middleware registration


eduardocampano:
`UseElasticApm` function relies on static `Agent.Setup(config)`, which throws if registering the middleware was already done. But this is not considering the scenario where more than one `IHost` or `IWebHost` instances can be running on the same process.

I think the common practice is registering a new `IApmAgent` instance as singleton to `IServiceCollection` and injecting it into the middleware. [AutoMapper](https://github.com/AutoMapper/AutoMapper.Extensions.Microsoft.DependencyInjection) had the same issue previously and they fixed it following this approach and they just removed the static API for good in version 9.

That said, having `IApmAgent` registered in DI allows injecting into controllers/services instead of relying on statics as [documentation recommends](https://www.elastic.co/guide/en/apm/agent/dotnet/current/public-api.html).

Furthermore, current configuration approach only allows configuring the Agent form `IConfiguration` and would be flexible to have an Options Builder approach like [EF Core](https://docs.microsoft.com/en-us/ef/core/miscellaneous/configuring-dbcontext#using-dbcontext-with-dependency-injection) has.

OlegUfaev:
I agree that UseElasticApm shouldn't rely on static fields and static API should be deprecated.

We already have an issue with integration tests in our Web application that uses APM agent. We are using WebApplicationFactory as explained here: https://docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-2.2#customize-webapplicationfactory

When we inject сustomized WebApplicationFactory into different test classes (via IClassFixture as explained in article) then multiple instances created and method UseElasticApm called multiple times. So we catch exception: "The singleton APM agent has already been instantiated and can no longer be configured".

We have that problem only with APM agent and other middlewares work just fine.

mterwoord:
What's the progress on this issue? To me this is a pretty big issue, as it bascally means all integration tests fail. 

Are there any known work-arounds?

pburrows:
This is a dumb workaround, but it works:

```
try
            {
                app.UseAllElasticApm(_configuration);
            }
            catch (Exception ex)
            {
                // ignore this error
            }
```
You could, of course, make it more sophisticated.

mterwoord:
Thanks. But that only enables it for the first integration test, not the rest. Also it can hide errors occurring in production. What I did now was to run integration tests with Environment=Test, and then use that to exclude registering ElasticApm

vhatsura:
[...]

PR #537, opened 2019-10-05
add extensions method on IHostBuilder for easy integration in .Net Core console apps

vhatsura:
it's very early draft request to point some issues.

@SergeyKleyman and @gregkalapos, I tried to implement extension method on IHostBuilder and has some difficulties.
First of all a few words about changes. I created the separate project `Elastic.Apm.NetCore`, which depends on `Elastic.Apm` project. `Elastic.Apm.AspNetCore` depends on `Elastic.Apm.NetCore` because conceptually asp net core is .net core console application. Also, it allow to share `MicrosoftExtensionsConfig` and some other shared stuff.

Topics which will be discussed:
1. `IHostBuilder` was introduced in 2.1.0 version of `Microsoft.Extensions.Hosting.Abstractions`. The package depends on `Microsoft.Extensions.Logging.Abstractions`, which means that `Microsoft.Extensions.Logging.Abstractions` must be also `2.1.0` version. Also, it means that `Microsoft.Extensions.Logging.Abstractions` and `Microsoft.Extensions.Configuration.Abstractions` packages for `Elastic.Apm.AspNetCore` should target `2.1.0` version, which can be an issue because such package was released with targeting of `2.0.0` packages version.However,  [EOL for .Net Core 2.0 was on October 1, 2018](https://dotnet.microsoft.com/platform/support/policy/dotnet-core) and I think increasing version of such dependenies is eligible. WDYT?
2. It's impossible to get `ILogger` and `IHostingEnvironment` from dependency injection container due to `IHostBuilder` doesn't have ability to get access for `IServiceProvider` as `IApplicationBuilder` has. It means that we need to return to discussion about integration with dependency injection which was raised in #441 issue. But, I recommend more softly changes: don't remove static API, just add additional ability to register apm components in dependency injection container. As a result, instead of static API we need to inject `IApmAgent` to classes where we want to use Public API (I believe after implement instrumentation for most of cases it will be rare case). As a bonus, proper integration with di container will give ability to auto reload (without service restart) agent dependencies on configuration update.

All changes you can find in this pull request. Due to not all components are designed to easy integration with default .Net Core DI container, some code is weird. However, if we align on future strategy I will polish code.


Mpdreamz:
I think this new package should be called `Elastic.Apm.IHostBuilderExtensions` or `Elastic.Apm.GernericHostBuilder`. `NetCore` is too vague and sends the wrong signal: the core libs are `netstandard2.0` and also work on `.NET Core`. 

Since in 3.0 `IHostBuilder` serves as the base for `IWebhostBuilder`I wonder if we could multitarget `Elastic.Apm.AspNetCore` instead to `netstandard2.0;netstandard2.1` and  `#ifdef` `Elastic.Net.AspNetCore` to take a dependency on `Elastic.Apm.IHostBuilderExtensions` when compiling against `netstandard2.1` but leave everything as is for `netstandard2.0` 


This would `Elastic.Apm.IHostBuilderExtensions` to be used by `netcoreapp2.1` console projects and in the context of `netcoreapp3.0` asp.net web apps.

vhatsura:
Hey, @Mpdreamz. Thank for feedback. I totally agree with you and such thought also was in my mind. However, first if all it would be nice to resolve raised questions to understand future steps. After defining strategy of implementation, I will implement mentioned concept.

Mpdreamz:
Glad were aspiring towards the same solution :smile:

To address your points more directly from my POV

1. This is less of an issue when we multitarget `Elastic.Apm.AspNetCore`, the `netstandard2.0` TFM stays on `2.0.0` and the new `netstandard2.1.` TFM can bump to `2.1.0 of the `Microsoft.Extensions.*` packages  

 `AspNetCore` apps targeting .NET 2.1 would resolve the `netstandard2` reference like they do today. .NET core 2.1 applications would be able to take a dependency on `Elastic.Apm.IHostBuilderExtensions`. AspNetCore 3.0 apps will get the best of both worlds.

2. I fully support #441 very keen to move that over 👍 Easiest would be to make `ApmAgent` public. 
This means you can initialize statically through `Agent` or through registering an instance of `ApmAgent` as singleton in apps or use many of them in your tests. The latter is what we also do and what was the driving motivation behind https://github.com/elastic/apm-agent-dotnet/pull/63. I'm keen to open this up to everyone. Looking forward to hear @gregkalapos and @SergeyKleyman thoughts here.

vhatsura:
Just a small update: the 1st topic will be resolved in #608

abilican:
Hi @vhatsura, what's the release date for this feature?

vhatsura:
Hey, @abilican.
It was just a concept to try and doesn't have a certain date to release. From my side, I'm going to return to the PR in 2 weeks as it needed in more projects.

gregkalapos:
I also think the direction here is good - so far this looks good to me.

2 things (both related to each other) I think we should discuss:

- What do we do about `IDiagnosticsSubscriber`s? 

Currently out-of-the-box nothing will be turned on and users are expected to pass those. This is fine as 1 option, but I think it’s realistic that people also would like to have the option to “just turn on everything” also when they use `IHostBuilder`. This would be similar to what we already have with `UseAllElasticApm` in `Elastic.Apm.NetCoreAll`. The thinking behind this is that a) I already see support cases stating the agent doesn’t capture anything if we don’t have it :) and b) we can’t expect people to know what `IDiagnosticsSubscriber` we have and expect them to discover those per library when they want to have auto instrumentation. Maybe as a solution for this we could add a `UseAllElasticApm` overload  in the `Elastic.Apm.NetCoreAll` package that passes all `IDiagnosticsSubscriber` implementations into `Elastic.Apm.Extensions.Hosting.HostBuilderExtensions.UseElasticApm`.

- Integrating this with ASP.NET Core

At the moment, I think this still won’t capture any ASP.NET Core calls, and I suspect there is no way to turn on the middleware here. In case of the suggestion above with `UseAllElasticApm`, it’d be also nice to turn on ASP.NET Core monitoring. We have the same problem in https://github.com/elastic/apm-agent-dotnet/pull/828 and to solve this I already started to rewrite the middleware based on DiagnosticSource. Once that’s done, we could just turn on the new `AspNetCorePageLoadDiagnosticListener` (or whatever we’ll call it - name is not final yet) as we do with any other `IDiagnosticListener` implementation.

So, the point I wanna make: I think it’d be nice to also have a “turn on everything for me” story also with `IHostBuilder`.

src/Elastic.Apm.Extensions.Hosting/HostBuilderExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+				services.AddSingleton<IConfigurationReader>(sp =>
+					new MicrosoftExtensionsConfig(ctx.Configuration, sp.GetService<IApmLogger>(), ctx.HostingEnvironment.EnvironmentName));
+
+				services.AddSingleton(sp =>
+				{
+					var components = new AgentComponents(sp.GetService<IApmLogger>(), sp.GetService<IConfigurationReader>());
+					UpdateServiceInformation(components.Service);
+					return components;
+				});
+				services.AddSingleton<IApmAgent, ApmAgent>(sp =>
```
gregkalapos, line 30: Should we maybe also add the `ITracer` here?

Reason I bring this up is that I see people a lot being focused on the Agent class itself (or `IApmAgent`), but what they typically want is just to start traces, spans, and capture errors and for that all they need is an `ITracer`.

vhatsura:
@gregkalapos, thanks for your feedback, it is much appreciated to me.
I fully agree with you regarding `just turn on everything`. I'll do it later, just want to stabilize changes inside `Elastic.Apm.Extensions.Hosting` project.

> Integrating this with ASP.NET Core
The main purpose of the changes is to be able to use Elastic .NET APM agent in console applications. However, ASP.NET Core integration can be done also at the moment without an additional diagnostic listener. It's just needed to add an additional extension method without APM agent configuration. The main issue is the `ApmMiddleware` relies on `ITracer` instead of `Tracer`. Also, having two extensions method will be a little bit confusing for users.
The only possible way is the breaking change of public interface (remove the extension method on `IApplicationBuilder`), which isn't very good.

src/Elastic.Apm.Extensions.Hosting/HostBuilderExtensions.cs
vhatsura: Added

vhatsura:
@gregkalapos, CI gets the following error during the build new console sample:

> error NETSDK1073: The FrameworkReference 'NETStandard.Library' was not recognized 

From discussions at https://github.com/dotnet/sdk/issues/10273#issuecomment-499988244 looks like it was an issue with referencing .netstandard2.1 libraries from .net core app 3.0 and above, however, some people still face issue in some cases.

So, the purpose of having `netstandard2.1` target to be able to use the new library in .net core 2.x and 3.x projects. What is the preferable way to solve it? Do we need to target `netcoreapp3.0` instead of `netstandard2.1` or trying to investigate the issue in CI?

gregkalapos:
> So, the purpose of having `netstandard2.1` target to be able to use the new library in .net core 2.x and 3.x projects. 

Probably I miss something, but why isn't `netstandard2.0` enough for that? Do we use anything from 2.1? 

Probably related, I just noticed in `Elastic.Apm.AspNetCore.csproj` you also added `netcoreapp3.0` - I'm also wondering about that one.

vhatsura:
> > So, the purpose of having `netstandard2.1` target to be able to use the new library in .net core 2.x and 3.x projects.
> 
> Probably I miss something, but why isn't `netstandard2.0` enough for that? Do we use anything from 2.1?
> 
> Probably related, I just noticed in `Elastic.Apm.AspNetCore.csproj` you also added `netcoreapp3.0` - I'm also wondering about that one.

The reason is in `IHostingEnvironment` marked as deprecated in .net core 3.0

gregkalapos:
> > > So, the purpose of having `netstandard2.1` target to be able to use the new library in .net core 2.x and 3.x projects.
> > 
> > 
> > Probably I miss something, but why isn't `netstandard2.0` enough for that? Do we use anything from 2.1?
> > Probably related, I just noticed in `Elastic.Apm.AspNetCore.csproj` you also added `netcoreapp3.0` - I'm also wondering about that one.
> 
> The reason is in `IHostingEnvironment` marked as deprecated in .net core 3.0

That explains the change in `Elastic.Apm.AspNetCore.csproj` - but does `Elastic.Apm.Extensions.Hosting.csproj` need `netstandard2.1` because of this?

Honestly I'm a but reluctant to change target framework unless we really need to - maybe a quick research to read EnvironmentName differently would be worth - I'll also try to take a look into that.

vhatsura:
So, regarding `HostingEnvironment` in `Elastic.Apm.AspNetCore.csproj` I see that it was done by mistake. The ASP .NET Core has [dedicated interface for `EnvironmentName`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.hosting.ihostingenvironment.environmentname?view=aspnetcore-3.1).

For `Elastic.Apm.Extensions.Hosting.csproj` without additional target, I got the following error:

```
System.MissingMethodException: Method not found: 'Microsoft.Extensions.Hosting.IHostingEnvironment Microsoft.Extensions.Hosting.HostBuilderContext.get_HostingEnvironment()'.
  at at Elastic.Apm.Extensions.Hosting.HostBuilderExtensions.<>c__DisplayClass0_1.<UseElasticApm>b__1(IServiceProvider sp)
  at at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitFactory(FactoryCallSite factoryCallSite, RuntimeResolverContext context)
  at at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteVisitor`2.VisitCallSiteMain(ServiceCallSite callSite, TArgument argument)
  at at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitCache(ServiceCallSite callSite, RuntimeResolverContext context, ServiceProviderEngineScope serviceProviderEngine, RuntimeResolverLock lockType)
  at at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitRootCache(ServiceCallSite singletonCallSite, RuntimeResolverContext context)
  at at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteVisitor`2.VisitCallSite(ServiceCallSite callSite, TArgument argument)
  at at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.Resolve(ServiceCallSite callSite, ServiceProviderEngineScope scope)
  at at Microsoft.Extensions.DependencyInjection.ServiceLookup.DynamicServiceProviderEngine.<>c__DisplayClass1_0.<RealizeService>b__0(ServiceProviderEngineScope scope)
  at at Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngine.GetService(Type serviceType, ServiceProviderEngineScope serviceProviderEngineScope)
  at at Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngineScope.GetService(Type serviceType)
  at at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetService[T](IServiceProvider provider)
  at Elastic.Apm.Extensions.Hosting.HostBuilderExtensions.<>c.<UseElasticApm>b__0_2(IServiceProvider sp) in D:\Projects\github\apm-agent-dotnet\src\Elastic.Apm.Extensions.Hosting\HostBuilderExtensions.cs:36
```

This is because, as I understand correctly, `IHostBuilderContext.HostingEnvironment` change type from `IHostingEnvironment` to `IHostEnvironment` in .Net Core 3.0.

> Honestly I'm a bit reluctant to change target framework unless we really need to - maybe a quick research to read EnvironmentName differently would be worth - I'll also try to take a look into that.

It would be nice, however, I think adding additional target worth it due to prevent any further issues with dependencies.

vhatsura:
I dig into environment configuration and got the following results:

So, the environment can be set in various ways:
1. Environment variables: `ASPNETCORE_ENVIRONMENT` or `DOTNET_ENVIRONMENT`
2. `--environment` console argument
3. Programmatically with `UseEnvironment` on `IWebHostBuilder`. 

It was a quick look into possible ways of configuration. So, to implement it in agent codebase isn't the option (It was just thought).
The search for other data types to get the environment name has no results.

Also, a little bit details of why the additional target is needed for `Elastic.Apm.AspNetCore`. There is need to not package downgrade in `NetCoreAll`.

mccow002:
I pulled down this fork and tried it out locally, and I get this error:

`System.MethodAccessException: Attempt by method 'Elastic.Apm.Extensions.Hosting.HostBuilderExtensions+<>c__DisplayClass0_0.<UseElasticApm>b__0(Microsoft.Extensions.Hosting.HostBuilderContext, Microsoft.Extensions.DependencyInjection.IServiceCollection)' to access method 'System.Func`2<System.__Canon,System.__Canon>..ctor(System.Object, IntPtr)' failed`

I can only resolve the issue by changing this:

`services.AddSingleton<IApmAgent, ApmAgent>(sp =>
				{
					var apmAgent = new ApmAgent(sp.GetService<AgentComponents>());
					if (subscribers != null && subscribers.Any()) apmAgent.Subscribe(subscribers);
					return apmAgent;
				});`

To this:

`services.AddSingleton<IApmAgent>(sp =>
				{
					var apmAgent = new ApmAgent(sp.GetService<AgentComponents>());
					if (subscribers != null && subscribers.Any()) apmAgent.Subscribe(subscribers);
					return apmAgent;
				});`

The fix is to remove the second generic parameter, ApmAgent.

I'm adding this to my project using the `.UseAllElasticApm()` method. Is there something else that needs to happen to prevent this error?

mccow002:
This will be extremely useful to us. Any idea when this will be merged?

fsqBr:
I Agree with @mccow002 , there is any idea when this will merged and the pack is available to use?

gregkalapos:
Thanks @vhatsura  - let's merge.

mccow002:
Any idea when this will be available through Nuget?

gregkalapos:
@mccow002 we don't publicly communicate specific release dates. It'll be out soon.

PR #539, opened 2019-10-09
Request body does not need old copy

iquirino:
The result of a EnableRewind is a valid Request.Body

All tests passed.

https://devblogs.microsoft.com/aspnet/re-reading-asp-net-core-request-bodies-with-enablebuffering/

Issue #460, opened 2019-08-26
ASP.NET Core request body capturing erases the body


gregkalapos:
Reported [here](https://github.com/elastic/apm-agent-dotnet/pull/402#issuecomment-523290643) by @brezaie. 

Reproducer also included in the comment - copying here:

```
[HttpPost("send")]
public JsonResult Send([FromBody] BaseReportFilter<SendMessageFilter> filter)
{
	//Do something
	return new JsonResult("");
}


public class BaseReportFilter<T>
{
    public T ReportFilter { get; set; }
}

public class SendMessageFilter{
	public string SenderApplicationCode { get; set; }
	public string MediaType { get; set; }
	public string Body { get; set; }
	public List<string> Recipients { get; set; }
}

//appsettings.json:
"ElasticApm": {
    "ServerUrls": "http://localhost:8200",
    "ServiceName": "ServiceName",
    "CaptureHeaders": true,
    "CaptureBody": "all",
    "CaptureBodyContentTypes": "application/x-www-form-urlencoded*, text/*, application/json*, application/xml*",
    "LogLevel": "Info",
    "StackTraceLimit": -1
  }
```

```"CaptureBodyContentTypes": "application/x-www-form-urlencoded*, text/*, application/json*, application/xml*",```

Without agent the `filter` parameter is populated, once the agent is activated it became `null`.

gregkalapos:
[Here](https://github.com/elastic/apm-agent-dotnet/blob/master/test/Elastic.Apm.AspNetCore.Tests/AspNetCoreMiddlewareTests.cs#L187) we have the test that covers this feature - the test case is a bit simpler, I don't see yet how it makes a difference, but as it seems it does make a difference.

minbyulcat:
This works fine if I comment out `request.Body = initialBody` in `ExtractRequestBody` method in `RequestExtentions.cs`. I hope this helps solve the problem.

iquirino:
Hi!

I am using the latest version and continues with this problem.

  "ElasticApm": {
    "ServerUrls": "#####",
    "TransactionSampleRate": 1.0,
    "CaptureBody": "all",
    "CaptureBodyContentTypes": "application/x-www-form-urlencoded*, text/*, application/json*, application/xml*"
  },

iquirino:
You are right!

All tests passed when comment this line.

I think is not necessary to do this and an accepted pull request.

[...]

PR #542, opened 2019-10-09
Add test covering #460 - making sure capturing request body does not erase the body

gregkalapos:
In #460 we had a bug which was that once the agent captures the request body, the body itself was not visible anymore in the MVC action method. This is fixed in https://github.com/elastic/apm-agent-dotnet/pull/539. 

This PR adds a test that reproduces and tests that scenario.

~The new test will fail until https://github.com/elastic/apm-agent-dotnet/pull/539 is not merged (atm. build is in-progress, I wait for it to finish, then I'll merge).~ Update: #539 is merged to `master`, I rebased this PR, all test should be green.

### DIFF
--- a/sample/SampleAspNetCoreApp/Controllers/HomeController.cs
+++ b/sample/SampleAspNetCoreApp/Controllers/HomeController.cs
@@ -31,6 +31,7 @@ namespace SampleAspNetCoreApp.Controllers
 				throw new ArgumentException($"{captureControllerActionAsSpanQueryStringKey} query string key should have at most one value" +
 					$", instead it's values: {captureControllerActionAsSpanQueryStringValues}",
 					captureControllerActionAsSpanQueryStringKey);
+
 			// ReSharper disable once SimplifyConditionalTernaryExpression
 			return captureControllerActionAsSpanQueryStringValues.Count == 0 ? false : bool.Parse(captureControllerActionAsSpanQueryStringValues[0]);
 		}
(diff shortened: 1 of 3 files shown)


sample/SampleAspNetCoreApp/Controllers/HomeController.cs
```diff
@@ -39,25 +40,25 @@ public Task<IActionResult> Index()
 		{
 			return SafeCaptureSpan<IActionResult>(GetCaptureControllerActionAsSpanFromQueryString(),
 				"Index_span_name", "Index_span_type", async () =>
-			{
```
gregkalapos: Just code style... we are working on avoiding this. https://github.com/elastic/apm-agent-dotnet/pull/536

gregkalapos:
There is 1 new failing test: `TestJsonBodyRetrievalOnRequestFailureInAspNetCore – Elastic.Apm.AspNetCore.Tests.AspNetCoreDiagnosticListenerTest` - investigating it.

test/Elastic.Apm.AspNetCore.Tests/Helper.cs
```diff
@@ -27,8 +27,6 @@ public static class Helper
 
 						app.UseDeveloperExceptionPage();
 
-						app.UseHsts();
```
gregkalapos: This is somewhat unrelated, but this HSTS is not needed, so I removed it.

iquirino:
SampleAspNetCoreApp -> we can reproduce trying to add a new row on "InsertDb Sample".

The request.Body was initially typed as: Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpRequestStream

After request.EnableRewind() it get the type changed to: Microsoft.AspNetCore.WebUtilities.FileBufferingReadStream

You consume all the request.Body and then return a copy of the stream alredy used.
But this will have the position to the end and will not able to rewind.

You need to mantain the rewinded version.

I think a good test is:

Check if the Stream Position=0 and CanSeek=true

gregkalapos:
> SampleAspNetCoreApp -> we can reproduce trying to add a new row on "InsertDb Sample".
> 
> The request.Body was initially typed as: Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpRequestStream
> 
> After request.EnableRewind() it get the type changed to: Microsoft.AspNetCore.WebUtilities.FileBufferingReadStream
> 
> You consume all the request.Body and then return a copy of the stream alredy used.
> But this will have the position to the end and will not able to rewind.
> 
> You need to mantain the rewinded version.
> 
> I think a good test is:
> 
> Check if the Stream Position=0 and CanSeek=true

I agree with your analysis, but I'm not sure I get what you suggest here. What I'm saying is that our original test did not repro this problem (probably due to the fact that we did not really have real HTTP requests going on), but the test in this PR does. The test we added here failed before I merged your fix.

>  Check if the Stream Position=0 and CanSeek=true

Yeah, that'd make sense, but given the current test also repros it I think it's fine. But feel free to submit a PR, I'm happy to take such a test, but I think the coverage with this new test is already good.

gregkalapos:
> Test itself LGTM:
> 
> https://github.com/elastic/apm-agent-dotnet/blob/1247981652264355d5a900fa9476f9503661e186/src/Elastic.Apm.AspNetCore/Extensions/RequestExtentions.cs#L33-L39
> 
> The truncate here seems less then optimal.
> 
> Is capturing a body something worth putting behind a configuration toggle? `EnableRewind()` does things like spooling to disk.

By default this feature is disabled - so it's already behind a configuration. Yeah, I'm also not a big fan of this, but I think given that it's not on by default I think it's fine... maybe we can research better ways to do this later, e.g. `HttpContext.Request.BodyReader` in ASP.NET Core 3.0.

Issue #544, opened 2019-10-09
Using Serilog with Elastic APM causes NullReferenceException


scotm-gxg:
**Describe the bug**
When using Serilog and Elastic APM together in Asp.Net COre, a NullReferenceException is thrown on start-up.

Versions:
netcoreapp 2.1
Elastic.Apm.NetCoreAll 1.1.1
Serilog.AspnetCore 3.0.0

**To Reproduce**
1. Clone this repo: https://github.com/scotmeiste/elastic-apm-serilog-error
2. in appsettings.json, replace ServerUrls and SecretToken with your instance of Elastic APM
3. Run Web Api project
4. See error in debug output (Debug Console tab in VS Code)
```
Elastic APM .NET Agent: [CRITICAL] Exception thrown by logging implementation. Log message: `The agent was started without a service name. The automatically discovered service name is {ServiceName}'. args.Length: 1. Current thread: managed ID: 1
Elastic APM .NET Agent: +-> Exception (exception): System.NullReferenceException: Object reference not set to an instance of an object.
Elastic APM .NET Agent:    at Serilog.Extensions.Logging.SerilogLogger.Log[TState](LogLevel logLevel, EventId eventId, TState state, Exception exception, Func`3 formatter)
   at Elastic.Apm.AspNetCore.AspNetCoreLogger.Log[TState](LogLevel level, TState state, Exception e, Func`3 formatter)
   at Elastic.Apm.Logging.LoggingExtensions.DoLog(IApmLogger logger, LogLevel level, String message, Exception e, Object[] args)e: <null>
```

**Expected behavior**
Either no error, or a more helpful description of what went wrong.

gregkalapos:
Hi @scotm-gxg,

thanks for the reproducer. Unfortunately it does not reproduce for me. I tried it on 2 machines, in both cases it worked, I don't see the log you pasted, the sample app worked fine (I tried the `http://localhost:5000/weatherforecast`) and the data is also visible in Kibana. Tested both with release and debug builds. Logs are at the end of this comment. 

Now to the error you see: what it says is that the logging library - in this case serilog - throws an exception when the agent tries to log. The `Log message: The agent was started...` you see is the log that we try to print when the `NullReferenceException` happens and this is the very first line the agent tries to log.

[...]

PR #545, opened 2019-10-09
Fix building log state - NullReferenceException with Serilog

gregkalapos:
Solves #544

As it seems when we parse the log state ourselves, everything is ok, but what we built in `LogValuesFormatter.GetState()` was not correct and this became a problem when we passed that state object to Serilog.

- The parameters were missing due to `<`, `>` mismatch. That caused items with `null` causing `NullReferenceException`s. 
- By adding `{` `}` to the scope, Serilog was unable to resolve the scope causing lines like this (so instead of the real scope `{{Scope}}` got printed):

```
{{Scope}} The agent was started without a service name. The service name will be automatically discovered.
```

### DIFF
--- a/src/Elastic.Apm/Logging/LogValuesFormatter.cs
+++ b/src/Elastic.Apm/Logging/LogValuesFormatter.cs
@@ -183,11 +183,11 @@ namespace Elastic.Apm.Logging
 			var args = new KeyValuePair<string, object>[values.Length + offset];
 			args[0] = new KeyValuePair<string, object>("{OriginalFormat}", OriginalFormat);
 			if (_scope != null)
-				args[1] = new KeyValuePair<string, object>("{Scope}", _scope);
+				args[1] = new KeyValuePair<string, object>("Scope", _scope);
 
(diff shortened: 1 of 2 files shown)


src/Elastic.Apm/Logging/LogValuesFormatter.cs
```diff
@@ -183,11 +183,11 @@ public LogValues GetState(object[] values)
 			var args = new KeyValuePair<string, object>[values.Length + offset];
 			args[0] = new KeyValuePair<string, object>("{OriginalFormat}", OriginalFormat);
 			if (_scope != null)
-				args[1] = new KeyValuePair<string, object>("{Scope}", _scope);
+				args[1] = new KeyValuePair<string, object>("Scope", _scope);
 
 			for (int i = 0, j = _scope != null ? 1 : 0; j < ValueNames.Count; i++, j++)
```
Mpdreamz: `j = offset - 1` and adding the condition to the for `j < ValueNames.Count && i < values.Length` reads slightly better to me. 

Unrelated to this PR though feel free to ignore 😄

Issue #300, opened 2019-06-18
Sanitize potential security sensitive info in HTTP headers, cookies and POST form fields


SergeyKleyman:
Sanitize (sanitise, scrub, redact, remove) potential security sensitive info in HTTP headers, cookies and POST form fields.

Implementations by other Elastic APM agents:

| Agent  | Configuration setting |
|---------|-------------------------|
| Java     | [`sanitize_field_names`](https://www.elastic.co/guide/en/apm/agent/java/current/config-core.html#config-sanitize-field-names)  |
| Go       | [`sanitize_field_names`](https://www.elastic.co/guide/en/apm/agent/go/current/configuration.html#config-sanitize-field-names) |
| node.js | [`filterHttpHeaders`](https://www.elastic.co/guide/en/apm/agent/nodejs/current/configuration.html#filter-http-headers) |
| Ruby    | [`custom_key_filters`](https://www.elastic.co/guide/en/apm/agent/ruby/current/configuration.html#config-custom-key-filters) |

gregkalapos:
In 1.0GA we don't implement it, since it can be turned off. We move this after 1.0GA.

gregkalapos:
Feedback and response, related to this: https://github.com/elastic/apm-agent-dotnet/issues/379#issuecomment-515235602

gregkalapos:
Implemented in #549.

PR #549, opened 2019-10-10
Add SanitizeFieldNames setting

gregkalapos:
Solves #300.

Introduces a new setting: `SanitizeFieldNames`. 

The [`WildcardMatcher`](https://github.com/elastic/apm-agent-java/blob/2c9cbfb02c28353384c5039444327c6a7b17cbbe/apm-agent-core/src/main/java/co/elastic/apm/agent/matcher/WildcardMatcher.java) which does the matching is coped (and C#ified) from the Java agent.

We do sanitization for 2 things:
 - `Request.Headers`
 - `Request.Body` if it's in `Key1=Value1&Key2=Value2...` format (typically contentype `application/x-www-form-urlencoded`)

All sanitization is done as part of the JSON serialization. The activation of the `JsonConverter`s ([here](https://github.com/elastic/apm-agent-dotnet/pull/549/files#diff-e221d69a7dd1034580084ff0635a75c7) and [here](https://github.com/elastic/apm-agent-dotnet/pull/549/files#diff-b2ad116536c2201f7bb4f4cb4166aac6)) responsible for the serialization is done within a [`ContractResolver`](https://github.com/elastic/apm-agent-dotnet/pull/549/files#diff-c5293f773a6dd162bd1c02319dbeb589) - huge thanks for @Mpdreamz  and @russcam for the ideas. 

I also ended up exposing the setting as `List<WildcardMatcher>` instead of `List<string>`. Below is a discussion that I original was not a fan of exposing `WildcardMatcher`, but a) it's also public in Java and b) it has a very clean and simple API and it's I think fairly nicely documented (stolen form Java ;) ) - this saves us some gymnastics later with central config.

Regarding `Request.Body`: It's not the nicest way to reconstruct the string in `BodyStringSanitizerConverter`. E.g. in ASP.NET Core we could use `HttpRequest.ReadFormAsync();` - I already have some ideas [here](https://github.com/gregkalapos/apm-agent-dotnet/blob/SanitizeFieldNamesTmp/src/Elastic.Apm.AspNetCore/Extensions/RequestExtentions.cs#L26). But that would change request body capturing too much, and this PR is already big, so in case needed I'd do that work outside this PR. The upside of the current solution is that it works everywhere - even if the users set the body manually.

Also added a whole bunch of tests - only thing I failed on is adding a Full Framework test - I was unable to make it work, I'll reiterate on it, but I'd also merge without that.


iquirino:
Nice PR.
This makes sense...

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
```diff
@@ -54,6 +54,22 @@ protected internal static bool TryParseLogLevel(string value, out LogLevel level
 			}
 		}
 
+		protected List<string> ParseSanitizeFieldNames(ConfigurationKeyValue kv)
+		{
+			if (kv == null || string.IsNullOrEmpty((kv.Value))) return DefaultValues.SanitizeFieldNames;
```
vhatsura: What if I don't want to use sanitization, e.g. during troubleshooting?

gregkalapos:
> Hey, @gregkalapos. It's the most awaitable feature which prevents using Elastic Apm in production. Nice to see that you're working on it, however I have a couple of questions about implementation.
> 
> 1. Why we cannot construct list of wildcard matchers inside configuration readers? It will be guaranteed that matchers are constructed only once and will be synchronized between all components inside agent. Also, it will simplify integration `SanitizeFieldNames` with central configuration.
> 2. What about integrate sanitization on serialization level? It allows to have only single place with sanitization logic (isolability), which can be used in .Net Core and .Net Framework (as I understand .Net Framework out of scope in this PR). In addition, maintainability of such code will be easier.

1. Yeah, it'd make things way easier. The only problem I see here is that currently on `IConfigurationReader` we have very simple types... every method returns `int`, `string`, `List<string>` and things like that - maybe `LogLevel` is an exception, but it's still a simple `enum`. If we return `List<WildcardMatcher>` from the config reader then we have to make `WildcardMatcher` also public and in reality the setting itself is `List<string>`, `WildcardMatcher` to me feels like an implementation detail. But as you correctly say this will make it harder to integrate with central config, that is definitely a valid point... as you can see some of my comments in the code I definitely don't like this aspect of the current status of this PR, on the other hand that is a future problem and no matter if we make it live changeable now or later, this setting is at the moment not configurable through central config. I'm still a little bit undecided on this one.

2. That is a good idea! We can easily do that for the headers. The body is a little bit more problematic. First of all we won't have the content type during serialization. We can just try to split the body on `&`s and then on `=` (so we'd basically reconstruct the `application/x-www-form-urlencoded` form) and then do the sanitization on that. But this is a little but unfortunate since ASP.NET Core already gives us the `IFormsCollection` with `HttpRequest.ReadFormAsync()`, so we'd basically do the parsing a second time just to sanitize the data. But for this we could introduce an internal field like `isSanitize` and in case of ASP.NET Core do the sanitization with  `HttpRequest.ReadFormAsync()` and set this flag, so we'd know that we can skip sanitization during serialization.

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
gregkalapos: Hmm... 1 option I see here is that we could allow `""`, so we'd only check for `kv == null` and in the `appsettings.json` you could add an empty string with `"SanitizeFieldNames": ""`.

gregkalapos:
> as I understand .Net Framework out of scope in this PR

I want to have this setting for both, it's just that the PR becomes better, so I thought it'd make sense to address Full Framework in an other PR to make reviews easier. If we go with sanitization during serialization then probably it'll be just 1 PR with some additional Full Framework tests here.

SergeyKleyman:
> If we return `List<WildcardMatcher>` from the config reader then we have to make `WildcardMatcher` also public and in reality the setting itself is `List<string>`, `WildcardMatcher` to me feels like an implementation detail.

A possible approach to address this concern is to deprecate `IConfigurationReader` and instead have agent's public API accept a simple interface with read-only `string` to `string` indexer as a custom configuration source. This change will also address https://github.com/elastic/apm-agent-dotnet/issues/553.

SergeyKleyman:
In addition if we want to make this more future proof and allow users to supply changeable custom configuration we can add an event to _raw-configuration_ interface.

src/Elastic.Apm/Helpers/WildcardMatcher.cs
```diff
@@ -0,0 +1,284 @@
...
+		/// Constructs a new <see cref="WildcardMatcher"/> via a wildcard string.
+		/// By default, matches are a case insensitive.
+		/// It supports the <code>*</code> wildcard which matches zero or more characters.
+		/// Prepend <code>(?-i)</code> to your pattern to make it case sensitive.
+		///
+		/// Example: <code>(?-i)foo*</code> matches the string <code>foobar</code> but does not match <code>FOOBAR</code>.
+		/// </summary>
+		/// <param name="wildcardString"></param>
+		/// <returns></returns>
+		public static WildcardMatcher ValueOf(string wildcardString)
```
Mpdreamz: I think a short circuit could be introduced for exact match cases that can defer to builtin string compare operations (case sensitive or not). e.g a third matcher `VerbatimMatcher`.

src/Elastic.Apm.AspNetCore/Config/MicrosoftExtensionsConfig.cs
```diff
@@ -46,7 +46,7 @@ public override LogLevel LogLevel
 		protected override ConfigurationKeyValue Read(string key, string fallBackEnvVarName)
 		{
 			var value = _configuration[key];
-			if (!string.IsNullOrWhiteSpace(value)) return Kv(key, value, Origin);
+			if (value != null) return Kv(key, value, Origin);
```
gregkalapos: With this we make a difference between configs with value `null` (these were not set) and configs with `""`(explicitly set to empty string). With this by setting `"SanitizeFieldNames": ""` you can turn off this feature.

src/Elastic.Apm/Helpers/WildcardMatcher.cs
gregkalapos: You mean something like [this](https://github.com/elastic/apm-agent-dotnet/pull/549/commits/20140b503584c3138ea0e210f7e2675baee2ef2b#diff-3d54c6ce902fd9ab1b0ccdd12bb278a9R219)? 

I ran some benchmarks (that is also in the PR) - 1000 long string, `MatcherVerbatim*` use the new `VerbatimWildcardMatcher`, `MatcherSimple*` use the `SimpleWildcardMatcher`:

```
|                         Method |            Mean |          Error |          StdDev |          Median | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|------------------------------- |----------------:|---------------:|----------------:|----------------:|------------:|------------:|------------:|--------------------:|
|   MatcherVerbatimCaseSensitive |        13.36 ns |      0.1536 ns |       0.1282 ns |        13.34 ns |      0.0051 |           - |           - |                32 B |
| MatcherVerbatimCaseInsensitive |        13.30 ns |      0.1715 ns |       0.1605 ns |        13.29 ns |      0.0051 |           - |           - |                32 B |
|     MatcherSimpleCaseSensitive | 2,992,389.11 ns | 59,774.9240 ns | 103,108.8070 ns | 2,941,651.39 ns |      7.8125 |           - |           - |             64032 B |
|   MatcherSimpleCaseInsensitive | 2,896,890.25 ns | 33,103.8968 ns |  29,345.7422 ns | 2,891,317.69 ns |      7.8125 |           - |           - |             64032 B |
```

Based on the numbers: yeah, this probably makes some sense 🤣 .

Mpdreamz: Sweet! I wonder if we can do better on the Simple/Composite matcher paths too. I am happy for that to happen on a follow up PR though

src/Elastic.Apm.AspNetCore/Config/MicrosoftExtensionsConfig.cs
Mpdreamz: How do we feel about an explicit `__empty__` key? Where `null` remains to fallback to the default values?

SergeyKleyman: Is using explicit `__empty__` key aligned with the rest of the agents?

gregkalapos: > How do we feel about an explicit __empty__ key? 

I'm not sure I totally understand the idea here. The key itself is the key of the config, like `ElasticApm:SanitizeFieldNames`. Or do you mean `__empty__` as the value for marking something as empty - in this case turning off the sanitization and not using the default list? Sorry if I misunderstood.

Mpdreamz: The latter as a marker value for setting it to empty, `""` seems like a weird way to reset. EIther that or another option that toggles (with the default being `true`) seem better to me. No clue what other agents do here :)

gregkalapos: Ok, I see. 

Yeah, `""` is indeed not the best, I agree on that.

> No clue what other agents do here :)

Me neither, but I'm trying to find out.

docs/configuration.asciidoc
```diff
@@ -262,6 +262,40 @@ If set to `true`, the agent makes periodic requests to the APM Server to fetch t
 | true    | Boolean 
 |============
 
+[float]
+[[config-sanitize-field-names]]
+==== `SanitizeFieldNames` (added[1.2])
+
+Sometimes it is necessary to sanitize the data sent to Elastic APM, e.g. remove sensitive data.
```
bmorelli25: e.g. --> for example (**egg** sample)
i.e. --> in other words (**i**n **e**ssence)

Is removing sensitive data an example of sanitizing data? Or is it the same thing? If it's the same thing, which I think it is, this should be an i.e. instead.
```suggestion
Sometimes it is necessary to sanitize, i.e., remove, sensitive data sent to Elastic APM.
```

docs/configuration.asciidoc
```diff
@@ -262,6 +262,40 @@ If set to `true`, the agent makes periodic requests to the APM Server to fetch t
...
+
+This option supports the wildcard `*`, which matches zero or more characters.
+Examples: `/foo/*/bar/*/baz*`, `*foo*`.
+Matching is case insensitive by default.
+
+Prepending an element with `(?-i)` makes the matching case sensitive.
+
+NOTE: Data in the query string is considered non-sensitive,
+as sensitive information should not be sent in the query string.
+See https://www.owasp.org/index.php/Information_exposure_through_query_strings_in_url for more information
```
```suggestion
See https://www.owasp.org/index.php/Information_exposure_through_query_strings_in_url[owasp.org] for more information.
```

docs/configuration.asciidoc
```diff
@@ -262,6 +262,40 @@ If set to `true`, the agent makes periodic requests to the APM Server to fetch t
...
+
+NOTE: Review the data captured by Elastic APM carefully to make sure it does not capture sensitive information.
+If you do find sensitive data in the Elasticsearch index,
+you should add an additional entry to this list (make sure to also include the default entries).
```
```suggestion
you should add to the default entries on this list.
```

docs/configuration.asciidoc
gregkalapos: Yeah, agreed, it's the same.

docs/configuration.asciidoc
gregkalapos: Hmm, this is interesting, if you don’t mind I’d discuss this a little bit.

So if you don’t configure anything, we have a default list of keys and based on those we’ll sanitize the HTTP headers and the body. If you still see some sensitive data, then you should use the setting. 

But the thing is, as soon as you use this setting the defaults will go away. So not configuring `SanitizeFieldNames` gives you a default list (list is above in this doc), but then adding `”SanitizeFieldNames”: “foo”` will only have `”foo`”, the defaults are gone - we'll only sanitize the `foo` HTTP header and form body values with `foo`. So you can’t really “add to the defaults”. 

I think having the `(make sure to also include the default entries)` makes this clear. Does this makes sense?

Btw. all this is copied from Java (or maybe the Go doc is mixed into it at some point) - so none of this text is really written by me.

gregkalapos: > Is this typical, i.e., will a developer expect this? Now I'm concerned that the docs don't really explain that the default goes away if you configure SanitizeFieldNames.

That is a very good point. I'm not sure that everyone expects this behaviour. We can add some more text to explain this, I think it'd make the docs better.

bmorelli25: 👍 Working on it

docs/configuration.asciidoc
```diff
@@ -262,6 +262,40 @@ If set to `true`, the agent makes periodic requests to the APM Server to fetch t
...
+
+Prepending an element with `(?-i)` makes the matching case sensitive.
+
+NOTE: Data in the query string is considered non-sensitive,
+as sensitive information should not be sent in the query string.
+See https://www.owasp.org/index.php/Information_exposure_through_query_strings_in_url[owasp.org] for more information.
+
+NOTE: Review the data captured by Elastic APM carefully to make sure it does not capture sensitive information.
+If you do find sensitive data in the Elasticsearch index,
+you should add an additional entry to this list (make sure to also include the default entries).
```
bmorelli25: Can I get a little wild? What about something like this?

```asciidoc
Sometimes it is necessary to sanitize, i.e., remove, sensitive data sent to Elastic APM.
This config accepts a list of wildcard patterns of field names which should be sanitized.
These apply for example to HTTP headers and `application/x-www-form-urlencoded` data.

The wildcard, `*`, matches zero or more characters, and matching is case insensitive by default.
Prepending an element with `(?-i)` makes the matching case sensitive.
Examples: `/foo/*/bar/*/baz*`, `*foo*`.

You should review the data captured by Elastic APM carefully to make sure it does not contain sensitive information.
If you do find sensitive data in your Elasticsearch index,
you should add an additional entry to this list.
Make sure to include the default entries as well, as setting a value here will overwrite the defaults. 

NOTE: Data in the query string is considered non-sensitive,
as sensitive information should not be sent in the query string.
See https://www.owasp.org/index.php/Information_exposure_through_query_strings_in_url[owasp.org] for more information.
```

You don't have to use all of this. The important changes are in paragraph three. I removed the admonition paragraph (NOTE) and clarified that the defaults get overwritten.

gregkalapos: I like it, it captures all the details nicely. I'm fine with taking this. 👍

gregkalapos: I like it, it captures all the details nicely. I'm fine with taking this. 👍

SergeyKleyman:
Regarding performance implications of this feature - current implementation tries to match the input string against the configured patterns one by one, right? Maybe it's worth considering using regular expressions? For example default configuration will result in the following regular expression 
```
password|passwd|pwd|secret|.*key|.*token.*|.*session.*|.*credit.*|.*card.*|authorization|set-cookie
```
Using regular expression has a potential performance advantage - it reads input string only once and matches it against all patterns in one pass. Of course, only benchmarks will show how the current implementation compares to the regular expression approach (for example on the default configuration patterns).

gregkalapos:
> #568 takes another approach (sanitize fields ASAP) while this PR defers sanitization until serialization.
> #568 approach is justified by having an advantage for logging (logs won't contain sensitive data). Wouldn't the same reasoning be applicable here as well? Or is the logging not an issue here?

I discussed this on slack with other agent devs, all of them do this during or before serialization, that is why I went with this solution. [This comment](https://github.com/elastic/apm-agent-dotnet/pull/549#issuecomment-541465077) also has a longer rant about this.

Some agents e.g. Go already do the sanitization on a background thread (...or goroutine). Long term we could also move to that approach.

gregkalapos:
> Regarding performance implications of this feature - current implementation tries to match the input string against the configured patterns one by one, right? if there are a lot of configured patterns the performance overhead might add up. Maybe it's worth considering using regular expressions? For example default configuration will result in the following regular expression
> 
> ```
> password|passwd|pwd|secret|.*key|.*token.*|.*session.*|.*credit.*|.*card.*|authorization|set-cookie
> ```
> 
> Using regular expression has a potential performance advantage - it reads input string only once and matches it against all patterns in one pass. Of course, only benchmarks will show how the current implementation compares to the regular expression approach (for example on the default configuration patterns).

[This file](https://github.com/elastic/apm-agent-java/blob/2c9cbfb02c28353384c5039444327c6a7b17cbbe/apm-agent-core/src/main/java/co/elastic/apm/agent/matcher/WildcardMatcher.java#L34) has some relevant comment. 

From what I saw, there was already some effort to unify this, [this is the PR](https://github.com/elastic/apm-agent-go/pull/261) in the Go agent turning regex implementation into using the same approach that we and Java does. 

> Of course, only benchmarks will show how the current implementation compares to the regular expression approach

I strongly agree with this.

So, given that other agents already follow the same approach we do, I'd suggest we also do that and we can reiterate and measure this.

docs/configuration.asciidoc
gregkalapos: Pushed it.

PR #562, opened 2019-10-17
Fix failing master

gregkalapos:
Currently `master` fails during the `Release to AppVeyor` stage. 

Logs:
```
[2019-10-16T17:46:01.383Z] + .ci/linux/release.sh
//..more logs
[2019-10-16T17:46:12.412Z] /var/lib/jenkins/workspace/tnet_apm-agent-dotnet-mbp_master/apm-agent-dotnet/sample/AspNetFullFrameworkSampleApp/AspNetFullFrameworkSampleApp.csproj : error MSB4057: The target "pack" does not exist in the project.
```

`AspNetFullFrameworkSampleApp.csproj` is a project which we can't `dotnet pack` plus it's a sample project (so not an agent project) that we don't want to publish anyway. So I excluded it.

Tested in 2f41af9 - that was green, so I assume `master` with this stage of the PR will be also green.

### DIFF
--- a/.ci/linux/release.sh
+++ b/.ci/linux/release.sh
@@ -5,9 +5,8 @@
 #
 set -euxo pipefail
 
-# Remove sample projects
+# Remove sample projects - and other we don't want to pack
 dotnet sln remove test/Elastic.Apm.PerfTests/Elastic.Apm.PerfTests.csproj
-dotnet sln remove src/Elastic.Apm.AspNetFullFramework/Elastic.Apm.AspNetFullFramework.csproj
 dotnet sln remove test/Elastic.Apm.AspNetFullFramework.Tests/Elastic.Apm.AspNetFullFramework.Tests.csproj
-
+dotnet sln remove sample/AspNetFullFrameworkSampleApp/AspNetFullFrameworkSampleApp.csproj
 dotnet pack -c Release
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -308,7 +308,6 @@ pipeline {
             when {
               beforeAgent true
               anyOf {
-                branch 'master'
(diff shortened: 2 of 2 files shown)


Mpdreamz:
For sdk style projects we can set the following to false:

https://docs.microsoft.com/en-us/dotnet/core/tools/csproj#ispackable

gregkalapos:
> For sdk style projects we can set the following to false:
> 
> https://docs.microsoft.com/en-us/dotnet/core/tools/csproj#ispackable

Good point, but I think we already use that where we can. E.g. [AspNetFullFramework.Tests](https://github.com/elastic/apm-agent-dotnet/blob/master/test/Elastic.Apm.AspNetFullFramework.Tests/Elastic.Apm.AspNetFullFramework.Tests.csproj#L4), [SampleAspNetCoreApp.csproj](https://github.com/elastic/apm-agent-dotnet/blob/master/sample/SampleAspNetCoreApp/SampleAspNetCoreApp.csproj#L8), [Elastic.Apm.Tests](https://github.com/elastic/apm-agent-dotnet/blob/master/test/Elastic.Apm.Tests/Elastic.Apm.Tests.csproj#L5).

This one is our full framework sample app - and that is [the old format](https://github.com/elastic/apm-agent-dotnet/blob/master/sample/AspNetFullFrameworkSampleApp/AspNetFullFrameworkSampleApp.csproj), so in this case we can't do other than `dotnet sln remove`, right? ...on the other hand I don't know why this did not fail earlier.

gregkalapos:
Merging now, since we have failing `master` and with high probability this will make it green. - happy to discuss remaining things and open follow ups if needed.

PR #565, opened 2019-10-17
Efcore/commandfailed

### LINKED ISSUE #400: Entity Framework Core's CommandError event is not handled
According to [this test](https://github.com/aspnet/EntityFrameworkCore/blob/master/test/EFCore.Relational.Tests/Storage/RelationalCommandTest.cs#L1149)  `CommandError` event is sent to the listeners instead of `CommandExecuted` which would mean we never `End()` that DB spans and it also causes a memory leak because the span stays in `Elastic.Apm.EntityFrameworkCore.EfCoreDiagnosticListener._spans`. I've opened a separate issue (https://github.com/elastic/apm-agent-dotnet/issues/387) to handle the latter problem (agent having collections that can grow indefinitely) in systematic cross-agent manner.

vhatsura:
closes #400 

should be reviewed after #560 merge


gregkalapos:
Nice @vhatsura! Would it be possible to have some test covering this case?

vhatsura:
I tried to write test for this case, however it was very hard to force EfCore to fail command (I want to have test with real scenario). I will look again to it

gregkalapos:
> I tried to write test for this case, however it was very hard to force EfCore to fail command (I want to have test with real scenario). I will look again to it

I haven't tried this (so take this with a grain of salt), but maybe 1 option would be to manually call `OnNext(KeyValuePair<string, object> kv)` and manually construct the parameter. Forcing EF Core to send the newly handled event would be of course better, but at least we could make sure the code always does what we expect it to do in with that.

In some [`HttpClient` tests](https://github.com/elastic/apm-agent-dotnet/blob/master/test/Elastic.Apm.Tests/HttpDiagnosticListenerTests.cs#L98) we already do this.

SergeyKleyman:
> I tried to write test for this case, however it was very hard to force EfCore to fail command (I want to have test with real scenario). I will look again to it

#400 has a link to [the test](https://github.com/aspnet/EntityFrameworkCore/blob/master/test/EFCore.Relational.Tests/Storage/RelationalCommandTest.cs#L1149) in EntityFrameworkCore repo that (I assume) simulates a scenario causing `CommandError` event to be sent to the listeners.

vhatsura:
@SergeyKleyman, @gregkalapos, I implemented tests with using in-memory database. PR is ready for review.

test/Elastic.Apm.EntityFrameworkCore.Tests/EfCoreDiagnosticListenerTests.cs
```diff
@@ -0,0 +1,107 @@
...
+					await _dbContext.Database.ExecuteSqlCommandAsync("SELECT * FROM FakeTable");
+				}
+				catch
+				{
+					// ignore
+				}
+			});
+
+			// Assert
+			_payloadSender.Spans.Count.Should().Be(1);
```
gregkalapos: I think we are totally fine now, maybe long term it'd be also nice to have some tests that assert on the properties of `_payloadSender.FirstSpan.Context.Db`. I think we have some tests doing so in one of the ASP.NET Core tests, but probably this new project would be a good place to have some basic tests on the db fields.

test/Elastic.Apm.EntityFrameworkCore.Tests/Elastic.Apm.EntityFrameworkCore.Tests.csproj
```diff
@@ -0,0 +1,31 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <IsPackable>false</IsPackable>
+    <RootNamespace>Elastic.Apm.EntityFrameworkCore.Tests</RootNamespace>
+    <AssemblyName>Elastic.Apm.EntityFrameworkCore.Tests</AssemblyName>
```
gregkalapos: Nice, we should have had this project a long time ago... :)

Issue #515, opened 2019-09-27
Agent's log should not contain potentially security sensitive info


SergeyKleyman:
Example from https://discuss.elastic.co/t/apm-server-error-connection-refused/200109/3:
```
{{Scope}} Failed capturing request (failed to remove from ProcessingRequests) - This Span will be skipped in case it wasn't captured before. Request: method: "POST", URL: http://elastic:PleaseChangeMe@localhost:9200/_bulk
```

Agent logs URL including username and password: http://**elastic:PleaseChangeMe**@localhost:9200/_bulk

gregkalapos:
This is kind of a duplicate of #202. Opened #568, which addresses also this problem.

gregkalapos:
The specific log from the description is addressed in #568.

I personally would be fine with this - given this is planned for 7.5 and - and just in general during reviews make sure we don't log sensitive data - at least data that is not directly entered into some config by our users.

There was a comment here to maybe do more https://github.com/elastic/apm-agent-dotnet/pull/568#discussion_r339418480

I leave this open for discussion for a while and will bring this up during our next planning to see what others think.

gregkalapos:
This was fixed in #568.

This was discussed in an agent weekly and no other agent does some magic filtering or so. It was agreed that on a default log level we should not log anything security sensitive, but if the log level is changed to a more detailed level it's fine to log security critical things, as this is something that the user actively triggers.

Closing this now, we can do another iteration later if we feel so.

Issue #202, opened 2019-04-25
Sanitise userinfo in URLs


SergeyKleyman:
Derived from https://github.com/elastic/apm-dev/issues/538

We need to sanitize URLs in

- Incoming requests (`transaction.context`)
- Outgoing requests (`span.context`)

and we need to it for

- Full .NET Framework
- .NET Core

gregkalapos:
I move this into the milestone `7.5`, since we moved #515 into `7.5` during planning, and I think we overlooked this one. This is very related to #515, it's almost like a duplicate.

gregkalapos:
Implemented in https://github.com/elastic/apm-agent-dotnet/pull/568.

PR #568, opened 2019-10-18
Sanitize URLs that contain username and password

gregkalapos:
Solves #202 and #515. 

- For outgoing HTTP requests we make sure that URLs in `http://[username]:[password]@example.com/` format are captured with sanitization - we turn those into `http://[REDACTED]:[REDACTED]@example.com/`
- We also make sure these don't end up in logs.

- For incoming requests we do the same for `Request.Url.Full` and `Request.Url.Raw`. 

## General idea: 
We store these things as `string`s, which for this problem isn't the best. I think, in general it'd be overkill to manually parse the string. Therefore the idea is to use `Uri` and `UriBuilder`, which helps us to check if user info data is included in the uri or not. So we turn our `string`s to `Uri`s and do the check&sanitization on those.

## The "`uri` -> `string` -> `uri` -> `string`" problem

In case there is nothing to sanitize, we basically allocate a `Uri` object just for the checking. On `Elastic.Apm.Api.Http` I added a `SetUrl(Uri uri)` method to avoid this - so that's good, but we need to keep our existing `string` properties, since those are public on the agent API. 

For `Elastic.Apm.Api.Request.Url` I was unable to add methods that directly take a `Uri` and set `Raw`and `Full` with those, because those are built from multiple types, so there isn't really a `Uri` object in those cases - but to be fair these are 2 objects per incoming request, which sounds ok for now.


src/Elastic.Apm/Api/Http.cs
```diff
@@ -9,12 +10,108 @@ namespace Elastic.Apm.Api
...
+		private string _url;
+
 		[JsonConverter(typeof(TrimmedStringJsonConverter))]
 		public string Method { get; set; }
 
 		[JsonProperty("status_code")]
 		public int StatusCode { get; set; }
 
-		public string Url { get; set; }
+		public string Url
```
axw: Maybe add docs to this as well, noting that the setter will parse and sanitise the value - and that SetUrl should be preferred if the overhead is unacceptable?

src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerImplBase.cs
```diff
@@ -97,7 +99,7 @@ public void OnNext(KeyValuePair<string, object> kv)
 
 			if (IsRequestFilteredOut(requestUrl))
 			{
-				Logger.Trace()?.Log("Request URL ({RequestUrl}) is filtered out - exiting", requestUrl);
+				Logger.Trace()?.Log("Request URL ({RequestUrl}) is filtered out - exiting", Http.Sanitize(requestUrl));
```
axw: `Http.Sanitize` is going to be called regardless of whether trace logging is enabled, right? That's going to add some overhead. Maybe worth adding some kind of lazy evaluation for logging?

test/Elastic.Apm.Tests/HttpDiagnosticListenerTests.cs
```diff
@@ -490,6 +523,40 @@ public async Task HttpCallWithoutRegisteredListener()
 			}
 		}
 
+		/// <summary>
+		/// Makes sure that in case of outgoing HTTP requests with a URL containing username and password those
+		/// are not showing up in the agent logs.
+		/// </summary>
+		[Fact]
+		public async Task NoUserNameAndPasswordInLogsForHttp()
```
axw: this is nice :)

src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerImplBase.cs
axw: @SergeyKleyman clued me in - I thought the arguments would still be evaluated, it just wouldn't call Log with them, but that is not the case.

src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerImplBase.cs
```diff
@@ -58,7 +58,9 @@ protected HttpDiagnosticListenerImplBase(IApmAgent agent)
 
 		public void OnNext(KeyValuePair<string, object> kv)
 		{
-			Logger.Trace()?.Log("Called with key: `{DiagnosticEventKey}', value: `{DiagnosticEventValue}'", kv.Key, kv.Value);
+			// We only print the key, we don't print the value, because it can contain a http request object, and its ToString prints
+			// the URL, which can contain username and password. See: https://github.com/elastic/apm-agent-dotnet/issues/515
+			Logger.Trace()?.Log("Called with key: `{DiagnosticEventKey}'", kv.Key);
```
SergeyKleyman: Maybe it's worth handling #515 in a separate PR? I don't think just not logging information that might be very helpful to us in support cases is the solution. Otherwise we will have to remove a lot of other useful data from logging. For example if I set `SampleAspNetCoreApp` configuration to:
```
{
  "Logging": {
    "LogLevel": {
      "Elastic.Apm": "Trace"
    }
  },
  "ElasticApm": {
    "ServerUrls": "http://my_username:my_password@localhost:8200",
  }
}
```

The logs will contain:
```
{BackendCommUtils} Building HTTP client with BaseAddress: http://my_username:my_password@localhost:8200/ for BackendCommComponentBase (PayloadSenderV2)...
```

```
{PayloadSenderV2} Using the following configuration options: Events intake API absolute URL: http://my_username:my_password@localhost:8200/intake/v2/events, FlushInterval: 10s, MaxBatchEventCount: 10, MaxQueueEventCount: 1000
```

```
{CentralConfigFetcher} Making HTTP request to APM Server... Request: Method: GET, RequestUri: 'http://my_username:my_password@localhost:8200/config/v1/agents?service.name=SampleAspNetCoreApp&service.environment=Production', Version: 2.0, Content: <null>, Headers:
{
}.
```

In a separate PR a possible approach might be to filter logged data (message text and contextual data) for patterns such as credentials in URL and maybe even using patterns configured via `SanitizeFieldNames`.

gregkalapos:
> looks good ~, apart from the fact that Http.Sanitize will be called for logging regardless of whether it's enabled (unless there's some C# magic I'm unaware of?)~
> 
> EDIT: there was C# magic that I was unaware of :)

Yes, that was a ting that we spent some time thinking about :) This PR (in the description) has all the details: https://github.com/elastic/apm-agent-dotnet/pull/146.

src/Elastic.Apm/Api/Http.cs
gregkalapos: 👍  good idea, I added docs to that property.

src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerImplBase.cs
gregkalapos: Sure, we can follow up on this. I can leave #515 open and sync during planning on what we would like to do.

I think we addressed the main issue here, but if something is missing, or some more thinking is needed we can definitely follow up on that.

Issue #572, opened 2019-10-23
[meta][docs] Documentation restructure


bmorelli25:
A meta issue for tracking documentation improvements outlined in https://github.com/elastic/apm-dev/issues/425.

### Doc restructure
- [x] master - https://github.com/elastic/apm-agent-dotnet/pull/574
- [x] 1.x backport - https://github.com/elastic/apm-agent-dotnet/pull/578

### Build release notes to docs: 
More details: https://github.com/elastic/apm/issues/132.
The PRs in each group must be merged at the same time to prevent breaking the doc build.

#### `master`
- [x] apm-agent-dotnet - https://github.com/elastic/apm-agent-dotnet/pull/575
- [x] docs - https://github.com/elastic/docs/pull/1365

#### `1.x`
- [x] apm-agent-dotnet pr - https://github.com/elastic/apm-agent-dotnet/pull/576
- [x] docs pr - https://github.com/elastic/docs/pull/1366

bmorelli25:
All done!

PR #574, opened 2019-10-24
[docs] Documentation restructure

bmorelli25:
For https://github.com/elastic/apm-agent-dotnet/issues/572. A live preview is available **[here](http://apm-agent-dotnet_574.docs-preview.app.elstc.co/guide/en/apm/agent/dotnet/master/index.html)**.
- [x] ~Top level heading and doc sync for log correlation: https://github.com/elastic/apm/issues/149~ Added in a comment for where this heading will go
- [x] Use shared version attributes: https://github.com/elastic/apm/issues/150
- [x] Add a section on how the Agent does instrumentation: https://github.com/elastic/apm/issues/146
- [x] Rename the APM UI to "APM app": https://github.com/elastic/apm-dev/issues/567
- [x] Use `apm-app-ref`: https://github.com/elastic/docs/pull/1276
- [x] Updates to the new documentation structure outlined [here](https://docs.google.com/spreadsheets/d/1P53LbNY7jFSLLJhOKw-nmDenQbrIFlY4EEFJGqT1kmY/edit#gid=0)
- [x] ~Sync Central config docs: https://github.com/elastic/apm-agent-dotnet/pull/497#discussion_r326303325~ Already done

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -134,7 +134,7 @@ Below is a sample `Web.config` configuration file for a ASP.NET application.
 This is used to keep all the errors and transactions of your service together
 and is the primary filter in the Elastic APM user interface.
 
-NOTE: The service name must conform to this regular expression: `^[a-zA-Z0-9_-]+$`. In less regexy terms: Your service name must only contain characters from the ASCII alphabet, numbers, dashes, underscores and spaces. Characters in service name which don't match regular expression will be replaced by `__` symbol.
+NOTE: The service name must conform to this regular expression: `^[a-zA-Z0-9 _-]+$`. In less regexy terms: Your service name must only contain characters from the ASCII alphabet, numbers, dashes, underscores and spaces. Characters in service name which don't match regular expression will be replaced by `_` symbol.
 
 [options="header"]
 |============
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -134,7 +134,7 @@ Below is a sample `Web.config` configuration file for a ASP.NET application.
 This is used to keep all the errors and transactions of your service together
 and is the primary filter in the Elastic APM user interface.
 
-NOTE: The service name must conform to this regular expression: ^[a-zA-Z0-9 _-]+$. In less regexy terms: Your service name must only contain characters from the ASCII alphabet, numbers, dashes, underscores and spaces. Characters in service name which don't match regular expression will be replaced by `__` symbol.
+NOTE: The service name must conform to this regular expression: `^[a-zA-Z0-9_-]+$`. In less regexy terms: Your service name must only contain characters from the ASCII alphabet, numbers, dashes, underscores and spaces. Characters in service name which don't match regular expression will be replaced by `__` symbol.
 
 [options="header"]
 |============
@@ -174,11 +174,11 @@ of the deployed revision, e.g. the output of git rev-parse HEAD.
 
 The name of the environment this service is deployed in, e.g. "production" or "staging".
 
-Environments allow you to easily filter data on a global level in the APM UI.
+Environments allow you to easily filter data on a global level in the APM app.
 It's important to be consistent when naming environments across agents.
-See {kibana-ref}/filters.html#environment-selector[environment selector] in the Kibana UI for more information.
+See {apm-app-ref}/filters.html#environment-selector[environment selector] in the Kibana UI for more information.
 
-NOTE: This feature is fully supported in the APM UI in Kibana versions >= 7.2.
+NOTE: This feature is fully supported in the APM app in Kibana versions >= 7.2.
 You must use the query bar to filter for a specific environment in versions prior to 7.2.
 
 [options="header"]
@@ -248,7 +248,7 @@ The Agent will revert to the default value if the value is set below `-1`.
 [[config-central-config]]
 ==== `CentralConfig` (added[1.1])
 
-If set to `true`, the agent makes periodic requests to the APM Server to fetch the latest {kibana-ref}/agent-configuration.html[APM Agent configuration].
+If set to `true`, the agent makes periodic requests to the APM Server to fetch the latest {apm-app-ref}/agent-configuration.html[APM Agent configuration].
 
 [options="header"]
 |============
--- a/docs/intro.asciidoc
+++ b/docs/intro.asciidoc
@@ -18,7 +18,7 @@ The agent auto-instruments <<supported-technologies,supported technologies>> and
 like HTTP requests and database queries. To do this, it uses built in capabilities of the instrumented frameworks like [Diagnostic Source](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.diagnosticsource?view=netcore-3.0), an HTTP module for IIS, or [IDbCommandInterceptor](https://docs.microsoft.com/en-us/dotnet/api/system.data.entity.infrastructure.interception.idbcommandinterceptor?view=entity-framework-6.2.0) for Entity Framework.
 This means that for the supported technologies, there are no code changes required.
 
-The Agent automatically {CONTENT_HERE}
+The Agent automatically registers callback methods for these built in Diagnostic Source events - or in case of the IIS HTTP module as described on the <<setup,Set up the Agent page>>, the user needs to register the module. With this, the supported frameworks trigger agent code for relevant events to measures their duration and collects metadata (like DB statements), as well as HTTP related information (like the URL, parameters, and headers).
 
 These events, called Transactions and Spans, are sent to the APM Server.
 The APM Server converts them to a format suitable for Elasticsearch, and sends them to an Elasticsearch cluster.
--- a/docs/intro.asciidoc
+++ b/docs/intro.asciidoc
@@ -6,17 +6,23 @@ endif::[]
 [[intro]]
 == Introduction
 
-Welcome to the APM .NET Agent documentation.
+The Elastic APM .NET Agent automatically measures the performance of your application and tracks errors.
+It has built-in support for the most popular frameworks and routers,
+as well as a simple API which allows you to instrument any application.
 
-Elastic APM automatically measures the performance of your application and tracks errors.
-For example, it records spans for database queries and transactions for incoming HTTP requests.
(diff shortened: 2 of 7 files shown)


docs/configuration.asciidoc
```diff
@@ -134,7 +134,7 @@ Below is a sample `Web.config` configuration file for a ASP.NET application.
 This is used to keep all the errors and transactions of your service together
 and is the primary filter in the Elastic APM user interface.
 
-NOTE: The service name must conform to this regular expression: ^[a-zA-Z0-9 _-]+$. In less regexy terms: Your service name must only contain characters from the ASCII alphabet, numbers, dashes, underscores and spaces. Characters in service name which don't match regular expression will be replaced by `__` symbol.
+NOTE: The service name must conform to this regular expression: `^[a-zA-Z0-9_-]+$`. In less regexy terms: Your service name must only contain characters from the ASCII alphabet, numbers, dashes, underscores and spaces. Characters in service name which don't match regular expression will be replaced by `__` symbol.
```
bmorelli25: Is this change ok? I removed the space between the `9` and the `_` because the formatting goes wild if there's a space. I also put this in a code snippet for easier readability.

docs/intro.asciidoc
```diff
@@ -6,17 +6,23 @@ endif::[]
...
 
-Spans are grouped in transactions - by default one for each incoming HTTP request.
-But it's possible to create custom transactions not associated with an HTTP request.
+The Agent automatically {CONTENT_HERE}
 
-By default the agent comes with support for ASP.NET Core out-of-the-box.
-To instrument other events, you can use custom spans.
+These events, called Transactions and Spans, are sent to the APM Server.
+The APM Server converts them to a format suitable for Elasticsearch, and sends them to an Elasticsearch cluster.
+You can then use the APM app in Kibana to gain insight into latency issues and error culprits within your application.
```
bmorelli25: One of the things this PR aims to do is add in a section on how the agent does instrumentation. I've added some boilerplate text here to hopefully assist with this. There are two {CONTENT_HERE} sections above that I'll need one of you to fill out. Feel free to write something different or amend this as necessary.

I've also updated elastic/apm#146 with links to examples in other Agent's docs.

docs/setup.asciidoc
```diff
@@ -80,7 +103,7 @@ You can also configure the agent using `web.config` as described at <<configurat
 
 [float]
 [[setup-ef6]]
-=== Entity Framework 6
+==== Entity Framework 6
 
 You can enable auto instrumentation for Entity Framework 6 by referencing the `Elastic.Apm.EntityFramework6` package
 and including the `Ef6Interceptor` interceptor in your application's `web.config`:
```
bmorelli25: I've redesigned the layout of the **Setup** page to *hopefully* be easier to digest. Please take a look and compare the changes: **NEW LAYOUT** /vs/ **OLD LAYOUT**

docs/intro.asciidoc
```diff
@@ -6,17 +6,23 @@ endif::[]
...
 
-Elastic APM automatically measures the performance of your application and tracks errors.
-For example, it records spans for database queries and transactions for incoming HTTP requests.
-See <<supported-technologies>> to learn about which frameworks and technologies are supported by Elastic APM .NET agent.
+[float]
+[[how-it-works]]
+=== How does the Agent work?
+
+The agent auto-instruments <<supported-technologies,supported technologies>> and records interesting events,
+like HTTP requests and database queries. To do this, it {CONTENT_HERE}
```
gregkalapos:
```suggestion
like HTTP requests and database queries. To do this, it uses built in capabilities of the instrumented frameworks like [Diagnostic Source](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.diagnosticsource?view=netcore-3.0), an HTTP module for IIS, or [IDbCommandInterceptor](https://docs.microsoft.com/en-us/dotnet/api/system.data.entity.infrastructure.interception.idbcommandinterceptor?view=entity-framework-6.2.0) for Entity Framework.
```

docs/intro.asciidoc
```diff
@@ -6,17 +6,23 @@ endif::[]
 [[intro]]
 == Introduction
 
-Welcome to the APM .NET Agent documentation.
+The Elastic APM .NET Agent automatically measures the performance of your application and tracks errors.
+It has built-in support for the most popular frameworks and routers,
```
gregkalapos: What is `routers`? I'm not sure it has a well known meaning in .NET. 
Update: Hah, by looking at other docs, I suspect that's some Node.js lingo :) Probably I'd still not add that word here.

docs/intro.asciidoc
```diff
@@ -6,17 +6,23 @@ endif::[]
...
-Elastic APM automatically measures the performance of your application and tracks errors.
-For example, it records spans for database queries and transactions for incoming HTTP requests.
-See <<supported-technologies>> to learn about which frameworks and technologies are supported by Elastic APM .NET agent.
+[float]
+[[how-it-works]]
+=== How does the Agent work?
+
+The agent auto-instruments <<supported-technologies,supported technologies>> and records interesting events,
+like HTTP requests and database queries. To do this, it {CONTENT_HERE}
+This means that for the supported technologies, there are no code changes required.
```
gregkalapos: Sooo... the `no code changes required` is not 100% accurate: there is a 1 liner (or you need to register the IIS module) in order to to turn on these auto instrumentations. Maybe we could say, `there are no code changes required, other than enabling auto-instrumentation as shown on the [setup page](LinkToSetupPage)`. With this we'd be on the safe side and not create the impression that adding the NuGet packages is enough.

docs/intro.asciidoc
```diff
@@ -6,17 +6,23 @@ endif::[]
...
 
-Spans are grouped in transactions - by default one for each incoming HTTP request.
-But it's possible to create custom transactions not associated with an HTTP request.
+The Agent automatically {CONTENT_HERE}
```
gregkalapos:
```suggestion
The Agent automatically registers callback methods for these built in Diagnostic Source events - or in case of the IIS HTTP module as described on the <<setup,Set up the Agent page>>, the user needs to register the module. With this, the supported frameworks trigger agent code for relevant events to measures their duration and collects metadata (like DB statements), as well as HTTP related information (like the URL, parameters, and headers).
```

gregkalapos: In the `<<setup,Set up the Agent page>>` link it'd be nice to link to the `setup.html#setup-asp-net` part - not sure how to do that 😿

docs/setup.asciidoc
gregkalapos: Awesome, I love it!

docs/intro.asciidoc
gregkalapos: Thx for the link - it was useful. I added suggestions to both.

docs/intro.asciidoc
bmorelli25: Ahhh, oops! Sorry about that

docs/intro.asciidoc
bmorelli25: Great idea, thank you!

docs/intro.asciidoc
bmorelli25: I'll fix the link

bmorelli25:
Thanks @gregkalapos! I've added in your feedback and done one more edit of the "How the Agent works" section.

docs/intro.asciidoc
```diff
@@ -6,20 +6,30 @@ endif::[]
...
+To do this, it uses built-in capabilities of the instrumented frameworks like
+https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.diagnosticsource?view=netcore-3.0[Diagnostic Source],
+an HTTP module for IIS, or
+https://docs.microsoft.com/en-us/dotnet/api/system.data.entity.infrastructure.interception.idbcommandinterceptor?view=entity-framework-6.2.0[IDbCommandInterceptor] for Entity Framework.
+This means that for the supported technologies, there are no code changes required beyond enabling <<setup,auto-instrumentation>>.
 
-By default the agent comes with support for ASP.NET Core out-of-the-box.
-To instrument other events, you can use custom spans.
+The Agent automatically registers callback methods for built-in Diagnostic Source events.
+The IIS HTTP module, as described in <<setup-asp-net, Set up ASP.net Core>>, will need to be registered.
```
gregkalapos: One last thing: The IIS HTTP module is only required for ASP.NET Classic - it's not needed for ASP.NET Core. So what I tried to say here is that for almost everything the agent registers for the built-in Diagnostic Source events, except in case of ASP.NET Classic - in that case you need to manually register the IIS module, because there is no Diagnostic Source for that - so once either the the Diagnostic Source listeners of the HTTP module is registered, agent code will be triggered and measure the duration, etc. 

I'm not even sure we need to go into these details, I just wanted to mention the IIS module for completeness, but maybe it was rather confusing and less useful.


I let you decide, but I think it'd make sense to either drop the IIS module sentence, or adapt it as described above.

bmorelli25: Thanks for the clarification. Let's remove the sentence. It's probably more detail than is necessary.

docs/configuration.asciidoc
SergeyKleyman: Could you clarify why space needs to be removed from `^[a-zA-Z0-9 _-]+$`? I see that other agents have it and in the paragraph above the following sentence says that spaces are allowed.

SergeyKleyman: I think we need to replace two underscores in "replaced by `__` symbol" above by just one underscore because while it does render as one underscore at github it renders as two underscores at https://www.elastic.co/guide/en/apm/agent/dotnet/current/config-core.html#config-service-name.

bmorelli25: My mistake -- thanks for catching this Sergey. You're right, removing the space from the regular expression doesn't make sense as then spaces wont match.
```suggestion
NOTE: The service name must conform to this regular expression: `^[a-zA-Z0-9 _-]+$`. In less regexy terms: Your service name must only contain characters from the ASCII alphabet, numbers, dashes, underscores and spaces. Characters in service name which don't match regular expression will be replaced by `_` symbol.
```

bmorelli25: Fixed in 1493271 (all numbers -- very cool)

PR #577, opened 2019-10-29
Support getting CAPTURE_BODY and TRANSACTION_MAX_SPANS options from central config

### LINKED ISSUE #534: Add support for the next batch of central config options (CAPTURE_BODY and TRANSACTION_MAX_SPANS)
Derived from https://github.com/elastic/apm/issues/138.

Add support for the next batch of central (AKA remote) config options. At the moment the next batch consists of the following options:
1. `CAPTURE_BODY`
2. `TRANSACTION_MAX_SPANS` (see https://github.com/elastic/apm/issues/138#issuecomment-535748667 regarding backend validation of this option's value)

We should also consider marking in docs which options are reloadable and via which sources (if we want to support reloadable sources other than central configuration - for example `Microsoft.Extensions.Configuration`)

SergeyKleyman:
Closes #534


docs/configuration.asciidoc
```diff
@@ -617,3 +626,36 @@ The default unit for this option is `ms`
...
+| <<config-sanitize-field-names,`SanitizeFieldNames`>> | No | Core
+| <<config-secret-token,`SecretToken`>> | No | Reporter
+| <<config-server-urls,`ServerUrls`>> | No | Reporter
+| <<config-service-name,`ServiceName`>> | No | Core
+| <<config-service-version,`ServiceVersion`>> | No | Core
+| <<config-span-frames-min-duration,`SpanFramesMinDuration`>> | No | Stacktrace, Performance
+| <<config-stack-trace-limit,`StackTraceLimit`>> | No | Stacktrace, Performance
+| <<config-transaction-max-spans,`TransactionMaxSpans`>>  | Yes | Performance, Core
+| <<config-transaction-sample-rate,`TransactionSampleRate`>> | Yes | Performance, Core
+|============
```
bmorelli25: I really like this idea. My only concern is that it's one more spot in the documentation that needs to be updated if a config changes or a new one is added. Let's keep it around, just wanted to share that thought with you. It would be awesome if we could someday automate these jump tables.

docs/configuration.asciidoc
```diff
@@ -617,3 +626,36 @@ The default unit for this option is `ms`
...
+
+NOTE: Option being reloadable means that the setting can be changed after agent's start.
```
bmorelli25: Let's simplify:

```suggestion
NOTE: Reloadable means that the setting can be changed after agent's start.
```

docs/configuration.asciidoc
```diff
@@ -617,3 +626,36 @@ The default unit for this option is `ms`
...
+For example some reloadable options can be changed
```
```suggestion
For example, some reloadable options can be changed
```

docs/configuration.asciidoc
```diff
@@ -617,3 +626,36 @@ The default unit for this option is `ms`
...
+via {apm-app-ref}/agent-configuration.html[APM Agent configuration]).
```
bmorelli25: Thanks for using `apm-app-ref` here!! 😊

docs/configuration.asciidoc
SergeyKleyman: Yes, I agree if we get to a situation when maintaining this summary becomes too error-prone we will consider how to automate it.

SergeyKleyman: BTW if you think we should have a different set of columns please let me know.

docs/configuration.asciidoc
SergeyKleyman: I just copied [your changes](https://github.com/elastic/apm-agent-dotnet/pull/574/files#diff-01c1a79a5a1e1b67003a1885365f2cf5R251) :)

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
```diff
@@ -651,22 +651,13 @@ protected string ParseCaptureBody(ConfigurationKeyValue kv)
 			return captureBodyInConfig.ToLowerInvariant();
 		}
 
-		protected List<string> ParseCaptureBodyContentTypes(ConfigurationKeyValue kv, string captureBody)
+		protected List<string> ParseCaptureBodyContentTypes(ConfigurationKeyValue kv)
```
gregkalapos: Sooo..what happens here is that since `CAPTURE_BODY` is now live changeable, we can't take that into consideration when we read `CaptureBodyContentTypes`, right?

test/Elastic.Apm.AspNetCore.Tests/BodyCapturingTests.cs
```diff
@@ -30,77 +34,258 @@ namespace Elastic.Apm.AspNetCore.Tests
...
+							IsSampled = isSampled
+						};
+					}
+				}
+			}
+		}
+
+		[Theory]
+		[MemberData(nameof(OptionsChangedAfterStartTestVariants))]
+		public async Task Options_changed_after_start(int startCfgVariantIndex, OptionsTestVariant startCfgVariant)
```
gregkalapos: It'd be really nice to add some comments to this one. To be honest, I already started going through this PR before EAH and this was the point where I gave up - some comments would really help here.

src/Elastic.Apm.AspNetCore/Extensions/TransactionExtentions.cs
```diff
@@ -1,36 +1,52 @@
...
-			}
+			string body = null;
+			if (IsCaptureRequestBodyEnabled(transaction, isForError) && IsCaptureRequestBodyEnabledForContentType(transaction, request))
+				body = request.ExtractRequestBody(logger);
+			// According to the documentation - the default value of 'body' is '[Redacted]'
+			transaction.Context.Request.Body = body ?? Apm.Consts.Redacted;
+		}
+
+		private static bool IsCaptureRequestBodyEnabled(Transaction transaction, bool isForError) =>
+			transaction.ConfigSnapshot.CaptureBody.Equals(ConfigConsts.SupportedValues.CaptureBodyAll)
```
gregkalapos: At least to me `IConfigurationReaderExtentions` with `ShouldExtractRequestBodyOnError` and `ShouldExtractRequestBodyOnTransactions` was easier to follow. - no action needed on this, just saying.

src/Elastic.Apm/Model/Transaction.cs
```diff
@@ -117,6 +116,9 @@ ICurrentExecutionSegmentsContainer currentExecutionSegmentsContainer
 		/// </summary>
 		public Context Context => _context.Value;
 
+		[JsonIgnore]
+		internal IConfigSnapshot ConfigSnapshot { get; }
```
gregkalapos: So, if I understand correctly, the plan is that this is the `ConfigSnapshot` instance that we pass to every other component like `Span` and also this instance is used by things like `TransactionExtensions`, right? If I'm right, then this seems to be a fairly important property and I'd comment it. If I'm wrong, then I'd still comment it, because after looking at the code long enough I got the wrong impression.

src/Elastic.Apm.AspNetCore/Extensions/TransactionExtentions.cs
```diff
@@ -1,36 +1,52 @@
...
 		{
-			var body = Consts.BodyRedacted; // According to the documentation - the default value of 'body' is '[Redacted]'
+			if (!transaction.IsSampled) return;
 
-			// We need to parse the content type and check it's not null and is of valid value 
-			if (!string.IsNullOrEmpty(httpContext?.Request?.ContentType))
-			{
-				var contentType = new ContentType(httpContext.Request.ContentType);
+			// Is request body already captured?
+			if (transaction.IsContextCreated
```
gregkalapos: Why does `transaction.IsContextCreated` matter here? 🤔 Is it because `transaction.Context.Request` would create one? If so, is there any case when only the `Body` is filled? I don't think so... so if we are here, I think we already have a context anyway, or not?

src/Elastic.Apm.AspNetCore/DiagnosticListener/AspNetCoreDiagnosticListener.cs
```diff
@@ -35,14 +33,15 @@ public void OnNext(KeyValuePair<string, object> kv)
 
 			var exception = kv.Value.GetType().GetTypeInfo().GetDeclaredProperty("exception").GetValue(kv.Value) as Exception;
 
-			var transaction = _agent.Tracer.CurrentTransaction;
+			var transaction = (Transaction)_agent.Tracer.CurrentTransaction;
```
gregkalapos: Hmm.. this cast is a little bit unfortunate - with this it MUST be a `Transaction`, otherwise it will throw an exception - so e.g. we can't just replace to `NoOpTransaction` later if needed.

I guess the root cause of the challenge here is that we don't dependency inject the configuration anymore, instead it comes through the `Transaction`. I'm not sure this is good - could we maybe find another way to flow the `ConfigSnapshot`?

src/Elastic.Apm.AspNetCore/Extensions/TransactionExtentions.cs
```diff
@@ -1,12 +1,11 @@
...
-using Elastic.Apm.Api;
-using Elastic.Apm.Config;
 using Elastic.Apm.Logging;
+using Elastic.Apm.Model;
 using Microsoft.AspNetCore.Http;
 
 namespace Elastic.Apm.AspNetCore.Extensions
 {
-	public static class TransactionExtensions
+	internal static class TransactionExtensions
```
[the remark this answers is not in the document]
SergeyKleyman: I agree with your assessment:

> Probably no one uses it

So is it worth spending any more time on it?

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
SergeyKleyman: That is correct.

src/Elastic.Apm.AspNetCore/Extensions/TransactionExtentions.cs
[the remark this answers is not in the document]
gregkalapos: > So is it worth spending any more time on it?

Yes, because it's just about changing `internal` back to `public` and we have 1 less breaking change. That can be done very quickly.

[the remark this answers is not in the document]
SergeyKleyman: Unfortunately it's not as simple as changing `internal` back to `public` because the signature of `CollectRequestInfo` changed.

test/Elastic.Apm.AspNetCore.Tests/BodyCapturingTests.cs
SergeyKleyman: Done.

src/Elastic.Apm/Model/Transaction.cs
SergeyKleyman: Done.

src/Elastic.Apm.AspNetCore/Extensions/TransactionExtentions.cs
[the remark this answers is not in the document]
gregkalapos: Oh, yeah I see. As said [here](https://github.com/elastic/apm-agent-dotnet/pull/577#discussion_r345905598), I'd left it as it is. I know, the `IConfigurationReader` went away, but technically we could still pass that as a parameter.

Anyway, ok, let's make it `internal` and hope no one uses it. But could we try to avoid these things in the future please?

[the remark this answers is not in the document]
SergeyKleyman: Definitely.

src/Elastic.Apm.AspNetCore/DiagnosticListener/AspNetCoreDiagnosticListener.cs
SergeyKleyman: It's not like it's the only place where we assume that `ITransaction` is actually `Transaction` - for example https://github.com/elastic/apm-agent-dotnet/blob/a73b10f0f6d61de45786f10501120644b9a82ff7/src/Elastic.Apm/Report/PayloadSenderV2.cs#L219)

If/when we need to have other implementations of `ITransaction` we will revisit it. For example we can introduce `TransactionBase` and move `ConfigSnapshot` there.

src/Elastic.Apm.AspNetCore/Extensions/TransactionExtentions.cs
SergeyKleyman: I didn't want to assume too much about the context in which this method is called - the purpose was to avoid creating empty `Context`. I've added a comment to the code.

src/Elastic.Apm.AspNetCore/DiagnosticListener/AspNetCoreDiagnosticListener.cs
gregkalapos: I'd consider a pattern matching based on types very different from a hard cast.

But I think we are good here. In https://github.com/elastic/apm-agent-dotnet/pull/608 I moved this part out from here, so probably we can also get rid of the cast once we merge that. 

Context: In #608 I realized that capturing request body here is not a very good thing, because this is a sync method and reading the body synchronously can cause thread starvation. So I'd say let's leave it as is, and in #608 this will evolve.

src/Elastic.Apm.AspNetCore/Extensions/TransactionExtentions.cs
gregkalapos: Ok, sounds good to me.

Issue #582, opened 2019-11-01
Public API Span Labels Length Issue 


soumithx:
Captured Span Label Values Show truncating String,

I am using dotnet apm public api.

my requirement is to assign a label to span which is large text( Generally its large XML String), 
if i do that in general way ,it shows truncated string instead of full string on kibanna dashboard.

if there any solution exist, Greatly Appreciated .

gregkalapos:
Hi @soumithx,

labels are truncated to 1024 characters, so I'm afraid you won't be able to store very long XMLs there. 

This is specified in the API between the agent and the APM server, so this statement holds to all Elastic APM agents, so this isn't specific to the .NET APM Agent - so what I want to say with this second part is that this is not something we could easily change in the agent.

gregkalapos:
Removed the bug label, because this is by design.

soumithx:
@gregkalapos , 

Hi Greg , then what is best way to display , is there any  custom instrumentation we can apply to achieve ?

gregkalapos:
> @gregkalapos ,
> 
> Hi Greg , then what is best way to display , is there any custom instrumentation we can apply to achieve ?

I'm not sure we have anything on the public API that you can use to transfer that amount of data. The big problem with this would be that this could lead to huge data usage by elasticsearch very easily. 

Off the top of my head, I'm not sure if we have any field where you could transfer unlimited data to the APM server, but I asked the rest of the team if we have something on the APM server API to this - I'll report back here once I know more. If there is something we could expose thorough the agent API then I'm totally open for considering it.

Could you maybe say a little bit more about your use case? Like, what kind of span is this? Where is the XML coming from?

soumithx:
@gregkalapos ,


its custom Captured Span through public Api, i can say  this is not pure XML ,but its an just XML look like string which created  by Stringbuilder

soumithx:
some how , i got breakthrough  by attaching that string as   Context.Db Database statement .....

still , that size is not yet completely fitted

gregkalapos:
> some how , i got breakthrough by attaching that string as Context.Db Database statement .....
> 
> still , that size is not yet completely fitted

:) the limit on that is 10000. Plus the other problem is that that is not intended for this use case.

Soo.. there is 1 field on the `Transaction` in the APM Server's intake API that we could use for this: https://github.com/elastic/apm-server/blob/master/docs/spec/context.json#L7 

[...]

PR #585, opened 2019-11-05
Add Custom to Context and expose it via public API

gregkalapos:
Addressing https://github.com/elastic/apm-agent-dotnet/issues/582.

Adding and exposing `Custom` which is not yet available in .NET.

Also added docs.


soumithx:
is build failed? @gregkalapos

gregkalapos:
> is build failed? @gregkalapos

We had some known CI issues which we already work on. The build fail is probably not related to this PR.

src/Elastic.Apm/Api/Context.cs
```diff
@@ -1,45 +1,50 @@
-using System;
```
gregkalapos: Sorry about the lots of changes here - this is again our line ending issue. The only real change here is that I added `public Dictionary<string, string> Custom => _custom.Value;` and `public bool ShouldSerializeCustom() => _custom.IsValueCreated && Custom.Count > 0;`

docs/public-api.asciidoc
```diff
@@ -302,6 +302,23 @@ NOTE: Return value of <<convenient-capture-span>> method overloads that accept T
 
 NOTE: Code samples above use `Elastic.Apm.Agent.Tracer.CurrentTransaction`. In production code you should make sure the `CurrentTransaction` is not `null`.
 
+[float]
+[[api-transaction-Custom]]
+==== `Dictionary<string,string> Custom`
+
+Custom context is used to add non-indexed, custom contextual information to transactions. Non-indexed means the data is not searchable or aggregatable in Elasticsearch, and you cannot build dashboards on top of the data. However, non-indexed information is useful for other reasons, like providing contextual information to help you quickly debug performance issues or errors.
+
+Unlike <<api-transaction-tags>> the data in this property is not trimmed.
```
bmorelli25:
```suggestion
Unlike <<api-transaction-tags>>, the data in this property is not trimmed.
```

docs/public-api.asciidoc
```diff
@@ -302,6 +302,23 @@ NOTE: Return value of <<convenient-capture-span>> method overloads that accept T
 
 NOTE: Code samples above use `Elastic.Apm.Agent.Tracer.CurrentTransaction`. In production code you should make sure the `CurrentTransaction` is not `null`.
 
+[float]
+[[api-transaction-Custom]]
+==== `Dictionary<string,string> Custom`
+
+Custom context is used to add non-indexed, custom contextual information to transactions. Non-indexed means the data is not searchable or aggregatable in Elasticsearch, and you cannot build dashboards on top of the data. However, non-indexed information is useful for other reasons, like providing contextual information to help you quickly debug performance issues or errors.
```
bmorelli25: Can we move each sentence to its own line?

gregkalapos: Sure, I did it.

docs/public-api.asciidoc
```diff
@@ -302,6 +302,25 @@ NOTE: Return value of <<convenient-capture-span>> method overloads that accept T
...
+Non-indexed means the data is not searchable or aggregatable in Elasticsearch, and you cannot build dashboards on top of the data.
+However, non-indexed information is useful for other reasons, like providing contextual information to help you quickly debug performance issues or errors.
+
+Unlike <<api-transaction-tags>>, the data in this property is not trimmed.
+
+[source,csharp]
+----
+Agent.Tracer.CaptureTransaction(transactionName, transactionType, (transaction) =>
+{
+	transaction.Custom.Add("foo", "bar");
```
SergeyKleyman: It might be more consistent to use indexer here just as in `Labels` example and also it might be worth using value that would not be possible to store as label (i.e., very long string).

src/Elastic.Apm/Api/Context.cs
```diff
@@ -1,45 +1,50 @@
...
+	{
+		private readonly Lazy<Dictionary<string, string>> _custom = new Lazy<Dictionary<string, string>>();
 		private readonly Lazy<Dictionary<string, string>> _labels = new Lazy<Dictionary<string, string>>();
 
-		/// <summary>
-		/// <seealso cref="ShouldSerializeLabels" />
-		/// </summary>
-		[JsonProperty("tags")]
-		[JsonConverter(typeof(LabelsJsonConverter))]
+		public Dictionary<string, string> Custom => _custom.Value;
```
SergeyKleyman: We need `JsonConverter` here as well to ensure the keys are valid.

docs/public-api.asciidoc
gregkalapos: Makes sense, changed it to indexer

gregkalapos: Makes sense, changed it to indexer

src/Elastic.Apm/Api/Context.cs
gregkalapos: Added a converter for this - also covered by tests now.

PR #587, opened 2019-11-08
Align naming of configuration readers between Full Framework and .NET Core modules

NeoXtreem:
This PR is an offshoot of #364, the core change of which was implemented in a separate PR #503.  However, #364 contained other changes that were not included in #503, which the title of this PR reflects.

The purpose of this PR is to align the names of corresponding classes whose purpose are the same, namely `MicrosoftExtensionsConfig` in `Elastic.Apm.AspNetCore` and `FullFrameworkConfigReader` in `Elastic.Apm.AspNetFullFramework`.

The two modules are to support each of .NET Core and .NET Framework.  It therefore makes sense to make the corresponding classes that each derive from the same base class (`AbstractConfigurationWithEnvFallbackReader`) have the same name.  For `FullFrameworkConfigReader` in particular, it makes little sense to have "FullFramework" as part of the class name when "FullFramework" is part of the assembly name and therefore the namespace.

`MicrosoftExtensionsConfig` is also unusual as it is not conventional to name a class according to the namespace of one of its dependencies.

The advantage of this change is that it will reduce confusion and assist the coder in understanding the purpose of the classes, especially when wishing to make changes in both modules that require changes to these classes.


SergeyKleyman:
@NeoXtreem We are in the process of refactoring our configuration related part of the agent to address the following issues:
1. Minimize amount of changes required when adding new configuration setting, especially if setting type is already supported (such as boolean, integer, simple "name like" string, etc.)
2. Minimize amount of changes required when making a configuration setting [reloadable](https://www.elastic.co/guide/en/apm/agent/dotnet/master/config-all-options-summary.html) (setting can be changed after agent’s start)
3. Allow users to supply custom configuration sources (as it stands now users can implement `IConfigurationReader` but we add methods to it very frequently which will break those interface implementations while it should not since it has nothing to do with configuration source).

The tentative plan is to have draft PR for the refactoring next week. I propose to get back to this PR after that refactoring is completed.

gregkalapos:
This got very outdated - closing now.

PR #588, opened 2019-11-09
move the same info from projects into props files

vhatsura:
When we need to create a new source or test projects, it's necessary to copy-paste a lot of information from other projects.

Besides, to release a new version [a `PackageVersion` and assembly info stuff need to be updated](https://github.com/elastic/apm-agent-dotnet/pull/528) for each project in a solution. With adding more and more projects, this kind of work became a routine that can be solved.

After merging #426, we can store package info in `Directory.Build.props`, which done in this PR.

@gregkalapos, I believe it will simplify the release process to you.

Also, I want to raise a question about `Elastic.Apm.PerfTests` project. The project isn't similar to the test project and I suggest to move it into the separate `benchmark` folder in the root of the repository. It will allow the pinning `xunit` package version for all test projects, because after adding a new test project, `xunit` version needs to be downgraded to be consistent with `xunit` version in other test projects. Besides, it will resolve the misunderstanding that the project automatically will notify about performance degradation.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -3,5 +3,9 @@
   <PropertyGroup>
     <LangVersion>latest</LangVersion>
     <NoWarn>NU1503</NoWarn>
+
+    <SignAssembly>true</SignAssembly>
+    <AssemblyOriginatorKeyFile>..\..\build\elasticapm.snk</AssemblyOriginatorKeyFile>
+    <DelaySign>false</DelaySign>
   </PropertyGroup>
 </Project>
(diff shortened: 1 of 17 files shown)


gregkalapos:
Oh, I was going to resolve the conflict, thanks for doing it :)

vhatsura:
> Oh, I was going to resolve the conflict, thanks for doing it :)

no problem. So, I have only the last question in my head: when do we need to merge this PR? As earlier as possible will introduce a lot of merge conflicts with other PRs, as later as possible means that the PR will be updated (at least because #592 adds new projects).

gregkalapos:
>  As earlier as possible will introduce a lot of merge conflicts

Only PRs that touch the `.csproj` and `.sln` files will conflict with this, right? I think these are very easy to resolve, usually changes in those files are only adding 1-2 lines, so you can just take the those files from this PR and manually add your small changes and with that the conflict is resolved. 

Or do I miss anything?

vhatsura:
> > As earlier as possible will introduce a lot of merge conflicts
> 
> Only PRs that touch the `.csproj` and `.sln` files will conflict with this, right? I think these are very easy to resolve, usually changes in those files are only adding 1-2 lines, so you can just take the those files from this PR and manually add your small changes and with that the conflict is resolved.
> 
> Or do I miss anything?
> 
> Update: by this I guess merging soon is my preference.

Sure, it's good enough for me. I will wait for merge of this PR to update #592.

gregkalapos:
Merged, yeah right  #592 is one that needs manual update. Thanks for updating that! 👍

PR #584, opened 2019-11-04
Reduce memory allocations for payload formatter

vhatsura:
After analysis of memory consumption, I found out that `PayloadSenderV2` allocates a lot of strings. This pull request should reduce memory allocations for payload formatter.
The main idea is to write serialized payload directly to `JsonTextWriter` (wraps `StringWriter`) without intermediate string allocation. To not create `StringWriter` every time, I introduce `StringWriterPool` - a lock-free concurrent pool of string writers. 

What was done:

1. Extracted formatting logic from `PayloadSenderV2` to separate class. `PayloadFormatterV2` class has the old logic of payload formatting. `EnhancedPayloadFormatter` has a new logic for formatting.
2. Introduced `ObjectPool` which allow to pool objects. `StringWriterPool` is a pool of string writers.
3. Added payload formatter benchmark to measure memory allocations.

NOTE: All class names aren't final ;)

What will be added, in case of interest from your side:

1. Rename `EnhancedPayloadFormatter` and remove old formatter
2. Add more tests for `EnhancedPayloadFormatter`

Result of benchmark run:

```
// * Summary *

BenchmarkDotNet=v0.11.4, OS=Windows 10.0.19002
Intel Core i7-7700 CPU 3.60GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=2.1.505
  [Host]     : .NET Core 2.1.9 (CoreCLR 4.6.27414.06, CoreFX 4.6.27415.01), 64bit RyuJIT
  DefaultJob : .NET Core 2.1.9 (CoreCLR 4.6.27414.06, CoreFX 4.6.27415.01), 64bit RyuJIT


|                   Method | CallAmount |        Mean |      Error |    StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|------------------------- |----------- |------------:|-----------:|----------:|------------:|------------:|------------:|--------------------:|
|       PayloadFormatterV2 |          1 |    11.73 us |  0.0492 us | 0.0460 us |      4.6539 |           - |           - |            19.11 KB |
| EnhancedPayloadFormatter |          1 |    10.23 us |  0.0704 us | 0.0658 us |      1.9379 |           - |           - |             7.98 KB |
|       PayloadFormatterV2 |         10 |   114.90 us |  0.9248 us | 0.8650 us |     42.8467 |           - |           - |           175.98 KB |
[...]

Issue #322, opened 2019-06-26
Manually send custom metrics (Agent API or integration with a Metrics Library)


gregkalapos:
We should offer users the ability to manually send metrics.

I think there would be 2 approaches:
- Integrate with an existing metrics library (e.g. [AppMetrics](https://github.com/AppMetrics/AppMetrics), [Metrics.NET](https://github.com/etishor/Metrics.NET)) 
- Extend the public agent API

1. step:  decide which approach to take 
2. step: implement.

oleh-zheleznyak:
For my project having custom metrics is really important, and it is also available in other APM products on the market. 
It would be really awesome to have this!

From a consumers point of view, in a ASP .NET Core application, I would expect to be able to register a source of `IMetricsProvider` (it could be a list, a factory, etc) with my DI container by using a simple extension method on `IServiceCollection` for example `services.AddElasticApmMetrics(IReadOnlyCollection<IMetricsProvider> metricProviders)`
(That collection/factory could later be injected into a new `MetricsCollector` class ctor)

Without custom metrics key stakeholders in our company will most likely choose another product.

Thanks a lot for the great effort you're doing with APM agent for dotnet!

maverix:
Could this be done using Agent.PayloadSender.QueueMetrics(IMetricSet metrics);

maverix:
Otherwise the IMetricsCollector.MetricsProviders needs to be exposed via a public API

maverix:
https://github.com/workbetter-com-au/apm-agent-dotnet-metrics
Attempted a quick manual 'send metric'.  Couldn't get it to work.  No errors were logged.
I will pull into the project the actual Elastic.Apm to troubleshoot why it didn't work

gregkalapos:
> Couldn't get it to work. No errors were logged.

Let me try to help.

The agent currently sends data to the APM Server when one of these 2 things happen:
- A transaction ends (`Transaction.End()` is called either manually or by auto instrumentation)
- There are at [least 20 events](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm/Report/PayloadSenderV2.cs#L29) recorded (1 event is e.g. a span, or a metrics with 1 timespan)

[...]

PR #590, opened 2019-11-09
Filter NaN and Infinite double value of metric set

### LINKED ISSUE #589: metricsets including NaN values rejected
**Describe the bug**
.NET agent producing invalid metricset payloads, apparently due to inclusion of `NaN` values.

**To Reproduce**
Per [this discuss issue](https://discuss.elastic.co/t/apm-dotnet-1-1-2-error-validating-json-document-against-schema-i-s-doesnt-validate-with-metricset/207101), run on a system where system process cpu isn't available?

**Expected behavior**
Omit `NaN` values.

vhatsura:
[A java agent consider `NaN` and `Infinite`](https://github.com/elastic/apm-agent-java/blob/151356f2ea2cb4fdaeefda40fe3310f26c7fc3ff/apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/MetricRegistrySerializer.java#L174) values as invalid.

The issue can be solved in two ways:
1. Add only valid metric samples to metric set
2. Filter metric samples during serialization

I decided to fix the issue with adding filtration of metric samples during instantiation `MetricsSet`, because, IMHO, it's better to filter such values as earlier as possible in a pipeline.

Regardless of the selected option, we can have a case, when all metric samples are filtered out. So, the main question what we need to do in such case? Do we need to send metrics set with empty samples or ignore such metrics set?


SergeyKleyman:
My concern is this solution effectively masks the bug in the way we obtain metrics instead of fixing the root cause for NaN.

vhatsura:
> My concern is this solution effectively hides the bug in the way we obtain metrics instead of fixing the root cause for NaN.

Reasonable concern. Right now, the agent doesn't allow to send custom metrics, however, there will able to send it after implementation of  #322. After this, a question about `NaN` and `Infinity` can be raised again. I don't sure whether `MetricSet` will be used or not, but I agreed with @gregkalapos that it's a good candidate to be public (https://github.com/elastic/apm-agent-dotnet/issues/322#issuecomment-507646456).

So, as a result I suggest two things:
1. Save filtration logic. Also, it would be nice to have a log record that some metrics values were filtered out.
2. Resolve the original issue.

weedaj:
> My concern is this solution effectively hides the bug in the way we obtain metrics instead of fixing the root cause for NaN.

hi @SergeyKleyman , just to remark that the issue manifests on dotnet APM agents implemented on Lambda environments so imo it would be logical that no CPU stats are provided. In that light filtering seems like a adequate approach.

SergeyKleyman:
@vhatsura I agree that we can add "filtering out abnormal floating point values" mechanism you propose in this PR independently of solving #589 - but I don't think we should mark this mechanism as the fix to #589.

SergeyKleyman:
@vhatsura regarding:

> The issue can be solved in two ways:
> 
> 1. Add only valid metric samples to metric set
> 2. Filter metric samples during serialization
> 
> I decided to fix the issue with adding filtration of metric samples during instantiation `MetricsSet`, because, IMHO, it's better to filter such values as earlier as possible in a pipeline.

I agree that there are certain advantages to enforcing Intake API constraints as early as possible but so far we always did it at the very last (serialization) stage so I think it would be preferable to be consistent with that approach.

SergeyKleyman:
@vhatsura regarding:

> Regardless of the selected option, we can have a case, when all metric samples are filtered out. So, the main question what we need to do in such case? Do we need to send metrics set with empty samples or ignore such metrics set?

I think we should not send empty metric sets to avoid potential issues with APM Server having issues handling this case.

SergeyKleyman:
@weedaj Could you please clarify as to why "it would be logical that no CPU stats are provided"? Is it because API we use to obtain CPU stats doesn't work on AWS? What about other process metrics (such as for memory)? Is there an alternative that should be used?

SergeyKleyman:
@vhatsura Another point regarding "filtering out abnormal floating point values" mechanism you propose in this PR - I think it is important to make it clearly visible when metric values are being discarded for example when we need to investigate a case with a user complaining that some metrics are not collected by .NET agent. The simplest solution would be to log it under high enough level (ERROR?) but if we do it on every metric collection time interval it will flood the logs so maybe we should extend [the existing MetricsCollector's mechanism that disables a metrics provider after it fails a predefined number of times in a row](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm/Metrics/MetricsCollector.cs#L120
) and consider returning `NaN` or `Infinity` values as a failure as well? Extending the existing MetricsCollector's mechanism will take care of logging and there won't be a need to filter at the later (serialization) stage. What do you think?

src/Elastic.Apm/Metrics/MetricSet.cs
```diff
@@ -8,8 +9,8 @@ namespace Elastic.Apm.Metrics
 	[JsonConverter(typeof(MetricSetConverter))]
 	internal class MetricSet : IMetricSet
 	{
-		public MetricSet(long timeStamp, List<MetricSample> samples)
-			=> (TimeStamp, Samples) = (timeStamp, samples);
+		public MetricSet(long timeStamp, IEnumerable<MetricSample> samples)
```
gregkalapos:
```suggestion
		public MetricSet(long timeStamp, IEnumerable<MetricSample> samples, IApmLogger logger)
		{
			var metricSamples = samples as MetricSample[] ?? samples.ToArray();
			(TimeStamp, Samples) = (timeStamp, metricSamples.Where(x => !double.IsNaN(x.KeyValue.Value) && !double.IsInfinity(x.KeyValue.Value)));
			if (Samples.Count() != metricSamples.Length)
			{
				logger.Debug()?
					.Log("Some values were filtered out from MetricSample due to IsNan && IsInfinity, key: {key}", metricSamples.First()?.KeyValue.Key);
			}
		}
```

src/Elastic.Apm/Agent.cs
```diff
@@ -47,7 +47,9 @@
 [assembly:
 	InternalsVisibleTo(
 		"Elastic.Apm.Tests.MockApmServer, PublicKey=002400000480000094000000060200000024000052534131000400000100010051df3e4d8341d66c6dfbf35b2fda3627d08073156ed98eef81122b94e86ef2e44e7980202d21826e367db9f494c265666ae30869fb4cd1a434d171f6b634aa67fa8ca5b9076d55dc3baa203d3a23b9c1296c9f45d06a45cf89520bef98325958b066d8c626db76dd60d0508af877580accdd0e9f88e46b6421bf09a33de53fe1")]
-
+[assembly:
+	InternalsVisibleTo(
+		"DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7")]
```
SergeyKleyman: Could you please clarify the purpose of this change?

vhatsura: Sure, it's needed to allow `Moq` library mocking internal interfaces.

src/Elastic.Apm/Metrics/MetricSet.cs
vhatsura: Sure, I also thought about logging, however, I don't sure that adding IApmLogger as a dependency as a good idea. Is it better to use `ApmAgent.Logger` for this?

vhatsura:
@SergeyKleyman, thanks a lot for your insights. I think all of them are valuable. Let's comment on some of them:
> but I don't think we should mark this mechanism as the fix to #589.

I removed the link to the issue.

> so maybe we should extend the existing MetricsCollector's mechanism that disables a metrics provider after it fails a predefined number of times in a row and considers returning NaN or Infinity values as a failure as well?

It's a good idea, however, I'd prefer to open an additional issue for that.

> I think we should not send empty metric sets to avoid APM Server potentially having issues handling this case.

I'll update my PR with such logic.

> I agree that there are certain advantages to enforcing Intake API constraints as early as possible but so far we always did it at the very last (serialization) stage so I think it would be preferable to be consistent with that approach.

From a consistency perspective, I agree with you, however, my experience of using agent shows that agent is using a lot of memory (this is a reason why I open #584). So, maybe we need to rethink this strategy to have filtration of invalid data on the serialization stage?

src/Elastic.Apm/Metrics/MetricSet.cs
gregkalapos: I agree that it's a little bit odd to pass the logger to `MetricSet`. On the other hand by doing filtering in `MetricSet` we made `MetricSet` into more than just a data transfer object, it has some filtering logic, so I think adding a logger to `MetricSet` is justified with that.

>  Is it better to use ApmAgent.Logger for this?

You mean `Agent.Instance.Logger`? I'd try to avoid using the static instance, because it causes all sorts of problems during testing - when possible I'd just pass an instance.

Overall I'd be fine to just pass a logger instance, another option could be to either just do the same check on `Count` [here](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm/Metrics/MetricsCollector.cs#L137) when we build the `MetricSet`, but that'd mean we have to remember to do this check in case we build `MetricSet`s at other places, or another approach would be to add a flag with `IsFiltered` to `MetricsSet` and check on it. I personally would just stick with this simple solution by passing a logger - or maybe even make it optional if we have too many usages without a logger.

SergeyKleyman:
@vhatsura 

> > so maybe we should extend the existing MetricsCollector's mechanism that disables a metrics provider after it fails a predefined number of times in a row and considers returning NaN or Infinity values as a failure as well?
> 
> It's a good idea, however, I'd prefer to open an additional issue for that.

After considering possible approaches I do think extending the existing MetricsCollector's "stop metrics provider after too many failures" mechanism is the best fit for this PR. I see the following pros:

1. `MetricSet` stays a simple data transfer object and there's no need to try and find a way to pass logger into it.
2. Flooding the log problem is taken care of by the fact that offending metrics provider is not invoked after it returned `NaN` or `Infinity` too many times.
3. We still don't know why CPU metrics provider returns `NaN` and it's possible that invoking it repeatedly has some other negative consequences such as performance impact, memory leak, etc. By not invoking it at all we reduce the risk.
4. If/when we add support for user supplied metric providers they will be invoked by `MetricsCollector` and thus their results will be automatically validated as well.

Please let me know if you see some cons to this approach.

vhatsura:
> @vhatsura
> 
> > > so maybe we should extend the existing MetricsCollector's mechanism that disables a metrics provider after it fails a predefined number of times in a row and considers returning NaN or Infinity values as a failure as well?
> > 
> > 
> > It's a good idea, however, I'd prefer to open an additional issue for that.
> 
> After considering possible approaches I do think extending the existing MetricsCollector's "stop metrics provider after too many failures" mechanism is the best fit for this PR. I see the following pros:
> 
> 1. `MetricSet` stays a simple data transfer object and there's no need to try and find a way to pass logger into it.
> 2. Flooding the log problem is taken care of by the fact that offending metrics provider is not invoked after it returned `NaN` or `Infinity` too many times.
> 3. We still don't know why CPU metrics provider returns `NaN` and it's possible that invoking it repeatedly has some other negative consequences such as performance impact, memory leaks, etc. By not invoking it at all we reduce the risk.
> 4. If/when we add support for user supplied metric providers they will be invoked by `MetricsCollector` and thus their results will be automatically validated as well.
> 
> Please let me know if you see some cons to this approach.

I got your point. Will proceed with changes. 
Thanks a lot for your feedback.  It's really helpful.

gregkalapos:
Hey, 

no pressure, but we'll have a release soon and it'd be nice to offer something for #589 in this new release - as already said above I personally think this PR in its current state is already an acceptable solution for now. 

Options:
- We merge it as it is now (I don't think this breaks anything, I think it's totally acceptable). - if there is agreement, I'm happy to do this. Of course we can follow up and do all the changes agreed above later. - I don't see big functional differences between the current PR and the proposed solution.
- We get a solution very soon as agreed [in this comment](https://github.com/elastic/apm-agent-dotnet/pull/590#issuecomment-552569494).
- We don't address #589 in the next release - which is not the end of the world, but I think we spent already enough effort on this, so probably this is the least favorable option.

Thoughts?

weedaj:
> Hey,
> 
> no pressure, but we'll have a release soon and it'd be nice to offer something for #589 in this new release - as already said above I personally think this PR in its current state is already an acceptable solution for now.
> 
> Options:
> 
> * We merge it as it is now (I don't think this breaks anything, I think it's totally acceptable). - if there is agreement, I'm happy to do this. Of course we can follow up and do all the changes agreed above later. - I don't see big functional differences between the current PR and the proposed solution.
> * We get a solution very soon as agreed [in this comment](https://github.com/elastic/apm-agent-dotnet/pull/590#issuecomment-552569494).
> * We don't address #589 in the next release - which is not the end of the world, but I think we spent already enough effort on this, so probably this is the least favorable option.
> 
> Thoughts?

+1 for merge as is :)

SergeyKleyman:
If there's no pressure then I think we should do it the way that better integrates with the current implementation.

vhatsura:
> Hey,
> 
> no pressure, but we'll have a release soon and it'd be nice to offer something for #589 in this new release - as already said above I personally think this PR in its current state is already an acceptable solution for now.
> 
> Options:
> 
> * We merge it as it is now (I don't think this breaks anything, I think it's totally acceptable). - if there is agreement, I'm happy to do this. Of course we can follow up and do all the changes agreed above later. - I don't see big functional differences between the current PR and the proposed solution.
> * We get a solution very soon as agreed [in this comment](https://github.com/elastic/apm-agent-dotnet/pull/590#issuecomment-552569494).
> * We don't address #589 in the next release - which is not the end of the world, but I think we spent already enough effort on this, so probably this is the least favorable option.
> 
> Thoughts?

@gregkalapos, give me one more day, please, and it will be fixed in a proper way, which was discussed with @SergeyKleyman

SergeyKleyman:
> @gregkalapos, give me one more day, please, and it will be fixed in a proper way, which was discussed with @SergeyKleyman

@vhatsura Vadim really, there's no pressure :) And again thank you very much for your contribution!.

src/Elastic.Apm/Metrics/MetricsCollector.cs
```diff
@@ -108,13 +130,18 @@ internal void CollectAllMetricsImpl()
 					if (samplesFromCurrentProvider != null)
 					{
 						var sampleArray = samplesFromCurrentProvider as MetricSample[] ?? samplesFromCurrentProvider.ToArray();
-						if (sampleArray.Any())
+						if (sampleArray.Length != 0 && sampleArray.All(x => !double.IsNaN(x.KeyValue.Value) && !double.IsInfinity(x.KeyValue.Value)))
```
vhatsura: @SergeyKleyman, @gregkalapos, do we need to filter all samples from metrics provider as it did now in case of any value is either `NaN` or `Infinite`?

src/Elastic.Apm/Metrics/MetricsCollector.cs
```diff
@@ -108,13 +130,18 @@ internal void CollectAllMetricsImpl()
...
 						{
 							_logger.Trace()?.Log("Collected {MetricsProviderName} - adding it to MetricSet", metricsProvider.DbgName);
-							samplesFromAllProviders.AddRange(sampleArray);
+							samples.AddRange(sampleArray);
+							metricsProvider.ConsecutiveNumberOfFailedReads = 0;
+						}
+						else
```
vhatsura: @SergeyKleyman, @gregkalapos, from comment to `ConsecutiveNumberOfFailedReads` property
https://github.com/elastic/apm-agent-dotnet/blob/c25254bd47b881abb0a91f38e882406f4afa6c45/src/Elastic.Apm/Metrics/IMetricsProvider.cs#L14-L18
an empty list is interpreted as fail read, however, it wasn't so
https://github.com/elastic/apm-agent-dotnet/blob/c25254bd47b881abb0a91f38e882406f4afa6c45/src/Elastic.Apm/Metrics/MetricsCollector.cs#L111-L117

What is the truth?

gregkalapos: I think the comment should be the truth, which I guess means we should change ``if (samplesFromCurrentProvider != null)` to `if (samplesFromCurrentProvider != null && samplesFromAllProviders.Any())`  and remove this `if (sampleArray.Any()) `.

Maybe 1 thing I can add here is that when we read CPU usage through perf counters then the 1. call does not return a value because it's based on diffing - I'm not sure, maybe this caused some issues.

PS: this was just a quick response, take it with a grain of salt.

vhatsura: > PS: this was just a quick response, take it with a grain of salt.

Good, this is what I did, however in a little bit different manner.

gregkalapos: > So, I believe it isn't an issue due to we need to return NaN values 5 times before we disable collector. 

Right, I think this is fine, the "1 call does not return value"-problem is not relevant here from what I understand.

src/Elastic.Apm/Metrics/MetricsCollector.cs
gregkalapos: >  do we need to filter all samples from metrics provider as it did now in case of any value is either NaN or Infinite?

Hmm, why don't we just filter out samples with `NaN` and `Infinity`? Maybe we can run into `IMetricsProvider` implementations where some numbers are wrong, but not all, in that case I'd still send the values that are correct numbers. No strong feeling though.

gregkalapos: Oh, I see, this would open up the question that if there is 1 `NaN` or `Infinite` in `sampleArray`, then is it a `ConsecutiveNumberOfFailedReads` or not, and how to handle that, right?

vhatsura: Yeap, I thought about it and this is a reason why I asked it. So, filter out only invalid samples is reasonable for me. Will do changes

vhatsura:
@SergeyKleyman, @gregkalapos, I believe I did all the necessary stuff. Only one thing is left in my mind: not send metric sets without samples. After a few digging it seems that can be done in payload sender. As the PR doesn't touch this functionality, would you mind if I create a separate issue for it?

gregkalapos:
> @SergeyKleyman, @gregkalapos, I believe I did all the necessary stuff. Only one thing is left in my mind: not send metric sets without samples. After a few digging it seems that can be done in payload sender. As the PR doesn't touch this functionality, would you mind if I create a separate issue for it?

That sounds good to me. I leave the final call on this PR to @SergeyKleyman, since technically he is the owner of this, I was just trying to help out.

src/Elastic.Apm/Metrics/MetricsCollector.cs
```diff
@@ -93,8 +93,30 @@ internal void CollectAllMetricsImpl()
 		{
 			_logger.Trace()?.Log("CollectAllMetrics started");
 
-			var samplesFromAllProviders = new List<MetricSample>();
+			var samplesFromAllProviders = CollectMetricsFromProviders();
```
SergeyKleyman: If we add here

```
if (samplesFromAllProviders.IsEmpty())
{
    _logger.Debug()?.Log("No metrics collected (no provider returned valid samples) - nothing to send to APM Server");
    return;
}
```
It will achieve the requirement of not sending empty metric sets. IMHO making payload sender filter out empty metric sets breaks encapsulation - all metrics related logic should stay in metrics related code.
I think this requirement is related to this PR so there is no need for a separate PR.

Of course, `CollectAllMetrics_ShouldDisableProvider_WhenSamplesAreInvalid` test will need to be updated:

```
mockPayloadSender.Metrics.Should().BeEmpty();
```

instead of

```
mockPayloadSender.Metrics.Count.Should().Be(iterations);
mockPayloadSender.Metrics.Should().OnlyContain(x => !x.Samples.Any());
```

vhatsura: done

src/Elastic.Apm/Metrics/MetricSet.cs
```diff
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using System.Linq;
```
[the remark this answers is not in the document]
vhatsura: fixed

Issue #583, opened 2019-11-03
Labels do not allow periods


rgmills:
**Describe the bug**
Setting a transaction/context label with a key that has a period in it, such as `transaction.Context.Labels["account.paid"]` results in a very difficult to parse error from the server.

**To Reproduce**
1. Using a transaction `t`, set a label `account.paid` with any value, say `true`
2. Call `t.End()`
3. See error in logs, something along the lines of 
```
[#/allOf/3/properties/context/properties/tags/additionalProperties] additionalProperties \"account.id\", \"route.controller\", \"route.action\", \"account.paid\" not allowed
```

**Expected behavior**
Ideally, periods would be allowed.  But if a limitation of APM, I'd prefer a more clear error from the client. A couple of options:

- Validation when serializing `Labels` that indicates `.` is not an allowed character.
- A method on `Context` or `ITransaction` to add some guards against this that can error out with an exception.
- An update to documentation that would be clear that `.` (and any other characters) are not allowed.

gregkalapos:
Thanks for reporting this @rgmills.

Just some update on this:
- The APM server currently does not allow `.`, so probably short/mid term we won't be able to allow this. 
- I'm trying to figure out what all other Elastic APM agents do in this case, we usually try to align on these things.

I'll update once I know how all other agents handle this. Overall I totally agree with you that this could be improved.

gregkalapos:
Turns out that other agents simple replace `.` by `_`. I implemented this in #591 and I also added this to the documentation.

rgmills:
Awesome, thanks @gregkalapos

gregkalapos:
Implemented in #591 - this will be in the next release.

PR #591, opened 2019-11-10
Dedotting keys in label

gregkalapos:
Addressing #583.

Asked other agents, they do the same.

I also added this to the docs.

### DIFF
--- a/docs/public-api.asciidoc
+++ b/docs/public-api.asciidoc
@@ -205,6 +205,8 @@ WARNING: Avoid defining too many user-specified labels.
 Defining too many unique fields in an index is a condition that can lead to a
 {ref}/mapping.html#mapping-limit-settings[mapping explosion].
 
+Dots (`.`) in the keys are not allowed. In case you have a `.` in your label key, it will be replaced by `_`. For example `foo.bar` will be stored as `foo_bar`.
+
 [source,csharp]
 ----
 Ageny.Tracer
@@ -346,6 +348,8 @@ TIP: Before using custom labels, ensure you understand the different types of
 WARNING: Avoid defining too many user-specified labels.
 Defining too many unique fields in an index is a condition that can lead to a
 {ref}/mapping.html#mapping-limit-settings[mapping explosion].
+
+Dots (`.`) in the keys are not allowed. In case you have a `.` in your label key, it will be replaced by `_`. For example `foo.bar` will be stored as `foo_bar`.
 
 [source,csharp]
 ----
--- a/docs/public-api.asciidoc
+++ b/docs/public-api.asciidoc
@@ -198,14 +198,14 @@ finally
 ==== `Dictionary<string,string> Labels`
 A flat mapping of user-defined labels with string values. 
 
+If the key contains any special characters (`.`,`*`, `"`), they will be replaced with underscores. For example `a.b` will be stored as `a_b`.
+
 TIP: Before using custom labels, ensure you understand the different types of
 {apm-overview-ref-v}/metadata.html[metadata] that are available.
 
 WARNING: Avoid defining too many user-specified labels.
 Defining too many unique fields in an index is a condition that can lead to a
 {ref}/mapping.html#mapping-limit-settings[mapping explosion].
-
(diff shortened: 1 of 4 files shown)


src/Elastic.Apm/Report/Serialization/LabelsJsonConverter.cs
```diff
@@ -11,7 +11,8 @@ public override void WriteJson(JsonWriter writer, Dictionary<string, string> lab
 			writer.WriteStartObject();
 			foreach (var keyValue in labels)
 			{
-				writer.WritePropertyName(SerializationUtils.TrimToPropertyMaxLength(keyValue.Key));
+				// Labels are trimmed and also de dotted in order to satisfy the Intake API
+				writer.WritePropertyName(SerializationUtils.TrimToPropertyMaxLength(keyValue.Key).Replace('.', '_'));
```
axw: We should also be replacing `*` and `"`, as they JSON Schema disallows those as well:

https://github.com/elastic/apm-server/blob/f633a79345d3333327d7de247fdf9901e0df2643/docs/spec/tags.json#L7

docs/public-api.asciidoc
```diff
@@ -205,6 +205,8 @@ WARNING: Avoid defining too many user-specified labels.
 Defining too many unique fields in an index is a condition that can lead to a
 {ref}/mapping.html#mapping-limit-settings[mapping explosion].
 
+Dots (`.`) in the keys are not allowed. In case you have a `.` in your label key, it will be replaced by `_`. For example `foo.bar` will be stored as `foo_bar`.
```
SergeyKleyman: Using `foo.bar` as an example of a label's key might be confusing as the following code sample has `foo` as the key and `bar` as the value.

src/Elastic.Apm/Api/IExecutionSegment.cs
```diff
@@ -31,6 +31,8 @@ public interface IExecutionSegment
 
 		/// <summary>
 		/// A flat mapping of user-defined labels with string values.
+		/// Dots (<code>.</code>) in the keys are not allowed. In case you have a <code>.</code> in your label key, it will be replaced by <code>_</code>.
+		/// For example <code>foo.bar</code> will be stored as <code>foo_bar</code>.
```
SergeyKleyman: It might be worth clarifying the term `stored` since the key is still "stored" with dots in `Labels` `Dictionary`. Maybe `stored in Elasticsearch`?

src/Elastic.Apm/Report/Serialization/LabelsJsonConverter.cs
gregkalapos: Oh, right! Added those, thanks.

docs/public-api.asciidoc
gregkalapos: Changed to `a.b`.

src/Elastic.Apm/Api/IExecutionSegment.cs
gregkalapos: 👍  - added it to the text.

src/Elastic.Apm/Report/Serialization/LabelsJsonConverter.cs
```diff
@@ -11,7 +11,11 @@ public override void WriteJson(JsonWriter writer, Dictionary<string, string> lab
 			writer.WriteStartObject();
 			foreach (var keyValue in labels)
 			{
-				writer.WritePropertyName(SerializationUtils.TrimToPropertyMaxLength(keyValue.Key));
+				// Labels are trimmed and also de dotted in order to satisfy the Intake API
+				writer.WritePropertyName(SerializationUtils.TrimToPropertyMaxLength(keyValue.Key)
+					.Replace('.', '_')
+					.Replace('*', '_')
```
SergeyKleyman: It might be worth updating the docs to mention all the characters that are replaced.

gregkalapos: I changed the doc - the Go doc had 1 nice sentence covering this. The Java doc does not even talk about this details.

PR #592, opened 2019-11-10
Add instrumentation for sqlclient

### LINKED ISSUE #114: Investigate how to monitor ADO.NET calls with “library-only” approach
Among data access layer technologies currently we only support Entity Framework (EF) Core which less popular in classic ASP.NET applications. ADO.NET is much more popular for classic ASP.NET applications - either direct ADO.NET calls or usage of classic Entity Framework which in turn is implemented on top of ADO.NET. So we would like to investigate:

- If  it's possible to monitor ADO.NET calls with “library-only” approach (i.e., without resorting to Profiling API, etc.) 
- How much details we can get employing “library-only” approach

vhatsura:
resolves #114 

`SqlClientDiagnosticListener` allows processing events from `Microsoft.Data.SqlClient` and `System.Data.SqlClient`. As payload data spreads as anonymous objects, it needed to use reflection to access properties. [Diagnostic source users guide suggests using property fetchers](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/DiagnosticSourceUsersGuide.md#decoding-payloads) to reduce the overhead of reflection. To measure the difference in performance between reflection and property fetchers I wrote a benchmark. You can run it on your PC, but I got the following results on my machine:

```
BenchmarkDotNet=v0.11.4, OS=Windows 10.0.19002
Intel Core i7-7700 CPU 3.60GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=2.1.505
  [Host]     : .NET Core 2.1.9 (CoreCLR 4.6.27414.06, CoreFX 4.6.27415.01), 64bit RyuJIT
  DefaultJob : .NET Core 2.1.9 (CoreCLR 4.6.27414.06, CoreFX 4.6.27415.01), 64bit RyuJIT


|             Method |      Mean |     Error |    StdDev |
|------------------- |----------:|----------:|----------:|
|         Reflection | 348.15 ns | 8.4263 ns | 8.6532 ns |
|    PropertyFetcher |  15.38 ns | 0.3566 ns | 0.3503 ns |
|     CascadeFetcher |  16.10 ns | 0.4180 ns | 0.7953 ns |
| CachedPropertyInfo | 146.53 ns | 2.9561 ns | 4.2396 ns |

```

Due to property fetcher needs to use specific type, we need to have two property fetcher sets: one for `System.Data.SqlClient` and another for `Microsoft.Data.SqlClient`

To test the listener I use [`TestEnvironment.Docker`](https://github.com/Deffiss/testenvironment-docker), which allows running docker images from tests. It can be an issue for your CI due to docker needs to be installed on CI machines. I know that some CI tools allow using services for running docker images during the build, however, I don't sure that it's an option in case of Jenkins. Also, it cannot be repeated on local machine without additional effort before each test run.
Another issue with `TestEnvironment.Docker` is missing a strong name.

P.S.: I think that http diagnostic listeners can be rewritten to use property fetchers instead of reflection.


vhatsura:
> Now, enabling SqlClientDiagnosticListener by default will introduce a new problem which is when EF or EF Core is used in combination with SQL Server

This is the exact reason why I didn't enable this listener by default.

tomap:
Hi, just to add my point of view regarding this:
I was using AppInsight, which automaticaly tacks SQL request (SQL Client) but noting particular is done for EntityFramework.
So, I I were to enable something by default, that would be SQLClient monitoring, and not entityframework (but that is a breacking change :( )
You could do it in a new major version ?

gregkalapos:
> Hi, just to add my point of view regarding this:
> I was using AppInsight, which automaticaly tacks SQL request (SQL Client) but noting particular is done for EntityFramework.
> So, I I were to enable something by default, that would be SQLClient monitoring, and not entityframework (but that is a breacking change :( )
> You could do it in a new major version ?

Thanks for chiming in @tomap. 

My concern with that approach is that in case an app uses EF Core, then it'll only work if it's uses MS SQL as the database - or maybe there are some others that trigger SqlClient, but definitely not all of them. So e.g. EF Core with SQLite, or Oracle would not work by default - so as you correctly point out this would be a breaking change, and I'm not sure it's a better default.

The way I think about solving this is to add some ways to these components to talk to each other: so it'd be nice if the EF Core monitoring part could know that SqlClient monitoring is turned on, so it'd not capture calls from SqlClient. I suspect we'll have the same problem with other parts (e.g. `HttpClient` and other HTTP based frameworks, let's say GraphQL, or we already have something similar with our Elasticsearch client integration (not merged yet), but in that case having the `HttpClient` calls actually make sense).

So, the point of my rant is: I think it'd be nicer to make these components work smartly when all of them are turned on.

gregkalapos:
Just a short update:

I think this is ok to merge in it's current stage - we can do the integration later. I'll wait with #647, that should land very soon and then merge this one.

SergeyKleyman:
Since `span.context.destination.*` fields are deduced for DB spans this feature will also affect https://github.com/elastic/apm-agent-dotnet/issues/650 and https://github.com/elastic/apm-agent-dotnet/issues/611.

src/Elastic.Apm.SqlClient/SqlClientDiagnosticListener.cs
```diff
@@ -0,0 +1,174 @@
...
+						break;
+				}
+			}
+		}
+
+		private void HandleStartCommand(object payloadData, PropertyFetcherSet propertyFetcherSet)
+		{
+			try
+			{
+				var transaction = _apmAgent.Tracer.CurrentTransaction;
```
SergeyKleyman: It seems most of this code is almost the same as [code in `DbSpanCommon` ](https://github.com/elastic/apm-agent-dotnet/blob/1.2.0/src/Elastic.Apm/Model/DbSpanCommon.cs). Refactoring all this code into one location will simplify future features such as https://github.com/elastic/apm-agent-dotnet/issues/650 and https://github.com/elastic/apm-agent-dotnet/issues/611.

vhatsura: I'll look into it

src/Elastic.Apm/Helpers/PropertyFetcher.cs
```diff
@@ -0,0 +1,76 @@
...
+			_propertyName = propertyName;
+		}
+
+		public virtual object Fetch(object obj)
+		{
+			if (_innerFetcher == null)
+			{
+				var type = obj.GetType().GetTypeInfo();
+				var property = type.DeclaredProperties.FirstOrDefault(p =>
+					string.Equals(p.Name, _propertyName, StringComparison.InvariantCultureIgnoreCase));
```
SergeyKleyman: I think [the recommendation](https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings) is to use `StringComparison.OrdinalIgnoreCase` to compare "symbols":
> Use the non-linguistic StringComparison.Ordinal or StringComparison.OrdinalIgnoreCase values instead of string operations based on CultureInfo.InvariantCulture when the comparison is linguistically irrelevant (symbolic, for example).

src/Elastic.Apm.SqlClient/SqlClientDiagnosticListener.cs
vhatsura: @SergeyKleyman, I changed it to use `DbSpanCommon`. But I have a question, Do we have any reason to fill span info from db command during end operation instead of start (as it's done right now in db span common)? In case of exception, the end event isn't called and we don't capture an invalid db statement (if it's the reason), which would be nice to have.

src/Elastic.Apm/Helpers/PropertyFetcher.cs
vhatsura: Thanks for pointing it. Fixed.

src/Elastic.Apm.SqlClient/SqlClientDiagnosticListener.cs
vhatsura: @SergeyKleyman, never mind. I don't see that you actually call `DbSpanCommon.End` from error  event

SergeyKleyman: I think capturing whatever possible in `Start` instead of waiting for `End` might be a good idea. I even considered if this change might affect unusual use cases such as [this one](https://discuss.elastic.co/t/get-currentspan-in-net-apm-agent/211741/5) and I think it shouldn't.

vhatsura:
@gregkalapos, I'm fine with it. Also, I have thoughts how to enable it's automatically without duplication. Will try it after our Christmas (Orthodox) holidays.

gregkalapos:
Oh, I totally overlooked this part from the description:

> To test the listener I use TestEnvironment.Docker, which allows running docker images from tests. It can be an issue for your CI due to docker needs to be installed on CI machines. I know that some CI tools allow using services for running docker images during the build, however, I don't sure that it's an option in case of Jenkins. Also, it cannot be repeated on local machine without additional effort before each test run.
Another issue with TestEnvironment.Docker is missing a strong name.

Tests are failing in CI due to ⬆️, we should figure this our be fore we merge the PR. I'll see if I can come up with something.

v1v:
Regarding the CI, I'm working on some kind of DinD approach for the linux build:
- https://github.com/elastic/apm-agent-dotnet/pull/678/commits/5eedc8291d38195f8b0d36e3e66704c7fb4b2270

I used this PR changeset to create https://github.com/elastic/apm-agent-dotnet/pull/678 and be able to test it in the CI. Once it works then I'll close it and cherry-pick my changes.

Does it sound like a good plan?

gregkalapos:
> Regarding the CI, I'm working on some kind of DinD approach for the linux build:
> 
> * [5eedc82](https://github.com/elastic/apm-agent-dotnet/commit/5eedc8291d38195f8b0d36e3e66704c7fb4b2270)
> 
> I used this PR changeset to create #678 and be able to test it in the CI. Once it works then I'll close it and cherry-pick my changes.
> 
> Does it sound like a good plan?

I'm ok with this plan. We talked about this with @v1v. 

@vhatsura: @v1v did some docker magic and we have [a PR](https://github.com/vhatsura/apm-agent-dotnet/pull/2) open against your PR - that'd solve the CI issues. Please take a quick look, if you are ok with that please merge and then I'll merge this PR into our master.

vhatsura:
@gregkalapos, I merged @v1v changes.

v1v:
I missed a particular commit in my end :S

v1v:
aha!

image

This particular PR got the changes for the Jenkinsfile but our CI doesn't allow to run the Jenkinsfile from this PR if the owner is not a member of the elastic organization. :(

This PR was already validated with the [build](https://apm-ci.elastic.co/job/apm-agent-dotnet/job/apm-agent-dotnet-mbp/view/change-requests/job/PR-678/3/). PR-678 was cloned from this PR to work in parallel with the CI setup.

I ran the below commands to validate if both branches are equivalent.

```bash

$ git branch
  feature/support-linux-with-DinD
* instrumentation/sqlclient
  master

$ git log -1
commit 03c2c30e181958ceedfa0914311f8903030447a3 (HEAD -> instrumentation/sqlclient, vhatsura/instrumentation/sqlclient)
Author: Victor Martinez <victormartinezrubio@gmail.com>
Date:   Wed Jan 8 17:12:17 2020 +0000

    ci(jenkins): support test containers in the CI (#2)

$ git diff feature/support-linux-with-DinD
```

Therefore, even though the build fails it will work as soon is merged into the master branch.  So I'd say we are good to go

gregkalapos:
Thanks @v1v, ok, I merge now.

Issue #596, opened 2019-11-13
Agent initialisation error due to Lazy logic in Full Framework Elastic APM module


NeoXtreem:
Using the current version (1.1.2) of `Elastic.Apm.AspNetFullFramework` and following the configuration instructions [here](https://www.elastic.co/guide/en/apm/agent/dotnet/current/setup.html#setup-asp-net), I get following exception on startup:

> The singleton APM agent has already been instantiated and can no longer be configured

I've stepped into the Elastic APM code, and it appears that `ElasticApmModule.Init` calls `Agent.Instance` which has the effect of causing `Lazy.IsValueCreated` to always return true thereafter.  `ElasticApmModule.Init` then continues on the next line to indirectly call `Agent.Setup` which throws an exception if `Lazy.IsValueCreated` returns true (which it does because of the previous line).  Therefore, this is a showstopping bug as the Full Framework Elastic APM module simply cannot be used for ASP.NET with this initialisation logic.

It appears @chrisvanderpennen [experienced a similar problem](https://github.com/elastic/apm-agent-dotnet/issues/244#issuecomment-541474012) on #244, and also @OlegUfaev [here](https://github.com/elastic/apm-agent-dotnet/issues/441#issuecomment-528323128) on #441, but that relates to ASP.NET Core and can easily be circumvented with a try-catch as @pburrows suggested [here](https://github.com/elastic/apm-agent-dotnet/issues/441#issuecomment-530522330).  But in my case, I'm using ASP.NET with .NET Framework, so I don't have that option as config is done through **Web.config**.

Please can this be fixed, and a workaround provided in the meantime?

**Stack trace:**

> [Exception: The singleton APM agent has already been instantiated and can no longer be configured]
>    Elastic.Apm.Agent.Setup(AgentComponents agentComponents) +139
>    Elastic.Apm.AspNetFullFramework.<>c__DisplayClass24_0.<InitOnceForAllInstancesUnderLock>b__0() +88
>    Elastic.Apm.Helpers.<>c__DisplayClass6_0.<Init>b__0() +78
>    System.Threading.LazyInitializer.EnsureInitializedCore(T& target, Boolean& initialized, Object& syncLock, Func`1 valueFactory) +139
>    Elastic.Apm.Helpers.LazyContextualInit.Init(Action initAction) +300
>    Elastic.Apm.Helpers.IfNotInitedHelper.Init(Action initAction) +56
[...]

PR #597, opened 2019-11-13
Make ElasticApmModule work with existing Agent.Instance

SergeyKleyman:
Even though we still don't know the root cause for #596 - this fix should address "agent crashing application" part of #596.

### DIFF
--- a/src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
+++ b/src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
@@ -35,6 +35,22 @@ namespace Elastic.Apm.AspNetFullFramework
 
 		public void Init(HttpApplication httpApp)
 		{
+			try
+			{
+				InitImpl(httpApp);
+			}
+			catch (Exception ex)
+			{
+				const string linePrefix = "Elastic APM .NET Agent: ";
+				System.Diagnostics.Trace.WriteLine($"{linePrefix}[CRITICAL] Exception thrown by {nameof(ElasticApmModule)}.{nameof(InitImpl)}."
+					+ Environment.NewLine + linePrefix + $"+-> Exception: {ex.GetType().FullName}: {ex.Message}"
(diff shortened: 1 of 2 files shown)


src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
```diff
@@ -305,5 +320,19 @@ private static AgentComponents BuildAgentComponents(string dbgInstanceName)
...
+
+		private static void SafeAgentSetup(AgentComponents agentComponents)
+		{
+			try
+			{
+				Agent.Setup(agentComponents);
+			}
+			catch (Agent.InstanceAlreadyCreatedException ex)
+			{
+				Agent.Instance.Logger.Error()?.LogException(ex, "Agent's singleton instance already exists before call to"
```
gregkalapos: I'm not sure users would understand this error message. Would maybe something like this easier to understand for someone who is not familiar with the agent's code? 

```
The Elastic APM agent was already initialized before call to {nameof(ElasticApmModule)}.{nameof(Init)} - {nameof(ElasticApmModule)} will use existing instance
```

Also, I'm not sure about the `even though it might lead to using incorrect configuration...` part. Maybe I'd just say 

```
This could lead to unexpected behaviour.
```

SergeyKleyman: How about?

>  This might lead to unexpected behavior (for example agent using incorrect configuration source such as environment variables instead of Web.config).

gregkalapos: That sounds good to me. 

Would we also change the 1. part? My thinking behind that is that explicitly saying "agent was already initialized" can be easier to understand than "Agent's singleton instance already exists" because that would make them wonder about what singleton it is and how they could change it.

SergeyKleyman: Agreed.

PR #605, opened 2019-11-15
Increase ASP.NET FullFramework version

gregkalapos:
Preparing `1.1.3` release to offer a workaround for #596. 

We only increment the version of `Elastic.Apm.AspNetFullFramework`, since the workaround is local to that project, it still depends on `Elastic.Apm` version `1.1.2`.

cc @SergeyKleyman jfyi.


gregkalapos:
Turns our #596 was specific to a fork which had code that is not in the released package. 

So we skip this patch release, all the improvements will be in the next minor release.

PR #606, opened 2019-11-15
Update ElasticApmAgent.sln

gregkalapos:
I messed up the merging of #425, there is a missing `EndProject` in the `sln` file, fixing here.


gregkalapos:
Merging without review, this is a 1 liner, and CI verified it fixes the problem.

Issue #604, opened 2019-11-15
ASP.NET Core 3.0 TypeLoadExcpetion


gregkalapos:
#546 is about moving to 3.0, but I did already some manual testing (also with ASP.NET Core 3.0) and did not run into any issues.

Still, we got this reported:

```
System.TypeLoadException: Could not load type 'Microsoft.AspNetCore.Http.Internal.BufferingHelper' from assembly 'Microsoft.AspNetCore.Http, Version=3.0.0.0"
```

tmksoft:
You can provoke the bug by setting the following in the appsettings.json:

  "ElasticApm": {
    "SecretToken": ....,
    "ServerUrls": ....,
    "ServiceName": ....,
    "CaptureHeaders": true,
    "CaptureBody": "all",
    "CaptureBodyContentTypes": "application/json"
  }

gregkalapos:
## Impact 
Request body capturing does not work on ASP.NET Core 3.0 - The good thing: we do not crash the application, the exception is handled within the agent - so no drastic impact on the observed application - only the overhead of the exception which is handled. Furthermore the request body is not captured. The transaction and every other field is still captured fine if this bug happens.

More logs with callstack:

```
  {ApmMiddleware} Exception thrown while trying to fill request context for sampled transaction e8ea233fcbc04caa
System.TypeLoadException: Could not load type 'Microsoft.AspNetCore.Http.Internal.BufferingHelper' from assembly 'Microsoft.AspNetCore.Http, Version=3.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'.
   at Elastic.Apm.AspNetCore.Extensions.HttpRequestExtensions.ExtractRequestBody(HttpRequest request, IApmLogger logger)
   at Elastic.Apm.AspNetCore.ApmMiddleware.GetRequestBody(HttpRequest request)
   at Elastic.Apm.AspNetCore.ApmMiddleware.FillSampledTransactionContextRequest(HttpContext context, Transaction transaction)
```

## Workaround 

Turn off request body capturing

## Why does this happen

We used the `HttpRequest.EnableRewind()` method to enable reading the request body, so we can capture it. Turns out that this method is gone from ASP.NET Core 3.0. 

## Solution 
Proposed solution in https://github.com/elastic/apm-agent-dotnet/pull/608

vhatsura:
[...]

PR #608, opened 2019-11-18
Request body capturing - use EnableBuffering instead of EnableRewind and make it async

gregkalapos:
Addressing #604 and making request body capturing async 

## EnableBuffering

Turns our that the `HttpRequest.EnableRewind()` method is gone in ASP.NET Core 3.0. We need that to be able to read the request body.

We pull the `HttpRequest` type through the `Microsoft.AspNetCore.Http.Abstractions` package. 

Now, `HttpRequest.EnableRewind()` is gone, but a new method, `HttpRequest.EnableBuffering()` was introduced and that is available since version `2.1` of the `Microsoft.AspNetCore.Http.Abstractions` package.

Until this point we were at `Microsoft.AspNetCore.Http.Abstractions` `2.0`, so we need to update in order to use the new `EnableBuffering` which works both on `2.1+` and `3.0`. 

With that we'd drop support for ASP.NET Core 2.0 and we'd only support it from `2.1`+. I think this is fine, since Microsoft also dropped support for it already (see [here](https://dotnet.microsoft.com/platform/support/policy/dotnet-core)) - `2.1` is the latest supported version.

## Making request body capturing async 

This was actually already a problem with `EnableRewind` - reading the request body happened synchronously and that can lead to thread starvation. The new method (`EnableBuffering`) by default throws an exception if you try to read the body synchronously - sync reading can be enabled (`AllowSynchronousIO`), but rather than "workarounding" I thought this is a good time to solve the real problem.

Therefore I also made reading of the request body async as it should be.

~I know #577 did some changes in this area (although as I mentioned in that PR, I think the best would be to not change anything on the capturing itself, since that PR should only be focused on remote config - at least that would make our life easier). Anyway, whichever is merged earlier, I'm happy to help with merging and resolving conflicts - it won't be that bad.~ 
Update: #577 merged to `master` and I merged `master` into this PR, all conflicts resolved ✅


src/Elastic.Apm.AspNetCore/DiagnosticListener/AspNetCoreDiagnosticListener.cs
```diff
@@ -28,7 +28,7 @@ public AspNetCoreDiagnosticListener(IApmAgent agent)
 
 		public void OnError(Exception error) { }
 
-		public void OnNext(KeyValuePair<string, object> kv)
+		public async void OnNext(KeyValuePair<string, object> kv)
```
gregkalapos: This is questionable. We need to call an async method here. 

Here is what I think:
This is a fire and forget call, meaning the caller will move on. 
2 things can happen: 
- this `OnNext` method finishes before the transaction is finished -> all good, we captured the body. 
- the transaction finishes before we set `Transaction.Request.Body` and in that case the body won't be captured. This is not cool. 

Alternatively we can revert back to async reading in this case, but that would be very unfortunate, because with that we again risk thread starvation. 

Another option would be to mark the transaction here and not do the reading and have a check in `ApmMiddleware.InvokeAsync` where we can call `await CollectRequestInfoAsync`. 

Happy to hear opinions.

gregkalapos:
We have some CI issues - when I build locally, everything is fine. 

```
[2019-11-18T15:31:37.409Z]          Restore failed in 86.46 ms for C:\Users\jenkins\workspace\tnet_apm-agent-dotnet-mbp_PR-608\apm-agent-dotnet\test\Elastic.Apm.Tests\Elastic.Apm.Tests.csproj.
[2019-11-18T15:31:37.409Z]          Resolving conflicts for .NETCoreApp,Version=v2.1...
[2019-11-18T15:31:37.409Z]          System.ArgumentNullException: Value cannot be null.
[2019-11-18T15:31:37.409Z]          Parameter name: path
[2019-11-18T15:31:37.409Z]             at System.IO.Path.GetFullPath(String path)
[2019-11-18T15:31:37.409Z]             at NuGet.Commands.RestoreCommand.GetAssetsFilePath(LockFile lockFile)
[2019-11-18T15:31:37.409Z]             at NuGet.Commands.RestoreCommand.ExecuteAsync(CancellationToken token)
[2019-11-18T15:31:37.409Z]             at NuGet.Commands.RestoreRunner.ExecuteAsync(RestoreSummaryRequest summaryRequest, CancellationToken token)
[2019-11-18T15:31:37.409Z]             at NuGet.Commands.RestoreRunner.ExecuteAndCommitAsync(RestoreSummaryRequest summaryRequest, CancellationToken token)
[2019-11-18T15:31:37.409Z]             at NuGet.Commands.RestoreRunner.CompleteTaskAsync(List`1 restoreTasks)
[2019-11-18T15:31:37.409Z]             at NuGet.Commands.RestoreRunner.RunAsync(IEnumerable`1 restoreRequests, RestoreArgs restoreContext, CancellationToken token)
[2019-11-18T15:31:37.409Z]             at NuGet.Commands.RestoreRunner.RunAsync(RestoreArgs restoreContext, CancellationToken token)
[2019-11-18T15:31:37.409Z]             at NuGet.Build.Tasks.RestoreTask.ExecuteAsync(ILogger log)
[2019-11-18T15:31:37.409Z]        Done executing task "RestoreTask" -- FAILED.
[2019-11-18T15:31:37.409Z]      1>Done building target "Restore" in project "ElasticApmAgent.sln" -- FAILED.
[2019-11-18T15:31:37.409Z]      1>Done Building Project "C:\Users\jenkins\workspace\tnet_apm-agent-dotnet-mbp_PR-608\apm-agent-dotnet\ElasticApmAgent.sln" (Restore target(s)) -- FAILED.
[2019-11-18T15:31:37.409Z] 
```

I'll dig deeper. Maybe related to the version bump - Probably it won't change the logic in the C# code.

xitaocrazy:
Hi @gregkalapos 

In issue #604 you suggested to turn off request body capturing, but I didn't find this configuration in the documentation, sincerely I did not know that the Agent already capture this information, in the past when I asked about it, the answer was that the Agent didn't capture it. I think that was added this behavior, but nothing was modified in the documentation. Can you help me? there is some key to set in environment variables as  ELASTIC_APM_CAPTURE_HEADERS ?

gregkalapos:
> Hi @gregkalapos
> 
> In issue #604 you suggested to turn off request body capturing, but I didn't find this configuration in the documentation, sincerely I did not know that the Agent already capture this information, in the past when I asked about it, the answer was that the Agent didn't capture it. I think that was added this behavior, but nothing was modified in the documentation. Can you help me? there is some key to set in environment variables as ELASTIC_APM_CAPTURE_HEADERS ?

Hi @xitaocrazy,

Oh, you are right, our documentation is not uptodate on this - but we fixed this recently, the documentation on `master` already has this, you can find it [here](https://www.elastic.co/guide/en/apm/agent/dotnet/master/config-http.html#config-capture-body). 

So, the current status is that it works on ASP.NET Core 2.0, 2.1, and 2.2, but does not work in 3.0, which should be fixed in this PR. The docs will be merged with the next release and then you'll find it also in the doc of the latest release.

[Here](https://github.com/elastic/apm-agent-dotnet/blob/master/sample/SampleAspNetCoreApp/appsettings.json#L12) you can find a sample `appsettings.json` file to turn on the setting. By default it's turned on. Also keep in mind that this adds overhead.

Sorry about causing confusion.

xitaocrazy:
> > Hi @gregkalapos
> > In issue #604 you suggested to turn off request body capturing, but I didn't find this configuration in the documentation, sincerely I did not know that the Agent already capture this information, in the past when I asked about it, the answer was that the Agent didn't capture it. I think that was added this behavior, but nothing was modified in the documentation. Can you help me? there is some key to set in environment variables as ELASTIC_APM_CAPTURE_HEADERS ?
> 
> Hi @xitaocrazy,
> 
> Oh, you are right, our documentation is not uptodate on this - but we fixed this recently, the documentation on `master` already has this, you can find it [here](https://www.elastic.co/guide/en/apm/agent/dotnet/master/config-http.html#config-capture-body).
> 
> So, the current status is that it works on ASP.NET Core 2.0, 2.1, and 2.2, but does not work in 3.0, which should be fixed in this PR. The docs will be merged with the next release and then you'll find it also in the doc of the latest release.
> 
> [Here](https://github.com/elastic/apm-agent-dotnet/blob/master/sample/SampleAspNetCoreApp/appsettings.json#L12) you can find a sample `appsettings.json` file to turn on the setting. By default it's turned on. Also keep in mind that this adds overhead.
> 
> Sorry about causing confusion.

Hi @gregkalapos 

Thank you for your attention, I will add this configuration here.

src/Elastic.Apm.AspNetCore/DiagnosticListener/AspNetCoreDiagnosticListener.cs
gregkalapos: Yeah, this was a problem. I moved the reading of the body to the `ApmMiddleware`. The CI fails mentioned [here](https://github.com/elastic/apm-agent-dotnet/pull/608#issuecomment-555076734) were caused by this.

src/Elastic.Apm.AspNetCore/DiagnosticListener/AspNetCoreDiagnosticListener.cs
```diff
@@ -33,15 +24,12 @@ public void OnNext(KeyValuePair<string, object> kv)
 
 			var exception = kv.Value.GetType().GetTypeInfo().GetDeclaredProperty("exception").GetValue(kv.Value) as Exception;
 
-			var transaction = (Transaction)_agent.Tracer.CurrentTransaction;
```
gregkalapos: No more cast 💪

src/Elastic.Apm.AspNetCore/Extensions/TransactionExtentions.cs
```diff
@@ -13,26 +14,34 @@ internal static class TransactionExtensions
 		/// </summary>
 		/// <param name="transaction">Transaction object</param>
 		/// <param name="isForError">Is request body being captured for error (otherwise it's for transaction)</param>
-		/// <param name="request">Current http request</param>
+		/// <param name="httpRequest">Current http context containing the http request</param>
```
SergeyKleyman: It sounds quite confusing - `httpRequest`'s type is `HttpRequest ` so why is it `http context`?

test/Elastic.Apm.AspNetCore.Tests/BodyCapturingTests.cs
```diff
@@ -193,7 +193,7 @@ async Task TestBodyCapture(string body, bool isError, bool shouldRequestBodyBeCa
 				var transaction = sutEnv.MockPayloadSender.FirstTransaction;
 				object capturedRequestBody = null;
 				transaction.IsSampled.Should().Be(isSampled);
-				transaction.IsContextCreated.Should().Be(isSampled, $"body: {body}");
+				//transaction.IsContextCreated.Should().Be(isSampled, $"body: {body}");
```
SergeyKleyman: I assume this is a temporary change before fixing the failing test, right?

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
```diff
@@ -42,21 +42,38 @@ public ApmMiddleware(RequestDelegate next, Tracer tracer, IApmAgent agent)
...
 				await _next(context);
 			}
-			catch (Exception e) when (transaction != null && ExceptionFilter.Capture(e, transaction, context, _logger))
-			{ }
+			catch (Exception e) when (transaction != null)
+			{
+				transaction.CaptureException(e);
+				// It'd be nice to have this in an exception filter, but that would force us capturing the request body synchronously.
+				// Therefore we rather unwind the stack in the catch block and call the async method.
+				if (context != null && transaction.IsCaptureRequestBodyEnabled(true))
```
SergeyKleyman: Isn't call to `IsCaptureRequestBodyEnabled` redundant here considering that `CollectRequestBody` calls `IsCaptureRequestBodyEnabled` among other checks before actually reading the body?

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
```diff
@@ -42,21 +42,38 @@ public ApmMiddleware(RequestDelegate next, Tracer tracer, IApmAgent agent)
...
+					await transaction.CollectRequestBody(true, context.Request, _logger);
+
+				throw;
+			}
 			finally
 			{
+				// In case an error handler middleware is registered, the catch block above won't be executed, because the
+				// error handler handles all the exceptions - in this case, based on the response code and the config, we may capture the body here
+				if (transaction != null && context?.Response.StatusCode >= 300 && transaction.Context?.Request?.Body is string body
```
SergeyKleyman: Is "not generating error when error handler middleware is registered" an issue in ASP.NET Core 2.0 + "agent before this PR's changes" as well? Is this change related to the purpose of this PR?

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
```diff
@@ -42,21 +42,38 @@ public ApmMiddleware(RequestDelegate next, Tracer tracer, IApmAgent agent)
...
 		public async Task InvokeAsync(HttpContext context)
 		{
-			var transaction = StartTransaction(context);
+			var transaction = await StartTransactionAsync(context);
 
 			try
 			{
+				// We need to reset the current transaction (which is async local), since the StartTransactionAsync call above ended with all its
+				// child spans at this point, therefore need to set the current transaction here again.
+				_tracer.CurrentExecutionSegmentsContainer.CurrentTransaction = transaction;
```
SergeyKleyman: IMHO this breaks encapsulation - a possible alternative would be to extract `await FillSampledTransactionContextRequest` from `StartTransactionAsync` which will revert `StartTransaction` to being sync.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
```diff
@@ -175,10 +192,10 @@ private void FillSampledTransactionContextRequest(HttpContext context, Transacti
 						Encrypted = context.Request.IsHttps, RemoteAddress = context.Connection?.RemoteIpAddress?.ToString()
 					},
 					HttpVersion = GetHttpVersion(context.Request.Protocol),
-					Headers = GetHeaders(context.Request.Headers, transaction.ConfigSnapshot),
+					Headers = GetHeaders(context.Request.Headers, transaction.ConfigSnapshot)
 				};
 
-				transaction.CollectRequestBody( /* isForError: */ false, context.Request, _logger);
+				await transaction.CollectRequestBody(false, context.Request, _logger);
```
SergeyKleyman: IMHO `/* isForError: */` is helpful - otherwise it's not clear what `false` argument means.

src/Elastic.Apm.AspNetCore/Extensions/TransactionExtentions.cs
```diff
@@ -13,26 +14,34 @@ internal static class TransactionExtensions
...
 		/// <param name="logger">Logger object</param>
-		internal static void CollectRequestBody(this Transaction transaction, bool isForError, HttpRequest request, IApmLogger logger)
+		internal static async Task CollectRequestBody(this Transaction transaction, bool isForError, HttpRequest httpRequest, IApmLogger logger)
 		{
 			if (!transaction.IsSampled) return;
 
+			if(httpRequest == null) return;
```
SergeyKleyman: IMHO we should avoid passing `null` as much as possible - it will make it easier for us to use C# 8 Nullable Reference Types (hopefully in the near future). It's worth checking if this method is ever called when `httpRequest` is `null`.

src/Elastic.Apm.AspNetCore/Elastic.Apm.AspNetCore.csproj
```diff
@@ -12,16 +12,17 @@
...
+    <PackageReference Include="Microsoft.AspNetCore.Http.Extensions" Version="2.1.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" Version="2.1.0" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="2.1.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Routing.Abstractions" Version="2.1.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.1.0" />
   </ItemGroup>
   <ItemGroup>
     <Folder Include="Config\" />
     <Folder Include="DiagnosticListener\" />
+    <Folder Include="Helpers" />
```
SergeyKleyman: There's no `Helpers` folder anymore.

src/Elastic.Apm.AspNetCore/Extensions/RequestExtentions.cs
```diff
@@ -16,26 +16,25 @@ public static class HttpRequestExtensions
 		/// <param name="request"></param>
 		/// <param name="logger"></param>
 		/// <returns></returns>
-		public static string ExtractRequestBody(this HttpRequest request, IApmLogger logger)
+		public static async Task<string> ExtractRequestBodyAsync(this HttpRequest request, IApmLogger logger)
```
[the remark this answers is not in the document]
gregkalapos: Right, it is :( 

This is very similar to the discussion [here](https://github.com/elastic/apm-agent-dotnet/pull/577#discussion_r340336900)

I made the class internal, with this we make 2 `public` classes to `internal` - one in #577, one here, which should have been `internal` from the beginning. I think it's reasonable to fix both.

src/Elastic.Apm.AspNetCore/Extensions/TransactionExtentions.cs
```diff
@@ -13,26 +14,34 @@ internal static class TransactionExtensions
...
+
+			if (transaction.IsContextCreated && transaction.Context.Request.Body != null
```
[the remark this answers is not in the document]
gregkalapos: 👍  Right, fixed that - as it seems it was not easy to resolve all the conflicts :)

src/Elastic.Apm.AspNetCore/Extensions/RequestExtentions.cs
```diff
@@ -16,26 +16,25 @@ public static class HttpRequestExtensions
...
 		{
-			string body = null;
+			var body = Apm.Consts.Redacted;
```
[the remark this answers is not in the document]
gregkalapos: I'm fine with that - to me this does not seem to be important.

src/Elastic.Apm.AspNetCore/Extensions/TransactionExtentions.cs
gregkalapos: That's merge leftover, fixed it - thanks.

test/Elastic.Apm.AspNetCore.Tests/BodyCapturingTests.cs
gregkalapos: Oh, this was a leftover.

I'm not sure if we changed the behaviour here - Anyway, I made the test pass and uncommented the line.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
gregkalapos: Right, yes, it is - (after I merged your change :) ) 👍 

I removed these checks.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
gregkalapos: Sorry, I don't get the first part of the question. 

> Is this change related to the purpose of this PR?

Good point, I appreciate we follow questioning these thing! 
Yes, it is very much related. I had a rant about it [here](https://github.com/elastic/apm-agent-dotnet/pull/608#discussion_r347429628). Since we read the request body asynchronously we can't do it in the `void AspNetCoreDiagnosticListener.OnNext(...)` method anymore - so the code that reads the body in case an error happened and is handled by an error handling middleware is moved here from the `AspNetCoreDiagnosticListener.OnNext` method.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
gregkalapos: No string feelings on this either. I suspect we'll run into this more later.

Changed this so we can move on.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
gregkalapos: I'm not sure if I manually turned this on, or it's the default, but most IDEs show this to you these days:

image

src/Elastic.Apm.AspNetCore/Extensions/TransactionExtentions.cs
gregkalapos: This parameter comes from the outside 

```
```

And we pass it like:

```
await transaction.CollectRequestBody(true, context.Request, _logger)
```

In general, the `HttpRequest` will always come from outside, it's not something we control, and I expect `HttpContext.Request` being nullable. 

I'd feel better with the check. I hope this is not a blocker.

src/Elastic.Apm.AspNetCore/Elastic.Apm.AspNetCore.csproj
gregkalapos: 👍  removed.

src/Elastic.Apm.AspNetCore/Extensions/RequestExtentions.cs
```diff
@@ -16,26 +16,25 @@ public static class HttpRequestExtensions
...
 
 			try
 			{
-				request.EnableRewind();
+				request.EnableBuffering();
```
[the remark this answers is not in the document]
gregkalapos: I did a little bit of a reading on this.

> I got an impression from this article that EnableBuffering should be called before any attempt to read the body

Why do you get that impression?  I did not think that by reading that article, nor has the [docs](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.httprequestrewindextensions.enablebuffering?view=aspnetcore-2.1) contain that info. 

Furthermore, we have tests (and I also tried manually) when we read the request body on errors, which means the body is first read by some ASP.NET Core component and then by us. And it works. 

> We should definitely add test for this use case.

So, cases I think are relevant for this are already covered. For example `TestJsonBodyRetrievalOnRequestFailureInAspNetCore` is 1 test where we definitely read the request body very late - basically right before the request ends and I also think that your `Options_changed_after_start` test covers that.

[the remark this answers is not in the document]
SergeyKleyman: I'll leave it for you to decide if it's worth further work - it's quite possible that tests pass because the body size is very small and thus arrives in one TCP packet and/or we don't send request body in multiple chunks (using `Transfer-Encoding: chunked`) or maybe some other reason.
I looked at the source code a little bit
- [`Microsoft.AspNetCore.Http.HttpRequestRewindExtensions.EnableBuffering` just calls `Microsoft.AspNetCore.Http.Internal.BufferingHelper.EnableRewind`](https://github.com/aspnet/AspNetCore/blob/v3.0.0/src/Http/Http/src/Extensions/HttpRequestRewindExtensions.cs#L23)
- [`EnableRewind` wraps `request.Body` into a buffering stream](https://github.com/aspnet/AspNetCore/blob/v3.0.0/src/Http/Http/src/Internal/BufferingHelper.cs#L24) which I don't think can do buffering retroactively

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
```diff
@@ -42,21 +42,37 @@ public ApmMiddleware(RequestDelegate next, Tracer tracer, IApmAgent agent)
...
 				await _next(context);
 			}
-			catch (Exception e) when (transaction != null && ExceptionFilter.Capture(e, transaction, context, _logger))
-			{ }
+			catch (Exception e) when (transaction != null)
+			{
+				transaction.CaptureException(e);
+				// It'd be nice to have this in an exception filter, but that would force us capturing the request body synchronously.
+				// Therefore we rather unwind the stack in the catch block and call the async method.
+				if (context != null && transaction.IsContextCreated)
```
SergeyKleyman: `transaction.IsContextCreated` is redundant since `CollectRequestBody` includes it.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
SergeyKleyman: I see. I have two points of concern:
1. This effectively changes the definition of an error. For example, do we want to treat HTTP status code 304 (Not Modified) as an error?
2. I assume `AspNetCoreDiagnosticListener.OnNext` is called before we get to this point of `ApmMiddleware.InvokeAsync` which means error event's context will not have request body. If we are going to align with other agents and send request body in error event's context even for non-sampled transaction wouldn't it be a problem?

Is there a reason you didn't take approach mentioned in [the comment](https://github.com/elastic/apm-agent-dotnet/pull/608#discussion_r347429628) 

> Another option would be to mark the transaction here and not do the reading and have a check in 

Generating error event detected in `AspNetCoreDiagnosticListener` can be deferred to this point of the flow thus solving both points above.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
gregkalapos: Removed. This property confuses me :)

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
gregkalapos: I think handling `context?.Response.StatusCode >= 300` as error was really not a good idea, I changed it to `context?.Response.StatusCode >= 400`. 

To your points:
1. After doing the change mentioned above, not anymore. What typically an error handling middleware does is that it handles the exception, sets the return code to some http 4xxx or 5xx and returns an HTML page.
2. Right, that makes it more challenging, I think what you say in your last sentence is a good approach. Nevertheless, I would not worry about it now, this is not something that the PR changes, this behaviour remains the same.

> Is there a reason you didn't take approach mentioned in the comment

I did not find a good way to do it - like, adding a flag to transaction and things like that does not sound very nice to me, and this approach looks cleaner.

Long term, when we align on the error context capturing behaviour we can revisit this, but for now I'd leave that out from this PR.

src/Elastic.Apm.AspNetCore/Extensions/RequestExtentions.cs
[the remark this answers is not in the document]
gregkalapos: I tested this with a very long request body, and it still seems to work. I leave it for now.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
SergeyKleyman: > I did not find a good way to do it - like, adding a flag to transaction and things like that does not sound very nice to me, and this approach looks cleaner.

I'm not sure what are the cons of adding a flag to transaction but please note that this change does break existing functionality - before this change error event [captured in `AspNetCoreDiagnosticListener.OnNext`](https://github.com/elastic/apm-agent-dotnet/pull/608/files#diff-178ddb4c5da109db16c57afee4635d9aR29) had request body (assuming configuration allowed it and transaction is sampled) but after this change it sometimes will and sometimes it won't. The tests pass because they check error event context after transaction is completed (and because we have this issue of not taking a snapshot of transaction context for error event) but in real life use it will be a data race between transaction thread and payload sender's thread - if error event is serialized to JSON before transaction completes serialized error event won't have request body.
I leave it to you to decide if the severity of this issue warrants additional work.

gregkalapos: Oh, I see what you mean, and yes that change is true indeed. But that code was kind of already ignoring this whole thing I feel (mainly because I think the error's context was not really thought through I feel).

We did basically this:
- Create an error and then set its context pointing to the transactions context
- Then, in a totally different call, set the body for the transaction's context, not thinking anything about the error's. 

I'd just do another iteration on when and how we capture errors context - and remove the sharing of that instance, I think that's worth a whole new PR.

PR #609, opened 2019-11-18
Add property to Agent to check configuration state

NeoXtreem:
As discussed [here](https://github.com/elastic/apm-agent-dotnet/issues/596#issuecomment-554445465) (under heading **3**), a use case can be where `Agent.Tracer` is called in a client application during startup before Elastic APM Agent has fully initialised. This may be for logging frameworks that log exceptions using `Agent.Tracer.CurrentTransaction.CaptureException` as advised [here](https://www.elastic.co/guide/en/apm/agent/dotnet/current/public-api.html). However, this means that `ApmAgent` is initialised from `Agent` with `_components` set to `null`. This results in `AgentComponents` setting its `ConfigurationReader` property to an instance of `EnvironmentConfigurationReader`. As a result, the Elastic APM config in **Web.config** is not read because this requires `FullFrameworkConfigReader` to be the `ConfigurationReader`.

By adding this property, it allows the calling code to first check that the `Agent` has been configured before accessing its `Instance` (or indeed `Tracer`).


global.json
```diff
@@ -1,5 +1,5 @@
 {
   "sdk": {
-    "version": "2.1.505"
+    "version": "2.1.802"
```
gregkalapos: Could you please revert this - this would break our CI. 

I suggest 2 ways to deal with this if this causes trouble during local development:
- The best is to install `2.1.505`, then all project in this repo will be able to pick up that version, all the other projects on your machine will use the newest version. 
- Do this change, but do not commit the `global.json` file.

src/Elastic.Apm/Agent.cs
```diff
@@ -94,13 +94,14 @@ public static class Agent
...
 
-
-		public static IConfigurationReader Config => Lazy.Value.ConfigurationReader;
+		public static IConfigurationReader Config => Instance.ConfigurationReader;
 
 		internal static ApmAgent Instance => Lazy.Value;
 
 		internal static bool IsInstanceCreated => Lazy.IsValueCreated;
 
+		public static bool IsConfigured { get; private set; }
```
gregkalapos: Isn't `IsInstanceCreated` basically the same? 🤔  - that is currently `internal`, so can't be used, but maybe we could just make it `public` and reuse that.

Also another thing: It'd be nice to have some unit tests for this.

NeoXtreem: Note quite the same. Similar to how I answered [here](https://stackoverflow.com/questions/58916834/#comment104099777_58917376) for the example code in that SO question, `IsConfigured` only returns `true` if `Setup` (the configuration method) has been called, while `IsInstanceCreated` would still return `false` because `Lazy` has not yet been evaluated. `IsInstanceCreated` will only return `true` after `Lazy` has been evaluated such as after `Agent.Tracer` has been called.

That said, having looked at how `IsInstanceCreated` is used in the code, I see no value in it versus `IsConfigured` whose logic is equally appropriate for all usages. Therefore, in my latest commit, I've removed `IsInstanceCreated` (it's internal, so not a breaking change) in favour of `IsConfigured` which is for public use. `IsInstanceCreated` was used by some internal tests (via `AgentSingletonUtils.EnsureInstanceCreated`), but these work fine without it (the test itself is unaffected).

Note that the solution wouldn't build locally because `AspNetFullFrameworkSampleApp` references `..\..\packages` which is not populated by any project in the solution including itself. I've therefore updated this project to get all its required dependencies from NuGet using `PackageReference` as it already seems to have been migrated to use this method (`packages.config` seemed redundant so I've deleted it).

global.json
NeoXtreem: Reverted.

NeoXtreem:
> Thanks @NeoXtreem,
> 
> Overall I'd be ok with adding this, I can see that such a property can be useful. Could you please take a look at my 2 comments?
> 
> Thanks.

@gregkalapos I've submitted some new commits based on your comments. Note that the solution wouldn't build locally because `AspNetFullFrameworkSampleApp` references `..\..\packages` which is not populated by any project in the solution including itself. I've therefore updated this project to get all its required dependencies from NuGet using `PackageReference` as it already seems to have been migrated to use this method (`packages.config` seemed redundant so I've deleted it).

gregkalapos:
@NeoXtreem  thanks for the update. 

This unfortunately seems breaking our CI, you can see all the info by clicking on failed steps. 

If I see correctly the `AspNetFullFrameworkSampleApp.csproj` file in the current form of the PR is missing. Also, the IIS tests are failing. As mentioned [here](https://github.com/elastic/apm-agent-dotnet/issues/596#issuecomment-555425515) you can run those locally as well. 


> I've therefore updated this project to

I appreciate the help, but I'd suggest addressing 1 issue in 1 PR. It would be easier to just focus on the `Agent` part, that is a couple of lines of change in a single file, the probability of breaking something is less if we just focus on that. 

Maybe I'd suggest reverting all changes and only focus on what the title of the PR says, that can be done safely.

sample/AspNetFullFrameworkSampleApp/AspNetFullFrameworkSampleApp.csproj
```diff
@@ -322,6 +216,44 @@
   <ItemGroup Condition="'$(OS)' != 'WINDOWS_NT'">
     <Content Include="LinuxReadme.md" />
   </ItemGroup>
+  <ItemGroup>
```
gregkalapos:
```suggestion
  <ItemGroup Condition="'$(OS)' == 'WINDOWS_NT'">
```

gregkalapos: We need this, otherwise the references will show up on Linux and macOS - we only want this project to do anything on Windows.

sample/AspNetFullFrameworkSampleApp/Web.config
```diff
@@ -1,317 +1,305 @@
...
+			<remove name="TelemetryCorrelationHttpModule"/>
+			<add name="TelemetryCorrelationHttpModule" type="Microsoft.AspNet.TelemetryCorrelation.TelemetryCorrelationHttpModule, Microsoft.AspNet.TelemetryCorrelation" preCondition="managedHandler"/>
+			<remove name="ApplicationInsightsWebTracking"/>
+			<add name="ApplicationInsightsWebTracking" type="Microsoft.ApplicationInsights.Web.ApplicationInsightsHttpModule, Microsoft.AI.Web" preCondition="managedHandler"/>
+		</modules>
+		<validation validateIntegratedModeConfiguration="false"/>
+	</system.webServer>
+	<system.codedom>
+		<compilers>
+		  <compiler language="c#;cs;csharp" extension=".cs" type="Microsoft.CodeDom.Providers.DotNetCompilerPlatform.CSharpCodeProvider, Microsoft.CodeDom.Providers.DotNetCompilerPlatform, Version=2.0.1.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35" warningLevel="4" compilerOptions="/langversion:default /nowarn:1659;1699;1701"/>
```
gregkalapos: I see you reformat - which is nice, 👍  for that, but the indention here is I think wrong - so I'd suggest either indent the whole file, or don't add indention changes at all.

sample/AspNetFullFrameworkSampleApp/Web.config
```diff
@@ -1,317 +1,305 @@
...
+			</parameters>
+		</defaultConnectionFactory>
+		<providers>
+			<provider invariantName="System.Data.SqlClient" type="System.Data.Entity.SqlServer.SqlProviderServices, EntityFramework.SqlServer"/>
+			<provider invariantName="System.Data.SQLite" type="System.Data.SQLite.EF6.SQLiteProviderServices, System.Data.SQLite.EF6"/>
+			<provider invariantName="System.Data.SQLite.EF6" type="System.Data.SQLite.EF6.SQLiteProviderServices, System.Data.SQLite.EF6"/>
+		</providers>
+		<interceptors>
+			<interceptor type="Elastic.Apm.EntityFramework6.Ef6Interceptor, Elastic.Apm.EntityFramework6"/>
+		</interceptors>
```
gregkalapos: nit: there are empty spaces after `>` - if you reformat the whole file that'll also go away.

src/Elastic.Apm/Agent.cs
gregkalapos: 👍 this sounds good to me. 

To the unit test part:
Here is one test:
```
		/// <summary>
		/// Makes sure Agent.IsConfigured only returns true after Setup is called
		/// </summary>
		[Fact]
		public void IsConfigured()
		{
			Agent.IsConfigured.Should().BeFalse();

			Agent.Setup(new AgentComponents());
			Agent.IsConfigured.Should().BeTrue();
		}
```

You can add this to the `BasicAgentTests.cs` [here](https://github.com/elastic/apm-agent-dotnet/blob/master/test/Elastic.Apm.Tests/BasicAgentTests.cs#L36).

sample/AspNetFullFrameworkSampleApp/Web.config
NeoXtreem: Done in 28aa9a2.

sample/AspNetFullFrameworkSampleApp/Web.config
NeoXtreem: Not sure how those spaces (instead of a tab) slipped through the net. Fixed in 28aa9a2.

src/Elastic.Apm/Agent.cs
NeoXtreem: Thanks. Have added this unit test in 28aa9a2.

NeoXtreem:
> Thanks for the updates @NeoXtreem.
> 
> The `Agent.IsConfigured` part is fine, other than the missing test - I added 1, which you can commit.
> 
> I also see some minor things in the `csproj` and `Web.config` files, could you update those?
> 
> I'd like to finish this review, but please do not do two or more things in PRs later - 1 PR should address 1 thing, I'd like to keep this rule in the future.

@gregkalapos I've made the requested changes, and I'm glad you accepted the fixes I made to the references for `AspNetFullFrameworkSampleApp`. I also prefer to have PRs that address one thing. However, a prerequisite of any submission should be that the code builds, and this code didn't build. I can only guess that your local machines and build server must've had the `..\..\packages` folder populated from a previous version of the code as this folder is no longer populated since the projects were updated to use `PackageReference` flavour of NuGet.

Issue #546, opened 2019-10-10
Move to .NET Core 3.0


gregkalapos:
- Both tooling and CI - especially for local tools mentioned [here](https://github.com/elastic/apm-agent-dotnet/pull/532#pullrequestreview-299706201)
- Also we should have sample apps exercising the agent on .NET Core 3.0. and also with ASP.NET Core 3.0

vhatsura:
Hey, @gregkalapos. Do we have any updates on it? .Net Core 3.0 will allow using C# 8.0 to simplify code in some places.

gregkalapos:
@vhatsura no update yet, we did not do any work yet.

vhatsura:
> @vhatsura no update yet, we did not do any work yet.

Do you have smth we, as a community, can do it to help your team?

gregkalapos:
> > @vhatsura no update yet, we did not do any work yet.
> 
> Do you have smth we, as a community, can do it to help your team?

@vhatsura I think a good first step would be to make sure we have test coverage for 3.0 - if we'd make both the `Elastic.Apm.Tests` and the `Elastic.Apm.AspNetCore.Tests` run both on `2.1` and `3.0` (actually `2.2` would be also nice) that would be already a good first step. 

I guess `Elastic.Apm.Tests` can be done by multitargeting, `Elastic.Apm.AspNetCore.Tests`  is more challenging, since that depends on the ASP.NET Core version.

If we would have that, we would already discover things like #604.

gregkalapos:
Let's start working on this! 

I see @vhatsura already opened some PRs, which is awesome.

I think there will be multiple steps needed.

1.
First we have to make sure CI is able to build 3.x - I started working on it in #638. 

2. 
Then we can update the projects to target netcoreapp3.x - that is done in #622 (thanks for that!)

3.
And then once that's done we need to fix all the tests and fix some bugs (we already know about #632 - and I think I also saw that on 3.x the default service name detection does not work either, and maybe there are other things. 

[...]

PR #616, opened 2019-11-19
Add ELASTIC_APM_SERVICE_NODE_NAME configuration option

### LINKED ISSUE #467: Add support for ELASTIC_APM_SERVICE_NODE_NAME configuration option
Add support for `ELASTIC_APM_SERVICE_NODE_NAME` configuration option and `service.node.configured_name` intake API field.

Derived from https://github.com/elastic/apm/issues/141

vhatsura:
closes #467

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -149,6 +149,27 @@ NOTE: The service name must conform to this regular expression: `^[a-zA-Z0-9 _-]
 |============
 
 [float]
+[[config-service-node-name]]
+==== `ServiceNodeName` (added[1.2])
+
+If set, this name is used to distinguish between different nodes of a service.
+If not set, data aggregations will be done based on a container ID (where valid) or on the reported hostname (automatically discovered).
+
+NOTE: Metrics views can utilize this configuration since APM Server 7.5
+
+[options="header"]
+|============
+| Environment variable name       | IConfiguration or Web.config key 
+| `ELASTIC_APM_SERVICE_NODE_NAME` | None
+|============
+
+[options="header"]
+|============
+| Default        | Type
+| `<none>`       | String 
+|============
+
+[float]
 [[config-service-version]]
 ==== `ServiceVersion`
 
@@ -648,6 +669,7 @@ Valid options: `Critical`, `Error`, `Warning`, `Info`, `Debug`, `Trace` and `Non
 | <<config-secret-token,`SecretToken`>> | No | Reporter
 | <<config-server-urls,`ServerUrls`>> | No | Reporter
 | <<config-service-name,`ServiceName`>> | No | Core
+| <<config-service-node-name, `ServiceNodeName`>> | No | Core
 | <<config-service-version,`ServiceVersion`>> | No | Core
 | <<config-span-frames-min-duration,`SpanFramesMinDuration`>> | No | Stacktrace, Performance
 | <<config-stack-trace-limit,`StackTraceLimit`>> | No | Stacktrace, Performance
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -152,7 +152,7 @@ NOTE: The service name must conform to this regular expression: `^[a-zA-Z0-9 _-]
 [[config-service-node-name]]
 ==== `ServiceNodeName` (added[1.2])
 
-If set, this name is used to distinguish between different nodes of a service.
+Optional name used to differentiate between nodes in a service.
 If not set, data aggregations will be done based on a container ID (where valid) or on the reported hostname (automatically discovered).
 
 NOTE: This feature requires APM Server versions >= 7.5
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -155,7 +155,7 @@ NOTE: The service name must conform to this regular expression: `^[a-zA-Z0-9 _-]
 If set, this name is used to distinguish between different nodes of a service.
 If not set, data aggregations will be done based on a container ID (where valid) or on the reported hostname (automatically discovered).
 
-NOTE: Metrics views can utilize this configuration since APM Server 7.5
+NOTE: This feature requires APM Server versions >= 7.5
 
 [options="header"]
 |============
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -160,7 +160,7 @@ NOTE: This feature requires APM Server versions >= 7.5
 [options="header"]
 |============
 | Environment variable name       | IConfiguration or Web.config key 
-| `ELASTIC_APM_SERVICE_NODE_NAME` | None
+| `ELASTIC_APM_SERVICE_NODE_NAME` | `ElasticApm:ServiceNodeName`
 |============
 
 [options="header"]
--- a/sample/AspNetFullFrameworkSampleApp/AspNetFullFrameworkSampleApp.csproj
+++ b/sample/AspNetFullFrameworkSampleApp/AspNetFullFrameworkSampleApp.csproj
@@ -4,7 +4,6 @@
   </Target>
   <Target Name="Clean" Condition="'$(OS)' != 'WINDOWS_NT'">
   </Target>
-  <Import Project="..\..\packages\MSBuild.Microsoft.VisualStudio.Web.targets.14.0.0.3\build\MSBuild.Microsoft.VisualStudio.Web.targets.props" Condition="Exists('..\..\packages\MSBuild.Microsoft.VisualStudio.Web.targets.14.0.0.3\build\MSBuild.Microsoft.VisualStudio.Web.targets.props')" />
   <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\TypeScript\Microsoft.TypeScript.Default.props" Condition="Exists('$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\TypeScript\Microsoft.TypeScript.Default.props')" />
   <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
   <PropertyGroup Condition="'$(OS)' != 'WINDOWS_NT'">
--- a/sample/AspNetFullFrameworkSampleApp/AspNetFullFrameworkSampleApp.csproj
+++ b/sample/AspNetFullFrameworkSampleApp/AspNetFullFrameworkSampleApp.csproj
@@ -4,6 +4,7 @@
   </Target>
   <Target Name="Clean" Condition="'$(OS)' != 'WINDOWS_NT'">
   </Target>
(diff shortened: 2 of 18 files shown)


src/Elastic.Apm/Config/AbstractConfigurationWithEnvFallbackReader.cs
```diff
@@ -41,6 +41,9 @@ internal AbstractConfigurationWithEnvFallbackReader(IApmLogger logger, string de
 		public virtual string Environment => ParseEnvironment(Read(ConfigConsts.KeyNames.Environment, ConfigConsts.EnvVarNames.Environment))
 			?? _defaultEnvironmentName;
 
+		public string ServiceNodeName => ParseServiceNodeName(Kv(ConfigConsts.EnvVarNames.ServiceNodeName,
```
gregkalapos: Here we force the `MicrosoftExtensionsConfig` to only read environment variables - what's the reason for this? I think it's reasonable that this setting can also be configured through a config file which contains different values for the different instances. Or did I miss something? 

I'd just do this here:

```
ParseServiceNodeName(Read(ConfigConsts.KeyNames.ServiceNodeName, ConfigConsts.EnvVarNames.ServiceNodeName));
```

_`ConfigConsts.KeyNames.ServiceNodeNames` also needs to be added in this case_

docs/configuration.asciidoc
```diff
@@ -148,6 +148,27 @@ NOTE: The service name must conform to this regular expression: `^[a-zA-Z0-9 _-]
...
+
+If set, this name is used to distinguish between different nodes of a service.
+If not set, data aggregations will be done based on a container ID (where valid) or on the reported hostname (automatically discovered).
+
+NOTE: Metrics views can utilize this configuration since APM Server 7.5
+
+[options="header"]
+|============
+| Environment variable name       | IConfiguration or Web.config key 
+| `ELASTIC_APM_SERVICE_NODE_NAME` | None
```
gregkalapos: The `IConfiguration or Web.config key ` part is TBD, if we all agree we want to have that, this needs to be updated. If not, then sorry for the noise :)

docs/configuration.asciidoc
```diff
@@ -148,6 +148,27 @@ NOTE: The service name must conform to this regular expression: `^[a-zA-Z0-9 _-]
 | Name of the entry assembly       | String 
 |============
 
+[float]
+[[config-service-node-name]]
+==== `ServiceNodeName` (added[1.2])
```
gregkalapos: I personally linked the [Go version of this doc](https://github.com/elastic/apm-agent-go/pull/651/files#diff-01c1a79a5a1e1b67003a1885365f2cf5R153) better, but this is also totally fine.

docs/configuration.asciidoc
```diff
@@ -148,6 +148,27 @@ NOTE: The service name must conform to this regular expression: `^[a-zA-Z0-9 _-]
...
+
+If set, this name is used to distinguish between different nodes of a service.
+If not set, data aggregations will be done based on a container ID (where valid) or on the reported hostname (automatically discovered).
+
+NOTE: Metrics views can utilize this configuration since APM Server 7.5
```
```suggestion
NOTE: This feature requires APM Server versions >= 7.5
```

docs/configuration.asciidoc
```diff
@@ -148,6 +148,27 @@ NOTE: The service name must conform to this regular expression: `^[a-zA-Z0-9 _-]
 | Name of the entry assembly       | String 
 |============
 
+[float]
+[[config-service-node-name]]
+==== `ServiceNodeName` (added[1.2])
+
+If set, this name is used to distinguish between different nodes of a service.
```
```suggestion
Optional name used to differentiate between nodes in a service.
```

src/Elastic.Apm/Config/AbstractConfigurationWithEnvFallbackReader.cs
vhatsura: done

docs/configuration.asciidoc
vhatsura: done

sample/AspNetFullFrameworkSampleApp/AspNetFullFrameworkSampleApp.csproj
```diff
@@ -4,6 +4,7 @@
   </Target>
   <Target Name="Clean" Condition="'$(OS)' != 'WINDOWS_NT'">
   </Target>
+  <Import Project="..\..\packages\MSBuild.Microsoft.VisualStudio.Web.targets.14.0.0.3\build\MSBuild.Microsoft.VisualStudio.Web.targets.props" Condition="Exists('..\..\packages\MSBuild.Microsoft.VisualStudio.Web.targets.14.0.0.3\build\MSBuild.Microsoft.VisualStudio.Web.targets.props')" />
```
vhatsura: It was deleted in #609, however, it cannot be built locally without such import

gregkalapos: Oh, you probably don't have ASP.NET installed locally, right?

vhatsura: If we talk about Windows features, I have `ASP`, `ASP.NET 3.5` and `ASP.NET 4.8` components

gregkalapos: I meant the ASP.NET workload - or to be precise "Web Development build tools" when you install Visual Studio. Like mentioned [here](https://stackoverflow.com/a/44064473)

I think the Windows feature only makes you able to host those project, I'm not sure it's the full SDK - but I'm not sure. 

I just would like to understand why this happened to you - it seems working for me locally and also in our CI (which I know has the web dev. build tools). 

Overall I'm ok with merging it - if that helps people without that workload to build then it's definitely better to have it.

NeoXtreem: That folder is not populated by anything in the solution, so if it exists locally for @vhatsura, then it must be because he doesn't have a clean workspace, and has this folder remaining from a previous version of the code. Dead paths should not be added into the project file. This project file was previously updated to use `PackageReference` flavour NuGet, but the work wasn't complete. #609 addresses this by pulling in all required dependencies properly via NuGet. As @gregkalapos mentioned, this works on the CI server.

@vhatsura I would suggest doing a complete clean and rebuild to begin with, and then addressing any missing workloads or SDKs on your system before proceeding.

vhatsura: @NeoXtreem, I deleted `packages` folder manually (clean target didn't help) and remove this line. After that I got the following error:
```
  The imported project "C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\MSBuild\Microsoft\VisualStudio\v16.0\WebApplications\Microsoft.WebApplication.targets" was not found. Also, tried to find "WebApplications\Microsoft.WebApplication.targets" in the fallback search path(s) for $(VSToolsPath) - "C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v16.0" . These search paths are defined in "C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\MSBuild\Current\Bin\MSBuild.exe.Config". Confirm that the path in the <Import> declaration is correct, and that the file exists on disk in one of the search paths.  C:\projects\apm-agent-dotnet\sample\AspNetFullFrameworkSampleApp\AspNetFullFrameworkSampleApp.csproj
```
Beside I found issue in developer community which is the same as for me: https://developercommunity.visualstudio.com/content/problem/414282/vs-2019-webapplication-projects-fail-to-load.html. So, I think it isn't related to missed some installation packages. BTW, "Web Development build tools" is installed in my setup (I use Rider, however VS 2019 is installed). Downgrade MSBuild version (Rider allow to do it for all solution) to 15.0 helps me to build the project, however, it will be an issue with moving to .Net Core 3.0 support.

gregkalapos: Thanks for the update! 
I'd say let's ignore it for now, if we ran into this in https://github.com/elastic/apm-agent-dotnet/issues/546 we can still make a change - usually we should prefer these build and project file changes outside of agent code changes anyway.

vhatsura: Never mind, it seems I fixed it (I just open VS installer to check installed components and close it). After that, all is working fine. Sorry for inconvenience.

gregkalapos:
Hmm, I see failing tests:

```
[2019-11-21T16:18:22.111Z] [xUnit.net 00:20:41.3470816]   Finished:    Elastic.Apm.AspNetFullFramework.Tests
[2019-11-21T16:18:22.111Z] Failed   Elastic.Apm.AspNetFullFramework.Tests.CentralConfigTests+MaxSpansAndSampleRateTests.MaxSpans_invalid_value
[2019-11-21T16:18:22.111Z] Error Message:
[2019-11-21T16:18:22.111Z]  Expected service.Node to be <null>, but found Service{ConfiguredName: null}.
```

Let's wait with the merge, I'll dig deeper - not sure I'll be able to do it today though.

vhatsura:
> Hmm, I see failing tests:
> 
> ```
> [2019-11-21T16:18:22.111Z] [xUnit.net 00:20:41.3470816]   Finished:    Elastic.Apm.AspNetFullFramework.Tests
> [2019-11-21T16:18:22.111Z] Failed   Elastic.Apm.AspNetFullFramework.Tests.CentralConfigTests+MaxSpansAndSampleRateTests.MaxSpans_invalid_value
> [2019-11-21T16:18:22.111Z] Error Message:
> [2019-11-21T16:18:22.111Z]  Expected service.Node to be <null>, but found Service{ConfiguredName: null}.
> ```
> 
> Let's wait with the merge, I'll dig deeper - not sure I'll be able to do it today though.

Sorry about that. Finally, I configured my machine to run AspNetFullFramework tests and fix the tests.

src/Elastic.Apm/Api/Node.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Elastic.Apm.Api
+{
+	public class Node
+	{
+		[JsonProperty("configured_name")]
+		[JsonConverter(typeof(TrimmedStringJsonConverter))]
+		public string ConfiguredName { get; set; }
+
+		public override string ToString() => new ToStringBuilder(nameof(Service)) { { nameof(ConfiguredName), ConfiguredName } }.ToString();
```
SergeyKleyman: It should be `nameof(Node)` instead of `nameof(Service)`, right?

vhatsura: yeap, you're totally right. That's what it means to copy-paste methods

vhatsura: Yeap, you're right. That's what it means to copy-paste methods. Fixed.

docs/configuration.asciidoc
vhatsura: @gregkalapos, I believe we need to change version here, as this feature isn't included in 1.2 version. Maybe, we need to think about some placeholders which will be replaced during release to avoid issues with documentation

gregkalapos: Thanks for tagging me @vhatsura! Opened https://github.com/elastic/apm-agent-dotnet/pull/627 - yeah we should came up with something for this.

Issue #555, opened 2019-10-15
Limit the type of Request.Body to string/Dictionary<string,string>, it's currently an object


gregkalapos:
I don't know why we did this, but `Request.Body` ended up being an `object`, which is very unfortunate. This is a public property, users can set it.

```
public class Request
{
    public object Body { get; set; }
    //other things...
}
```

- If something which we can't serialize ends up there, there'll be probably an exception thrown during serialization.
- For sanitization it'd also make our job easier 
- I think this can't really be anything other than `string`.  (Update: it can also be `Dictionary<string,string>`)

This is a simple, but breaking change. It'll be a good time to do this when we increase our major version.

axw:
Agents can send parsed `application/x-www-form-urlencoded` bodies as JSON objects to the APM Server. Perhaps that's the reason why?

FWIW, the Go agent has something like a discriminated union for this. If a form body is populated, it'll be sanitised according to the form keys.

gregkalapos:
> Agents can send parsed `application/x-www-form-urlencoded` bodies as JSON objects to the APM Server. Perhaps that's the reason why?
> 
> FWIW, the Go agent has something like a discriminated union for this. If a form body is populated, it'll be sanitised according to the form keys.

What we currently do (and now I realize that it's probably not good) is that we turn `application/x-www-form-urlencoded` bodies also into `string` - it'll be something like `Key1=Value1&Key2=Value2`. Basically we always capture the body as `string`. In case of `application/x-www-form-urlencoded` I guess this is bad for search-ability.

ASP.NET Core gives us the body as a stream, and we just turn it into a string. Unfortunately there isn't really a discriminated union in C#, but we can come up with something. 

So, after reading your comment, I'd say `Request.Body` can be either a `string` or a `Dictionary<string,string>` (like a keyvalue pair of strings). Is this assumption correct? I still think that having it as `object` is very dangerous and this could be improved.

axw:
> Unfortunately there isn't really a discriminated union in C#, but we can come up with something.

Same in Go :) I just meant conceptually the same.

> So, after reading your comment, I'd say Request.Body can be either a string or a Dictionary<string,string> (like a keyvalue pair of strings). Is this assumption correct?

Yes, that's correct; and limiting to those two types, rather than object, would be better.

Issue #28, opened 2018-12-14
Create performance benchmarks


gregkalapos:

SergeyKleyman:
One of the use cases worth testing - https://github.com/elastic/apm-agent-dotnet/pull/394#issuecomment-513362154. It would be interesting to measure a difference between agent implementation with and without `Interlocked` counters.

gregkalapos:
Done in #821.

PR #617, opened 2019-11-20
BodyStringSanitizerConverter - handling json and other strings without `&` and `=`

gregkalapos:
Noticed this during testing. 

In `BodyStringSanitizerConverter.cs` we look rebuild the `key=value` format and look for keys contained in `SanitizeFieldNames`. 

If we can't rebuilt the `key=value` format we just capture the body as it is. 

In the `FormatCheck` method we did not handle the case when there is no `&` or `=` in the body, therefore we serialized things like "foo" or a json string like "{"foo" : "bar"}" as empty string.

### DIFF
--- a/src/Elastic.Apm/Report/Serialization/BodyStringSanitizerConverter.cs
+++ b/src/Elastic.Apm/Report/Serialization/BodyStringSanitizerConverter.cs
@@ -62,23 +62,29 @@ namespace Elastic.Apm.Report.Serialization
 			if (!bodyValue.Contains("=") && !bodyValue.Contains("&"))
 				return false;
 
+			var numberOfEqual = 0;
+			var numberOfAnds = 0;
+
 			var i = 0;
 			foreach (var c in bodyValue)
 			{
 				if (c == '=')
 				{
+					numberOfEqual++;
 					i++;
 					if (i > 1)
 						return false;
 				}
 				if (c == '&')
 				{
+					numberOfAnds++;
 					i--;
 					if (i < 0)
 						return false;
 				}
 			}
-			return true;
+
+			return numberOfAnds != 0 || numberOfEqual != 0;
 		}
 
 		public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer) =>
--- a/src/Elastic.Apm/Report/Serialization/BodyStringSanitizerConverter.cs
+++ b/src/Elastic.Apm/Report/Serialization/BodyStringSanitizerConverter.cs
@@ -32,8 +32,8 @@ namespace Elastic.Apm.Report.Serialization
 					foreach (var formValue in formValues)
 					{
 						var formsValueSplit = formValue.Split('=');
-						if (formsValueSplit.Length != 2 || string.IsNullOrWhiteSpace(formsValueSplit[0])
-							|| string.IsNullOrWhiteSpace(formsValueSplit[1]))
+						if (formsValueSplit.Length != 2 || string.IsNullOrEmpty(formsValueSplit[0])
+							|| string.IsNullOrEmpty(formsValueSplit[1]))
 						{
 							//in this case it's not in the expected format, so we won't (and can't) sanitize, we just write it and move on
 							writer.WriteValue(strValue);
--- a/src/Elastic.Apm/Report/Serialization/BodyStringSanitizerConverter.cs
+++ b/src/Elastic.Apm/Report/Serialization/BodyStringSanitizerConverter.cs
@@ -7,9 +7,9 @@ using Newtonsoft.Json;
 namespace Elastic.Apm.Report.Serialization
 {
 	/// <summary>
-	/// Sanitizes request body based on the config passed to the constructor, in case it's in  `Key1=Value1& Key2=Value2` format.
+	/// Sanitizes request body based on the config passed to the constructor, in case it's in `Key1=Value1& Key2=Value2` format.
 	/// Ideally this would inherit from <code> JsonConverter{string} </code>.
-	/// Until  https://github.com/elastic/apm-agent-dotnet/issues/555 is not done, we roll with a generic JsonConverter.
+	/// Until https://github.com/elastic/apm-agent-dotnet/issues/555 is not done, we roll with a generic JsonConverter.
 	/// </summary>
 	internal class BodyStringSanitizerConverter : JsonConverter
 	{
@@ -22,12 +22,12 @@ namespace Elastic.Apm.Report.Serialization
 		{
 			if (value is string strValue)
 			{
-				var formValues = strValue.Split('&');
-
 				if (!FormatCheck(strValue))
 					writer.WriteValue(strValue);
 				else
 				{
+					var formValues = strValue.Split('&');
+
 					var sb = new StringBuilder();
 					foreach (var formValue in formValues)
 					{
@@ -59,6 +59,9 @@ namespace Elastic.Apm.Report.Serialization
 		/// <returns></returns>
 		private static bool FormatCheck(string bodyValue)
 		{
+			if (!bodyValue.Contains("=") && !bodyValue.Contains("&"))
+				return false;
+
 			var i = 0;
 			foreach (var c in bodyValue)
 			{
--- a/src/Elastic.Apm/Report/Serialization/BodyStringSanitizerConverter.cs
+++ b/src/Elastic.Apm/Report/Serialization/BodyStringSanitizerConverter.cs
@@ -3,7 +3,6 @@ using System.Text;
 using Elastic.Apm.Config;
 using Elastic.Apm.Helpers;
 using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
 
 namespace Elastic.Apm.Report.Serialization
 {
@@ -33,7 +32,8 @@ namespace Elastic.Apm.Report.Serialization
 					foreach (var formValue in formValues)
 					{
 						var formsValueSplit = formValue.Split('=');
-						if (formsValueSplit.Length != 2 || string.IsNullOrWhiteSpace(formsValueSplit[0]) || string.IsNullOrWhiteSpace(formsValueSplit[1]) )
+						if (formsValueSplit.Length != 2 || string.IsNullOrWhiteSpace(formsValueSplit[0])
+							|| string.IsNullOrWhiteSpace(formsValueSplit[1]))
 						{
 							//in this case it's not in the expected format, so we won't (and can't) sanitize, we just write it and move on
 							writer.WriteValue(strValue);
--- a/src/Elastic.Apm/Report/Serialization/BodyStringSanitizerConverter.cs
+++ b/src/Elastic.Apm/Report/Serialization/BodyStringSanitizerConverter.cs
@@ -3,6 +3,7 @@ using System.Text;
 using Elastic.Apm.Config;
 using Elastic.Apm.Helpers;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 
 namespace Elastic.Apm.Report.Serialization
 {
@@ -32,8 +33,12 @@ namespace Elastic.Apm.Report.Serialization
 					foreach (var formValue in formValues)
 					{
 						var formsValueSplit = formValue.Split('=');
-						if (formsValueSplit.Length != 2)
-							continue;
+						if (formsValueSplit.Length != 2 || string.IsNullOrWhiteSpace(formsValueSplit[0]) || string.IsNullOrWhiteSpace(formsValueSplit[1]) )
+						{
+							//in this case it's not in the expected format, so we won't (and can't) sanitize, we just write it and move on
+							writer.WriteValue(strValue);
+							return;
+						}
 
 						if (sb.Length != 0) sb.Append("&");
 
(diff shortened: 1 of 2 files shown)


SergeyKleyman:
Is there a reason there is no check if the `Content-Type` is actually a form data (namely `multipart/form-data` or `application/x-www-form-urlencoded`) before trying to parse body as such? It would avoid both:
- Correctness (sanitizing some content that is not form data but has `&` and `=` characters in it)
- Performance overhead of parsing content that is not form data

src/Elastic.Apm/Report/Serialization/BodyStringSanitizerConverter.cs
```diff
@@ -59,6 +59,9 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
 		/// <returns></returns>
 		private static bool FormatCheck(string bodyValue)
 		{
+			if (!bodyValue.Contains("=") && !bodyValue.Contains("&"))
```
SergeyKleyman: It seems to be very inefficient - with this change in the worst case `WriteJson` will iterate over the body 4 times.

SergeyKleyman: IMHO this check introduces a regression - after this change `"password=qwerty"` will not be sanitized because it doesn't contain `&`.

gregkalapos:
> Is there a reason there is no check if the `Content-Type` is actually a form data (namely `multipart/form-data` or `application/x-www-form-urlencoded`) before trying to parse body as such? It would avoid issues with:
> 
> * Correctness (sanitizing some content that is not form data but has `&` and `=` characters in it)
> * Performance overhead of parsing content that is not form data

The main reason we don't have that check is that we don't have that information during serialization.

The advantage of doing the sanitization during serialization is that with this, if users set the body manually through the API, we'll still serialize it. Additionally it also out of the box applies to all our implementations and to any other implementations where we capture body. This was discussed in one of the PRs introducing sanitization.

But I totally agree with your concerns and it'd be nice to address those long term. First the problem is that currently the `Request.Body` is an `object`, but the [server could handle](https://github.com/elastic/apm-server/blob/master/docs/spec/request.json#L9) `string` and `dictionary<string,string>` (which is `object` in json, but that's not a C# object). This is what I'd like to tackle first - there is already an issue #555. With this the serialization will automatically work with either `Dictionary<string,string>` or `string`. In case of `Dictionary<string,string>` we can apply serialization simply based on the dictionary keys (maybe here the question comes up if we only sanitize `application/x-www-form-urlencoded` as other agent, if yes, then we still somehow need to get the content type). 

Alternatively we can just sanitize during the reading of the body (e.g ASP.NET Core can give it as a key value pair in case it's `application/x-www-form-urlencoded`, do the sanitization during reading, and mark the body as already sanitize and not run this logic. But hopefully we don't need this, the solution above would be better.

src/Elastic.Apm/Report/Serialization/BodyStringSanitizerConverter.cs
gregkalapos: On it

gregkalapos: I don't think so 🤔 
There is an `&&` between to two `bodyValue.Contains` checks, so it's enough to have a `=` in it to not return. I pushed 2 new tests, one tests against your scenario - seems working and the code also makes sense to me.

test/Elastic.Apm.Tests/SerializationTests.cs
```diff
@@ -378,6 +378,20 @@ public void CustomDeDotting()
...
 
+		/// <summary>
+		/// Makes sure that the request bodies that contain json are not changed by serialization.
+		/// It basically tests <see cref="BodyStringSanitizerConverter"/>.
+		/// </summary>
+		[Fact]
+		public void RequestBodyWithJson()
+		{
+			var context = new Context();
+			context.Request = new Request("GET", new Url()) { Body = "{\"foo\": \"bar\"}" };
```
SergeyKleyman: With `Body = "{\"foo\": \"k1=v1&k2\"}"` is still fails because `&k2` part is dropped.

gregkalapos: Oh, my understanding was that something like that we definitely can't and don't want to serialize. 

I thought we only serialize based on the root keys (I'm not sure what the right term here is... so on the thing that is "foo". 

Furthermore this is also json, which I thought we don't serialize. My understanding was that we only serialize formats like `key1=value1&key2=value2`.

src/Elastic.Apm/Report/Serialization/BodyStringSanitizerConverter.cs
gregkalapos: How about https://github.com/elastic/apm-agent-dotnet/pull/617/commits/46e97f06851f097cbfa660e9d487a89a9b371c21? 

Overall this whole thing is not efficient - reading the body will add lots of overhead, and returning and parsing the keys adds even more I think. I feel what I talk about [here](https://github.com/elastic/apm-agent-dotnet/pull/617#issuecomment-557001669) could make a real perf. difference. 

This discussion is yet another reason to finish https://github.com/elastic/apm-agent-dotnet/issues/28 :)

test/Elastic.Apm.Tests/SerializationTests.cs
gregkalapos: Oh, I see, on it!

gregkalapos: I think this made it: https://github.com/elastic/apm-agent-dotnet/pull/617/commits/9be210c15f74f02f38c29a7f29facdea29c89d49 

Thought about check specifically for JSON, but that won't help, because at the end of the day, this could be anything, HTML, XML, or something else.

Now, if as a side-effect we may still sanitize something (I guess that could happen) when it's not in the expected format, then I'd say... then so be it 🤷‍♂ 

After #555 we'll do this better and we can get rid of this while re-parsing thing, which is currently a pain, but I don't see any good way to avoid it with having the body as `object`. (happy to take suggestions though! )

src/Elastic.Apm/Report/Serialization/BodyStringSanitizerConverter.cs
```diff
@@ -22,18 +22,23 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
...
 					foreach (var formValue in formValues)
 					{
 						var formsValueSplit = formValue.Split('=');
-						if (formsValueSplit.Length != 2)
-							continue;
+						if (formsValueSplit.Length != 2 || string.IsNullOrWhiteSpace(formsValueSplit[0])
+							|| string.IsNullOrWhiteSpace(formsValueSplit[1]))
+						{
+							//in this case it's not in the expected format, so we won't (and can't) sanitize, we just write it and move on
+							writer.WriteValue(strValue);
```
SergeyKleyman: After this change `FormatCheck` becomes redundant - making it always return `true` doesn't cause any tests to fail...

src/Elastic.Apm/Report/Serialization/BodyStringSanitizerConverter.cs
SergeyKleyman: Please see - https://github.com/elastic/apm-agent-dotnet/pull/617/files#r349052916
I don't think benchmarks are necessary to avoid iterating multiple times over potentially quite a big buffer.

SergeyKleyman: You are right - my mistake. Do we have tests that invoke sanitization and serialization together (as they are invoked in production code) to verify serialization doesn't neutralize sanitization by mistake?

test/Elastic.Apm.Tests/SerializationTests.cs
SergeyKleyman: I think we should sanitize only known data formats (at this point only form data and we can ask other agents if they sanitize JSON, XML, etc. as a future items) because as it stands now if it's not a valid encoded form data we don't sanitize it that well - for example `Body = "password=my password with equal (=) sign"` will expose `my password with equal (=) sign` as is.

SergeyKleyman:
> The main reason we don't have that check is that we don't have that information during serialization.

IMHO it's not that hard to propagate knowledge about `Content-Type` to serialization stage. For example by adding some `enum` `RequestContentType` property somewhere near `Body`. This `enum` can include `Unknown` (or `UnsupportedForSanitization` - name doesn't matter) value to skip sanitization for unsupported formats. We can make this new `RequestContentType` property public for users who would like to manually capture `Body` in supported format and have it automatically sanitized.
As for supporting sanitization of unknown formats - I am not even sure what it means even from the requirements point of view (please see an example at https://github.com/elastic/apm-agent-dotnet/pull/617/files#r349062715)

src/Elastic.Apm/Report/Serialization/BodyStringSanitizerConverter.cs
```diff
@@ -22,18 +22,23 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
...
 				{
+					var formValues = strValue.Split('&');
+
 					var sb = new StringBuilder();
 					foreach (var formValue in formValues)
 					{
 						var formsValueSplit = formValue.Split('=');
-						if (formsValueSplit.Length != 2)
-							continue;
+						if (formsValueSplit.Length != 2 || string.IsNullOrWhiteSpace(formsValueSplit[0])
```
SergeyKleyman: According to [String.Split docs](https://docs.microsoft.com/en-us/dotnet/api/system.string.split?redirectedfrom=MSDN&view=netframework-4.8#System_String_Split_System_Char___System_Int32_System_StringSplitOptions_) returned array should not contain `nulls` - it can be empty or contain empty strings but it cannot contain `null`s.

src/Elastic.Apm/Report/Serialization/BodyStringSanitizerConverter.cs
gregkalapos: Well, it is not so easy.

Here is my thinking:
Iterating through a string is very fast - and we can find formats that we can't sanitize with an iteration, but as it turned out, it's not 100% correct.

On the other hand after the `FormatCheck` and line 39 (where we'd end up without `FormatCheck`) there is an allocation, which is very very bad. 

To show this, here is a benchmark I wrote:

```
[Benchmark]
public void Sanitizer()
{
	var converter = new BodyStringSanitizerConverter(new EnvironmentConfigurationReader());
	var sb = new StringBuilder();
	for (var i = 0; i < 10000; i++)
	{
		sb.Append("aa");
	}
	sb.Append("=bb&&c=dd");
	converter.WriteJson(new JsonTextWriter(new DebugTextWriter()), sb.ToString(), null );
}
```

Here is the result with the PR as it is (including `CheckFormat`):

```

|    Method |     Mean |    Error |   StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|---------- |---------:|---------:|---------:|------------:|------------:|------------:|--------------------:|
| Sanitizer | 98.58 us | 1.527 us | 1.428 us |     15.1367 |      0.1221 |           - |             93.2 KB |
```

And here is the result without the "redundant" `CheckGormat`):

```
|    Method |    Mean |    Error |   StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|---------- |--------:|---------:|---------:|------------:|------------:|------------:|--------------------:|
| Sanitizer | 2.182 s | 0.0148 s | 0.0131 s |           - |           - |           - |              5.1 MB |
```

As you see the difference is big.

You can argue that my sample data is kind of an edge case, but my point with this is that 1 additional iteration in `CheckFormat` adds very little overhead, while the additional allocation adds way more overhead.

src/Elastic.Apm/Report/Serialization/BodyStringSanitizerConverter.cs
gregkalapos: > I don't think benchmarks are necessary to avoid iterating multiple times over potentially quite a big buffer.

:) Sure - benchmarks don't avoid iterations, they deliver data. 

What I meant is that benchmarks are useful to discuss performance. Discussing performance without benchmark is what I don't like doing.

gregkalapos:
@SergeyKleyman I really appreciate the feedback on this, I feel we improved this tremendously. 

I'd say this reached a point where the PR improves the current state - specifically: it does not cut the request body accidentally anymore.

I'd handle questions like completely getting rid of the key rebuilding logic and what to sanitize or not in follow ups - I still feel #555 would make a difference here.

My reason for this is that this PR does not change much on it - maybe there is 1 aspect, which is that in some cases we may redact things in non `application/x-www-form-urlencoded` bodies, but previously we completely removed those (so users saw nothing from that, nothing was captured), now we may redact things in those that would normally only be redacted in `application/x-www-form-urlencoded`. I think this is still an improvement.

src/Elastic.Apm/Report/Serialization/BodyStringSanitizerConverter.cs
SergeyKleyman: `WhiteSpace` part of `IsNullOrWhiteSpace` will cause empty strings and strings white space only to be exposed instead of being replaced by "[REDACTED]".

gregkalapos: Changed it.

gregkalapos: Appreciate the info, but I'm not sure what you suggest here or what action I should take.

You suggest checking on formsValueSplit[0] just for the length? Honestly I feel it just makes the code look wired - would that make any significant difference?

gregkalapos:
Closing this over https://github.com/elastic/apm-agent-dotnet/pull/712.

PR #618, opened 2019-11-20
Fix IsConfigured test

gregkalapos:
The new `IsConfigured` test is failing - only in CI, only on Windows.

It uses the static `Agent`, which I think causes the problem - maybe another test runs in parallel and uses `Agent`. Update: yeah, this theory seems to be verified.

Probably a strategy like this would be ok:
- No tests should use the static agents (except tests that are in the next category), basically all our tests already do this
- Tests that must use the static agent property (I think `IsConfigured` is the first and only) must be in the same category, so they won't run in parallel - I'm doing this now by putting all these tests into the same class - those tests will run sequentially.

### DIFF
--- a/test/Elastic.Apm.Tests/ConfigTests.cs
+++ b/test/Elastic.Apm.Tests/ConfigTests.cs
@@ -303,7 +303,7 @@ namespace Elastic.Apm.Tests
 		}
 
 		[Fact]
-		public void DefaultLogLevelTest() => Agent.Config.LogLevel.Should().Be(LogLevel.Error);
+		public void DefaultLogLevelTest() => new ApmAgent(new TestAgentComponents()).ConfigurationReader.LogLevel.Should().Be(LogLevel.Error);
 
 		[Theory]
 		[InlineData("Trace", LogLevel.Trace)]
--- a/test/Elastic.Apm.Tests/StaticTests.cs
+++ b/test/Elastic.Apm.Tests/StaticTests.cs
@@ -1,0 +1,25 @@
+using FluentAssertions;
+using Xunit;
+
+namespace Elastic.Apm.Tests
+{
+	/// <summary>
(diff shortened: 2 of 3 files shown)


test/Elastic.Apm.Tests/ConfigTests.cs
```diff
@@ -303,7 +303,7 @@ public void TransactionMaxSpansTest(string configurationValue, int expectedValue
 		}
 
 		[Fact]
-		public void DefaultLogLevelTest() => Agent.Config.LogLevel.Should().Be(LogLevel.Error);
```
gregkalapos: This was accessing `Agent.Config` and that caused the failing tests in CI.

test/Elastic.Apm.Tests/StaticTests.cs
```diff
@@ -0,0 +1,25 @@
+using FluentAssertions;
+using Xunit;
+
+namespace Elastic.Apm.Tests
+{
+	/// <summary>
+	/// These tests access the static <see cref="Agent"/> instance.
+	/// All other tests should have their own <see cref="ApmAgent"/> instance and not rely on anything static.
+	/// Tests accessing the static <see cref="Agent"/> instance can not run in parallel with tests that also access the static instance.
```
SergeyKleyman:
```suggestion
	/// Tests accessing the static <see cref="Agent"/> instance cannot run in parallel with tests that also access the static instance.
```

SergeyKleyman: BTW the problem is not only running tests using agent singleton in parallel - even when run sequentially the tests can affect each other because the tests (at least the one below) don't reset changes to the global state.

gregkalapos: Right, the order also matters. I really hope we won't have more such tests. We can control the ordering easily.

PR #621, opened 2019-11-21
Make IsConfigured thread safe

SergeyKleyman:
In #609 `Agent.IsInstanceCreated` was replaced `Agent.IsConfigured` introducing a regression - `IsInstanceCreated` was thread safe (in the sense that code checking if `IsInstanceCreated` is `true` before accessing `Agent.Instance` was guaranteed to see `Agent.Instance` constructed with  `AgentComponents` passed to `Agent.Setup()`) while [newly introduced `IsConfigured = true;` in `Agent.Setup()`](https://github.com/elastic/apm-agent-dotnet/pull/609/files#diff-2165dddf8d0c70f9a5985092338383bdR132) doesn't have memory release fence after it so .NET CLR memory model allows various optimization phases (compiler, JITer, CPU pre-fetching/caching, etc.) to effectively swap the order of `_components = agentComponents;` and `IsConfigured = true;` thus potentially causing situation where other thread see `Agent.IsConfigured` as `true` then calls `Agent.Instance` and creates agent singleton based on `_components` still being `null`.

Making `IsConfigured` baking field `volatile` introduces memory release fence after `_isConfigured = true;` (and, which just as important, memory acquire fence for reads from `IsConfigured`) thus disallowing optimizations to change the order of `_components = agentComponents;` and `_isConfigured = true;` from the point of view of other threads (i.e., threads other than the one that actually executed `Agent.Setup()`.

### DIFF
--- a/src/Elastic.Apm/Agent.cs
+++ b/src/Elastic.Apm/Agent.cs
@@ -96,12 +96,13 @@ namespace Elastic.Apm
 	{
 		private static readonly Lazy<ApmAgent> LazyApmAgent = new Lazy<ApmAgent>(() => new ApmAgent(_components));
 		private static AgentComponents _components;
+		private static volatile bool _isConfigured;
 
 		public static IConfigurationReader Config => Instance.ConfigurationReader;
 
 		internal static ApmAgent Instance => LazyApmAgent.Value;
 
-		public static bool IsConfigured { get; private set; }
+		public static bool IsConfigured => _isConfigured;
 
 		/// <summary>
 		/// The entry point for manual instrumentation. The <see cref="Tracer" /> property returns the tracer,
@@ -129,7 +130,7 @@ namespace Elastic.Apm
 				throw new InstanceAlreadyCreatedException("The singleton APM agent has already been instantiated and can no longer be configured.");
 
 			_components = agentComponents;
-			IsConfigured = true;
+			_isConfigured = true;
 		}
 
 		internal class InstanceAlreadyCreatedException : Exception


gregkalapos:
Ignore the failing test, that's already ok on `master`.

src/Elastic.Apm/Agent.cs
```diff
@@ -96,12 +96,13 @@ public static class Agent
...
 		private static readonly Lazy<ApmAgent> LazyApmAgent = new Lazy<ApmAgent>(() => new ApmAgent(_components));
 		private static AgentComponents _components;
+		private static volatile bool _isConfigured;
 
 		public static IConfigurationReader Config => Instance.ConfigurationReader;
 
 		internal static ApmAgent Instance => LazyApmAgent.Value;
 
-		public static bool IsConfigured { get; private set; }
+		public static bool IsConfigured => _isConfigured;
```
NeoXtreem: Adding `volatile` backing field achieves absolutely nothing as described in the PR.  From [here](http://www.albahari.com/threading/part4.aspx#_The_volatile_keyword):
> As it happens, Intel’s X86 and X64 processors always apply acquire-fences to reads and release-fences to writes — whether or not you use the **volatile** keyword — so this keyword has no effect on the _hardware_ if you’re using these processors. However, **volatile** _does_ have an effect on optimizations performed by the compiler and the CLR — as well as on 64-bit AMD and (to a greater extent) Itanium processors. This means that you cannot be more relaxed by virtue of your clients running a particular type of CPU.
> 
> (And even if you _do_ use volatile, you should still maintain a healthy sense of anxiety, as we’ll see shortly!).

I see no point in adding this backing field for the above reasons.  It will not solve anything to do with thread-safety, and will only add complexity and overhead to the code.

NeoXtreem: Adding `volatile` backing field achieves absolutely nothing as described in the PR.  From [here](http://www.albahari.com/threading/part4.aspx#_The_volatile_keyword):
> As it happens, Intel’s X86 and X64 processors always apply acquire-fences to reads and release-fences to writes — whether or not you use the **volatile** keyword — so this keyword has no effect on the _hardware_ if you’re using these processors. However, **volatile** _does_ have an effect on optimizations performed by the compiler and the CLR — as well as on 64-bit AMD and (to a greater extent) Itanium processors. This means that you cannot be more relaxed by virtue of your clients running a particular type of CPU.
> 
> (And even if you _do_ use volatile, you should still maintain a healthy sense of anxiety, as we’ll see shortly!).

I see no point in adding this backing field for the above reasons.  It will not solve anything to do with thread-safety, and will only add complexity and overhead to the code.

SergeyKleyman: I don't think I follow your arguments:
 - Are you saying that the problem described in this PR doesn't actually occur so there is no need to fix anything?
- Or do you agree that there's a problem but the fix implemented by this PR is incorrect?

NeoXtreem: I'm not familiar with the problem you describe (reordering lines of code in optimisation), but making the backing field `volatile` seems immaterial as the implicit backing field created by the compiler in the case of an auto-property already has an acquire-fence to reads and a release-fence to writes (as per the link I mentioned).

Do you have a link to anything that suggests such reordering would take place (that would be prevented by explicit fencing) as this would be new to me, but I'm open to the possibility of course (always willing to learn something new).

SergeyKleyman: > Do you have a link to anything that suggests such reordering would take place (that would be prevented by explicit fencing) as this would be new to me, but I'm open to the possibility of course (always willing to learn something new).

If you search the Internet for ".NET memory model volatile" you will find a huge amount of information about it - for example [here](https://docs.microsoft.com/en-us/archive/msdn-magazine/2013/january/csharp-the-csharp-memory-model-in-theory-and-practice-part-2
). But even the quote from the link you mentioned says

> However, volatile does have an effect on optimizations performed by the compiler and the CLR — as well as on 64-bit AMD and (to a greater extent) Itanium processors. This means that you cannot be more relaxed by virtue of your clients running a particular type of CPU.

The only factual issue with this quote is that it contrasts x64 and 64-bit AMD which are actually [the same thing](https://en.wikipedia.org/wiki/X86-64) (unlike let's say Itanium which is indeed completely different beast).
Putting x64 vs 64-bit AMD issue aside the quote correctly says that compiler, CLR and some hardware can change the observable order of execution and `volatile` is there to give the programmer control over that reordering.

SergeyKleyman: @NeoXtreem I am going to merge the PR as it is now. If you have any further objections please let us know (by opening a new issue or any other means you find convenient)

NeoXtreem: @SergeyKleyman I've added issue #623 as I read the link you sent me and it states:

> The x86-x64 processor will not reorder two writes

As for the JIT compiler, ordering is guaranteed on the x86-x64 processor as also mentioned in that link:

> the JIT compiler doesn’t need to use any special instructions on the x86-x64 to achieve volatile semantics

It might've been better to hold off for more than an hour (maybe a day) before merging to allow me to reply (to avoid having to create an issue if my findings are correct), but issue is created nonetheless.  Can continue discussion [there](https://github.com/elastic/apm-agent-dotnet/issues/623) if you like.

PR #625, opened 2019-11-22
Update supported-technologies.asciidoc

gregkalapos:
We don't support ASP.NET Core 2.0 anymore due to #608  (more details in the issue description) - Microsoft already dropped support a long time ago. 

Updated the docs accordingly, so from the current release we support ASP.NET Core 2.1 and later


gregkalapos:
Changing `2.x` to `2.1 and later` would also suggest we support `3.0`, which is true - we are not good with testing it, that should be addressed in https://github.com/elastic/apm-agent-dotnet/issues/546, but I tested 3.0 manually and we have 1 fix in the release specifically addressing a `3.0` issue so I think we are good with saying `2.1 and later`.

PR #629, opened 2019-11-28
Try fix broken 'GetNameFromRouteContext' in AspNetCore 3.0 by upgrading dependencies

yhvicey:
In AspNetCore 3.0 `IRouteFeature` is replaced by `IEndpointFeature` (see [doc](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-3.0#endpoint-routing-differences-from-earlier-versions-of-routing)) and not accessible in `context.Features`. To access route values now we should use `IRouteValuesFeature` but it's only available in AspNetCore 2.2 and later.

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -150,7 +150,7 @@ NOTE: The service name must conform to this regular expression: `^[a-zA-Z0-9 _-]
 
 [float]
 [[config-service-node-name]]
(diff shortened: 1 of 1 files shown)


yhvicey:
I'll create another PR exploring using `TargetFrameworks`

yhvicey:
Looks like there's already a PR (#632) fixing it, i'll wait for that one

PR #630, opened 2019-11-28
ci(jenkins): trigger opbeans release process

v1v:
## What does this pull request do?

Add a final stage in the release process to bump the opbeans agent dependency. This is triggered when a tag is created and matches the pattern `[0-9]+.[0-9]+.[0-9]+`

## Why is it important?

When a new tag, aka a new release, is created then, the automation will kick the opbeans build which consists of the following tasks:
- [x] Bump dependency versions in the opbeans-dotnet
- [x] Push changes. (Implementation within the opbeans-dotnet but the call it's done within this pipeline)
- [x] Create a tag to trigger the release pipeline in the opbeans-dotnet. That particular tag will match with the same one that the one in the agent itself.

This particular pipeline will push the changes into the `master` branch for the opbeans-dotnet and tag it with the name of the tag version too. Therefore, the opbeans-dotnet pipeline will trigger two builds, the one for its master branch and the other one for the just created tag. This particular flow is implemented within the opbeans-dotnet.

## Questions
- [x] Clarify with the apm-agent-dotnet whether the `Release to AppVeyor` stage should only happen when there is commit into master and nothing when there is a tagged release.

## Follow-ups
- Implement the release process to NuGet as it's done manually yet.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -333,38 +333,17 @@ pipeline {
               }
             }
           }
-          stage('Release') {
+          stage('AfterRelease') {
             options {
               skipDefaultCheckout()
-              timeout(time: 12, unit: 'HOURS')
             }
             when {
-              beforeInput true
               anyOf {
                 tag pattern: '\\d+\\.\\d+\\.\\d+', comparator: 'REGEXP'
                 expression { return params.Run_As_Master_Branch }
               }
             }
             stages {
-              stage('Notify') {
-                steps {
-                    emailext subject: '[apm-agent-dotnet] Release ready to be pushed',
-                            to: "${NOTIFY_TO}",
-                            body: "Please go to ${env.BUILD_URL}input to approve or reject within 12 hours."
-                }
-              }
-              stage('Release to NuGet') {
-                steps {
-                  input(message: 'Should we release a new version on NuGet?', ok: 'Yes, we should.')
-                  withGithubNotify(context: 'Release NuGet', tab: 'artifacts') {
-                    deleteDir()
-                    unstash 'source'
-                    dir("${BASE_DIR}"){
-                      release('secret/apm-team/ci/elastic-observability-nuget')
-                    }
-                  }
-                }
-              }
               stage('Opbeans') {
                 environment {
                   REPO_NAME = "${OPBEANS_REPO}"
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,6 +13,7 @@ pipeline {
     CODECOV_SECRET = 'secret/apm-team/ci/apm-agent-dotnet-codecov'
     GITHUB_CHECK_ITS_NAME = 'Integration Tests'
     ITS_PIPELINE = 'apm-integration-tests-selector-mbp/master'
+    OPBEANS_REPO = 'opbeans-dotnet'
   }
   options {
     timeout(time: 1, unit: 'HOURS')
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -331,22 +332,53 @@ pipeline {
               }
             }
           }
-          stage('Release to NuGet') {
-            options { skipDefaultCheckout() }
+          stage('Release') {
```
v1v: Created a top-level release stage which contains:
- a nested `Notify` stage to send an email with the link of the input approval. (NEW)
- a nested `Release to NuGet` stage as used to be.
- a new nested `opbeans` stage to bump and release the opbeans-dotnet project. (NEW)

Jenkinsfile
```diff
@@ -331,22 +332,53 @@ pipeline {
...
+            options {
+              skipDefaultCheckout()
+              timeout(time: 12, unit: 'HOURS')
```
v1v: Ensure there is 12 hours timeout for this particular release stage

Jenkinsfile
```diff
@@ -331,22 +332,53 @@ pipeline {
...
-                  release('secret/apm-team/ci/elastic-observability-nuget')
+            stages {
+              stage('Notify') {
+                steps {
+                    emailext subject: '[apm-agent-dotnet] Release ready to be pushed',
+                            to: "${NOTIFY_TO}",
+                            body: "Please go to ${env.BUILD_URL}input to approve or reject within 12 hours."
+                }
+              }
+              stage('Release to NuGet') {
```
gregkalapos: Soo... to be honest we never really used this stage - we usually manually deployed the packages (I can give more info on why... the point is: this is a little bit untested :) )

I think ideally there would be 2 nuget feeds that we use:
- One for every daily build (or on every commit to `master` - I'm fine with anything) - this would be an unofficial release mainly for internal usage - this could be AppVeyor - or something else.
- And one for the official releases - this would be nuget.org and we need to be careful with this one - we only should push official releases there.

Now, to your question:
>  Clarify with the apm-agent-dotnet whether the Release to AppVeyor stage should only happen when there is commit into master and nothing when there is a tag release.

Yes, that sounds good to me - if I understand correctly with that, we'd test all commit to master, no reason to test a tag release in my opinion. - But the important point (and this is the reason for my rant above: this should go into AppVayor (or some other nuget feed) and not into nuget.org). Then I'd handle the "official releases" differently.

I also tried to understand what this stage exactly does, and I'm a little bit unsure. The `def release(secret){` part that this uses calls `.ci/linux/deploy.sh`, which does a `nuget push`. But that pushes to a nuget feed that is configured somewhere -or uses the default, which is nuget.org.

So... I think we should discuss to which nuget feed we want to push for this (probably not nuget.org) and configure accordingly.

Jenkinsfile
```diff
@@ -332,22 +333,33 @@ pipeline {
               }
             }
           }
-          stage('Release to NuGet') {
-            options { skipDefaultCheckout() }
+          stage('AfterRelease') {
```
v1v: I renamed the tag as suggested in:
- https://github.com/elastic/apm-agent-java/pull/940#discussion_r354818054

This might be changed in the future once the release is fully automated, but to avoid any misleading let's use another name

Jenkinsfile
```diff
@@ -332,22 +333,32 @@ pipeline {
...
+            }
             when {
-              beforeAgent true
               anyOf {
-                tag "\\d+\\.\\d+\\.\\d+*"
+                tag pattern: '\\d+\\.\\d+\\.\\d+', comparator: 'REGEXP'
                 expression { return params.Run_As_Master_Branch }
               }
             }
-            steps {
```
v1v: This is not required for time being

Jenkinsfile
v1v: Just had a chat with @gregkalapos. This particular PR will delete the section regarding the NuGet release as it's not used yet. Then we can think about the release automation based on the above requirements. There will be some changes in the upcoming weeks to define this particular release process.

To keep it simple, this particular PR cares for all the tags that matches the `X.Y.Z` pattern, then, it will bump the agent version of the opbeans-dotnet and create a tag in that particular repo that matches the tag in this repo.

The current release process does create the tagged release in this repo after the artifacts have been uploaded to NuGet. There is also a requirement of matching the tagged releases with the version of the apm-agent-dotnet that has been uploaded to NuGet.

PR #632, opened 2019-12-01
Fix extracting route data for net core 2.2 and 3.0

### LINKED ISSUE #628: [AspNetCore 3.0] 'IRoutingFeature' cannot be extracted from context features
**Describe the bug**
`IRoutingFeature` cannot be extracted from context features, leading to `GetNameFromRouteContext(RouteData)` not work (`/user/profile/1` cannot be replaced by `/user/profile/{id}`)

**To Reproduce**
1. Create AspNet Core 3.0 project from template
2. Add Elastic.Apm.AspNetCore package reference
3. Add `app.UseElasticApm(Configuration);` in Startup.cs
4. You'll see transaction name won't be replaced by `RouteData`

**Expected behavior**
Transaction name should be replaced by route data

vhatsura:
fixes #628, related to #546 

It's a result of #622  PR

/cc @gregkalapos, @SergeyKleyman


src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
```diff
@@ -138,10 +138,15 @@ private void StopTransaction(Transaction transaction, HttpContext context)
 				if (!transaction.HasCustomName)
 				{
 					//fixup Transaction.Name - e.g. /user/profile/1 -> /user/profile/{id}
-					var routeData = (context.Features[typeof(IRoutingFeature)] as IRoutingFeature)?.RouteData;
+#if NETCOREAPP2_2
```
lafriks: This should probably be:
```suggestion
#if NETCOREAPP3_0
```

vhatsura: No, this is right due to [`IRouteValuesFeature` was occurred in ASP.NET Core 2.2](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.features.iroutevaluesfeature?view=aspnetcore-3.0#applies-to)

lafriks: sry, yes you are right

PR #634, opened 2019-12-02
Add Serilog Enricher for logcorrelation 

gregkalapos:
Work in progress.


src/Elastic.Apm.SerilogEnricher/ElasticApmEnricherExtension.cs
```diff
@@ -0,0 +1,26 @@
...
+	public static class ElasticApmEnricherExtension
+	{
+		/// <summary>
+		/// Enrich log events with a trace and transaction id properties containing the
+		/// current ids that the Elastic APM .NET Agent generated.
+		/// </summary>
+		/// <param name="enrichmentConfiguration">Logger enrichment configuration.</param>
+		/// <returns>Configuration object allowing method chaining.</returns>
+		/// <exception cref="ArgumentNullException">If <paramref name="enrichmentConfiguration"/> is null.</exception>
+		public static LoggerConfiguration WithElasticApmTraceId(
```
SergeyKleyman: Since it doesn't just add trace ID maybe a better name would be something like `WithElasticApmCorrelationInfo`?

test/LogCorrelation.Tests/SerilogTests.cs
```diff
@@ -0,0 +1,70 @@
...
+	public class SerilogTests
+	{
+		/// <summary>
+		/// Creates 1 simple transaction and makes sure that the log line created within the transaction has
+		/// the transaction and trace ids, and logs prior to and after the transaction do not have those.
+		/// </summary>
+		[Fact]
+		public void SerilogEnricherWithTransaction()
+		{
+			Agent.Setup(new AgentComponents(payloadSender: new NoopPayloadSender()));
```
SergeyKleyman: It might clash with other tests using agent's singleton and tests changing environment variables.
The former can be mitigated by the change proposed in https://github.com/elastic/apm-agent-dotnet/pull/636#discussion_r353559868.
The latter can be mitigated by using `TestAgentComponents` instead of `AgentComponents`.

Mpdreamz:
Superseded by: https://github.com/elastic/ecs-dotnet/pull/10/files

src/Elastic.Apm.SerilogEnricher/ElasticApmEnricherExtension.cs
Mpdreamz: 👍 `WithElasticApmCorrelationInfo` seems better. How about just `WithElasticApm()` ?

SergeyKleyman: I think `WithElasticApm` might be too broad. For example, some might infer from it that we can generate APM events from log entries (we have a few requests to support this use case).

gregkalapos: Changed it to `WithElasticApmCorrelationInfo`

gregkalapos:
Closing this, this moved to https://github.com/elastic/ecs-dotnet/pull/10/files.

PR #635, opened 2019-12-02
Logcorrelation with NLog and Log4Net

gregkalapos:
Work in Progress

Most of this will be moved to [ecs-dotnet](https://github.com/elastic/ecs-dotnet).

### DIFF
--- a/src/Elastic.Apm.Log4Net/Log4NetCorrelation.cs
+++ b/src/Elastic.Apm.Log4Net/Log4NetCorrelation.cs
@@ -1,0 +1,25 @@
+﻿using Elastic.Apm.Api;
+using log4net;
+
+namespace Elastic.Apm.Log4Net
+{
+	public class Log4NetCorrelation : ITransactionObserver
+	{
+		public void TransactionStarted(ITransaction transaction)
+		{
+			if (!Agent.IsConfigured) return;
+
+			var currentTransaction = Agent.Tracer.CurrentTransaction;
+
+			if (currentTransaction == null)
+			{
+				LogicalThreadContext.Properties.Remove("Transaction.Id");
+				LogicalThreadContext.Properties.Remove("Trace.Id");
+				return;
+			}
+
+			LogicalThreadContext.Properties["Transaction.Id"] = currentTransaction.Id;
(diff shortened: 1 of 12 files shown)


SergeyKleyman:
> Most of this will be moved to [ecs-dotnet](https://github.com/elastic/ecs-dotnet).

I probably miss some context but does it mean that this feature can be used without the agent?

src/Elastic.Apm.Log4Net/Log4NetCorrelation.cs
```diff
@@ -0,0 +1,25 @@
...
+			var currentTransaction = Agent.Tracer.CurrentTransaction;
+
+			if (currentTransaction == null)
+			{
+				LogicalThreadContext.Properties.Remove("Transaction.Id");
+				LogicalThreadContext.Properties.Remove("Trace.Id");
+				return;
+			}
+
+			LogicalThreadContext.Properties["Transaction.Id"] = currentTransaction.Id;
```
SergeyKleyman: Maybe it's worth using more globally unique names (for example `"Elastic.Apm.Transaction.Id"`) to avoid clashing with other libraries and/or the application?

gregkalapos: This follows the elastic common schema.

gregkalapos:
> > Most of this will be moved to [ecs-dotnet](https://github.com/elastic/ecs-dotnet).
> 
> I probably miss some context but does it mean that this feature can be used without the agent?

Not really - The idea is to have these in separate libraries (as we have our EF Core, ASP.NET Core, etc. things). But these libraries will reference the agent (that is how they read the current transaction). 

Yesterday in the .NET sync we agreed we move these log correlation related libraries to the `ecs-dotnet` repo. So these very small projects you see here will live there and my plan is to release those as separate libraries and users can reference them according to the logging framework they use.

Technically you'll be able to add these without having a configured agent, in that case the library will pull the agent, but it'll remain unconfigured and the library won't do anything.

gregkalapos:
Closing this for https://github.com/elastic/ecs-dotnet/pull/22

PR #643, opened 2019-12-08
Update .NET SDK to 3.1.100 in CI

gregkalapos:
Update `global.net` and all CI files to use the 3.1 SDKs. Additionally we'll also need a 2.x SDK, since we also want to keep targeting `netcoreapp2.x` additional to 3.x.

- The `Windows .NET Core` stage is green.
- The `Linux` stage breaks, because I'm unable to install the 2.0 SDK within the container we use - I'll ping someone from the automation team tomorrow, once someone helps me with jenkins, I think we can make it work
- The `Windows .NET Framework` stage breaks, because we have a too old MSBuild version - already pinged automation for this.


gregkalapos:
Merging this to `NetCore3` branch (not into `master`), since, at least 1 stage already works with this in CI.

PR #645, opened 2019-12-08
Add targetframework netcoreapp3.1

gregkalapos:
Adds targetframework `netcoreapp3.1`.

### DIFF
--- a/sample/WebApiSample/WebApiSample.csproj
+++ b/sample/WebApiSample/WebApiSample.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.1;netcoreapp2.2;netcoreapp3.0;netcoreapp3.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.1;netcoreapp2.2;netcoreapp3.0;netcoreapp3.1</TargetFrameworks>
     <SignAssembly>false</SignAssembly>
     <IsPackable>false</IsPackable>
   </PropertyGroup>
--- a/sample/WebApiSample/WebApiSample.csproj
+++ b/sample/WebApiSample/WebApiSample.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.1;netcoreapp2.2;netcoreapp3.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.1;netcoreapp2.2;netcoreapp3.0;netcoreapp3.0</TargetFrameworks>
     <SignAssembly>false</SignAssembly>
     <IsPackable>false</IsPackable>
   </PropertyGroup>
--- a/test/Elastic.Apm.AspNetCore.Tests/Helper.cs
+++ b/test/Elastic.Apm.AspNetCore.Tests/Helper.cs
@@ -32,7 +32,7 @@ namespace Elastic.Apm.AspNetCore.Tests
 						app.UseStaticFiles();
 						app.UseCookiePolicy();
 
-#if NETCOREAPP3_0
+#if NETCOREAPP3_0 || NETCOREAPP3_1
 						app.UseRouting();
 
 						app.UseAuthentication();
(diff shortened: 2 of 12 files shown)


test/Elastic.Apm.AspNetCore.Tests/Helper.cs
```diff
@@ -144,14 +144,16 @@ internal static void ConfigureServices(IServiceCollection services)
 			Startup.ConfigureServicesExceptMvc(services);
 			services
 #if NETCOREAPP3_0
-				.AddRazorPages()
+				.AddMvc()
```
vhatsura: In case of using `AddMvc()` for .netcoreapp3.0 and 3.1 it's make sense to move `AddMvc()` before preprocessor directives.

test/Elastic.Apm.AspNetCore.Tests/Helper.cs
```diff
@@ -144,14 +144,16 @@ internal static void ConfigureServices(IServiceCollection services)
...
 				.SetCompatibilityVersion(CompatibilityVersion.Version_3_0)
 #elif NETCOREAPP2_2
 				.AddMvc()
 				.SetCompatibilityVersion(CompatibilityVersion.Version_2_2)
-#else
+#elif NETCOREAPP2_1
 				.AddMvc()
 				.SetCompatibilityVersion(CompatibilityVersion.Version_2_1)
+#else
+				.AddMvc()
```
vhatsura: Does it make sense to set the compatibility version to `3_1`?

test/Elastic.Apm.AspNetCore.Tests/Helper.cs
```diff
@@ -32,7 +32,7 @@ public static class Helper
 						app.UseStaticFiles();
 						app.UseCookiePolicy();
 
-#if NETCOREAPP3_0
+#if NETCOREAPP3_0 || NETCOREAPP3_1
```
vhatsura: I thought about it when I was working for adding the `netcoreapp3.0` target, and it seems it's also relevant for the `netcoreapp3.1` target. What you think about extracting routing/mvc registration to a separate method?

sample/WebApiSample/WebApiSample.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.1;netcoreapp2.2;netcoreapp3.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.1;netcoreapp2.2;netcoreapp3.0;netcoreapp3.0</TargetFrameworks>
```
vhatsura: Seems like typo

test/Elastic.Apm.AspNetCore.Tests/Helper.cs
gregkalapos: I think there is no `CompatibilityVersion.Version_3_1`. The doc page also does not exist for 3.1: https://docs.microsoft.com/en-us/aspnet/core/mvc/compatibility-version?view=aspnetcore-3.1 - if you switch to [3.0](https://docs.microsoft.com/en-us/aspnet/core/mvc/compatibility-version?view=aspnetcore-3.0) it already says:

> The SetCompatibilityVersion method is a no-op for ASP.NET Core 3.0 apps. That is, calling SetCompatibilityVersion with any value of CompatibilityVersion has no impact on the application.

test/Elastic.Apm.AspNetCore.Tests/Helper.cs
gregkalapos: Agreed. In fact - after reading the docs of `SetCompatibilityVersion`  I'm not even sure we actually need it - all tests pass with those, as it seems the sample app does not use anything that'd depend on this - which is not a surprise, it's a fairly simple app. I'm thinking about removing all `SetCompatibilityVersion` calls - it makes no difference in our case.

test/Elastic.Apm.AspNetCore.Tests/Helper.cs
gregkalapos: > What you think about extracting routing/mvc registration to a separate method?

Makes sense, we repeated this code a lot. I moved it into a separate method.

gregkalapos:
Merging this now into `NetCore3` - we'll do a bigger review when we merge to `master`.

PR #646, opened 2019-12-09
ci(jenkins): support different targets

v1v:
## What?

Simplify the build and test stages by using Docker as much as possible or be able to configure the required tools in the CI workers.

## Why?

Simplify the requirements to run the builds, so, docker is the one in charge to drive what SDK is required.

image

## Tasks
- [x] Cache docker images in the Workers. 
- [x] Linux stages
- [x] Windows dotnet stages
- [x] Windows msbuild stages. Requirements [here](https://docs.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools?view=vs-2019)
- [x] Enable dotnet for windows - msbuild. Beta ONLY 

### MSBuild dependencies
- `includeOptionals` is required to get rid of the below error

```
C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\Microsoft.CSharp.CurrentVersion.targets(324,5): 
error MSB4019: The imported project "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\Roslyn\Microsoft.CSharp.Core.targets" was not found. 
Confirm that the expression in the Import declaration "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\Roslyn\Microsoft.CSharp.Core.targets" is correct, and that the file exists on disk. 
[C:\Users\jenkins.PACKER-5DA3F2D9\workspace\tnet_apm-agent-dotnet-mbp_PR-646\a\apm-agent-dotnet\test\Elastic.Apm.EntityFrameworkCore.Tests\Elastic.Apm.EntityFrameworkCore.Tests.csproj]
```
- `visualstudio2019-workload-webbuildtools` is not required for the MSBuild.

- `.ci/windows/prepare-test.bat` was created to run the tests in the MSBuild

image

otherwise, there will be an error

```
[...]

PR #647, opened 2019-12-09
Add targetframework netcoreapp3.0 netcoreapp3.1 netcoreapp2.2

gregkalapos:
Work in progress. 

Addresses #546.

This PR adds multiple target frameworks to test and sample projects. With this, we test the agent on:
- netcoreapp2.1 (it was already there)
- netcoreapp2.2
- netcoreapp3.0
- netcoreapp3.1

It also adapts code for ASP.NET Core 3+ and makes CI build all this. 

It'll be merged once all builds are green and after a final review.


Mpdreamz:
`netcoreapp2.2` is EOLnetcoreapp3.0 will be in march, do we want to support these?

https://dotnet.microsoft.com/platform/support/policy/dotnet-core

.ci/windows/tools.ps1
```diff
@@ -8,7 +8,8 @@ Install-WindowsFeature -Name Web-Server, Web-Mgmt-Tools ;
 Add-WindowsFeature NET-Framework-45-ASPNET ;
 Add-WindowsFeature Web-Asp-Net45 ;
 
-# Install .Net SDK'
+# Install .Net SDKs
+& choco install dotnetcore-sdk -m -y --no-progress -r --version 3.1.100
 & choco install dotnetcore-sdk -m -y --no-progress -r --version 2.1.505
```
Mpdreamz: Do we still need this version (` 2.1.505`) given we are installing `3.1.100`

gregkalapos: I'm unsure on this one. When I only had 3.1 SDK installed I saw failing builds on Linux telling that 2.1 SDK is missing and some nuget pacakges can't be found - adding 2.1 SDK solved the problem there. 

Then I tested on a clean windows machine, in that case, only with 3.1 SDK the `netcoreapp2.2` builds were skipped telling that some 2.2 stuff is missing. 

We can remove and give it a try, but as it seems, somehow only installing 3.1 is not enough.

gregkalapos:
> netcoreapp2.2 is EOL netcoreapp3.0 will be in march, do we want to support these?

I think we can drop those when they reach end of life. For 2.2 it'd be Dec. 23. I'd definitely keep `netcoreapp3.0` for now.

.ci/windows/tools.ps1
gregkalapos: @v1v works in CI in #646  and we saw this:

```
2019-12-10T13:07:49.913Z] A total of 1 test files matched the specified pattern.
[2019-12-10T13:07:50.175Z] Testhost process exited with error: It was not possible to find any compatible framework version
[2019-12-10T13:07:50.175Z] The framework 'Microsoft.AspNetCore.App', version '2.2.0' was not found.
[2019-12-10T13:07:50.175Z]   - The following frameworks were found:
[2019-12-10T13:07:50.175Z]       2.1.9 at [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
[2019-12-10T13:07:50.175Z]       2.1.13 at [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
[2019-12-10T13:07:50.175Z]       3.1.0 at [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
[2019-12-10T13:07:50.175Z] You can resolve the problem by installing the specified framework and/or SDK.
[2019-12-10T13:07:50.175Z] The specified framework can be found at:
[2019-12-10T13:07:50.175Z]   - https://aka.ms/dotnet-core-applaunch?framework=Microsoft.AspNetCore.App&framework_version=2.2.0&arch=x64&rid=win10-x64
[2019-12-10T13:07:50.175Z] . Please check the diagnostic logs for more information.
[2019-12-10T13:07:50.175Z] Results File: C:\Users\jenkins\workspace\tnet_apm-agent-dotnet-mbp_PR-646\apm-agent-dotnet\test\Elastic.Apm.AspNetCore.Tests\TestResults.xml
[2019-12-10T13:07:50.175Z] 
[2019-12-10T13:07:50.175Z] Test Run Aborted.
[2019-12-10T13:07:50.175Z
```

Mpdreamz: Aye netcoreapp needs the sdks, somewhere I knew this and forgot :)

src/Elastic.Apm.AspNetCore/Elastic.Apm.AspNetCore.csproj
```diff
@@ -1,16 +1,16 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
     <AssemblyName>Elastic.Apm.AspNetCore</AssemblyName>
     <RootNamespace>Elastic.Apm.AspNetCore</RootNamespace>
     <PackageId>Elastic.Apm.AspNetCore</PackageId>
     <Description>Elastic APM for ASP.NET Core. This package contains auto instrumentation for ASP.NET Core. See: https://github.com/elastic/apm-agent-dotnet/tree/master/docs</Description>
     <PackageTags>apm, monitoring, elastic, elasticapm, analytics, aspnetcore</PackageTags>
+    <TargetFrameworks>netstandard2.0;netcoreapp2.2</TargetFrameworks>
```
gregkalapos: 🤔 this is not ok - we can't publish this as `netcoreapp2.2`.  I'll come back and think about this.

vhatsura: > 🤔 this is not ok - we can't publish this as `netcoreapp2.2`. I'll come back and think about this.

What is the reason?

gregkalapos: > > 🤔 this is not ok - we can't publish this as `netcoreapp2.2`. I'll come back and think about this.
> 
> What is the reason?

Because we can't publish a single nuget package that has targets `netcoreapp2.2.` and `netstandard2.0` at the same time. The `dotnet publish` command also fails with this `.csproj`:

```
/local/share/dotnet/sdk/3.1.100/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.CrossTargeting.targets(27,5): error : The 'Publish' target is not supported without specifying a target framework. The current project targets multiple frameworks, please specify the framework for the published application. 
```

 which is not the case when we publish the `Elastic.Apm` package, which targets `netstandard2.0` and `net461`. 

The fact that `dotnet publish` fails without specifying a target kind of makes sense if you think about it: let's say you have an app that targets `netcoreapp2.2.`, or `netcoreapp3.0` - if you only add this to its `.csproj` file:

```
 <PackageReference Include="Elastic.Apm.AspNetCore" ...>
```

and `Elastic.Apm.AspNetCore` has both `netcoreapp` and `netstandard` target, then what target (meaning what `.dll` file in the nuget pacakge) should the `netcoreapp2.2` or `netcoreapp3.0` app pick? Both could be good, both are supported.

Overall I think the default approach should be to target `netstandardx` if we are a library.

But.. I think we don't even need `netcoreapp2.0`, you eliminated the need here: https://github.com/elastic/apm-agent-dotnet/pull/632/commits/7a09246f72adb8d1b9a8fa3b798a9ffa33e2308b , right?

vhatsura: Hm, interesting. I don't remember case when I need to target both netstandard and netcoreapp in nuget package, and your explanation sounds reasonable.

However, why are you using `publish` instead of `pack`? Also, I thought that package referencing works in a little bit different manner, at least if we have two different netstandard targets, e.g. `netstandard1.6` and `netstandard2.0`, the latest possible target will be used. In case of `netcoreapp2.2` it will be `netstandard2.0`. Also, such targets can be packed inside one nuget package.

gregkalapos: > However, why did you use publish instead of pack?

Yeah, you are right - and I have no idea why I tested with `dotnet publish` - sorry, that was a quick investigation on a train 🚋 ... And yes, both `netstandard2.0` and `netcoreapp2.2` can be in the same package, so that statement was definitely not true.

Nevertheless:

> BTW, it seems you're right and we don't need netcoreapp2.2 target.

Yeah, I think we both agree on this one, so, although the discussion is interesting, ultimately I guess we can just remove `netcoreapp2.2`.

gregkalapos:
I think this is ready to review.  I see some strange build fails during chocolatey installation, hopefully it's just some temporary thing in CI.

I tagged same people, please have a quick look at your corresponding part of this PR. 

@vhatsura I can't add you as a reviewer (no idea why it's not allowed here) but feel free to chime in.

sample/SampleAspNetCoreApp/SampleAspNetCoreApp.csproj
```diff
@@ -20,4 +31,21 @@
   <ItemGroup>
     <ProjectReference Include="..\..\src\Elastic.Apm.NetCoreAll\Elastic.Apm.NetCoreAll.csproj" />
   </ItemGroup>
+  <ItemGroup>
```
gregkalapos: We had a discussion about this here: https://github.com/elastic/apm-agent-dotnet/pull/622#discussion_r354009316 

I ran into that problem on netcoreapp2.x and adding this back solved the problem. It's still strange to me that we need this, but I'd suggest keeping it, otherwise we seem to run into the problem mentioned in https://github.com/elastic/apm-agent-dotnet/pull/622#discussion_r355208499.

test/Elastic.Apm.AspNetCore.Tests/DistributedTracingAspNetCoreTests.cs
```diff
@@ -38,9 +38,20 @@ public Task InitializeAsync()
 					{
 						Startup.ConfigureServicesExceptMvc(services);
 
-						services.AddMvc()
-							.AddApplicationPart(Assembly.Load(new AssemblyName(nameof(SampleAspNetCoreApp))))
-							.SetCompatibilityVersion(CompatibilityVersion.Version_2_1);
+						services
+#if NETCOREAPP3_0
+							.AddMvc()
+							.SetCompatibilityVersion(CompatibilityVersion.Version_3_0)
```
vhatsura: @gregkalapos, As I remember we discussed it and aligned on deleting `SetCompatibilityVersion`.

gregkalapos: Thanks! Yes, these seem to be leftovers - removed them.

.ci/docker/buildtools-windows/Dockerfile
```diff
@@ -0,0 +1,26 @@
+# escape=`
+
+# Use the latest Windows Server Core image with .NET Framework 4.8.
+FROM mcr.microsoft.com/dotnet/framework/sdk:4.8-windowsservercore-1909
```
v1v: For the time being this particular implementation is only for beta purposes, it's not yet ready but the framework is in place so we are working on supporting it but it might happen after this particular PR is merged.

gregkalapos:
Merging now.

Huge thanks to @v1v to help with CI! Also thanks to @vhatsura for a significant portion of this work.

Issue #651, opened 2019-12-10
APM .NET Agent don’t Capture Http Status 2xx 


luisfernandomoraes:
**Describe the bug**
When uploading the application by running docker-compose up, a request that returns status 2xx such as http://localhost:xxxx/hc is not being captured by the APM agent but a request that returns a 404 for example http://localhost:xxxx/foo/bar is being logged by APM.

**To Reproduce**
1. clone repository from https://github.com/luisfernandomoraes/Way2CodeChallenge
2. Then call http://localhost:xxxx/hc 
3. Then do http://localhost:xxxx/foo/bar
4. See Kibana APM

**Expected behavior**
Show the http requests with status 2xx.

rabberbock:
Same experience here, but I am using .net core 3.1.

luisfernandomoraes:
Yeah, I've tried the .net core 3.1 too 😑

willsilvano:
Same here.
I also noticed that POST requests are not being captured.

gregkalapos:
@luisaantonio I looked at your sample app and I think the problem is that the APM Agent is not registered as the 1. middleware [here](https://github.com/luisfernandomoraes/Way2CodeChallenge/blob/master/Way2.CodeChallenge.Api/Startup.cs#L100). 

The `app.UseAllElasticApm(Configuration);` should be the 1. line in the `Configure` method, otherwise the APM middleware does not have a chance to capture the request before it hits MVC, or any other Middleware.

willsilvano:
@gregkalapos Thanks! 
Your tip solved the problem here.

gregkalapos:
Great, thanks for the quick feedback.

I looked into our [docs](https://www.elastic.co/guide/en/apm/agent/dotnet/current/setup.html), and as it seems we don't document this properly. I'll add this to the docs!

luisfernandomoraes:
Thanks @gregkalapos!! 
Problem solved here too.

PR #654, opened 2019-12-11
[Docs] Update ASP.NET Core configuration instructions

gregkalapos:
Triggered by https://github.com/elastic/apm-agent-dotnet/issues/651.

The order in the `Configure` method when the agent is registered matters - this was not explained in the docs, so adding a note about this. 

Once we are ok with this, I'll also open a PR against `1.x`.

### DIFF
--- a/docs/setup.asciidoc
+++ b/docs/setup.asciidoc
@@ -56,6 +56,8 @@ public class Startup
 }
 ----
 
+NOTE: Make sure the `app.UseAllElasticApm(...)` line is the first line in the `Configure` method, otherwise the agent won't be able to properly measure the timing of your requests and potentially complete requests may be missed by the agent. 
+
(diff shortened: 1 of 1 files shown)


docs/setup.asciidoc
```diff
@@ -56,6 +56,8 @@ public class Startup
 }
 ----
 
+NOTE: Make sure the `app.UseAllElasticApm(...)` line is the first line in the `Configure` method, otherwise the agent won't be able to properly measure the timing of your requests and potentially complete requests may be missed by the agent.
```
bmorelli25:
```suggestion
The `app.UseAllElasticApm(...)` line must be the first line in the `Configure` method, otherwise the agent won't be able to properly measure the timing of your requests, and potentially complete requests may be missed by the agent. 
```

gregkalapos:
> I think this is more useful as a regular paragraph. WDYT? 

Sounds ok to me. 👍

PR #658, opened 2019-12-16
ci(jenkins): let's build those samples

v1v:


v1v:
I'm not sure what kind of dependencies are required, so closing this PR for the time being as the previous workaround is sufficient.

PR #665, opened 2019-12-19
Implement context.destination.address/port for DB spans

### LINKED ISSUE #611: Add span.context.destination.* properties for SIEM integration
See https://github.com/elastic/apm/issues/115#issuecomment-555814374

- [x] HTTP spans - #655
- [ ] DB spans - #665

SergeyKleyman:
Closes #611.
Implements the remaining part of #611 (context.destination.address/port for DB spans).

As a follow up PR it might be worth considering to improve `DbConnectionStringParser` cache strategy (currently it's just caches up to 100 first connection strings) - maybe LRU cache would be better?


src/Elastic.Apm/Model/DbSpanCommon.cs
```diff
@@ -2,17 +2,33 @@
...
+			internal const int MySql = 3306;
+			internal const int PostgreSql = 5432;
+		}
+
 		internal static Span StartSpan(IApmAgent agent, IDbCommand dbCommand) =>
 			(Span)ExecutionSegmentCommon.GetCurrentExecutionSegment(agent).StartSpan(dbCommand.CommandText.Replace(Environment.NewLine, " ")
 				, ApiConstants.TypeDb);
 
-		internal static void EndSpan(Span span, IDbCommand dbCommand, TimeSpan? duration = null)
+		internal static void EndSpan(Span span, IDbCommand dbCommand, IApmLogger logger, DbConnectionStringParser dbConnectionStringParser
```
vhatsura: Why we need a logger there if we aren't using it?

src/Elastic.Apm.EntityFrameworkCore/EfCoreDiagnosticListener.cs
```diff
@@ -12,9 +11,14 @@ namespace Elastic.Apm.EntityFrameworkCore
...
 	{
 		private readonly IApmAgent _agent;
+		private readonly DbConnectionStringParser _dbConnectionStringParser;
 		private readonly ConcurrentDictionary<Guid, Span> _spans = new ConcurrentDictionary<Guid, Span>();
 
-		public EfCoreDiagnosticListener(IApmAgent agent) => _agent = agent;
+		public EfCoreDiagnosticListener(IApmAgent agent)
+		{
+			_agent = agent;
+			_dbConnectionStringParser = new DbConnectionStringParser(agent.Logger);
```
vhatsura: Maybe it's make sense to have `DbConnectionStringParser` at `DbSpanCommon` class?
At first, it will allow not to instantiate parser in each `DbSpanCommon` consumer. Second, it makes sense from responsibility perspective: `DbSpanCommon` is responsible for parsing data from db command and enrich apm span. Why consumers dictate how to parse connection strings?

SergeyKleyman: > Maybe it's make sense to have `DbConnectionStringParser` at `DbSpanCommon` class?

Yes, I agree that current implementation is somewhat of hack just to work around `DbSpanCommon` being `static class`. We can make `DbSpanCommon` a regular class, keep an instance of it as a member of some other shared instance (for example `agent.Tracer`) and have an instance of `DbConnectionStringParser` to be a member of `DbSpanCommon`. WDYT?

vhatsura: Sound good to me

test/Elastic.Apm.EntityFrameworkCore.Tests/ExternalDbTests.cs
```diff
@@ -0,0 +1,345 @@
...
+using MySql.Data.MySqlClient;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Elastic.Apm.EntityFrameworkCore.Tests
+{
+	/// <summary>
+	/// Tests using external DB servers.
+	/// Tests will not run (even though they will show as passed) if any of the following environment variables is not set:
+	///  	- ELASTIC_APM_TESTS_XYZ_HOST
```
vhatsura: First of all, it would be nice to have documentation for configuring the test environment because no one looks through tests to find out information configuration. It was also my mistake in https://github.com/elastic/apm-agent-dotnet/pull/616#issuecomment-557233519 and the worst thing you don't have indicates about misconfiguration.
Second, it would be nice to automate such a process, as an example, with `TestEnvironment.Docker` package, which is used by me in #592. It will allow you to forget about the configuration of your environment, you need only run docker, which is not a big deal.

SergeyKleyman: I agree that this whole area of testing agent with in real environment - real databases, real application servers, etc., should be improved - I've opened #667 to follow up.

src/Elastic.Apm/Model/DbSpanCommon.cs
```diff
@@ -21,43 +37,80 @@ internal static void EndSpan(Span span, IDbCommand dbCommand, TimeSpan? duration
...
 				default:
-					span.Subtype = providerType; //TODO, TBD: this is an unknown provider
+					spanSubtype = dbConnectionClassName; //TODO, TBD: this is an unknown provider
+					defaultPort = null;
 					break;
 			}
+		}
 
-			switch (dbCommand.CommandType)
+		private static string GetSpanAction(CommandType dbCommandType)
```
vhatsura: As an alternative [`switch expression`](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#switch-expressions) from C# 8. Fortunately, it's possible after #647. It will be shorter and can be used without an additional method. Smth like that:

```csharp
span.Action = dbCommand.CommandType switch
{
    CommandType.Text => ApiConstants.ActionQuery,
    CommandType.StoredProcedure => ApiConstants.ActionExec,
    CommandType.TableDirect => "tabledirect",
    _ => dbCommand.CommandType.ToString()
};
```

src/Elastic.Apm/Helpers/UrlUtils.cs
```diff
@@ -7,26 +8,23 @@ internal static class UrlUtils
 	{
 		private const string ThisClassName = nameof(UrlUtils);
 
-		internal static bool TryExtractDestinationInfo(Uri url, out string host, out int? port, IApmLogger logger)
+		/// <returns><c>Destination</c> if successful and <c>null</c> otherwise</returns>
+		internal static Destination TryExtractDestination(Uri url, IApmLogger logger)
```
vhatsura: I know, we don't have an official convention about `TryParse` or `TryExtract` method signatures, however, for C# world is more convenient to return `bool` value and pass `out` parameter. Maybe it would be better to rename this method to not confuse people. WDYT?

src/Elastic.Apm/Helpers/DbConnectionStringParser.cs
```diff
@@ -0,0 +1,236 @@
...
+
+namespace Elastic.Apm.Helpers
+{
+	internal class DbConnectionStringParser
+	{
+		private const string ThisClassName = nameof(DbConnectionStringParser);
+		internal const int MaxCacheSize = 100;
+
+		private readonly IApmLogger _logger;
+		private readonly ThreadLocal<Dictionary<string, Destination>> _cache =
```
vhatsura: I am surprised by using `ThreadLocal` here. Diagnostic source listeners are executed in the thread from the thread pool, which is not the same each time. I run a small snippet to check it and got the following results (it was release configuration):

```
Thread Id: 7, Is background: True, Is from thread pool: True
Request '6' started: { "isMaster" : 1, "client" : { "driver" : { "name" : "mongo-csharp-driver", "version" : "2.9.2.0" }, "os" : { "type" : "Windows", "name" : "Microsoft Windows 10.0.18362", "architecture" : "x86_64", "version" : "10.0.18362" }, "platform" : "
.NET Core 3.0.1" }, "compression" : [] }
---   ---   ---   ---   ---   ---   ---   ---   ---
Thread Id: 8, Is background: True, Is from thread pool: True
Request '6' succeeded for isMaster in 14.1601ms with { "ismaster" : true, "maxBsonObjectSize" : 16777216, "maxMessageSizeBytes" : 48000000, "maxWriteBatchSize" : 100000, "localTime" : ISODate("2019-12-20T10:36:57.418Z"), "logicalSessionTimeoutMinutes" : 30, "co
nnectionId" : 8, "minWireVersion" : 0, "maxWireVersion" : 8, "readOnly" : false, "ok" : 1.0 }
---   ---   ---   ---   ---   ---   ---   ---   ---
Thread Id: 8, Is background: True, Is from thread pool: True
Request '7' started: { "buildInfo" : 1 }
---   ---   ---   ---   ---   ---   ---   ---   ---
Thread Id: 8, Is background: True, Is from thread pool: True
Request '7' succeeded for buildInfo in 1.8988ms with { "version" : "4.2.2", "gitVersion" : "a0bbbff6ada159e19298d37946ac8dc4b497eadf", "modules" : [], "allocator" : "tcmalloc", "javascriptEngine" : "mozjs", "sysInfo" : "deprecated", "versionArray" : [4, 2, 2, 0
], "openssl" : { "running" : "OpenSSL 1.1.1  11 Sep 2018", "compiled" : "OpenSSL 1.1.1  11 Sep 2018" }, "buildEnvironment" : { "distmod" : "ubuntu1804", "distarch" : "x86_64", "cc" : "/opt/mongodbtoolchain/v3/bin/gcc: gcc (GCC) 8.2.0", "ccflags" : "-fno-omit-fr
ame-pointer -fno-strict-aliasing -ggdb -pthread -Wall -Wsign-compare -Wno-unknown-pragmas -Winvalid-pch -Werror -O2 -Wno-unused-local-typedefs -Wno-unused-function -Wno-deprecated-declarations -Wno-unused-const-variable -Wno-unused-but-set-variable -Wno-missing
-braces -fstack-protector-strong -fno-builtin-memcmp", "cxx" : "/opt/mongodbtoolchain/v3/bin/g++: g++ (GCC) 8.2.0", "cxxflags" : "-Woverloaded-virtual -Wno-maybe-uninitialized -fsized-deallocation -std=c++17", "linkflags" : "-pthread -Wl,-z,now -rdynamic -Wl,--
fatal-warnings -fstack-protector-strong -fuse-ld=gold -Wl,--build-id -Wl,--hash-style=gnu -Wl,-z,noexecstack -Wl,--warn-execstack -Wl,-z,relro", "target_arch" : "x86_64", "target_os" : "linux" }, "bits" : 64, "debug" : false, "maxBsonObjectSize" : 16777216, "st
orageEngines" : ["biggie", "devnull", "ephemeralForTest", "wiredTiger"], "ok" : 1.0 }
---   ---   ---   ---   ---   ---   ---   ---   ---
Thread Id: 8, Is background: True, Is from thread pool: True
Request '8' started: { "getLastError" : 1 }
---   ---   ---   ---   ---   ---   ---   ---   ---
Thread Id: 8, Is background: True, Is from thread pool: True
Request '8' succeeded for getLastError in 1.3274000000000001ms with { "connectionId" : 8, "n" : 0, "syncMillis" : 0, "writtenTo" : null, "err" : null, "ok" : 1.0 }
---   ---   ---   ---   ---   ---   ---   ---   ---
Thread Id: 8, Is background: True, Is from thread pool: True
Request '9' started: { "find" : "people", "filter" : { }, "$db" : "diagnosticsource", "lsid" : { "id" : CSUUID("654c3f97-6643-4fa5-828e-ab474f8807fa") } }
---   ---   ---   ---   ---   ---   ---   ---   ---
Thread Id: 7, Is background: True, Is from thread pool: True
Request '9' succeeded for find in 8.997200000000001ms with { "cursor" : { "firstBatch" : [{ "_id" : ObjectId("5dfca35bb523d240ef428a13"), "name" : "Joe Smith", "email" : "jsmith@gmail.com", "age" : 40, "admin" : false }, { "_id" : ObjectId("5dfca35bb523d240ef42
8a14"), "name" : "Jen Ford", "email" : "jford@gmail.com", "age" : 45, "admin" : true }], "id" : NumberLong(0), "ns" : "diagnosticsource.people" }, "ok" : 1.0 }
---   ---   ---   ---   ---   ---   ---   ---   ---
Thread Id: 7, Is background: True, Is from thread pool: True
Request '10' started: { "$db" : "diagnosticsource", "lsid" : { "id" : CSUUID("654c3f97-6643-4fa5-828e-ab474f8807fa") } }
---   ---   ---   ---   ---   ---   ---   ---   ---
Thread Id: 8, Is background: True, Is from thread pool: True
Request '10' failed with 'MongoDB.Driver.MongoCommandException: $db command failed'
---   ---   ---   ---   ---   ---   ---   ---   ---
```

So, using `ThreadLocal` is less efficient for a caching scenario in our use case than `ConcurrentDictionary`. What was your motivation?

gregkalapos: Just chiming in quickly: I haven't look at the whole PR yet, just parsed it a little bit last evening and I also had the same question... I also thought that I'd just use `static ConcurrentDictionary`, unless there is a good reason not to.

src/Elastic.Apm/Model/DbSpanCommon.cs
SergeyKleyman: You are right - we don't need to pass a logger here. I'll fix it.

src/Elastic.Apm/Helpers/DbConnectionStringParser.cs
SergeyKleyman: It's an interesting point - I'll compare performance characteristics of using `ThreadLocal` vs. `ConcurrentDictionary`.

src/Elastic.Apm/Helpers/UrlUtils.cs
SergeyKleyman: I'm not sure it's that clear cut - IMHO `Try` prefix means that error is signaled by using return value and not by using exceptions. Exceptions is much better error handling mechanism (at least IMHO - for example Rust and Go people disagree :) so the motivation for using a different (and worse) error handling mechanism is performance. Throwing an exception in .NET [has significant performance cost](https://mattwarren.org/2016/12/20/Why-Exceptions-should-be-Exceptional/) so it should be reserved for truly exceptional situations. Of course, sometimes it's not easy to decide if certain situations are "exceptional" or "expected". That is why some libraries provide both `TryXyz()` and `Xyz()` APIs if there's a chance that `Xyz()` will be used as part of hot code path.
So assuming we agree that `Try` prefix just means error via return value then it's just a question of whether return value type has a special value that is never returned in case of success. For APIs such as `int.TryParse()` or `Dictionary.TryGetValue()` there is no special value to signal error so you have to use `bool` to signal error and `out` for actual return value. But in `TryExtractDestination` case `null` cannot be returned in case of success so it can be used to signal error.
I do agree with you that it becomes a question of choosing between consistency with already existing APIs and making APIs more concise by using modern features such as non-nullable references (or maybe tuples). Regarding consistency: for example `System.Uri.TryCreate()` uses `bool/out` even though it could have returned `null` to signal error. But OTOH sometimes BCL sacrifices consistency to make API easier to use - for example `Enumerable.FirstOrDefault` returns null if the sequence is empty. Regarding making API easier to use - especially now with C# 8  non-nullable references having `SomeClass? TryXyz()` should make it quite clear that API can return `null` and most likely to signal a error. You can see discussions about it [here](https://softwareengineering.stackexchange.com/questions/387674/c-8-non-nullable-references-and-the-try-pattern) and [here](https://stackoverflow.com/questions/3619691/check-reference-parameter-value-or-return-bool).

In the end I decided to split the difference by removing `Try` prefix and keeping the doc comment clarifying that `null` is returned on error.

src/Elastic.Apm/Model/DbSpanCommon.cs
SergeyKleyman: Done

src/Elastic.Apm.EntityFrameworkCore/EfCoreDiagnosticListener.cs
SergeyKleyman: Done.

src/Elastic.Apm/Helpers/UrlUtils.cs
vhatsura: Great explanation of your thoughts.

>In the end I decided to split the difference by removing Try prefix and keeping the doc comment clarifying that null is returned on error.

This is exactly what I meant, but maybe I shared it in a complicated way.

src/Elastic.Apm.EntityFrameworkCore/EfCoreDiagnosticListener.cs
```diff
@@ -11,10 +9,13 @@ namespace Elastic.Apm.EntityFrameworkCore
...
 	internal class EfCoreDiagnosticListener : IDiagnosticListener
 	{
-		private readonly IApmAgent _agent;
+		private readonly ApmAgent _agent;
 		private readonly ConcurrentDictionary<Guid, Span> _spans = new ConcurrentDictionary<Guid, Span>();
 
-		public EfCoreDiagnosticListener(IApmAgent agent) => _agent = agent;
+		public EfCoreDiagnosticListener(IApmAgent agent)
+		{
+			_agent = (ApmAgent)agent;
```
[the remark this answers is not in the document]
SergeyKleyman: > Overall I think having `DbSpanCommon` static and pass an `IApmAgent` to it still feels better than this hard cast.

The assumption that `IApmAgent agent` is actually `ApmAgent` was already there before this cast - [`DbSpanCommon.StartSpan` casts `ISpan` to `Span`](https://github.com/elastic/apm-agent-dotnet/blob/1.2.0/src/Elastic.Apm/Model/DbSpanCommon.cs#L11). In addition making  `DbSpanCommon` static will significantly complicate the tests as we saw it with the agent's singleton. I propose to leave as it is in this PR and re-evaluate this approach later if necessary.
I have an alternative in mind - please let me know if you would like to discuss it.

test/Elastic.Apm.EntityFrameworkCore.Tests/ExternalDbTests.cs
```diff
@@ -0,0 +1,349 @@
...
+					, () => { ExecuteTestCrudSequence(() => externalDbType.DbContextBuilder(connectionDetails)); });
+			}
+			mockPayloadSender.Transactions.Should().HaveCount(1);
+			mockPayloadSender.Spans.ForEach(span =>
+			{
+				span.Type.Should().Be(ApiConstants.TypeDb);
+				span.Subtype.Should().Be(externalDbType.SpanSubtype);
+				span.Action.Should().Be(ApiConstants.ActionQuery);
+				span.Context.Db.Type.Should().Be(Database.TypeSql);
+				span.Context.Destination.Address.Should().Be(connectionDetails.Host);
```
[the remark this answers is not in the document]
SergeyKleyman: IIUC `SQLEXPRESS` part is DB instance name - this test uses default DB instance. I've added a comment clarifying this assumption.

test/Elastic.Apm.EntityFrameworkCore.Tests/ExternalDbTests.cs
```diff
@@ -0,0 +1,349 @@
...
+		}
+
+		public class DbContextImplBase : DbContext
+		{
+			protected const string DbInstance = "ElasticApmExternalDbTests";
+
+			internal readonly ConnectionDetails ConnectionDetails;
+
+			protected DbContextImplBase(ConnectionDetails connectionDetails)
+			{
```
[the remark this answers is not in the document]
SergeyKleyman: Done

src/Elastic.Apm/Model/DbSpanCommon.cs
```diff
@@ -2,17 +2,39 @@
...
+			internal const int MsSql = 1433;
+			internal const int Oracle = 1521;
+			internal const int MySql = 3306;
+			internal const int PostgreSql = 5432;
+		}
+
+		private readonly DbConnectionStringParser _dbConnectionStringParser;
+
+		internal DbSpanCommon(IApmLogger logger)
+		{
```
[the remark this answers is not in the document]
SergeyKleyman: Done

src/Elastic.Apm/Helpers/DbConnectionStringParser.cs
```diff
@@ -0,0 +1,234 @@
...
+			new ThreadLocal<Dictionary<string, Destination>>(() => new Dictionary<string, Destination>());
+
+		internal DbConnectionStringParser(IApmLogger logger)
```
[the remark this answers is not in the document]
SergeyKleyman: Done

src/Elastic.Apm.EntityFrameworkCore/EfCoreDiagnosticListener.cs
```diff
@@ -11,10 +9,13 @@ namespace Elastic.Apm.EntityFrameworkCore
 {
 	internal class EfCoreDiagnosticListener : IDiagnosticListener
 	{
-		private readonly IApmAgent _agent;
+		private readonly ApmAgent _agent;
 		private readonly ConcurrentDictionary<Guid, Span> _spans = new ConcurrentDictionary<Guid, Span>();
 
-		public EfCoreDiagnosticListener(IApmAgent agent) => _agent = agent;
+		public EfCoreDiagnosticListener(IApmAgent agent)
```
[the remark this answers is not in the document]
SergeyKleyman: Done

src/Elastic.Apm/Helpers/DbConnectionStringParser.cs
SergeyKleyman: Done

src/Elastic.Apm/Helpers/DbConnectionStringParser.cs
```diff
@@ -0,0 +1,234 @@
...
+			var valueToParse = TrimDiscardable(valueToParseArg);
+
+			if (valueToParse.StartsWith(SqlServerLocalDbPrefix, StringComparison.OrdinalIgnoreCase)
+				|| valueToParse.StartsWith(SqlServerExpressUserInstancePrefix, StringComparison.OrdinalIgnoreCase))
+			{
+				destination.Address = "localhost";
+				return;
+			}
+
+			var dbInstanceSeparatorIndex = valueToParse.IndexOf(ServerNameDbInstanceSeparator);
```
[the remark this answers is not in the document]
SergeyKleyman: I assume you refer to the test failing because `ELASTIC_APM_TESTS_MS_SQL_HOST` contained DB instance name in addition to the host name - I've added a comment above.

src/Elastic.Apm.EntityFrameworkCore/EfCoreDiagnosticListener.cs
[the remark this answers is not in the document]
gregkalapos: Ok, let's keep it.

test/Elastic.Apm.EntityFrameworkCore.Tests/ExternalDbTests.cs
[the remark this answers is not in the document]
gregkalapos: Ok, now I understand your intent. Unfortunately leaving `SQLEXPRESS` does not make the test pass on my machine - as it seems it is not a default DB instance name and under the generated connection string `Data Source=GREGWINDOWSDEV;Initial Catalog=ElasticApmExternalDbTests;User ID=greg;Password=1234` (this password is not a secret ;) ) the database is not available. 

Anyways, there is already a follow up issue for this.

gregkalapos:
I saw 1 test is failing in CI:  `Elastic.Apm.SqlClient.Tests.SqlClientDiagnosticListenerTests`

https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-dotnet%2Fapm-agent-dotnet-mbp/detail/PR-665/16/tests/

Can't repro locally - I see the build was already re-triggered. 

If it's green I'm ok with merging it.

PR #696, opened 2020-01-20
ci(jenkins): Release to NuGet when tagged release

### LINKED ISSUE #619: Automate Agent release process
## Current process

The release process was automated to release builds from the master branch, see https://github.com/elastic/apm-agent-dotnet/issues/102

It actually supported two different central repos:
- AppVeyor - https://ci.appveyor.com/nuget/elastic
- NuGet - https://www.nuget.org/packages/Elastic.Apm.AspNetCore/

See https://github.com/elastic/apm-agent-dotnet/pull/172

Although the NuGet has been disabled for the time being. See https://github.com/elastic/apm-agent-dotnet/pull/630

## Actions
- [ ] Enable the release to NuGet when a tagged release is created.
- [ ] Document if required.

v1v:
## What does this PR do?

1. Enable the release automation to NuGet only when a tagged released is created with the format `[0-9].[0-9].[0-9](-alpha[0-9]*|-beta[0-9]*|-rc[0-9]*)?`.

1. The tagged release is created by the owner of the repo and pushed to the upstream, then the automation will happen afterward.

1. Tidy up some unnecessary notifications to GH as it does only work for PRs and simplify the when condition for the ITs.

1. Send an email when the release to NuGet is ready to be pushed since it's required an input approval.

1. Add RELEASING.md docs.

## Why is it important?

Automate the release process that was already created sometime ago.

## Related issues

Closes #619
Related to #172

## Questions
- [x] Should it be a fully automated process or a one-button release process? 
Just agreed to use a one-button release process.
- [x] If one button-release process then an email with the link to the input approval could help. See https://github.com/elastic/apm-agent-python/pull/609 as an example.

## Test Cases
Let's use my forked repo and leave the tag in place for future references. For such, let's skip all the stages but the release and change the deploy to be DRY based :)

#### Regex work as expected with the format `0.0.1`, `1.2.0-rc`, `1.2.0-alpha123` and `1.2.0-bad`

How did I prepare the test environment?
1. Created an MBP locally similar to https://github.com/elastic/apm-agent-dotnet/blob/master/.ci/jobs/apm-agent-dotnet-mbp.yml
1. `git checkout test/release-nuget` is the one with the changes of this particular PR but with the `when` condition to skip all the stages that are not related to the release to NuGet.
1. `git tag 0.0.1` and `git push origin 0.0.1` (the same for 1.2.0-rc and 1.2.0-bad)

<details><summary>Expand to view the whole tags status</summary>
<p>

Expected behavior:
- Stable builds for the unmatched patterns
- Aborted builds for the matched patterns, since their input approval was aborted

image

</p>
</details>

<details><summary>Expand to view 0.0.1 release</summary>
<p>

Expected behavior: 
* Input approval.
* Email notification.
* If approved then a release to NuGet will happen

image

image

image

image

image

Tag -> https://github.com/v1v/apm-agent-dotnet/tree/0.0.1
</p>
</details>

<details><summary>Expand to view 1.2.0-bad release</summary>
<p>

Expected behavior: 
* No release.

image

Tag -> https://github.com/v1v/apm-agent-dotnet/tree/1.2.0-bad
</p>
</details>

<details><summary>Expand to view 1.2.0-rc release</summary>
<p>

Expected behavior: 
* Input approval.
* Email notification.
* If approved then a release to NuGet will happen

image

Tag -> https://github.com/v1v/apm-agent-dotnet/tree/1.2.0-rc
</p>
</details>

#### Email notification when a tag released is ready to be pushed.

I'm afraid I do need to create a PR to test this particular scenario, as my local Jenkins instance has not got any email system in place.

How did I test the email notification?
1. https://github.com/elastic/apm-agent-dotnet/pull/699

<details><summary>Expand to view email release input approval notification</summary>
<p>

Expected behavior: 
* Input approval.
* Email notification.

image

image

</p>
</details>

### DIFF
--- a/ElasticApmAgent.sln
+++ b/ElasticApmAgent.sln
@@ -43,6 +43,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "files", "files", "{B406113B
 		CONTRIBUTING.md = CONTRIBUTING.md
 		LICENSE = LICENSE
 		README.md = README.md
+		RELEASING.md = RELEASING.md
 		global.json = global.json
 		Directory.Build.props = Directory.Build.props
 	EndProjectSection
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -363,7 +363,7 @@ pipeline {
               beforeInput true
               beforeAgent true
               // Tagged release events ONLY
-              tag pattern: '\\d+\\.\\d+\\.\\d+(-alpha|-beta|-rc)?', comparator: 'REGEXP'
+              tag pattern: '\\d+\\.\\d+\\.\\d+(-(alpha|beta|rc)\\d*)?', comparator: 'REGEXP'
             }
             stages {
               stage('Notify') {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -401,12 +401,9 @@ pipeline {
           stage('Integration Tests') {
             agent none
             when {
-              beforeAgent true
-              allOf {
-                anyOf {
-                  environment name: 'GIT_BUILD_CAUSE', value: 'pr'
-                  expression { return !params.Run_As_Master_Branch }
-                }
+              anyOf {
+                changeRequest()
+                expression { return !params.Run_As_Master_Branch }
               }
             }
             steps {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -341,12 +341,10 @@ pipeline {
               }
             }
             steps {
-              withGithubNotify(context: 'Release AppVeyor', tab: 'artifacts') {
-                deleteDir()
-                unstash 'source'
-                dir("${BASE_DIR}"){
-                  release('secret/apm-team/ci/elastic-observability-appveyor')
-                }
+              deleteDir()
+              unstash 'source'
+              dir("${BASE_DIR}"){
+                release('secret/apm-team/ci/elastic-observability-appveyor')
               }
             }
             post{
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -354,17 +354,36 @@ pipeline {
               }
             }
           }
-          stage('Release to NuGet') {
-            options { skipDefaultCheckout() }
+          stage('Release') {
+            options {
+              skipDefaultCheckout()
+              timeout(time: 12, unit: 'HOURS')
+            }
             when {
+              beforeInput true
               // Tagged release events ONLY
               tag pattern: '\\d+\\.\\d+\\.\\d+(-alpha|-beta|-rc)?', comparator: 'REGEXP'
             }
-            steps {
-              deleteDir()
-              unstash 'source'
-              dir("${BASE_DIR}") {
-                release('secret/apm-team/ci/elastic-observability-nuget')
+            stages {
+              stage('Notify') {
+                steps {
+                  emailext subject: '[apm-agent-dotnet] Release ready to be pushed',
+                           to: "${NOTIFY_TO}",
+                           body: "Please go to ${env.BUILD_URL}input to approve or reject within 12 hours."
+                }
+              }
+              stage('Release to NuGet') {
+                input {
+                  message 'Should we release a new version?'
+                  ok 'Yes, we should.'
+                }
+                steps {
+                  deleteDir()
+                  unstash 'source'
+                  dir("${BASE_DIR}") {
+                    release('secret/apm-team/ci/elastic-observability-nuget')
+                  }
+                }
               }
             }
           }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -356,6 +356,20 @@ pipeline {
               }
             }
           }
+          stage('Release to NuGet') {
+            options { skipDefaultCheckout() }
+            when {
+              // Tagged release events ONLY
+              tag pattern: '\\d+\\.\\d+\\.\\d+', comparator: 'REGEXP'
+            }
+            steps {
+              deleteDir()
+              unstash 'source'
+              dir("${BASE_DIR}") {
+                release('secret/apm-team/ci/elastic-observability-nuget')
+              }
+            }
+          }
           stage('AfterRelease') {
             options {
               skipDefaultCheckout()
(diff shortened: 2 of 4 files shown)


gregkalapos:
Thanks @v1v 

I'd prefer a one-button release process - I think a release is a "big enough deal" to have someone clicking a button for it.

 > a tagged released is created with the format v[0-9].[0-9].[0-9]

To this I have a small comment: in NuGet we can also add strings to the version numbers. Typical example is a pre-release package. [Here are](https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#pre-release-versions) is the docs.

So if possible I'd also enable: `-alpha`, `-beta `, `-rc` after the last number.

ElasticApmAgent.sln
```diff
@@ -43,6 +43,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "files", "files", "{B406113B
 		CONTRIBUTING.md = CONTRIBUTING.md
 		LICENSE = LICENSE
 		README.md = README.md
+		RELEASING.md = RELEASING.md
```
v1v: Not sure whether this is required though

Jenkinsfile
```diff
@@ -356,6 +354,41 @@ pipeline {
               }
             }
           }
+          stage('Release') {
+            options {
+              skipDefaultCheckout()
+              timeout(time: 12, unit: 'HOURS')
```
v1v: We wanna last the input approval for 12 hours, to let enough time for the input approval to be accepted or rejected

Jenkinsfile
```diff
@@ -356,6 +354,41 @@ pipeline {
...
+              stage('Notify') {
+                steps {
+                  emailext subject: '[apm-agent-dotnet] Release ready to be pushed',
+                           to: "${NOTIFY_TO}",
+                           body: "Please go to ${env.BUILD_URL}input to approve or reject within 12 hours."
+                }
+              }
+              stage('Release to NuGet') {
+                input {
+                  message 'Should we release a new version?'
```
v1v: Let's start with the simple use case when the versioning is created beforehand as this is a tagged release event-based.

Jenkinsfile
```diff
@@ -389,12 +422,9 @@ pipeline {
...
             agent none
             when {
-              beforeAgent true
-              allOf {
-                anyOf {
-                  environment name: 'GIT_BUILD_CAUSE', value: 'pr'
-                  expression { return !params.Run_As_Master_Branch }
-                }
+              anyOf {
+                changeRequest()
```
v1v: Cosmetic change as there was a nested condition that was unnecessary.

Jenkinsfile
```diff
@@ -356,6 +354,39 @@ pipeline {
               }
             }
           }
+          stage('Release') {
+            options {
+              skipDefaultCheckout()
+              timeout(time: 12, unit: 'HOURS')
+            }
+            when {
+              beforeInput true
```
kuisathaverat: I do not remember if it will provision an agent, I think so, I'd add the `beforeAgent true` too

v1v: It does use the top-level agent. But I think it's good to add it explicitly, in case we change it in the future to use a specific worker.

Mpdreamz:
Alpha/rc etcetera releases should ideally be suffixed with number e.g `rc1` and `alpha2` this is pattern the stack also uses. Not checked but will the regex pick this up?

Jenkinsfile
mdelapenya: Nit: could we add the version to the message? I.e.: `Should we release "v1.0.0"?` 🤔 

Whenever seeing this kind of screens (do you like to remove the item?), as a user I like having more information in the screen.

v1v:
> Alpha/rc etcetera releases should ideally be suffixed with number e.g rc1 and alpha2 this is pattern the stack also uses. Not checked but will the regex pick this up?

Good point, this pattern does not allow those figures :(

IIUC, the regex for that requirement could be `[0-9].[0-9].[0-9](-alpha[0-9]*|-beta[0-9]*|-rc[0-9]*)?`, therefore the below matches should be allowed:
- 1.2.3
- 1.2.3-rc
- 1.2.3-beta
- 1.2.3-alpha
- 1.2.3-alpha1
- 1.2.3-alpha99999

I definitely could use a more simplistic pattern, but don't wanna be sure it's also easy to read. It's case sensitive, btw, is that an issue?

RELEASING.md
```diff
@@ -0,0 +1,60 @@
...
+ git reset --hard upstream/master
+
+ ## <major>, <minor>, <bug> and <suffix> should be replaced accordingly. <suffix> is an optional one.
+ git tag <major>.<minor>.<bug>(-<suffix>)?
+
+ ## Push commit to the usptream repo.
+ git push upstream <major>.<minor>.<bug>(-<suffix>)?
+ ```
+
+The above commands will push the GitHub tag and will trigger the corresponding [Build pipeline](Jenkinsfile) which will run all the required stages to satisfy the release is in a good shape, then it will ask for approval to release to the NuGet repo.
```
cachedout: It's not totally clear from reading this how this input is asked for or where a release manager can go to approve or deny it. Perhaps a link and some additional instructions would help here?

cachedout: Great improvement here! Thanks!

Jenkinsfile
v1v: If we do something like that, it might be required to grab the version from the properties file, `./src/Directory.Build.props` and for such, I see two concerns:
- If that file change somehow in the future, it will require to change the CI pipeline accordingly. I'm not a huge fan of duplicating that.
- It might require some overhead with environment variables to be created on the fly at the very beginning to be able to change the message with that. So a bit hard to maintain it.

On the other hand, the input approval will always refer to the tag release, therefore the CI job name will always contain the version. And also the email notification with the link to the input approval, see https://github.com/elastic/apm-agent-dotnet/pull/696/files#diff-58231b16fdee45a03a4ee3cf94a9f2c3R373 that refers to the `BUILD_URL` env variable.

I'm not saying we should not do it, what do you think if we move forward and then review this release automation process after a few releases?

Issue #321, opened 2019-06-26
Send agent specific runtime metrics


gregkalapos:
In https://github.com/elastic/apm-agent-dotnet/pull/248 the 1. phase of the work related to metrics landed with basic metrics.

Next step: we should send runtime specific metrics (GC, threads, etc).

Tasks: 
- Decide what metrics to collect and send
- Implement

gregkalapos:
#697 introduced GC metrics - we can add more metrics later based on feedback.

PR #697, opened 2020-01-21
Add runtime metrics - GC

gregkalapos:
Addressing https://github.com/elastic/apm-agent-dotnet/issues/321. 

In this round, I add GC metrics. 

Currently planned:

- `clr.gc.count`
- `clr.gc.gen0size`
- `clr.gc.gen1size`
- `clr.gc.gen2size`
- `clr.gc.gen3size`

The agent collects these metrics through EventSource on .NET Core and through Microsoft.Diagnostics.Tracing.TraceEvent (ETW) on Full Framework.

Changes in tests: We had lots of test that instantiated an `AgentComponents` and that started collecting all metrics and subscribed to `EventSource`. The tests run in parallel, so we had multiple `AgentComponents` subscribing and unsubscribing to `EventSource`, which does not seem to work. To avoid this (which does not happen in prod, so no reason to test for it at the moment) I adapted all tests to either use `TestAgentComponents` which does not collect metrics, or skip metrics collection in case a real `AgentComponents` is needed.

Some Lens visualization:

Heap sizes:

image

image

GC count over time:

image

### DIFF
--- a/docs/metrics.asciidoc
+++ b/docs/metrics.asciidoc
@@ -100,3 +100,66 @@ platform: all
 
 The total physical memory the process has.
 --
+
+[float]
+[[metrics-runtime]]
+=== Runtime metrics
+
+
+*`clr.gc.count`*::
++
+--
+type: long
+
+Platform: all.
+
+The total number of GC collections that have occurred.
+--
+
+*`clr.gc.gen0size`*::
++
+--
+type: long
+
+format: bytes
+
+Platform: all.
+
+The size of the generation 0 heap.
+--
+
+*`clr.gc.gen1size`*::
++
+--
+type: long
+
+format: bytes
+
+Platform: all.
+
+The size of the generation 1 heap.
+--
+
+*`clr.gc.gen2size`*::
++
+--
+type: long
+
+format: bytes
+
+Platform: all.
+
+The size of the generation 2 heap.
+--
+
+*`clr.gc.gen3size`*::
++
+--
+type: long
+
+format: bytes
+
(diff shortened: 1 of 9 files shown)


vhatsura:
@gregkalapos, as I understand, such metrics are accessible only for .Net Core 2.2 and higher. So, it would be nice to have a mechanism, which allows enabling specific metrics to provide by framework condition.

gregkalapos:
> @gregkalapos, as I understand, such metrics are accessible only for .Net Core 2.2 and higher. So, it would be nice to have a mechanism, which allows enabling specific metrics to provide by framework condition.

That's a good point, but new PRs solved these: For Full Framework I added `Microsoft.Diagnostics.Tracing.TraceEvent` and we simply collect these through ETW - at least that's my current plan.

gregkalapos:
@bmorelli25 jfyi there are some new doc lines here in `docs/metrics.asciidoc `.

gregkalapos:
Failing test due to `The remote name could not be resolved: 'httpstat.us'`

 🤷‍♂ 😭 😢

WillGunn:
I'm in the middle of running down a memory leak issue, and I believe its related to this change.

It looks like ETW tracing in full .net framework never releases the GCPerHeapHistory objects for the lifetime of the Trace, causing memory to grow everytime a GC ends.
image


As APM uses ETW tracing for garbage collection metrics, and keeps an ETW Tracing session active for the lifetime of the process, this causes the memory usage of the process using APM to keep growing.

Issue #700, opened 2020-01-21
Elastic.Apm.AspNetFullFramework not showing database calls


sawo1337:
Only the HTTP calls are shown in Kibana in the Trace sample, there is nothing on database calls
There is nothing in the documentation specific to database queries interception for the full .net framework - https://www.elastic.co/guide/en/apm/agent/dotnet/current/setup.html#setup-asp-net

1. Create a new project ASP.NET Web Application (.NET Framework) with MVC and add the Elastic.Apm.AspNetFullFramework nugget (1.2)
2. Web.config:

```
<appSettings>
    <add key="ElasticApm:ServerUrls" value="http://IP.ADDR:8200/" />
  </appSettings>
    <modules>
      <add name="ElasticApmModule" type="Elastic.Apm.AspNetFullFramework.ElasticApmModule, Elastic.Apm.AspNetFullFramework" />
    </modules>
```

3. Execute any type of query using System.Data.SqlClient, for example

4. See the output in Kibana - only HTTP call is shown

**Expected behavior**
Trace sample to include SQL calls as well.

![kibana](https://user-images.githubusercontent.com/31248804/72808217-c0102800-3c61-11ea-99f2-1a9bbd022087.png)

gregkalapos:
Hi, 

by default db monitoring is not turned on in the Full Framework setup.

If you use Entity Framework 6, then here is the doc on how to turn it on: https://www.elastic.co/guide/en/apm/agent/dotnet/current/setup.html#setup-ef6

If you use EF Core, then simply execute this code once when you app starts up:

```
Agent.Subscribe(new EfCoreDiagnosticsSubscriber())
```

`EfCoreDiagnosticsSubscriber` lives in this NuGet package: https://www.nuget.org/packages/Elastic.Apm.EntityFrameworkCore/

gregkalapos:
Currently we only have EF6 and EFCore official support. 

We also have something for `SqlClient` - it's not released yet and not documented, but if that's what you use I can help you to setup.

So these are the 2 (+1) database libraries we support at the moment by auto instrumentation. 

Alternatively, you can also manually track the db calls by the [public agent API](https://www.elastic.co/guide/en/apm/agent/dotnet/current/public-api.html) - although I admit this is less convenient. 

What database do you use?

sawo1337:
Its MSSQL, if I can get this running for SQLClient, that would be great!

Attached is a piece of code to illustrate typical db call:
[...]

PR #703, opened 2020-01-23
add net461 target for Elastic.Apm.Tests project

vhatsura:
During work under #700, I need to have the ability to run tests under the net461 target framework.

Changes are very simple:
1. Add net461 target for `Elastic.Apm.Tests`
2. Add `IsCompletedSuccessfully()` extensions method for `Task` due to `IsCompletedSuccessfully` property is missed in net461
3. Fix other small method calls

So, I understand that infrastructure needs to be adjusted in such case and very appreciative to hear feedback on how it can be done in a more efficient way without a lot of changes in CI infrastructure.

A little bit of information, why we need to target a net461 framework. The `SqlClient` spread execution events through different approaches in .Net Framework and .Net Core: [`EventSource` in .Net Framework](https://github.com/dotnet/SqlClient/blob/master/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlEventSource.cs) and [`DiagnosticSource` in .Net Core](https://github.com/dotnet/SqlClient/blob/master/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlClientDiagnosticListenerExtensions.cs). As we have different behaviour for different frameworks (`HttpClientDiagnosticListener` also has different implementations) it would be nice cover this by tests at appropriate platform.

/cc, @gregkalapos

### DIFF
--- a/test/Elastic.Apm.Tests/ApiTests/ApiTests.cs
+++ b/test/Elastic.Apm.Tests/ApiTests/ApiTests.cs
@@ -10,6 +10,7 @@ using Elastic.Apm.Logging;
 using Elastic.Apm.Tests.Mocks;
 using FluentAssertions;
 using Xunit;
+
 // ReSharper disable AccessToDisposedClosure
 
 namespace Elastic.Apm.Tests.ApiTests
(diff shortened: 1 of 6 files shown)


gregkalapos:
@vhatsura some of this was I guess already done in https://github.com/elastic/apm-agent-dotnet/pull/480. 

We had some discussion if we solve some things differently, but I guess the easiest would be to just merge #480 .

vhatsura:
> @vhatsura some of this was I guess already done in #480.
> 
> We had some discussion if we solve some things differently, but I guess the easiest would be to just merge #480.

Auch, I didn't see it before. I need to refresh my memory with all PRs and issue in the repository ;)

gregkalapos:
No worries. I also need testing 461 - especially with metrics it'll be very useful. 

I'll spend time on #480 tomorrow and possible merge it - that'll hopefully also unblock you.

Issue #702, opened 2020-01-21
Add documentation for `Elastic.Apm.SqlClient` package


vhatsura:

vhatsura:
it would be nice to mention that instrumentation will be worked for `Microsoft.Data.SqlClient` on .Net Framework starting from `1.1.1` version. https://github.com/elastic/apm-agent-dotnet/pull/704#discussion_r380955327

PR #704, opened 2020-01-26
Listener for events from SqlClient for .Net Framework

vhatsura:
solves #700 

documentation will be added in scope of #702 

What needs to be done:
- [x] try to enable instrumentation engine and check SQL command text (by default, it isn't collected)
- [x] fill destination object of span (after merging #665)
- [x] check other provider types. Can they spread events via EventSource?
`System.Data.SQLite` and `Microsoft.Data.Sqlite` don't spread events via EventSource, however, does it means that other providers also don't do it? Can we set `Subtype` property of span to `mssql`?
- [x] handle sqlExceptionNumber

Unfortunately, it's not possible to capture command text with an event listener. It's possible to do it with [instrumentation engine](https://docs.microsoft.com/en-us/azure/azure-monitor/app/asp-net-dependencies#advanced-sql-tracking-to-get-full-sql-query), however, it seems in such case command text is captured from another place (I've checked command text in the payload with enabled instrumentation engine and hosting in IIS, no luck):

>In the above cases, the correct way of validating that instrumentation engine is correctly installed is by validating that the SDK version of collected DependencyTelemetry is 'rddp'. 'rdddsd' or 'rddf' indicates dependencies are collected via DiagnosticSource or EventSource callbacks, and hence full SQL query won't be captured.

Capture span representation:

```json
{
  "_index": "apm-7.5.2-span-000001",
  "_type": "_doc",
  "_id": "n25DJnABZF4_3hDy4m6U",
  "_version": 1,
  "_score": null,
  "_source": {
    "parent": {
      "id": "128e06672de0f0bb"
    },
    "observer": {
      "hostname": "ad6e0ae55174",
      "id": "a6272462-6b0f-4b04-9081-08f05f8037dc",
      "type": "apm-server",
      "ephemeral_id": "4ea50d25-d305-4afb-a034-3a903da8bbd4",
      "version": "7.5.2",
      "version_major": 7
    },
    "agent": {
      "name": "dotnet",
      "version": "1.2"
    },
    "trace": {
      "id": "4ed85f88e65ae40377fc1161cb094ebe"
    },
    "@timestamp": "2020-02-08T19:25:38.752Z",
    "ecs": {
      "version": "1.1.0"
    },
    "service": {
      "name": "WebApplication1"
    },
    "processor": {
      "name": "transaction",
      "event": "span"
    },
    "transaction": {
      "id": "128e06672de0f0bb"
    },
    "timestamp": {
      "us": 1581189938752426
    },
    "span": {
      "duration": {
        "us": 92956
      },
      "name": "model",
      "id": "a5d61d8d6c16b1e5",
      "type": "db",
      "db": {
        "instance": "model",
        "statement": "",
        "type": "sql"
      }
    }
  },
  "fields": {
    "@timestamp": [
      "2020-02-08T19:25:38.752Z"
    ]
  },
  "highlight": {
    "span.id": [
      "@kibana-highlighted-field@a5d61d8d6c16b1e5@/kibana-highlighted-field@"
    ]
  },
  "sort": [
    1581189938752
  ]
}
```


src/Elastic.Apm.SqlClient/SqlClientDiagnosticSubscriber.cs
```diff
@@ -9,13 +10,20 @@ public class SqlClientDiagnosticSubscriber : IDiagnosticsSubscriber
 		public IDisposable Subscribe(IApmAgent agentComponents)
 		{
 			var retVal = new CompositeDisposable();
-			var initializer = new DiagnosticInitializer(agentComponents.Logger, new[] { new SqlClientDiagnosticListener(agentComponents) });
+			if (PlatformDetection.IsDotNetCore)
```
gregkalapos: Are we totally sure about this? I tried to research and understand this and this seems to be correct, but is it always the case that on Full Framework all `SqlClient` versions emit through EventListener and on .NET Core through DiagnosticSource? 

Do you know anything about this?

src/Elastic.Apm.SqlClient/SqlEventListener.cs
```diff
@@ -0,0 +1,139 @@
...
+			var dataSource = Convert.ToString(payload[1]);
+			var database = Convert.ToString(payload[2]);
+			var commandText = Convert.ToString(payload[3]);
+			var start = Stopwatch.GetTimestamp();
+
+			_logger?.Trace()
+				?.Log("Process BeginExecute event. Id: {Id}. Data source: {DataSource}. Database: {Database}. CommandText: {CommandText}.", id,
+					dataSource, database, commandText);
+
+			var spanName = !string.IsNullOrWhiteSpace(commandText)
```
gregkalapos: Nice, as said [here](https://github.com/elastic/apm-agent-dotnet/pull/704#pullrequestreview-360691579) I think we'll have the `commandText` with newer versions - so over time this'll hopefully capture the CommandText more often.

src/Elastic.Apm.SqlClient/SqlEventListener.cs
```diff
@@ -0,0 +1,139 @@
...
+					Statement = !string.IsNullOrWhiteSpace(commandText)
+						? commandText.Replace(Environment.NewLine, "")
+						: string.Empty,
+					Instance = database,
+					Type = Database.TypeSql
+				};
+
+				span.Context.Destination = _apmAgent.TracerInternal.DbSpanCommon.GetDestination($"Data Source={dataSource}", false, null);
+
+				// System.Data.SQLite and Microsoft.Data.Sqlite don't spread events via EventSource, however,
```
gregkalapos: I'm not sure, but given the evensource name is `Microsoft-AdoNet-SystemData` I'd be very surprised if any other provider would trigger this code. I assume this is tied to the assembly that hosts `SqlClient` by Microsoft, therefore I think subtype mssql would be ok here. 

Of course I could be wrong - nevertheless I'd just set it here.

test/Elastic.Apm.SqlClient.Tests/SqlClientListenerTests.cs
```diff
@@ -41,10 +41,12 @@ public static IEnumerable<object[]> Connections
 			get
 			{
 				yield return new object[] { new Func<string, DbConnection>(connectionString => new SqlConnection(connectionString)) };
+#if !NETFRAMEWORK
```
gregkalapos: If we update `"Microsoft.Data.SqlClient"` to `Version="1.1.1"` I think this'll also work on Full Framework. I think there was a bug in the older version that made it not working - I was able to repro the problem that [this PR](
https://github.com/dotnet/SqlClient/pull/241) fixes, after that this worked for me on FullFramework with the `SqlEventListener`. 

Once this works, some of the `#if`s can be also updated.

test/Elastic.Apm.SqlClient.Tests/Elastic.Apm.SqlClient.Tests.csproj
```diff
@@ -1,15 +1,16 @@
...
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
     <RootNamespace>Elastic.Apm.SqlClient.Tests</RootNamespace>
     <AssemblyName>Elastic.Apm.SqlClient.Tests</AssemblyName>
+    <TargetFrameworks>net461;netcoreapp2.1</TargetFrameworks>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="FluentAssertions" Version="5.6.0" />
     <PackageReference Include="Microsoft.Data.SqlClient" Version="1.0.19269.1" />
```
gregkalapos: Let's update to `1.1.1` - that'll solve some of the things here.

test/Elastic.Apm.SqlClient.Tests/Elastic.Apm.SqlClient.Tests.csproj
```diff
@@ -1,15 +1,16 @@
...
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.7.0" />
+    <PackageReference Include="StrongNamer" Version="0.2.5" />
```
gregkalapos: Why do we need this? 🤔

src/Elastic.Apm.SqlClient/SqlEventListener.cs
```diff
@@ -0,0 +1,139 @@
...
+
+		public SqlEventListener(IApmAgent apmAgent)
+		{
+			_apmAgent = (ApmAgent)apmAgent;
+			_logger = _apmAgent.Logger.Scoped(nameof(SqlEventListener));
+		}
+
+		protected override void OnEventSourceCreated(EventSource eventSource)
+		{
+			if (eventSource != null && eventSource.Name == "Microsoft-AdoNet-SystemData")
```
gregkalapos: I saw some strange behavior around here. The `OnEventSourceCreated` method got triggered twice for me resulting with the following `Paylaod` in the `OnEventWritten` method:

```
[0] = "ERROR: Exception in Command Processing for EventSource Microsoft-AdoNet-SystemData: An instance of EventSource with Guid 6a4dfe53-eb50-5332-8473-7b7e10a94fd1 already exists."
```

So, in the `OnEventWritten` we never did anything since the payload was not something we expect there. 

My hack to make that work was to disable the eventsource and enable it again - I don't know why this happens and I don't really like this code - I'm not sure what should be the solution here. 

```
if (eventSource != null && eventSource.Name == "Microsoft-AdoNet-SystemData")
{
	if (_eventSource != null)
	{
		DisableEvents(_eventSource);
		_eventSource.Dispose();
	}

	EnableEvents(eventSource, EventLevel.Informational, (EventKeywords)1);
	_eventSource = eventSource;
}
```
(`_eventSource` is a field on `SqlEventListener`) 

One suspect is that both `System.Data` and `Microsoft.Data` triggers this.

src/Elastic.Apm.SqlClient/SqlClientDiagnosticSubscriber.cs
vhatsura: So, what I see in [`ApplicationInsights` repo](https://github.com/microsoft/ApplicationInsights-dotnet/blob/58618d7aaad5407d8fb5f6837fdc8e1096f88394/WEB/Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs#L27-L36). `EventListener` is used only in `net45` target. `DiagnosticSource` is used in both - `net45` and `netstandard` as well as for `HttpClient`.

As I understand correctly [the comment](https://github.com/microsoft/ApplicationInsights-dotnet/blob/58618d7aaad5407d8fb5f6837fdc8e1096f88394/WEB/Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs#L138)

src/Elastic.Apm.SqlClient/SqlEventListener.cs
vhatsura: The only one question is left: is it enough to use database name in case of command text is missing? Maybe we need somehow combine data source with database name to span name?

src/Elastic.Apm.SqlClient/SqlEventListener.cs
vhatsura: I agree with you and this is why I left a comment: to decide what to do it with on review.
I don't know why I didn't point to it in the PR description. BTW, set span subtype to `mssql`

test/Elastic.Apm.SqlClient.Tests/SqlClientListenerTests.cs
vhatsura: Nice catch. I'll update it

test/Elastic.Apm.SqlClient.Tests/Elastic.Apm.SqlClient.Tests.csproj
vhatsura: It's needed to solve the limitation of strong-named assembly in .Net Framework: signed assembly cannot load unsigned assemblies. .Net Core doesn't have such limitation and only raise warning during the build. This is why there wasn't a need it before. With adding `net461` target it was an issue

src/Elastic.Apm.SqlClient/SqlEventListener.cs
vhatsura: Thanks for your findings. I'll look deeper into it.

vhatsura: @gregkalapos, your assumption is right. There are 2 event sources with the same GUIDs (GUID is generated based on event source name), one from `System.Data.SqlClient` and another one from `Microsoft.Data.SqlClient`.
This is the exact place where the exception is thrown: https://github.com/microsoft/referencesource/blob/08b84d13e81cfdbd769a557b368539aac6a9cb30/mscorlib/system/diagnostics/eventing/eventsource.cs#L3029-L3033
With this place, it's understandable why disposing of previous event source help to solve the issue, however, it's quite unstable, sometimes it doesn't help.
I've created [an issue in `Microsoft.Data.SqlClient` repo](https://github.com/dotnet/SqlClient/issues/436). Will look, maybe they have a proper solution.

For now, I suggest keeping it as is with small hacking for our tests because I don't think that a lot of people use both libraries in one application (maybe I'm wrong). WDYT?

src/Elastic.Apm.SqlClient/SqlClientDiagnosticSubscriber.cs
gregkalapos: I see, thanks for the info.

Well, to me that comment in the AI code is a but cryptic :) I know that `HttpClient` was updated outside the framework with a [NuGet pacakge](https://www.nuget.org/packages/System.Net.Http/) - maybe that's what they mean by "referencing .NET Core System.Net.Http" - I suspect that package was built from the prior CoreFX repo - same for .NET Core. 

There isn't really such thing for `SqlClient` (unless you manually copy assemblies) - or actually  `Microsoft.Data.SqlClient` would be similar, but that seems to use EventListener on Full Framework - which surprised me - I though it'd use DiagnosticSource on both.

Is that maybe why Application Insights subscribe to DiagnosticSource on both?  I was unable to get DiagnosticSource out of this on Full Framework.

Nevertheless this is I'd say a minor thing. I think the current code is fine - if someone runs into a situation where we don't turn on the right stuff we can easily find that this is the root cause and we can adapt this part of the code.

What do you think? Keep it as it is?

src/Elastic.Apm.SqlClient/SqlEventListener.cs
gregkalapos: This could be relevant for searching and maybe grouping - but given that we won't have any dynamic data here I think it would not make a big difference if we add the database source. I'd keep it as it is.

src/Elastic.Apm.SqlClient/SqlEventListener.cs
```diff
@@ -0,0 +1,149 @@
...
+
+		private readonly ConcurrentDictionary<int, (Span Span, long Start)> _processingSpans = new ConcurrentDictionary<int, (Span, long)>();
+
+		public SqlEventListener(IApmAgent apmAgent)
+		{
+			_apmAgent = (ApmAgent)apmAgent;
+			_logger = _apmAgent.Logger.Scoped(nameof(SqlEventListener));
+		}
+
+		private EventSource _eventSource;
```
gregkalapos: nit: I think this is a leftover - could you please remove?

src/Elastic.Apm.SqlClient/SqlEventListener.cs
gregkalapos: >  I suggest keeping it as is with small hacking for our tests because I don't think that a lot of people use both libraries in one application

Yes, I agree that the point of this discussion is not a very "real world" scenario - I don't think either people would have both in 1 app. 

On the other hand - what's the small hack for the tests? Did you already push something for this? 🤔  Currently the tests on full framework fail to me due to this problem.

gregkalapos:
Oh, and I think we'll also need `Microsoft.NETFramework.ReferenceAssemblies` in the `Elastic.Apm.SqlClient.Tests` project in order to build on Linux. 

```
[2020-02-21T22:24:53.931Z] /usr/share/dotnet/sdk/3.1.100/Microsoft.Common.CurrentVersion.targets(1175,5): error MSB3644: The reference assemblies for .NETFramework,Version=v4.6.1 were not found. To resolve this, install the Developer Pack (SDK/Targeting Pack) for this framework version or retarget your application. You can download .NET Framework Developer Packs at https://aka.ms/msbuild/developerpacks [/var/lib/jenkins/workspace/tnet_apm-agent-dotnet-mbp_PR-704/apm-agent-dotnet/test/Elastic.Apm.SqlClient.Tests/Elastic.Apm.SqlClient.Tests.csproj]
[2020-02-21T22:24:53.931Z]     2 Warning(s)
[2020-02-21T22:24:53.931Z]     1 Error(s)
[2020-02-21T22:24:53.931Z] 
[2020-02-21T22:24:53.931Z] Time Elapsed 00:01:20.51
```

src/Elastic.Apm.SqlClient/SqlEventListener.cs
vhatsura: Yes, I haven't committed the hack yet. An idea is pretty simple. As `SqlEventSource` initializes statically the only one possible approach is to have three solutions. One with a base abstract class which will have test definitions, another one with referencing `System.Data.SqlClient` and the last one with `Microsoft.Data.SqlClient`.
I know, it's not a very good solution, but we will have it until [version `2.0.0` released with a fix for event source name](https://github.com/dotnet/SqlClient/issues/436#issuecomment-588434098)

vhatsura:
> Oh, and I think we'll also need `Microsoft.NETFramework.ReferenceAssemblies` in the `Elastic.Apm.SqlClient.Tests` project in order to build on Linux.
> 
> ```
> [2020-02-21T22:24:53.931Z] /usr/share/dotnet/sdk/3.1.100/Microsoft.Common.CurrentVersion.targets(1175,5): error MSB3644: The reference assemblies for .NETFramework,Version=v4.6.1 were not found. To resolve this, install the Developer Pack (SDK/Targeting Pack) for this framework version or retarget your application. You can download .NET Framework Developer Packs at https://aka.ms/msbuild/developerpacks [/var/lib/jenkins/workspace/tnet_apm-agent-dotnet-mbp_PR-704/apm-agent-dotnet/test/Elastic.Apm.SqlClient.Tests/Elastic.Apm.SqlClient.Tests.csproj]
> [2020-02-21T22:24:53.931Z]     2 Warning(s)
> [2020-02-21T22:24:53.931Z]     1 Error(s)
> [2020-02-21T22:24:53.931Z] 
> [2020-02-21T22:24:53.931Z] Time Elapsed 00:01:20.51
> ```

I moved reference to `Microsoft.NETFramework.ReferenceAssemblies` to root directory build properties as samples and source code already had such reference in their build props. I think it will fix all such issue 😃

src/Elastic.Apm.SqlClient/SqlEventListener.cs
vhatsura: @gregkalapos, I tried to implement solution with 3 projects but it also doesn't work. The issue is the `SqlEventSource` for `System.Data.SqlClient` is situated in the `System.Data` assembly. It means that `SqlEventSource` is always instantiated even if we don't reference `System.Data.SqlClient`.
So, my final suggestion (I believe) is to skip instrumentation for `Microsoft.Data.SqlClient` in .Net Framework. WDYT?

src/Elastic.Apm.SqlClient/SqlEventListener.cs
vhatsura: fixed

src/Elastic.Apm.SqlClient/SqlClientDiagnosticSubscriber.cs
vhatsura: > I know that HttpClient was updated outside the framework with a NuGet pacakge - maybe that's what they mean by "referencing .NET Core System.Net.Http"

I also thought so, however, when I tried to do it, there was loaded usual NetFx version of `HttpClient`.

> I though it'd use DiagnosticSource on both
 
As I understand correctly, `DiagnoticSource` is built-in in .Net Core runtime, and as for .Net Framework, it can be added only from the NuGet package. It seems it's one of the main reasons, why `DiagnosticSource` isn't widely used in .Net Framework.

> What do you think? Keep it as it is?

I suggest keeping it as is, at least for now.

src/Elastic.Apm.SqlClient/SqlEventListener.cs
gregkalapos: > So, my final suggestion (I believe) is to skip instrumentation for Microsoft.Data.SqlClient in .Net Framework. WDYT?

Ok, fine by me. 

Overall, I think having 3 projects wouldn't even be a hack, more like a legitimate solution - it's unfortunate that it does not work.

Anyways, let's skip it for Full Framework and see if someone needs it in the future.

Thanks for opening [the issue](https://github.com/dotnet/SqlClient/issues/436) in the SqClient repo.

Nice to have the response there, but having yet another name in the future (as the comment says) makes this thing super complicated - anyways this is outside our control, so we use whatever they offer...

vhatsura: > but having yet another name in the future (as the comment says) makes this thing super complicated

gregkalapos:
> > Oh, and I think we'll also need `Microsoft.NETFramework.ReferenceAssemblies` in the `Elastic.Apm.SqlClient.Tests` project in order to build on Linux.
> > ```
> > [2020-02-21T22:24:53.931Z] /usr/share/dotnet/sdk/3.1.100/Microsoft.Common.CurrentVersion.targets(1175,5): error MSB3644: The reference assemblies for .NETFramework,Version=v4.6.1 were not found. To resolve this, install the Developer Pack (SDK/Targeting Pack) for this framework version or retarget your application. You can download .NET Framework Developer Packs at https://aka.ms/msbuild/developerpacks [/var/lib/jenkins/workspace/tnet_apm-agent-dotnet-mbp_PR-704/apm-agent-dotnet/test/Elastic.Apm.SqlClient.Tests/Elastic.Apm.SqlClient.Tests.csproj]
> > [2020-02-21T22:24:53.931Z]     2 Warning(s)
> > [2020-02-21T22:24:53.931Z]     1 Error(s)
> > [2020-02-21T22:24:53.931Z] 
> > [2020-02-21T22:24:53.931Z] Time Elapsed 00:01:20.51
> > ```
> 
> I moved reference to `Microsoft.NETFramework.ReferenceAssemblies` to root directory build properties as samples and source code already had such reference in their build props. I think it will fix all such issues 😃

Cool, nice 👍 .


We have new test fails in CI:

```
Error
Expected span.Context.Destination.Address to be 
"localhost" with a length of 9, but 
"172.17.0.3" has a length of 10, differs near "172" (index 0).
 ```

Also, I'm not totally sure this project properly runs in `net461` - we should also look into that, but I don't see that as a blocker for the PR - if it needs more CI adaption we can do it later - Given that this technically should run on a Windows machine that should spawn a SQL Server Linux container - I'm not sure we can do that currently.

**Comment by vhatsura** [comment]
> Also, I'm not totally sure this project properly runs in net461

Yes, I see that only tests from `Elastic.Apm.Tests` are included to run on `net461`. However, it seems there is also an issue with that:

```
A total of 1 test files matched the specified pattern.

[2020-02-22T13:14:55.618Z] Logging TestHost Diagnostics in file: C:\Users\jenkins\workspace\tnet_apm-agent-dotnet-mbp_PR-704\apm-agent-dotnet\target\diag.host.20-02-22_13-14-54_52382_1.log

[2020-02-22T13:15:00.895Z] [xUnit.net 00:00:04.3089285] Skipping: Elastic.Apm.Tests (could not find dependent assembly 'Elastic.Apm, Version=1.3.0')

[2020-02-22T13:15:00.895Z] No test is available in C:\Users\jenkins\workspace\tnet_apm-agent-dotnet-mbp_PR-704\apm-agent-dotnet\test\Elastic.Apm.Tests\bin\Debug\net461\Elastic.Apm.Tests.dll. Make sure that test discoverer & executors are registered and platform & framework version settings are appropriate and try again.
```

> We have new test fails in CI:

I finally found the place, where I can find it. I tried to fix it but I need to restart CI from you

gregkalapos:
> > Also, I'm not totally sure this project properly runs in net461
> 
> Yes, I see that only tests from `Elastic.Apm.Tests` are included to run on `net461`. However, it seems there is also an issue with that:
> 
> ```
> A total of 1 test files matched the specified pattern.
> 
> [2020-02-22T13:14:55.618Z] Logging TestHost Diagnostics in file: C:\Users\jenkins\workspace\tnet_apm-agent-dotnet-mbp_PR-704\apm-agent-dotnet\target\diag.host.20-02-22_13-14-54_52382_1.log
> 
> [2020-02-22T13:15:00.895Z] [xUnit.net 00:00:04.3089285] Skipping: Elastic.Apm.Tests (could not find dependent assembly 'Elastic.Apm, Version=1.3.0')
> 
> [2020-02-22T13:15:00.895Z] No test is available in C:\Users\jenkins\workspace\tnet_apm-agent-dotnet-mbp_PR-704\apm-agent-dotnet\test\Elastic.Apm.Tests\bin\Debug\net461\Elastic.Apm.Tests.dll. Make sure that test discoverer & executors are registered and platform & framework version settings are appropriate and try again.
> ```
> 
> > We have new test fails in CI:
> 
> I finally found the place, where I can find it. I tried to fix it but I need to restart CI from you

I'll make sure the CI runs for you build now - I'll take a look at the remaining stuff later.

PR #707, opened 2020-01-27
Configuration value SpanFramesMinDuration not set

marcoregueira:
Simple case of mismatched parameters. Passing StackTraceLimit instead of SpanFramesMinDuration to parser.

Test and fix included.

image

### DIFF
--- a/src/Elastic.Apm/Config/EnvironmentConfigurationReader.cs
+++ b/src/Elastic.Apm/Config/EnvironmentConfigurationReader.cs
@@ -18,7 +18,7 @@ namespace Elastic.Apm.Config
 		{
 			_spanFramesMinDurationInMilliseconds
 				= new Lazy<double>(() =>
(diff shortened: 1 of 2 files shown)


marcoregueira:
Consider that if I modify the test SpanFramesMinDurationDefaultValuesInSync and remove the first line the tests still succeeds, but it shouldn't. 

image

The problem seems to be that these tests are setting variables to default values, while the variables themselves have those values from beginning. Therefore, if the code setting the value fails the test still succeeds.

gregkalapos:
> Consider that if I modify the test SpanFramesMinDurationDefaultValuesInSync and remove the first line the tests still succeeds, but it shouldn't.
> 
> image
> 
> The problem seems to be that these tests are setting variables to default values, while the variables themselves have those values from beginning. Therefore, if the code setting the value fails the test still succeeds.

I think `SpanFramesMinDurationInMilliseconds` is the more important test here, but that mocks `EnvironmentConfigurationReader.cs` 🤦‍♂

PR #711, opened 2020-01-30
Fix flaky SqlClientDiagnosticListenerTests

gregkalapos:
I saw these tests failing in CI sporadically - I also increased `Task.Delay` to 10 sec, but did not help. 

Let's change the db call to a sync call, which will cause that the event listener will be finished before the db call itself is finished.


gregkalapos:
cc @vhatsura fyi.

Issue #710, opened 2020-01-29
Agent doesn't send body to APM Server


vhatsura:
**Describe the bug**
The captured request body is reported to the APM server as empty string.

**To Reproduce**
It's reproducible in #661 
1. Run `SampleHttpSelfHostApp` project
2. Execute the following command:
```
curl --location --request POST 'http://localhost:12345/api/values/' \
--header 'Content-Type: application/json' \
--data-raw '{
	"id": 1,
	"name": "sample"
}'
```
3. To APM server the `{\n\t"id": 1,\n\t"name": "sample"\n}` body needs to be reported, however, I got the following from logs:
```
[2020-01-30 00:01:45.536 +03:00][Trace] - {PayloadSenderV2} Serialized item to send: Transaction{Id: a157529ca7a9c0d7, TraceId: aebc002e58df3557b1bd85090ae02957, ParentId: null, Name: POST /api/values/, Type: request, IsSampled: True} as {"context":{"request":{"body":"","headers":{"Postman-Token":"[REDACTED]","Cache-Control":"no-cache","Connectio
n":"keep-alive","Accept":"*/*","Accept-Encoding":"gzip, deflate, br","Host":"localhost:12345","User-Agent":"PostmanRuntime/7.22.0"},"http_version":"1.1","method":"POST","socket":{"encrypted":false},"url":{"full":"http://localhost:12345/api/values/","hostname":"localhost","pathname":"/api/values/","protocol":"http","raw":"http://localhost:12345/ap
i/values/","search":""}},"response":{"finished":true,"headers":{},"status_code":200}},"duration":653.856,"id":"a157529ca7a9c0d7","sampled":true,"name":"POST /api/values/","result":"http 2xx","span_count":{"dropped":0,"started":0},"timestamp":1580331686757135,"trace_id":"aebc002e58df3557b1bd85090ae02957","type":"request"}
```

It seems that the issue in
https://github.com/elastic/apm-agent-dotnet/blob/59b527ed12a97e5ef6eef2ed8470b4c637039ad2/src/Elastic.Apm/Report/Serialization/BodyStringSanitizerConverter.cs#L55-L79
The possible fix is to change `return true` at the end of the method to `return i != 0`, however, what if we will have smth like `{ "Data": "x=1&y=2" }` in the POST body? So, in this case, it would be nice to have an additional technical field inside `Context` class, which will point to the original `Content-Type` of the request body.

**Expected behavior**
[...]

Issue #656, opened 2019-12-13
Body sentinizer ignores everyting except string


skynet2:
**Describe the bug**
Starting from 1.2.0 "Sentinize" attribute was added to **Body**.
So now, it starts to use BodyStringSanitizerConverter.

According to logic inside that class "if not string - write null" but Request.Body is object, so value of Request.Body can be anything (as a rule, in my case Dictionary<string,object>)

https://github.com/elastic/apm-agent-dotnet/blob/ac1b55cf937e15016fd939dc85ecf8aba367efcf/src/Elastic.Apm/Report/Serialization/BodyStringSanitizerConverter.cs#L52

**Expected behavior**
Request.Body should not be limited to be string only.

cc @gregkalapos 
Thanks,
Stas

gregkalapos:
Hi Stas,

we know the current sanitizer logic is not correct - [This PR](https://github.com/elastic/apm-agent-dotnet/pull/617) would improve it, but it did not make it into 1.2. 

> Expected behavior
> Request.Body should not be limited to be string only.

There is already an issue for this: https://github.com/elastic/apm-agent-dotnet/issues/555 

The only thing to keep in mind is that having `object` on the public API can be super dangerous. If you put something into it that cannot be json serialized then there will be an exception thrown and we won't be able to send that data to the server - or we need to handle it somehow in a converter, and having `object` is basically impossible to handle.

As discussed in #555 probably `string` and `Dictionary<string, string>` would be the way to go - so, yes, we should not limit it to string, but unfortunately we won't be able to allow anything (meaning having `object` there) without having all the problems.

Now, to sanitization: if you look at #617 I think it's clear that sanitizing the body during serialization is super challenging - I think all other agents only sanitize `application/x-www-form-urlencoded` body. When we read the body we have this info, and also in ASP.NET Core, `application/x-www-form-urlencoded` body is already parsed into key value pairs, so it'd be very easy to do the sanitization when we read the data - it's hard when we try to sanitize during serialization, because we don't have the  content type.

So, the idea is that we could either flow the content type into the sanitization logic, or do the sanitization before we set the content type - the second is I think easier. This'd mean that if someone sets the body manually through the public API and we don't have the content type then realistically we won't be able to correctly sanitize the body - at least after some iteration I'd say implementing it is fairly hard and it'll definitely cause overhead.

[...]

PR #712, opened 2020-02-03
Sanitize request body during capturing instead of serialization #656

gregkalapos:
We have some issues with our request body sanitization logic - currently the code has bugs and in some cases the whole body is incorrectly "sanitized away".

Prior to this PR we sanitized the request body during serialization. The reason for doing so was that if we sanitize during serialization then it applies to everything - including body captured manually through public API, ASP.NET Core and old ASP.NET - so 1 code for all. The challenge is that with this we only had a string and we tried to parse the keys and values from a single string - very resource consuming and hard to do it correctly. I tried to fix it in #617, but that still is not 100% correct.

In the meantime errors were reported multiple times: #656, #710. 

So, after some iteration, I decided that the best is to just sanitize the body in ASP.NET Core during capturing - this is trivial, there is a nice API for doing this, in that case we have the keys and values from the form encoded body - and then once the body is set we leave it as it is, so nothing happens during serialization.

Of course it'd be nice to have 1 implementation for body sanitization that is shared across all technologies, but I feel it's not worth the very high effort to make that implementation bug free.


gregkalapos:
@bmorelli25 here we go again 😆; there is a new note in `docs/configuration.asciidoc`.

gregkalapos:
Alternatively we could store the content type and do the sanitization based on that during serialization.

Reason we I did not do that:
- with public API it's not guaranteed that it's set 
- still during serialization with that we have to re-parse the body, which is resource consuming

Doing it during capturing is significantly simpler.

docs/configuration.asciidoc
```diff
@@ -298,6 +298,8 @@ Sometimes it is necessary to sanitize, i.e., remove, sensitive data sent to Elas
 This config accepts a list of wildcard patterns of field names which should be sanitized.
 These apply for example to HTTP headers and `application/x-www-form-urlencoded` data.
 
+NOTE: In case you capture the request body manually through the public API, the agent won't sanitize the body - this setting only applies to values that are captured automatically by the agent.
```
bmorelli25: Let's bump this from a NOTE to IMPORTANT. Also, I think it makes sense to swap the sentences. What do you think of this?
```suggestion
IMPORTANT: This setting only applies to values that are captured automatically by the agent. If you capture the request body manually with the public API, this configuration doesn't apply, and the agent won't sanitize the body. 
```

gregkalapos: Agreed 👍

gregkalapos:
Failing test is due to unavailable `http://httpstat.us` - merging now.

PR #713, opened 2020-02-04
GC metrics test fixes

gregkalapos:
Some cleanup and fixing 1 IIS test.

To the IIS test:
In `MetricsAssertValid.AssertValid` we iterate through a fixed set of metrics and only made the test pass when the metric keys sent by the agent 100% matched the fixed set of metrics. 

Problem is that with GC this is flaky - GC metrics are only captured when there is a GC, so for old metrics tests that don't care about the GC itself it should not matter if GC metrics are captured or not. 

So adapted the test accordingly - old tests don't care about gc metrics and pass regardless gc metrics are captured or not.


gregkalapos:
Green 🎉

PR #724, opened 2020-02-12
[Docs] Update All options summary

gregkalapos:
Found that some configs are missing from this list.


gregkalapos:
Docs build green. 

image


I'm merging now, I'm working on the release and I think this is small enough to progress.

Of course we can reopen or revisit in case something is not ok here

bmorelli25:
Great catch! Thanks

PR #727, opened 2020-02-13
Extract action from SOAP body for version 1.2

marcoregueira:
Sorry if you were working on it already, but I needed it.

This adds support for extracting action name for SOAP 1.2 requests. Action name, in this case, needs to be extracted from the document body.

Extraction is done using a lightweight parser that doesn't need to read the entire stream.

Unit test is included, but being sincere, I have not been able to run it at second try. While ISS was running and the pool was started it complained about come configuration section I had not time to dig in; it should work, anyway, in your testing environment.  But I have tested it with a real service and it works.

Hope it helps..


src/Elastic.Apm.AspNetFullFramework/Extensions/RequestExtensions.cs
```diff
@@ -32,5 +33,49 @@ public static string ExtractSoapAction(this HttpRequest request, IApmLogger logg
...
+			if (!request.InputStream.CanSeek) return null;
+
+			try
+			{
+				var xmlReader = new XmlTextReader(request.InputStream);
+				xmlReader.WhitespaceHandling = WhitespaceHandling.None;
+				xmlReader.Read();
+				xmlReader.ReadStartElement(); // if (xmlReader.LocalName != "Body") return null;
+				xmlReader.ReadStartElement();
+				var action = xmlReader?.LocalName;
```
gregkalapos: Unfortunately I don't think this logic is enough to parse the Body. 

For example [on this page](https://www.w3.org/TR/soap12-part1/) there is a sample, which this would not parse correctly.

I'd say 1 `Read()` and 2 `ReadStartElement()` is not guaranteed to hit the `Body` part in the xml.

test/Elastic.Apm.AspNetFullFramework.Tests/DistributedTracingTests.cs
```diff
@@ -152,6 +152,42 @@ public async Task CallSoapRequest()
...
+				{
+					RequestUri = new Uri(fullUrl),
+					Method = HttpMethod.Post,
+					Content = httpContent
+				};
+
+				client.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("text/xml"));
+				request.Content.Headers.ContentType = new MediaTypeHeaderValue("application/soap+xml; charset=utf-8");
+
+				var response = client.SendAsync(request).Result;
```
gregkalapos: Let's use `await` here.

test/Elastic.Apm.AspNetFullFramework.Tests/DistributedTracingTests.cs
```diff
@@ -152,6 +152,42 @@ public async Task CallSoapRequest()
...
+			{
+				var request = new HttpRequestMessage()
+				{
+					RequestUri = new Uri(fullUrl),
+					Method = HttpMethod.Post,
+					Content = httpContent
+				};
+
+				client.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("text/xml"));
+				request.Content.Headers.ContentType = new MediaTypeHeaderValue("application/soap+xml; charset=utf-8");
```
gregkalapos: This throws an exception for me: 

```
Unhandled Exception: System.FormatException: The format of value 'application/soap+xml; charset=utf-8' is invalid.
```

I don't think the `charset=utf-8` part is needed.

```suggestion
				request.Content.Headers.ContentType = new MediaTypeHeaderValue("application/soap+xml");
```

test/Elastic.Apm.AspNetFullFramework.Tests/DistributedTracingTests.cs
```diff
@@ -152,6 +152,42 @@ public async Task CallSoapRequest()
...
 		}
 
+		[AspNetFullFrameworkFact]
+		public async Task CallSoap12Request()
+		{
+			var rootTxData = SampleAppUrlPaths.CallSoapServiceProtocolV1_2;
+			var fullUrl = Consts.SampleApp.RootUrl + "/" + rootTxData.RelativeUrlPath;
+			var action = "Ping";
+
+			var httpContent = new StringContent($@"<?xml version=""1.0"" encoding=""utf-8""?>
```
gregkalapos: We need more tests here with more sample data. [Wikipedia](https://en.wikipedia.org/wiki/SOAP) has 1 example, and I'd also add the one from [here](https://www.w3.org/TR/soap12-part1/) (section 1.4 Example SOAP Message).

marcoregueira:
Hi @gregkalapos 

My bad. I should have been more exhaustive searching for use cases.

There are not many different possibilities, but definitely more than I considered.

In short the soap message defines 

An Envelope element that identifies the XML document as a SOAP message
A Header element that contains header information
A Body element that contains call and response information
A Fault element containing errors and status information

According to W3 the header element is optional and can be skipped for our purposes.

The actual method called is the first element in the body (and should be the only).

And finally the Fault element is for response messages only.

In many cases only the body element will be available. 

Here is a typical random sample from the web. Clicking on the methods will show a soap message sample you can use. 
https://secure.logmeinrescue.com/API/API.asmx
My previous implementation was based on a sample like this.

And this is from your own sample
image

Here is the proposed algorithm, a new implementation is in place.
* Go to content, it should be Envelope. Exit if it is not.
* Skip the Header if any. 
* Get into the Body and get the first element.

There are some new tests in the code testing all variations I could think of...

For some extensions like WS security it will not give the desired result. It will give a formally correct result anyway, since WS-Security will call a generic method and pass its encrypted payload. A WCF (or so) extension would be needed to properly log the action called, not just a simple module. There is a test for this too.

Parsing the entire body is not needed. The XmlReader is lightweight and won't do that. How much reads from the stream is unknown for me. I've checked the stream when sending a huge message and it reads a bit more than needed but less that it would require to reach the end of the element. It seems to read 4kb chunks or so. According to many references it does not deserialize the content internally anyway.

Aside of that, I've removed the UTF-8 header you pointed out. It shouldn't be passed that way, but it can be passed. 

I'm sorry I'm unable to run the iis tests in my machine. I have a few issues more I've solved for myself and I would like to send you a couple of test cases if you have a moment to tell me how to run them.

marcoregueira:
Completing my previous comment, some of the tests include ill-formed or incomplete messages that demonstrate that xmlreader doesn't read the entire message. When it retrieves an element in the document if doesn't check if it is closed or if the document is still valid after that... 

I can contribute the test on the stream position but it is not quite descriptive as I would like.

gregkalapos:
Hey @marcoregueira

Thank you so much for picking this up.

Just a quick comment to try to unblock you with running the tests.

> I'm sorry I'm unable to run the iis tests in my machine. I have a few issues more I've solved for myself and I would like to send you a couple of test cases if you have a moment to tell me how to run them.

The [readme file here](https://github.com/elastic/apm-agent-dotnet/tree/master/test/Elastic.Apm.AspNetFullFramework.Tests) has an overview on how to run tests from `Elastic.Apm.AspNetFullFramework.Tests`. What I usually do: open a cmd prompt in admin mode, set env var `ELASTIC_APM_TESTS_FULL_FRAMEWORK_ENABLED` to `true` (like execute `set ELASTIC_APM_TESTS_FULL_FRAMEWORK_ENABLED=true`), then open Visual Studio from the admin cmd prompt (for this, just check the location that your VS icon points to and copy it there). After that all tests from `Elastic.Apm.AspNetFullFramework.Tests` should run from VS as any other tests.

Regarding the point from your last message
> ill-formed or incomplete messages 

I feel we don't need to over-tests this. I think it's enough to have tests that make sure that in case such an invalid soap message is passed to the application then the app behaves the same with and without the agent - so no exception bubbles up to the app from the agent, and also make sure we have at least some meaningful transaction name (obviously not based on the soap action in this case, but at least something that helps users to identify the request).

I hope this'll help - I'll come back to this and take a closer look.

Again, huge thanks for the PR! 👏

marcoregueira:
About the ill-formed messages, I feel the same. It wouldn't be necessary to test them aside of a base case, but since one of the concerns (mine too) is that the entire message should not be processed that's one way of testing that. I intentionally added the errors after the method name element. You can safely remove them once reviewed.

kodigo35:
Hello, i had the same need for my WCF services and did it using WCF  behaviors
and  IDispatchMessageInspector in an external component.

Issue #734, opened 2020-02-20
HttpDiagnosticListenerImplBase logs warning when CurrentTransaction is null


gregkalapos:
We have 2 entries on discuss ([1](https://discuss.elastic.co/t/elastic-apm-plus-ms-application-insights-spamms-logs-in-net-core-2-1-web-app/), [2](https://discuss.elastic.co/t/application-wont-load-when-using-useallelasticapm/219336) )

One is caused by a request to an external WCF service, the other one is cause by Application Insights ( Never thought we'll have a collision with AI  :trollface: 😄 ) 

I managed to repro the issue with application insights. Just by putting it into the project it repros. 

Logs:

```
dbug: Elastic.Apm[0]
      {HttpDiagnosticListenerImplBase} No current transaction, skip creating span for outgoing HTTP request
warn: Elastic.Apm[0]
      {HttpDiagnosticListenerImplBase} Failed capturing request (failed to remove from ProcessingRequests) - This Span will be skipped in case it wasn't captured before. Request: method: POST, URL: https://rt.services.visualstudio.com/QuickPulseService.svc/ping?ikey=6c76da4f-e040-480d-bf77-955aab698537
```

The debug log in itself is acceptable, but then we have a `warn`, which is not ok. 

So, the issue here is that in the `Start` part of the HTTP request we stop processing it due to `No current transaction` (that's how it should be), but then in the end part we somehow end up in the line where we'd enqueue the http request - but it's not in the queue, since we did not put that into the queue in `Start`.

gregkalapos:
Fixed in https://github.com/elastic/apm-agent-dotnet/pull/736

PR #736, opened 2020-02-20
Exiting early in HttpDiagnosticListenerImplBase in case CurrentTransaction is null

gregkalapos:
Solves #734

### DIFF
--- a/src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerImplBase.cs
+++ b/src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerImplBase.cs
@@ -98,12 +98,6 @@ namespace Elastic.Apm.DiagnosticListeners
 				return;
 			}
 
-			if (_agent.Tracer.CurrentTransaction == null)
-			{
-				Logger.Debug()?.Log("No current transaction, skip creating span for outgoing HTTP request, {url}", requestUrl);
-				return;
-			}
-
 			if (IsRequestFilteredOut(requestUrl))
 			{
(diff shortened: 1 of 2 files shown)


src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerImplBase.cs
```diff
@@ -98,6 +98,12 @@ public void OnNext(KeyValuePair<string, object> kv)
 				return;
 			}
 
+			if (_agent.Tracer.CurrentTransaction == null)
```
vhatsura: As I see, `ProcessStartEvent` already has such a check.

https://github.com/elastic/apm-agent-dotnet/blob/460b0078f3d56f677d194c5cb650293dac7ba2db/src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerImplBase.cs#L121-L126

So, it means that your change will affect Stop and Exception events. If it's possible that the current transaction is null during the HTTP start event (this is mentioned in issue), I think it can be possible that the current transaction isn't null during the start event but will be null during the stop event?

gregkalapos: Interesting point. 

> I think it can be possible that the current transaction isn't null during the start event but will be null during the stop event.

Since the current transaction is in the async local storage, this would sounds like a scenario where the `HttpClient` class has a bug - otherwise it'd be impossible to stop a transaction in the middle of an http call. Or is there any way to execute code on the given Task during a method call of the `HttpClient`?

Nevertheless this may indeed happen - technically it's definitely possible - I agree on that.

I change the code - now all the changes are in `ProcessStopEvent` - if the transaction is null, then we just print a debug log.

PR #753, opened 2020-03-06
Remove logging in CurrentExecutionSegmentsContainer

gregkalapos:
Solves https://github.com/elastic/ecs-dotnet/issues/58. 

This property is used in lots of cases - also in the serilog enricher. In that case creating a new log line within an enricher calls the enricher itself, which leads to stackoverflow.

Alternative solution would be to introduce a property that does not log.

2 options I see here:
- Add an internal property that does not log and can be called from the serilog enricher: The less good thing here is that we aimed to only use public things in the serilog enricher, since that is something that should be doable in any other logging framework - also by users without using anything internal.
- Add a public property that does not log: I feel for a verbose log that is an overkill.

Also considered adding a method - something like `SuspendLogging()`, but that again feel an overkill and makes things more complicated.

Overall I feel the simplest is to just not log there.

### DIFF
--- a/src/Elastic.Apm/AgentComponents.cs
+++ b/src/Elastic.Apm/AgentComponents.cs
@@ -44,7 +44,7 @@ namespace Elastic.Apm
 			CentralConfigFetcher = centralConfigFetcher ?? new CentralConfigFetcher(Logger, ConfigStore, Service);
 
 			TracerInternal = new Tracer(Logger, Service, PayloadSender, ConfigStore,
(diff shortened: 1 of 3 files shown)


snakefoot:
If you remove all logging, then I would completely remove inner-class `ContextLocalHolder`. Instead have the `AsyncLocal` as direct members on `CurrentExecutionSegmentsContainer`

gregkalapos:
> If you remove all logging, then I would completely remove inner-class `ContextLocalHolder`. Instead have the `AsyncLocal` as direct members on `CurrentExecutionSegmentsContainer`

Agreed - removed those.

PR #752, opened 2020-03-05
WIP ci(jenkins): fix some issues

v1v:
### What

- Use https://github.com/spekt/junit.testlogger rather than `nunit`, therefore the JUnit reporting supports multi-platform reports.
- Bump test versions for the codecov running as required, see https://github.com/tonerdo/coverlet#requirements

### Why

Let's put back the CI for the apm-agent-dotnet green and stable

### Historical changes

- This will help to report the test output for all the stages in Windows and Linux.
- Use https://github.com/spekt/junit.testlogger rather than `nunit`, therefore the JUnit reporting supports multi-platform reports.
- Bump test versions for the codecov running as required, see https://github.com/tonerdo/coverlet#requirements
- Fail if any failures in the test execution in Linux.
- Filter target platforms for Linux.

PR #763, opened 2020-03-12
ci(jenkins): enable test reporting for all the test stages

v1v:
This is a subset of https://github.com/elastic/apm-agent-dotnet/pull/752 

I raised a new PR to decouple what changes are required to be done for enabling the test reporting for all the stages versus the ones that will care of fixing some warning/errors that are already happening in some scripts in the CI.


v1v:
> In this PR we still have the error when we try adding XunitXml.TestLogger, I assume that'll be done in #752.

Yep, I split the actions I took in #752 in small PRs to help with the iterations

PR #767, opened 2020-03-13
ci(jenkins): use JUnit testlogger to support multi-platform projects

v1v:
### What

- Use https://github.com/spekt/junit.testlogger rather than `nunit`, therefore the JUnit reporting supports multi-platform reports.
- Report Test IIS reports too
- JUnit reporting will allow showing all the executed tests for all the target platforms.

### Why

Simplify the CI Pipeline and being able to see all the tests.

### Details

There are more tests reported based on the number fo target-frameworks and OS:

#### Same OS with different target-frameworks

<details><summary>Expand to view screenshot</summary>
<p>

image

</p>
</details>

There is not a way to `easily` filter what of those tests refer to a specific framework.

#### Same tests for different OS versions

The test output files (xml format) do contain the target-plaform:

<details><summary>Expand to view screenshot</summary>
<p>

image

</p>
</details>

#### IIS reporting

<details><summary>Expand to view screenshot</summary>
<p>

image

</p>
</details>

### Questions

What are your thoughts?


v1v:
I'll then merge it now and let's see how we can enable the target framework to be shown in the test results in another iteration :)

PR #766, opened 2020-03-13
ci(jenkins): run net461 in the CI

v1v:
### Actions
- https://github.com/elastic/apm-agent-dotnet/pull/766/commits/608a2e2994ff6b5b2ce0721b5a997fa8a6eb64e6 did not work
- https://github.com/elastic/apm-agent-dotnet/pull/766/commits/1116bcea66a4e09353bb417b22632747e29aa037 did not work

PR #768, opened 2020-03-16
ci(jenkins): run net461 in the CI - another try

gregkalapos:
Separating building and test run for `net461`. 

With `dotnet publish --framework net461` we publish the test project to `net461` and then with `dotnet vstest Elastic.Apm.Tests.dll` we run the tests. This seems working in CI.

The `dotnet vstest` command will run on whatever .NET Framework version is installed. So we won't actually run on 4.6.1, we run on a version that is at least 4.6.1 (or newer), but I'd say this is fine, this whole thing is about making sure that we run these tests on at least 1 full framework version. This is due to the fact that .NET Framework only has in-place updates, you can basically only have a single version installed on a machine.

I still don't really know why it did not work with `dotnet test -f net461`. Maybe because it'll try to only run it on 4.6.1 and newer version is there, which is not ok for the test runner... no idea, this is just a theory 🤷‍♂ 

~Install .NET Framework 4.6.1~ Update: this is not needed

I also killed code coverage with this PR - as it seems `vstest` does not pick up the codecov settings.

### DIFF
--- a/.ci/docker/sdk-linux/README.md
+++ b/.ci/docker/sdk-linux/README.md
@@ -21,6 +21,7 @@ docker run --rm -ti \
 docker run --rm -ti \
        --name sdk \
        -v $(pwd):/src \
+       -v /var/run/docker.sock:/var/run/docker.sock \
        -w /src \
        sdk:latest \
        /bin/bash -c '.ci/linux/test.sh'
--- a/.ci/linux/test.sh
+++ b/.ci/linux/test.sh
@@ -11,29 +11,45 @@ dotnet sln remove src/Elastic.Apm.AspNetFullFramework/Elastic.Apm.AspNetFullFram
 dotnet sln remove test/Elastic.Apm.AspNetFullFramework.Tests/Elastic.Apm.AspNetFullFramework.Tests.csproj
 
 # Configure the projects for coverage and testing
-for i in $(find . -name '*.csproj')
+while IFS= read -r -d '' file
 do
-	if [[ $i == *"AspNetFullFrameworkSampleApp.csproj"* ]]; then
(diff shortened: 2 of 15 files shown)


gregkalapos:
Accidentally pushed #766 :s 

So, `choco install dotnet4.6.1` says:

```
[2020-03-16T11:50:08.085Z] Installing the following packages:
[2020-03-16T11:50:08.085Z] dotnet4.6.1
[2020-03-16T11:50:08.085Z] By installing you accept licenses for the packages.
[2020-03-16T11:50:08.085Z] DotNet4.6.1 v4.6.01055.20170308 already installed.
[2020-03-16T11:50:08.085Z]  Use --force to reinstall, specify a version to install, or try upgrade.
[2020-03-16T11:50:08.085Z] 
[2020-03-16T11:50:08.085Z] Chocolatey installed 0/1 packages. 
```

But I don't believe it, so I'll try it also with `--force`.

gregkalapos:
Also testing this package: https://chocolatey.org/packages/netfx-4.6.2-devpack

The other one was I guess only the runtime, which is not enough.

gregkalapos:
@v1v could you take a look - this seems to make the test run in CI.

We can do this also over in #766... whichever is easier.

v1v:
I just merged with the upstream and added the `test` folder in the `LogFilePath` to be able to report the executed tests for net461.

Regarding the coverage, there is no coverage report at the moment for the windows, but only the linux stage, if required we can work on supporting the other OS in a follow-up.

gregkalapos:
> Regarding the coverage, there is no coverage report at the moment for the windows, but only the linux stage, if required we can work on supporting the other OS in a follow-up.

I think this is fine now - we can reiterate this later.

PR #769, opened 2020-03-17
Update release process - only release fix set of packages 

gregkalapos:
I'd like to adapt our deploy script.

## Why? 
Currently the `deploy.sh` script iterates through all `*.nupkg` files and pushes them. The problem with this is that we can have packages there that we don't want to release yet. Currently we have the `Elastic.Apm.SqlClient` package, which is already on `master`, but we don't want to create an official release package yet (it's not ready for that) - with the current script we'd push this package and release it. I assume this situation is not special, we'll have similar packages later.

## Alternatives
I know this is not so good from the automation point of view, but given this part really pushes official packages to our users, I feel it's better to be very careful here.

Worst case that can happen with this is that we forget to update the `deploy.sh` file when we release a new package, in that case we can still push it later, or rerun the script. The other way around (pushing something we don't want) is worse I feel.

### DIFF
--- a/.ci/linux/deploy.sh
+++ b/.ci/linux/deploy.sh
@@ -4,8 +4,9 @@
 #
 set -euo pipefail
 
-for nupkg in $(find . -type f -not -path './.nuget/*' -name '*.nupkg')
+for nupkg in Elastic.Apm Elastic.Apm.AspNetCore Elastic.Apm.EntityFrameworkCore Elastic.Apm.NetCoreAll Elastic.Apm.EntityFramework6 Elastic.Apm.AspNetFullFramework
 do
+	nupkg+=".nupkg"
 	echo "dotnet nuget push ${nupkg}"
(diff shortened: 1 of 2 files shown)


vhatsura:
I think we can have the alternative approach to mark projects as not packable (IsPackable=false). In such case we don’t need to adjust deploy.sh script.

gregkalapos:
@vhatsura that's a good idea - but still we can forget to add `IsPackable=false` - if a package ends up on master without it, we'll still push that.

kuisathaverat:
I guess it needs the full path so it does not find the packages

```
[2020-03-17T22:33:16.697Z] + .ci/linux/deploy.sh ******** ********
[2020-03-17T22:33:16.697Z] dotnet nuget push Elastic.Apm.nupkg
[2020-03-17T22:33:17.264Z] error: File does not exist (Elastic.Apm.nupkg).
```

PR #771, opened 2020-03-17
Add ITransaction.EnsureParentId() method - RUM correlation with dynamic HTML pages

gregkalapos:
Mostly triggered by the fact that in ASP.NET Core transactions from the RUM agent are not correlated to transactions created by the .NET agent on page load - found by @alex-fedotyev 

Turned out the `EnsureParentId()` was missing in the .NET Agent which is needed for this.

Also adding docs - mostly stolen from Ruby.

Page load with RUM + .NET with this PR and the setup described in the docs:
image

### DIFF
--- a/docs/public-api.asciidoc
+++ b/docs/public-api.asciidoc
@@ -305,6 +305,33 @@ NOTE: Return value of <<convenient-capture-span>> method overloads that accept T
 NOTE: Code samples above use `Elastic.Apm.Agent.Tracer.CurrentTransaction`. In production code you should make sure the `CurrentTransaction` is not `null`.
 
 [float]
+[[api-transaction-ensure-parent-id]]
+==== `EnsureParentId`
+
+If the transaction does not have a ParentId yet, calling this method generates a new ID, sets it as the ParentId of this transaction, and returns it as a `string`.
+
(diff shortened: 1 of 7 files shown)


docs/public-api.asciidoc
```diff
@@ -304,6 +304,33 @@ NOTE: Return value of <<convenient-capture-span>> method overloads that accept T
 
 NOTE: Code samples above use `Elastic.Apm.Agent.Tracer.CurrentTransaction`. In production code you should make sure the `CurrentTransaction` is not `null`.
 
+[float]
+[[api-transaction-ensure-parent-id]]
+==== `EnsureParentId`
```
gregkalapos: Do we want to add the version where we introduce this? We do that for configs, not sure if it's needed here.

bmorelli25: If we're going to introduce version tags to the API documentation, we should do it for all methods, not just this one. It's probably better to do that in a separate PR if so.

gregkalapos:
> Doc changes look good on my end. I opened [elastic/apm-agent-rum-js#710](https://github.com/elastic/apm-agent-rum-js/pull/710) to link to this setting like I've done with the other agents.

Hahh, yeah - nice catch - I wanted to have that link...

Issue #775, opened 2020-03-19
Stop using extension methods


ffMathy:
It is sooo annoying.

You're using extension methods on the `object` type? That's super bad practise in a library for internally-used methods. At least make them `internal` and not `public`.

I'm referring to the class `ObjectExtensions`.

ffMathy:
Thanks for the quick response to this <3

PR #777, opened 2020-03-19
Making some helper classes internal

gregkalapos:
Addressing #775

These classes were not meant to be public.


ffMathy:
Thanks man!

PR #778, opened 2020-03-19
Prepare Release 1.4.0

gregkalapos:
Preparing the new release - `1.4.0`.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -21,6 +21,24 @@ endif::[]
 [[release-notes-1.x]]
 === .NET Agent version 1.x
 
+[[release-notes-1.4.0]]
+==== 1.4.0
+
+[float]
+===== Features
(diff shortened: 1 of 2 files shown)


gregkalapos:
Ahh... Integration test fails, because the package version is already increased, so the integration test want to fetch that from nuget.org, but we have not yet pushed the new version, so it's not there. This is a 🐔 🥚  problem. 

```

[2020-03-19T13:01:11.326Z] error: Unable to find package Elastic.Apm.NetCoreAll with version (>= 1.4.0)
[2020-03-19T13:01:11.326Z] error:   - Found 8 version(s) in nuget.org [ Nearest version: 1.3.1 ]
[2020-03-19T13:01:11.326Z] error:   - Found 0 version(s) in local
```

kuisathaverat:
chicken egg problem 😄 I think does not have a good solution 1.4.0 release doesn't exist at this point

kuisathaverat:
ops!! moving the stage could make the trick

gregkalapos:
The integration test issue is fixed in https://github.com/elastic/apm-integration-testing/pull/793. 

Once that's merged I'll re-trigger this.

v1v:
Ready to be released now :)

PR #779, opened 2020-03-19
Revert "ci(jenkins): apm-integration-testing stage earlier"

v1v:
Reverts elastic/apm-agent-dotnet#770

Reason:
- https://github.com/elastic/apm-agent-dotnet/pull/778#issuecomment-601170586

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -116,120 +116,121 @@ pipeline {
                       archiveArtifacts(allowEmptyArchive: true,
                         artifacts: "${MSBUILDDEBUGPATH}/**/MSBuild_*.failure.txt")
                     }
-                  }
-                }
-              }
-            }
-            stage('Windows .NET Framework'){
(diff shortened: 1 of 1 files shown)


gregkalapos:
@v1v I think this won't fix the problem. 

I know I mentioned on slack that we changed something around this earlier, but #770 is not the one I referred to (will try to find the one, that was an earlier change). 

The problem here is that the Integration Test wants to already fetch the newest version (`1.4.0` in this case) from nuget.org, but that'll only be available once we push that to nuget.org, which'd happen after I merge #778, create a tag and let CI execute `dotnet nuget push` for all pacakges.

So I think we should adapt the Integration Test and make it build this repo and not try to install the agent from nuget.

v1v:
After some further analysis, the issue was related to some changes in the sdk requirements for the apm-its dotnet app:
- https://github.com/elastic/apm-integration-testing/pull/793

The above PR should fix this issue

PR #787, opened 2020-03-23
[Doc] ApplicationNamespaces and ExcludedNamespaces

gregkalapos:
Follow-up from #740

- Adding docs
- Moving tests to a separate file
- Minor tweak: handling empty string in comma separated lists (also added test covering this)

Also happy to take opinions on the naming of these two settings - we still can change those. Some discussion happened here: https://github.com/elastic/apm-agent-dotnet/pull/740#issuecomment-594202641


docs/configuration.asciidoc
```diff
@@ -674,6 +674,48 @@ When this setting is `true`, the agent will also add the header `elasticapm-trac
...
+| `ELASTIC_APM_APPLICATION_NAMESPACES` | `ElasticApm:ApplicationNamespaces`
+|============
+
+[float]
+[[config-excluded-namespaces]]
+==== `ExcludedNamespaces (added[1.5])`
+
+A list of namespaces to exclude when reading an exception's StackTrace to determine the culprit.
+
+Namespaces are checked with `string.StartsWith()` so "System." matches all System namespaces
```
```suggestion
Namespaces are checked with `string.StartsWith()`, so "System." matches all System namespaces.
```

docs/configuration.asciidoc
```diff
@@ -674,6 +674,48 @@ When this setting is `true`, the agent will also add the header `elasticapm-trac
 [[config-stacktrace]]
 === Stacktrace configuration options
 [float]
+
+[[config-application-namespaces]]
+==== `ApplicationNamespaces (added[1.5])`
+
+Used to determine whether a stack trace frame is an in-app frame or a library frame. When defined, all namespaces that do not start with one of the values of this collection are ignored when determining error culprit.
+
+Multiple namespaces can be configured as a comma separated list. E.g.: `"MyAppA, MyAppB"`.
```
bmorelli25: "e.g." is acceptable, but when it's the only part of the sentence, "for example" should be preferred.

```suggestion
Multiple namespaces can be configured as a comma separated list. For example: `"MyAppA, MyAppB"`.
```

Issue #758, opened 2020-03-10
Make stack trace public on span


BellBat:
**Is your feature request related to a problem? Please describe.**
Implementing elastic apm tracing on internal framework. We use an event pattern and this ends up adding framework frames to the trace. I would like to trim them prior to sending them on. Plan to use a decorator on the payload sender.

**Describe the solution you'd like**
Expose a way to manipuate the stacktrace captured on the span.

**Describe alternatives you've considered**
I have not found / thought any others.

**Additional context**

gregkalapos:
We can definitely make `StackTrace` public. The main reason it's not public yet is that there was some change in the API between the APM Server and the agent. This is related: https://github.com/elastic/apm-agent-dotnet/issues/677 

But I'm not sore if that'd be enough here - in order change the stacktrace something like a filter that other agents have would be useful.

gregkalapos:
This is implemented now. In #847 I made stack trace public - this will be released in the next version. Additionally in [Version 1.5](https://github.com/elastic/apm-agent-dotnet/releases/tag/1.5.0) we already shipped filters (more details on it in the [docs](https://www.elastic.co/guide/en/apm/agent/dotnet/current/public-api.html#filter-api)).

So in the next release in combination with filters this will be possible.

Issue #633, opened 2019-12-02
Ignore error reporting


ffMathy:
Can we control that APM should not report errors, or at least control what errors are reported?

I already have error logging in place directly to Elasticsearch, and it is properly configured to filter out specific errors that are out of my control (by other libraries etc).

APM picks up all these errors, which spams my cluster, and makes it unavailable.

gregkalapos:
Hi @ffMathy 

Unfortunately there is nothing in the APM Agent that'd give you this functionality at the moment. Let's keep this issue open and see what we can come up with.

gregkalapos:
#792 added an API to register filters - with that you'll be able to drop all errors that the agent'd send. Will be out with the next agent release. Docs will follow soon, too.

Issue #418, opened 2019-07-30
filter resource file request


hiwjcn:
> drop those

![](https://github.com/hiwjcn/file-store/raw/master/TIM%E6%88%AA%E5%9B%BE20190730160502.png)

gregkalapos:
You don't want the agent to capture those, right?

Some options we can do here:
- Add a config to disable capturing those (e.g. based on extension)
- Not capturing those et all (probably not the way to go)
- Do something on the UI/APM Server 

I'm also asking on other channels what other agents do.

gregkalapos:
Here is what the Java agent does: https://www.elastic.co/guide/en/apm/agent/java/current/config-http.html#config-ignore-urls

SergeyKleyman:
@hiwjcn Could you add a little bit more about your use case? I have a feeling that you might be using `ElasticApmModule` from `Elastic.Apm.AspNetFullFramework` - is it correct? If so please take a look at https://github.com/elastic/apm-agent-dotnet/issues/346 - by adding [`managedHandler` to modules `precondition`](https://docs.microsoft.com/en-us/iis/configuration/system.webserver/modules/add) monitoring can be limited to dynamic pages (URLs handled by ASP.NET) excluding static content.

hiwjcn:
@SergeyKleyman hey bro,im using asp.net core

``` c#
Elastic.Apm.All, Version=1.0.0.0, Culture=neutral, PublicKeyToken=ae7400d2c189cf22

app.UseElasticApm(config);
```
and configuration like this:
``` json
  "ElasticApm": {
    "LogLevel": "Debug",
    "ServerUrls": "http://3h.elk.hiwj.cn:8200",
    "TransactionSampleRate": 1.0
  }
```

hiwjcn:
@SergeyKleyman @gregkalapos is there any api like this:
``` c#
app.UseElasticApm(config,option=>
{
option.filter.add(url=>!url.endwith(".css"));
option.filter.add(url=>!url.endwith(".js"));
option.filter.add(url=>regex.match(url));
});
```

gregkalapos:
> @SergeyKleyman @gregkalapos is there any api like this:
> 
> ```cs
> app.UseElasticApm(config,option=>
> {
> option.filter.add(url=>!url.endwith(".css"));
> option.filter.add(url=>!url.endwith(".js"));
> option.filter.add(url=>regex.match(url));
> });
> ```

[...]

Issue #361, opened 2019-07-05
Agent's public API should allow user to register filters similar to those of node.js agent


SergeyKleyman:
[Documentation for node.js agent filters feature](https://www.elastic.co/guide/en/apm/agent/nodejs/current/agent-api.html#apm-add-filter)

We should consider invoking registered filter(s) earlier than "just before sending to APM server" and with more context the just event object that we captured (span, transaction, etc.). For example in order to support solution proposed at https://github.com/elastic/apm-agent-dotnet/issues/360#issuecomment-508797877 filter will have to receive both DB span instance (to which `Context` the filter will add custom tag/label) but also filter needs to receive instance of `Microsoft.EntityFrameworkCore.Diagnostics.CommandExecutedEventData` because the filter needs the value of `CommandExecutedEventData.Result` property to set the value for the above mentioned custom tag/label.

gregkalapos:
Done in #792.

PR #792, opened 2020-03-26
Add filters API

gregkalapos:
Introduce an API to register filters https://github.com/elastic/apm-agent-dotnet/issues/361

Similar to what the [Node.js agent has](https://www.elastic.co/guide/en/apm/agent/nodejs/current/agent-api.html#apm-add-filter)

## Use case

Users can register filters and these filters will be called each time before an event (transaction, span, or error) is sent to APM Server. This allows users to manipulate data before it leaves the process - e.g. sensitive user related data can be removed. Furthermore complete events could be also dropped and not be sent to the sever.

This would already help with some of the issues here in the repo: https://github.com/elastic/apm-agent-dotnet/issues/633, https://github.com/elastic/apm-agent-dotnet/issues/418, https://github.com/elastic/apm-agent-dotnet/issues/758 (partly).

## API (updated)
_Could change, this is the current idea_

`Agent.AddFilter(filter)` registers a filter - a filter can be:
- `Func<ITransaction, ITransaction>`
- `Func<ISpan, ISpan>`
- `Func<IError, IError>`

So it's a simple callback that receives either an `ITransaction`, or an `ISpan`, or an `IError` as input parameter and returns the event itself or `null`. The input parameter is always the instance that the agent captures and prepares to send to the server. These instances can be changed in the filter. If the filter returns an instance (in other words something that is not `null`), the event will be sent, if it returns `null`, the event won't be sent (it'll be dropped).

### Samples

Drop all spans for a specific database:

```
Agent.AddFilter((ISpan span) =>
{
	if (span.Context?.Db?.Instance == "VerySecretDb")
		return null;
	return span;
});
```

Hide some data:
```
Agent.AddFilter((ITransaction transaction) =>
{
	transaction.Context.Request.Url.Protocol = "[HIDDEN]";
	return transaction;
});
```

## Open issues/questions

- [x] Maybe similarly to Node.Js we could just use the input parameter as the return type - so `Func<ITransaction, ITransaction>` instead of `Func<ITransaction, bool>` (I personally would not do that for C#) - This was done, the API is updated to match other agents.
- [ ] What do we do if the change causes inconsistency in data? - some things (like Ids) are read-only on the public API, so that's not an issue, but still dropping spans could change metrics and we also deduct some data (like destination).  - One option here is to just document it and still let users do it - I personally don't think we have to "protect" users here that much. Update: asked regarding other agents, from what it seems, this is something we just need to document.

### DIFF
--- a/sample/ApiSamples/Program.cs
+++ b/sample/ApiSamples/Program.cs
@@ -224,25 +224,25 @@ namespace ApiSamples
 		{
 			Agent.AddFilter((ITransaction transaction) => {
 				transaction.Name = "NewTransactionName";
-				return true;
+				return transaction;
 			});
 
 			Agent.AddFilter((ITransaction transaction) =>
 			{
 				transaction.Type = "NewSpanName";
-				return true;
+				return transaction;
 			});
 
 			Agent.AddFilter((ISpan span) =>
 			{
 				span.Name = "NewSpanName";
-				return true;
+				return span;
 			});
 
 			Agent.AddFilter((IError error) =>
 			{
 				Console.WriteLine($"Error id printed in a filter: {error.Id}");
-				return true;
+				return error;
 			});
 		}
 	}
--- a/sample/ApiSamples/Program.cs
+++ b/sample/ApiSamples/Program.cs
@@ -167,7 +167,7 @@ namespace ApiSamples
 				Thread.Sleep(1100);
 
 				WriteLineToConsole("Waiting for callee process to exit...");
-				calleeProcess.WaitForExit();
+				calleeProcess?.WaitForExit();
 				WriteLineToConsole("Callee process exited");
 			}
 			finally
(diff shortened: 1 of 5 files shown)


SergeyKleyman:
Discarding transactions/spans when other events in the trace reference their IDs might also have unexpected effect on UI trying to render those traces because all the links based on discarded IDs will be broken.

One possible approach to avoid inconsistency is to introduce two stages for filters: one right after an event is created and another stage for filters is after an event has ended. Only filters for "after an event is created" stage will be allowed to discard the even while filters for "after an event has ended" stage will be allowed to modify the event but not discard it.

gregkalapos:
> Discarding transactions/spans when other events in the trace reference their IDs might also have unexpected effect on UI trying to render those traces because all the links based on discarded IDs will be broken.

Correct - at least we need to document this. Looked at other agents' docs: I think none of them mentions this problem - nevertheless I think at least we need to document this.

> One possible approach to avoid inconsistency is to introduce two stages for filters: one right after an event is created and another stage for filters is after an event has ended. Only filters for "after an event is created" stage will be allowed to discard the event while filters for "after an event has ended" stage will be allowed to modify the event but not to discard it.

If possible I'd like to avoid this - having 2 types of filters would cause more questions - if possible I'd go with the same approach as other agents.

Issue #563, opened 2019-10-17
dotnet format creates different line endings on Windows vs Linux (and macOS)


gregkalapos:
More details on the original problem: https://github.com/elastic/apm-agent-dotnet/pull/560#issuecomment-543239630 

Turned out that we had some files with windows style line ending (`CRLF`) and some with linux/macOS style ending (`LF`). My theory is that this triggered the bug above described in the comment which is that `dotnet format` on macOS and Linux mixed line endings within a single file - so prior to `dotnet format` we had different files with different line endings, now we have different line endings within a single file.

After changing all line endings to `LF` in the whole repo the `dotnet format` bug went away. In #566 everything is `LF` and running `dotnet format` on macOS, Linux, and Windows does not reformat anything and our code style check command also passes.

TsakiDev:
I'm having this bug too. I just added this to my CI (Drone) and experienced the same results.

david-sevcik:
It behaves similarly on Windows in files with `LF` line endings, when the file has `CRLF` it's without a problem, but when the file has `LF` then often after running `dotnet format` it gets the mixed line endings. Not every file, but it's consistent, when some file has this issue, it gets the mixed line endings every time I run `dotnet format`.

PR #799, opened 2020-03-30
ApmLoggerToNLog - Sample code optimized for NLog

snakefoot:
Ensure implementation of `IsEnabled` is working. Do not format log-message unless needed.

### DIFF
--- a/sample/AspNetFullFrameworkSampleApp/App_Start/ApmLoggerToNLog.cs
+++ b/sample/AspNetFullFrameworkSampleApp/App_Start/ApmLoggerToNLog.cs
@@ -8,17 +8,33 @@ namespace AspNetFullFrameworkSampleApp
 {
 	internal class ApmLoggerToNLog : IApmLogger
 	{
-		public bool IsEnabled(ApmLogLevel level) => true;
+		private readonly Lazy<Logger> _logger;
+
+		public bool IsEnabled(ApmLogLevel level) => _logger.Value.IsEnabled(ConvertLevel(level));
+
+		public ApmLoggerToNLog()
+			:this(string.Empty)
+		{
+
+		}
+
+		public ApmLoggerToNLog(string loggerName)
+		{
+			_logger = new Lazy<Logger>(() => LogManager.GetLogger(loggerName ?? string.Empty), true);
+		}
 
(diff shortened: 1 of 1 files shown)


snakefoot:
Cannot figure out how your sign-process works. Have signed it 3 times, and no change.

Please choose another sign-process-service, or improve the guidance.

sample/AspNetFullFrameworkSampleApp/App_Start/ApmLoggerToNLog.cs
```diff
@@ -8,17 +8,33 @@ namespace AspNetFullFrameworkSampleApp
...
+		public ApmLoggerToNLog()
+			:this(string.Empty)
+		{
+
+		}
+
+		public ApmLoggerToNLog(string loggerName)
+		{
+			_logger = new Lazy<Logger>(() => LogManager.GetLogger(loggerName ?? string.Empty), true);
+		}
```
gregkalapos: nit: [we use](https://github.com/elastic/apm-agent-dotnet/blob/master/.editorconfig#L70) expression bodied .ctor for 1 liners

```suggestion
		public ApmLoggerToNLog(string loggerName) => _logger = new Lazy<Logger>(() => LogManager.GetLogger(loggerName ?? string.Empty), true);
```

snakefoot: Looks more convoluted , but your wish, my command.

snakefoot:
Looks like the suggestion-commit changed all newline-characters. Not sure why.

gregkalapos:
> Looks like the suggestion-commit changed all newline-characters. Not sure why.

Yeah, we struggle with "CRLF vs LF" hell. https://github.com/elastic/apm-agent-dotnet/issues/563 

I suspect the file was CRLF (because it was created on Windows) but now GitHub made it LF.

snakefoot:
I can redo the commit suggestion without newlines change. If you want?

gregkalapos:
> I can redo the commit suggestion without newlines change. If you want?

That'd be awesome. I looked at it and actually it's the other way - it seems it was LF (that's what we want) and became CRLF (which we don't like) - so if it was LF I'd prefer it to remain all LF. 

#563 should take care of this long term. Sorry about the trouble.

snakefoot:
Forced commit executed

Issue #691, opened 2020-01-16
APM does not seem to use Asp.Net 3.0+ Tracing Context (W3C Trace Context/Activity)?


ckarcz:
In Asp.Net Core 3.0, they adopted the [W3C Trace Context standard](https://www.w3.org/TR/trace-context-1/)

https://devblogs.microsoft.com/aspnet/improvements-in-net-core-3-0-for-troubleshooting-and-monitoring-distributed-apps/

The AspNetCore APM agent does not seem to be using the same tracing context as the framework, which I believe is ```Activity.Current```.

The trace.id logged to ES is not the same TraceId that the framework uses (and logs in a scope).

This means we cannot view logs (in Kibana) of a trace you are viewing in Kibana APM app.

I'm not sure if something is not set up correctly, or the APM agent is not using the same tracing context as the framework.

😿

gregkalapos:
> I'm not sure if something is not set up correctly, or the APM agent is not using the same tracing context as the framework.

I assume your setup correct, we create our own trace id and not use the one that is created by `Activity`. 

The main reason is that when we started (a year ago) the `Acitivty` id wasn't logged anywhere, and I believe that the feature that ASP.NET Core creates a trace id and `HttpClient` puts that into the `traceparent` header was also introduced in 3.0 - plus we also support Full Framework.

Nevertheless I agree that we should think about reusing the id from `Activity`, if there is one. So I leave this issue here, definitely something worth investigating.

Regarding this:

> This means we cannot view logs (in Kibana) of a trace you are viewing in Kibana APM app.

Look at our almost brand new log correlation story [here](https://www.elastic.co/guide/en/apm/agent/dotnet/master/log-correlation.html). That's something we continuously work on and plan to add support for more scenarios.

ckarcz:
thanks for the prompt reply @gregkalapos. that is great to hear. and thanks for linking to that enricher package, as I did not know about it and it should solve my problem of correlating the APM and application logs. 👍

gregkalapos:
We had some progress on this and I feel everything which is possible is already covered. 

- W3C TraceContext support is already released - this means the Elastic agent uses the header format with the same header name as defined in the standard - this is already out
- #800 Makes sure that the trace id of Activity is the same as the trace id of the transactions/spans captured by the Elastic agent - this will be released with the next release.

[...]

Issue #303, opened 2019-06-19
Support one or more of OpenTracing, OpenCensus, OpenTelemetry technologies


SergeyKleyman:
The request to support OpenTracing was mentioned and now after OpenTracing and OpenCensus merged into OpenTelemetry maybe we should shoot directly OpenTelemetry?

It's indirectly related to #229.

gregkalapos:
I think we should go with OpenTelemetry. OpenTelemetry plans to be backwards compatible with OpenCensus and OpenTracing (it'd be some bridge or something like that, I don't know the details, but this is definitely on their roadmap). Also as it seems in case we'd start working on either OpenCensus or OpenTracing, those may be already deprecated when we release.

So with the future of those 3 projects plus the plan OpenTelemetry offering support for the 2 others ones I'd vote for OpenTelemetry.

SergeyKleyman:
To expand a little bit on 

> It's indirectly related to #229

It seems that the main motivation for the users to use one of OpenTracing, OpenCensus, OpenTelemetry API instead of Elastic .NET agent public API directly is to keep application code APM vendor neutral. `DiagnosticSource` API achieves this goal as well and from our experience it seems better adopted by .NET ecosystem than any of OpenTracing, OpenCensus, OpenTelemetry because `DiagnosticSource` is backed by Microsoft. So it is reasonable to conclude that starting with #229 `DiagnosticSource` will reach more users and allow them to keep application code APM vendor neutral.

yhvicey:
https://medium.com/opentracing/a-roadmap-to-convergence-b074e5815289 looks like OpenTracing and OpenCensus plans to merge together.

justinaspeciura:
I understand you want to maximize reach by waiting more broad APIs but OpenTracing is de-facto standard at this moment and no-one knows how long it will take to for them to merge or provide any packages/APIs.

You already provide OpenTracing bridges in most [other major languages](https://www.elastic.co/guide/en/apm/get-started/current/opentracing.html), why not C# too?

gregkalapos:
> You already provide OpenTracing bridges in most other major languages, why not C# too?

This has some practical reasons - .NET is the newest agent, and by the time we started with .NET the OpenTelemetry project was started and we already knew that OpenTracing won't be continued.

megakid:
I can understand your reasoning but I think it's a mistake to not put the effort into building an OpenTracing bridge for .NET.  

[...]

PR #800, opened 2020-03-31
Reuse TraceId from Activity.Current for new transactions 

gregkalapos:
Addressing #691. 

[`System.Diagnostics.Activity`](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.activity?view=netcore-3.1) implements W3C TraceContext. Once an activity is started, it generates an id, and if it uses `ActivityIdFormat.W3C` it also generates a `TraceId`.

Even better ASP.NET Core itself also creates an Activity for incoming calls which means it creates a TraceId. Prior to this PR we ignored the generated TraceId and the Elastic APM Agent created its own TraceId - or used the one passed in the `traceparent` header.

Now instead of generating TraceId for every transaction the agent will just use `Activity.Current.TraceId` if it exists and it's in `ActivityIdFormat.W3C`.

## Why we do this? 
`Activity` starts getting promoted a lot by Microsoft. E.g. the default ASP.NET Core template generates a controller method like this:

```
public IActionResult Error()
{
	return View(new ErrorViewModel {RequestId = Activity.Current?.Id ?? HttpContext.TraceIdentifier});
}
```

So for the error page, the current activity id is printed.

With this PR, once users print the `Activity.Current.TraceId` into a log or add it to an error page, it'll match the `TraceId` that the Elastic APM Agent uses - furthermore this also means that this `TraceId` is the same across:
- `Activity`
- Elastic APM Traces
- Logs once the APM Log correlation is enabled

## Limitations

`Activity` is not new and originally they used a hierarchical format - this can be checked with `Activity.IdFormat` - [docs](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.activityidformat?view=netcore-3.1).

Hierarchical id can't be reused for TraceId, since it is not compatible with W3C TraceContext (it uses invalid characters (not hex), etc.). Unfortunately it's still the default. Therefore: This feature only works if user sets `Activity.DefaultIdFormat = ActivityIdFormat.W3C;` or the service is called with `traceparent` header in ASP.NET Core 3+ (in this case `Activity` autodetects and uses W3C format unless forced with `Activity.DefaultIdFormat` to not do so).

## What this PR does not do

This does not mean we capture `Activity` instances or map it to a transaction or span - that is out of scope here. This is about making sure the traceid is the same across `Activity` and our transactions/spans. Potential follow-up: https://github.com/elastic/apm-agent-dotnet/issues/303

### DIFF
--- a/src/Elastic.Apm/Elastic.Apm.csproj
+++ b/src/Elastic.Apm/Elastic.Apm.csproj
@@ -11,15 +11,11 @@
   </PropertyGroup>
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="System.Diagnostics.PerformanceCounter" Version="4.5.0" />
-    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.0.0" />
-  </ItemGroup>
-  <ItemGroup Condition="'$(TargetFramework)' == 'net461'">
-    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.4.0" />  
   </ItemGroup>
   <ItemGroup>
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.6.0" />
     <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="2.0.2" />
     <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
-    <!-- TODO: Are we ok with this reference/version? -->
     <PackageReference Include="System.Threading.Tasks.Dataflow" Version="4.9.0" />
   </ItemGroup>
 </Project>
--- a/src/Elastic.Apm/Model/Transaction.cs
+++ b/src/Elastic.Apm/Model/Transaction.cs
@@ -73,9 +73,23 @@ namespace Elastic.Apm.Model
 			var isSamplingFromDistributedTracingData = false;
 			if (distributedTracingData == null)
 			{
-				var traceIdBytes = new byte[16];
-				TraceId = RandomGenerator.GenerateRandomBytesAsString(traceIdBytes);
+				// Here we ignore Activity.Current.ActivityTraceFlags because it starts out without setting the IsSampled flag, so relying on that would mean a transaction is never sampled.
 				IsSampled = sampler.DecideIfToSample(idBytes);
+
+				if (Activity.Current != null && Activity.Current.IdFormat == ActivityIdFormat.W3C)
+				{
+					TraceId = Activity.Current.TraceId.ToString();
+					ParentId = Activity.Current.ParentId;
+
+					// Also mark the sampling decision on the Activity
+					if(IsSampled)
+						Activity.Current.ActivityTraceFlags |= ActivityTraceFlags.Recorded;
+				}
+				else
+				{
+					var traceIdBytes = new byte[16];
+					TraceId = RandomGenerator.GenerateRandomBytesAsString(traceIdBytes);
+				}
 			}
 			else
 			{
--- a/src/Elastic.Apm/Model/Transaction.cs
+++ b/src/Elastic.Apm/Model/Transaction.cs
@@ -70,6 +70,8 @@ namespace Elastic.Apm.Model
 			HasCustomName = false;
 			Type = type;
 
+			StartActivity();
+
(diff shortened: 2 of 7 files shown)


src/Elastic.Apm/Elastic.Apm.csproj
```diff
@@ -11,15 +11,11 @@
   </PropertyGroup>
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="System.Diagnostics.PerformanceCounter" Version="4.5.0" />
-    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.0.0" />
-  </ItemGroup>
-  <ItemGroup Condition="'$(TargetFramework)' == 'net461'">
-    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.4.0" />  
   </ItemGroup>
   <ItemGroup>
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.6.0" />
```
gregkalapos: This is TBD - we recently changed this and increased to `4.4.0` only for `net461`. We should double-check if using `4.6.0` does not cause trouble to apps using the agent.

The newer version is needed, because `Activity.IdFormat` does not exist in older versions.

src/Elastic.Apm/Model/Transaction.cs
```diff
@@ -70,12 +70,25 @@ ICurrentExecutionSegmentsContainer currentExecutionSegmentsContainer
...
+
 			var isSamplingFromDistributedTracingData = false;
 			if (distributedTracingData == null)
 			{
-				var traceIdBytes = new byte[16];
-				TraceId = RandomGenerator.GenerateRandomBytesAsString(traceIdBytes);
+				// Here we ignore Activity.Current.ActivityTraceFlags because it starts out without setting the IsSampled flag, so relying on that would mean a transaction is never sampled.
 				IsSampled = sampler.DecideIfToSample(idBytes);
+
+				if (Activity.Current != null && Activity.Current.IdFormat == ActivityIdFormat.W3C)
```
SergeyKleyman, line 81: It might be worth considering that this change breaks backward compatibility - before this change transactions started using public API with `null` `DistributedTracingData` created new trace ID.

gregkalapos: I'm not sure this is a problem and I'm also not sure I'd consider this as breaking backward compatibility.

I think this only means that prior to this a new TraceId was generated and now the agent picks up the TraceId from Activity. I think this only means the agent ignored the TraceId on the Activity previously but now it does not - I'm not sure this is something a user would rely on.

Or is there anything else I miss here?

SergeyKleyman: I cannot say for sure that there are real users that might rely on public API generating different trace IDs for transactions started using public API with `null` `DistributedTracingData`.

I can imagine a scenario where an application reads a batch of tasks from somewhere and kicks off distributed processing of each task. Before it was possible to start a separate trace for each task but now if this processing happens in the context of an existing Activity (which might have been created by some library over which user doesn't have any control and doesn't care about Activities it creates) it's not possible to start a separate trace for each task anymore.

One way to solve this issue is to make this change as opt-in - agent can opt-in for transactions it creates automatically but public API will not opt-in by default to keep backward compatibility.

gregkalapos: Oh, I see the use-case you mean. Definitely a valid concern and indeed in that case the 2 transactions would be on the same trace now.

But why is this a backward compatibility issue? 🤔 Definitely changes how the agent behaves - and after the update the transactions will be part of the same trace (unlike before), but to me the question is more like if this is considered a breaking change or not.

Regarding making it opt-in: could be an option, but not sure we need to go into that complication.

One thing to consider here is that this only changes the behaviour if there is an active transaction and the user starts multiple separate transactions within the same chain of activities. E.g. in case of an ASP.NET Core app that would mean 2 transactions within the same HTTP request - and similar examples. One could argue that in this case making those multiple transactions as part of the same trace is actually correct and not having them in the same trace would be incorrect.

So usually when a library starts an activity it also ends it when the "unit of work" (like an HTTP request) is finished (and with that basically "resets" the TraceId). So creating multiple transactions with the same Activity really seems like they should be in the same trace.

To me this feels like something we could accept - but we can follow up on this in a planning.

Issue #802, opened 2020-04-02
Handle competing instrumentations


gregkalapos:
_Important for releasing `Elastic.Apm.SqlClient` - and also for future instrumentation._ 

There are situations when multiple components would crate spans for the same action. 

E.g.:
- EF Core is used with MS SQL. -> Both EF Core instrumentation and `SqlClient` instrumentation would capture the db calls
- An HTTP based library with some specific instrumentation (potentially GraphQL or lots of Azure libraries): Both the library specific instrumentation and the `HttpClient` instrumentation will capture the call - in some cases this may be want we want, but in some cases it's not and we don't want to show the pure HTTP call.
- And probably many more.

We should come up with a solution where 1 instrumentation component can inform other instrumentation components that it already captured a span for a specific action.

gregkalapos:
Implemented in  #809.

PR #809, opened 2020-04-03
Competing instrumentations - Introduce ActiveInstrumentationFlags on Transaction

gregkalapos:
Solves #802

(updated to store the info on a given span instead of storing it on the transaction)

Adds a flag to detect competing instrumentations. Each instrumenting module can set its own flag and with that, instrumenting modules can know which other module created a given span.

After this we can turn on both `SqlClient` and EFCore (or EF Classic) instrumentation. This means `Elastic.Apm.SqlClient` will know if `Elastic.Apm.EntityFrameworkCore` or `Elastic.Apm.EntityFramework6` created the currently active span and vice versa, so we avoid capturing the same db call twice.

Long term we can refine this and e.g. store this flag cover other scenarios. In this PR the focus is to create the base to turn on `SqlClient` instrumentation by default in additional to EFCore instrumentation.

### DIFF
--- a/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
+++ b/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
@@ -100,7 +100,7 @@ namespace Elastic.Apm.AspNetCore
 						transaction = _tracer.StartTransactionInternal(
 							transactionName,
 							ApiConstants.TypeRequest,
-							tracingData);
+							tracingData, InstrumentationFlag.AspNetCore);
 					}
 					else
 					{
@@ -110,14 +110,14 @@ namespace Elastic.Apm.AspNetCore
 								DistributedTracing.TraceContext.TraceParentHeaderNamePrefixed, headerValue);
 
 						transaction = _tracer.StartTransactionInternal(transactionName,
-							ApiConstants.TypeRequest);
+							ApiConstants.TypeRequest, instrumentationFlag: InstrumentationFlag.AspNetCore);
 					}
 				}
 				else
 				{
 					_logger.Debug()?.Log("Incoming request. Starting Trace.");
 					transaction = _tracer.StartTransactionInternal(transactionName,
-						ApiConstants.TypeRequest);
+						ApiConstants.TypeRequest, instrumentationFlag: InstrumentationFlag.AspNetCore);
 				}
 
 				return transaction;
--- a/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
+++ b/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
@@ -8,6 +8,7 @@ using System.Threading.Tasks;
 using Elastic.Apm.Api;
 using Elastic.Apm.AspNetCore.Extensions;
 using Elastic.Apm.Config;
+using Elastic.Apm.DistributedTracing;
 using Elastic.Apm.Logging;
 using Elastic.Apm.Model;
 using Microsoft.AspNetCore.Http;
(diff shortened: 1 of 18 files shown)


SergeyKleyman:
Isn't keeping track on transaction level be a problem if a transaction has more than one concurrent span sub-tree? Concurrent spans will be considered conflicting even though they are unrelated.

SergeyKleyman:
Wouldn't making each instrumentation know about all the rest create a combinatorial explosion? Why not just collapse nested spans of with the same (type, subtype, action)?

gregkalapos:
> Isn't keeping track on transaction level be a problem if a transaction has more than one concurrent span sub-tree? Concurrent spans will be considered conflicting even though they are unrelated.

For the current use-case it's not a problem. All I want to achieve now is that on a given transaction either the `SqlClient` instrumentation or the EFCore/Ef6 instrumentation captures db spans. Of course there can be use cases where this is not enough - that is what I meant when I added to the end of the description that "Long term we can refine this and e.g. store this flag also on every single span to cover other scenarios." Technically there could be an app with a mix of executing SQL statement with `SqlClient` and also using EFCore/EF6, but as the first step let's just not try to cover that use case - everything in this PR is internal, with this I feel we can be confident and release the SqlClient instrumentation (we have a ton of question about when we finally release that) - and if we have feedback that we need a more fine-granular competing instrumentation handling we can just extend this implementation.

The PR at this point is draft, I plan some other things here, one would be a setting to turn off this and ignore the flags. In apps with mixed db access (e.g. both EFCore and direct SqlClient) they can just turn this off and then we capture some db calls twice which for that use-case would be in my opinion acceptable in the first iteration.

> Wouldn't making each instrumentation know about all the rest create a combinatorial explosion? 

No, it's not that every single instrumentation checks for all other instrumentation - only potentially competing ones check for each other's flag. E.g. EFCore capturing won't care about HTTP or ASP.NET Core capturing.

> Why not just collapse nested spans of with the same (type, subtype, action)?

I'm not totally sure what this would be, but in the case of db capturing all 3 of these fields are the same both for the SqlClient and the EfCore/Ef6 instrumentation - so we can have multiple spans with the same type, subtype, action, and they can come from either all from the SqlClient instrumentation or some of them from SqlClient and some of them from EfCore/Ef6.

SergeyKleyman:
> > Why not just collapse nested spans of with the same (type, subtype, action)?
> 
> I'm not totally sure what this would be, but in the case of db capturing all 3 of these fields are the same both for the SqlClient and the EfCore/Ef6 instrumentation - so we can have multiple spans with the same type, subtype, action, and they can come from either all from the SqlClient instrumentation or some of them from SqlClient and some of them from EfCore/Ef6.

Now I am not totally sure what you mean :)

What I suggest is to check parent span when creating DB span - if the parent span is also DB then we don't create nested DB span - we just merge all the captured info into the parent span to handle the case when the child span is captured by instrumentation that gets more info than the one that captured the parent span. We can add a flag to mark spans automatically captured by agent and do this nested spans collapsing only for automatically captured spans so this change doesn't affect spans created by using public API.

gregkalapos:
> > > Why not just collapse nested spans of with the same (type, subtype, action)?
> > 
> > 
> > I'm not totally sure what this would be, but in the case of db capturing all 3 of these fields are the same both for the SqlClient and the EfCore/Ef6 instrumentation - so we can have multiple spans with the same type, subtype, action, and they can come from either all from the SqlClient instrumentation or some of them from SqlClient and some of them from EfCore/Ef6.
> 
> Now I am not totally sure what you mean :)
> 
> What I suggest is to check parent span when creating DB span - if the parent span is also DB then we don't create nested DB span - we just merge all the captured info into the parent span to handle the case when the child span is captured by instrumentation that gets more info than the one that captured the parent span. We can add a flag to mark spans automatically captured by agent and do this nested spans collapsing only for automatically captured spans so this change doesn't affect spans created by using public API.

Ok, now I understand what you mean. Yes, I was also thinking about something similar - this would also work.

With this we assume DiagnosticSource events (and in general span creation) in this order (some pseudo-code):

```
EFCore.StartDbCommand
SqlClient.StartDbCommand
SqlClient.StopDbCommand
EFCore.StopDbCommand
```

And we assume that in case of a single db call an order like this never happens:
```
SqlClient.StartDbCommand
SqlClient.StopDbCommand
EFCore.StartDbCommand
EFCore.StopDbCommand
```

In this case the SqlClient span would not be the parent of the EFCore span.

In the current situation I think this is a safe assumption - I debugged this and this seems to be the case. But we should keep in mind that this is only about nested spans.

gregkalapos:
Pushed an update with not doing this on the transaction, but instead storing this info on the span.

PR #811, opened 2020-04-06
[Doc] Create child spans on spans

gregkalapos:
Added 2 things to the API docs:

- Add `DistributedTracingData` as optional parameter when a transaction is started
- Document creating child spans on spans 

Both of these were added a very long time ago. We were asked about creating child spans on spans [here](https://discuss.elastic.co/t/create-span-as-a-child-of-a-span-asp-net-apm-agent/226678), wanted to link the docs, but realized it's missing, so adding now.

### DIFF
--- a/docs/public-api.asciidoc
+++ b/docs/public-api.asciidoc
@@ -19,7 +19,7 @@ You can access the API by using the static property on the Agent: `Elastic.Apm.A
 
 [float]
 [[api-start-transaction]]
-==== `ITransaction StartTransaction(string name, string type)`
+==== `ITransaction StartTransaction(string name, string type, DistributedTracingData = null)`
 Use this method to create a custom transaction.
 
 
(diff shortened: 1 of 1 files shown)


docs/public-api.asciidoc
```diff
@@ -82,7 +82,7 @@ var span = Elastic.Apm.Agent.Tracer.CurrentSpan;
 ==== `CaptureTransaction`
 
 This is a convenient method which starts and ends a transaction and captures unhandled exceptions.
-It has 3 parameters:
+It has 3 required parameters:
```
bmorelli25, line 85: One note: it's interesting that you separated required and optional parameters for `CaptureTransaction`, but for `CpatureSpan`, you grouped required and optional together (see https://github.com/elastic/apm-agent-dotnet/pull/811/files#diff-98d2de361a542d2ae87910714e3ddb27R508). For consistency, we should probably document the parameters in the same way for both methods

gregkalapos: Good catch, thanks. Changed `CaptureSpan`, so now they are separated for both methods.

Issue #483, opened 2019-09-10
Add Troubleshooting section to the docs


SergeyKleyman:
For example [Java agent's Troubleshooting section](https://www.elastic.co/guide/en/apm/agent/java/current/trouble-shooting.html))

Also see https://github.com/elastic/apm-dev/issues/425#issuecomment-490757036

Derived from #341

Candidates to be included in Troubleshooting section:
- [SystemTotalCpuProvider fails with InvalidOperationException: Category does not exist #501](https://github.com/elastic/apm-agent-dotnet/issues/501#issuecomment-534193901)

gregkalapos:
@bmorelli25 is this something we still want to have? If not, I'd suggest that we close this.

bmorelli25:
All other agents (except Go) have a troubleshooting section. I think we should still plan on adding one to the .NET agent docs. Are there any other common questions you can think of that might be good additions?

gregkalapos:
Ok, sure - makes sense, let's keep this issue.

Logging is definitely 1 topic we could add.

bmorelli25:
Great. Once we get the Troubleshooting section up, you can do what other agent devs do:
1. Notice that the same question has been asked in discuss/github/support multiple times
2. Open an issue with links to the discuss/github/support cases
3. I'll turn your responses into a documentation PR
4. No one asks that question any more 😄

gregkalapos:
Merged #814 - closing this now and we'll roll with the suggestion above in the future :)

PR #814, opened 2020-04-07
[Doc] Add troubleshooting section to docs

gregkalapos:
#483

Mostly covering logging. I guess this will be a continuous effort...

### DIFF
--- a/docs/troubleshooting.asciidoc
+++ b/docs/troubleshooting.asciidoc
@@ -54,7 +54,7 @@ internal class ApmLoggerBridge : IApmLogger
 		// to the input parameter of this method and return if it's the same/higher or not
 	}
 
-	public void Log<TState>(ApmLogLevel apmLogLevel, TState state, Exception e, 	Func<TState, Exception, string> formatter)
+	public void Log<TState>(ApmLogLevel apmLogLevel, TState state, Exception e, Func<TState, Exception, string> formatter)
 	{
 		// You can log the given log into your logging system here.
 	}
--- a/docs/troubleshooting.asciidoc
+++ b/docs/troubleshooting.asciidoc
@@ -74,7 +74,8 @@ For this in ASP.NET Core you need to place the following code into the `Applicat
 AgentDependencies.Logger = new ApmLoggerBridge();
 ----
 
-The `AgentDependencies` class lives in the `Elastic.Apm.AspNetFullFramework` namespace and during initialization the agent checks if an additional logger was configured - the agent only does this once, therefore it's important to set it as early in the process as possible (typically in the `Application_Start` method).
+The `AgentDependencies` class lives in the `Elastic.Apm.AspNetFullFramework` namespace.
+During initialization, the agent checks if an additional logger was configured--the agent only does this once, so it's important to set it as early in the process as possible (typically in the `Application_Start` method).
 
 
 ==== General .NET applications
--- a/docs/troubleshooting.asciidoc
+++ b/docs/troubleshooting.asciidoc
@@ -58,6 +58,7 @@ internal class ApmLoggerBridge : IApmLogger
 	{
 		// You can log the given log into your logging system here.
 	}
+}
 ----
 
 An example implementation for NLog can be seen https://github.com/elastic/apm-agent-dotnet/blob/master/sample/AspNetFullFrameworkSampleApp/App_Start/ApmLoggerToNLog.cs[in our GitHub repository].
--- a/docs/troubleshooting.asciidoc
+++ b/docs/troubleshooting.asciidoc
@@ -26,7 +26,7 @@ The logs should include everything from the application startup up until the fir
 If neither errors nor performance metrics are being sent to the APM Server,
 it's a good idea to first check your logs and look for output just as the app starts.
 
-In case you don't see anything suspicious in the agent logs (no warning or error) it's recommended to turn the log level to `Trace` for further investigation.
+If you don't see anything suspicious in the agent logs (no warning or error), it's recommended to turn the log level to `Trace` for further investigation.
 
 === Collecting agent logs
 
--- a/docs/troubleshooting.asciidoc
+++ b/docs/troubleshooting.asciidoc
@@ -10,9 +10,10 @@ Use the information in this section to troubleshoot common problems and find
 answers for frequently asked questions.
 As a first step, ensure your stack is compatible with the Agent's <<supported-technologies,supported technologies>>.
 
-Don't worry if you can't figure out what the problem is.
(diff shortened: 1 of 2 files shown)


gregkalapos:
Now done 👼  - promise I stop pushing more commits until there is a review.

docs/troubleshooting.asciidoc
```diff
@@ -0,0 +1,89 @@
...
+
+==== ASP.NET Core 
+
+In case you activated the agent with the `UseAllElasticApm` or the `UseElasticApm` method as described in <<setup-asp-net-core,the setup section>> the agent will integrate with the https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-3.1[ASP.NET Core logging infrastructure] and it'll log as any other component logs. This means, the agent will pick up the configured logging provider and log into that.
+
+==== ASP.NET Classic
+
+Unlike ASP.NET Core, ASP.NET (classic) does not have a predefined logging system. In case you already have a logging system in place (like NLog, Serilog, or similar) you can direct the agent logs into your logging system.
+
+For this you need to create a bridge between the agent's internal logger and your logging system.
```
bmorelli25, line 43:
```suggestion
Unlike ASP.NET Core, ASP.NET (classic) does not have a predefined logging system.
However, if you have a logging system in place, like NLog, Serilog, or similar, you can direct the agent logs into your
logging system by creating a bridge between the agent's internal logger and your logging system. 
```

docs/troubleshooting.asciidoc
```diff
@@ -0,0 +1,89 @@
...
+	public void Log<TState>(ApmLogLevel apmLogLevel, TState state, Exception e, Func<TState, Exception, string> formatter)
+	{
+		// You can log the given log into your logging system here.
+	}
+}
+----
+
+An example implementation for NLog can be seen https://github.com/elastic/apm-agent-dotnet/blob/master/sample/AspNetFullFrameworkSampleApp/App_Start/ApmLoggerToNLog.cs[in our GitHub repository].
+
+Then you need to tell the agent to use the `ApmLoggerBridge`.
```
bmorelli25, line 68: Remove filler words

```suggestion
Then tell the agent to use the `ApmLoggerBridge`.
```

docs/troubleshooting.asciidoc
```diff
@@ -0,0 +1,89 @@
...
+
+For this in ASP.NET Core you need to place the following code into the `Application_Start` method in the `HttpApplication` implementation of your app which is typically in the `Global.asx.cs` file:
```
bmorelli25, line 70: I'm confused by this sentence. Are we back to ASP.NET Core? I thought we were still talking about classic?

docs/troubleshooting.asciidoc
```diff
@@ -0,0 +1,89 @@
...
+
+[source,csharp]
+----
+AgentDependencies.Logger = new ApmLoggerBridge();
+----
+
+The `AgentDependencies` class lives in the `Elastic.Apm.AspNetFullFramework` namespace and during initialization the agent checks if an additional logger was configured - the agent only does this once, therefore it's important to set it as early in the process as possible (typically in the `Application_Start` method).
```
bmorelli25, line 77:
```suggestion
The `AgentDependencies` class lives in the `Elastic.Apm.AspNetFullFramework` namespace.
During initialization, the agent checks if an additional logger was configured--the agent only does this once, so it's important to set it as early in the process as possible (typically in the `Application_Start` method).
```

docs/troubleshooting.asciidoc
```diff
@@ -0,0 +1,89 @@
...
+[float]
+[[no-data-sent]]
+=== No data is sent to the APM Server
+
+If neither errors nor performance metrics are being sent to the APM Server,
+it's a good idea to first check your logs and look for output just as the app starts.
+
+In case you don't see anything suspicious in the agent logs (no warning or error) it's recommended to turn the log level to `Trace` for further investigation.
+
+=== Collecting agent logs
```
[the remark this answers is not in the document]
gregkalapos: Well, this was actually what triggered me to write the whole thing :)

Problem is, people sometimes struggle with collecting logs on ASP.NET classic. As you can see it's not trivial, but I feel this is very much needed in a troubleshooting scenario - otherwise they have no logs. Also, we had lots of questions on how to do this.

I totally agree this is way too complicated. By seeing this I think 1 step we can do mid/long term is to ship `IApmLogger` implementations for popular logging systems as part of the agent. With that the part where we show the `ApmLoggerBridge` (I feel that's the most complicated part) will go away and we just say something like "If you use NLog, just enable it with the code snippet below: (and we put a 1 liner there)"  - and same for Serilog, etc.

So, given that I see lots of users failing in the 1. step of the troubleshooting (collecting logs) I'd keep this and try to simplify this part later by shipping some code that they can just plug in easily. 

What do you think?

docs/troubleshooting.asciidoc
```diff
@@ -0,0 +1,89 @@
+ifdef::env-github[]
+NOTE: For the best reading experience,
+please view this documentation at https://www.elastic.co/guide/en/apm/agent/dotnet[elastic.co]
+endif::[]
+
+[[troubleshooting]]
+== Troubleshooting
```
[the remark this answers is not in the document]
gregkalapos: Yes, the first part of this is copy-paste from Java :)

docs/troubleshooting.asciidoc
gregkalapos: Good catch, my mistake - this should have been "ASP.NET (classic)". Fixed it.

gregkalapos:
> A few recommendations for you... Sorry, I probably went a bit overboard on this one 😇. Mostly, I tried to cut out the filler words and combine or separate sentences to improve the flow. Don't feel like you have to commit my suggestions as-is; feel free to change them.
> 
> My biggest concern is [#814 (comment)](https://github.com/elastic/apm-agent-dotnet/pull/814#discussion_r405156665).


Thanks @bmorelli25. I added a comment to [#814 (comment)](https://github.com/elastic/apm-agent-dotnet/pull/814#discussion_r405156665).

docs/troubleshooting.asciidoc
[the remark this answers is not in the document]
bmorelli25: > Problem is, people sometimes struggle with collecting logs on ASP.NET classic. As you can see it's not trivial, but I feel this is very much needed in a troubleshooting scenario - otherwise they have no logs.

> I'd keep this and try to simplify this part later by shipping some code that they can just plug in easily.

I appreciate the explanation! This makes sense now, and I agree, we should keep it.

PR #817, opened 2020-04-08
ci(docs): docs folder to filter

v1v:
Rather than searching for files with the extensions let's search for files under the `docs` folder.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -43,7 +43,7 @@ pipeline {
             stash allowEmpty: true, name: 'source', useDefaultExcludes: false
             script {
               dir("${BASE_DIR}"){
-                // Skip all the stages except docs for PR's only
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -43,8 +43,8 @@ pipeline {
             stash allowEmpty: true, name: 'source', useDefaultExcludes: false
             script {
               dir("${BASE_DIR}"){
-                // Skip all the stages except docs for PR's with asciidoc and md changes only
-                env.ONLY_DOCS = isGitRegionMatch(patterns: [ '.*\\.(asciidoc|md)' ], shouldMatchAll: true)
+                // Skip all the stages except docs for PR's only
```
v1v, line 46:
```suggestion
                // Skip all the stages for PRs with changes in the docs only
```

PR #826, opened 2020-04-20
ci(jenkins): opbeans bump from the tag releases

v1v:
## What

Disable opbeans for PRs or branches from the UI

## Why

When running the build from the Jenkins UI then this particular stage used to run and caused failures.
The stage is a post-release event therefore it should not be executed from PRs or branches that are not releases though.


v1v:
`Timeout has been exceeded` since the CI was a bit temperamental.

image


I'll merge this PR then

Issue #71, opened 2019-01-21
Load agent without code change


gregkalapos:
By implementing `IHostingStartup` and using the following two environment variables we could load the agent into an ASP.NET Core app without any code change or even recompilation:
- `DOTNET_ADDITIONAL_DEPS`
- `ASPNETCORE_HOSTINGSTARTUPASSEMBLIES`

I blogged about this feature [here](https://kalapos.net/Blog/ShowPost/IHostingStartupInAspNetCore20).

NeilCross:
Oh snap!

Mpdreamz:
That's a really cool feature! Would this provide us a hook to register `PayloadSender` as `IHostedService` rather than it starting a background thread?

gregkalapos:
> Would this provide us a hook to register PayloadSender as IHostedService rather than it starting a background thread?

Not directly. This feature basically enables the application to pick up a dll and register a middleware by code from that dll. And all that without changing the application's code. 

Independently we can still think about registering the `PayloadSender` as a `IHostedService`, but for that I think we don't really need this feature.

gregkalapos:
Another question about this in discuss: https://discuss.elastic.co/t/instrument-any-net-core-application/207680

gregkalapos:
> Another question about this in discuss: https://discuss.elastic.co/t/instrument-any-net-core-application/207680

gregkalapos:
Implemented in #828.

PR #828, opened 2020-04-22
Load agent without code change in .NET Core

gregkalapos:
Solves #71 

.NET Core 2.2 introduced [`DOTNET_STARTUP_HOOKS`](https://github.com/dotnet/runtime/blob/master/docs/design/features/host-startup-hook.md) environment variable which we can utilize in order to load the agent into any .NET Core process without any code change or without the need of adding a NuGet package or recompiling the application.

### How to use it

The agent assemblies need to be loaded onto the disk where the .NET app is running, and then set environment variable to point to the agent:

```
set DOTNET_STARTUP_HOOKS=[pathToAgent]\ElasticApmAgentStartupHook.dll
```

That's it - from that point the app will just "pick up" the agent.

### Changes done in the agent

- Added ElasticApmStartupHook project with `StartupHook.cs` which can be loaded with `DOTNET_STARTUP_HOOKS` which loads and initializes the agent.
- Since there is no way to register an ASP.NET Core middleware that early in the startup process I created a Diagnostic Source listener to create transactions for incoming ASP.NET Core HTTP calls, it does the same as the middleware, but only with DiagnosticSource.

### Limitations 

This only work with .NET Core 2.2+, but .NET Core 2.1 is an LTS version, so it's still important. So for .NET Core 2.1 this does not help. An option would be to use `IHostingStartup` and `ASPNETCORE_HOSTINGSTARTUPASSEMBLIES` (see #71), this would be an additional small project in the solution to cover that. `IHostingStartup` is a solution in ASP.NET Core to load code before the app starts. With that we can support ASP.NET Core 2.1 (no plain .NET Core 2.1 support). So, important to note here is that `DOTNET_STARTUP_HOOKS` is supported by .NET Core itself, regardless of the app type, but `ASPNETCORE_HOSTINGSTARTUPASSEMBLIES` is ASP.NET Core only.

### Things left 

- [ ] Create tests to cover this
- [x] Have a plan on how to release (we need to drop assemblies to the target machine)
- [x] Loading dependencies - currently I load them manually, one option would be to merge those into a single assembly.
- [x] DiagnosticSource is a sync API (`void OnNext(KeyValuePair<string,object>`), but as it seems we can only capture HTTP request body with an async API. That works in our current middleware, but it's not trivial to move the async body capturing into sync DiagnosticSource code.


src/Elastic.Apm.AspNetCore/DiagnosticListener/AspNetCoreErrorDiagnosticsSubscriber.cs
```diff
@@ -4,18 +4,18 @@
 namespace Elastic.Apm.AspNetCore.DiagnosticListener
 {
 	/// <summary>
-	/// Activates the <see cref="AspNetCoreDiagnosticListener" /> which enables
+	/// Activates the <see cref="AspNetCoreErrorDiagnosticListener" /> which enables
 	/// capturing errors within an ASP.NET Core application.
 	/// </summary>
-	public class AspNetCoreDiagnosticsSubscriber : IDiagnosticsSubscriber
+	public class AspNetCoreErrorDiagnosticsSubscriber : IDiagnosticsSubscriber
```
gregkalapos, line 10: This is technically a breaking change but we could merge `AspNetCoreErrorDiagnosticsSubscriber` and `AspNetCorePageLoadDiagnosticListener` into a single one under the old name (`AspNetCoreDiagnosticsSubscriber`)  - that would not be a breaking change.

gregkalapos:
Ah, the integration test fails due to [this](https://github.com/elastic/apm-integration-testing/blob/master/docker/dotnet/run.sh#L42). This is the first release where I use `<VersionSuffix>` to mark it as beta, but when we add the agent into the integration test we ignore the suffix, we only use `VersionPrefix`, so the beta marked package is not found:

```
[2020-09-01T13:24:55.741Z] + DOTNET_AGENT_VERSION=1.7.0
[2020-09-01T13:24:55.741Z] + '[' -z 1.7.0 ']'
[2020-09-01T13:24:55.741Z] + dotnet add package Elastic.Apm.NetCoreAll -v 1.7.0
[2020-09-01T13:24:57.130Z]   Writing /tmp/tmppUt2Kg.tmp
[2020-09-01T13:24:57.130Z] info : Adding PackageReference for package 'Elastic.Apm.NetCoreAll' into project '/src/aspnetcore/TestAspNetCoreApp.csproj'.
[2020-09-01T13:24:57.391Z] info : Restoring packages for /src/aspnetcore/TestAspNetCoreApp.csproj...
[2020-09-01T13:24:57.654Z] info :   GET https://api.nuget.org/v3-flatcontainer/elastic.apm.netcoreall/index.json
[2020-09-01T13:24:58.228Z] info :   OK https://api.nuget.org/v3-flatcontainer/elastic.apm.netcoreall/index.json 424ms
[2020-09-01T13:24:58.228Z] error: Unable to find package Elastic.Apm.NetCoreAll with version (>= 1.7.0)
[2020-09-01T13:24:58.228Z] error:   - Found 13 version(s) in nuget.org [ Nearest version: 1.6.1 ]
[2020-09-01T13:24:58.229Z] error:   - Found 1 version(s) in local [ Nearest version: 1.7.0-beta1 ]
[2020-09-01T13:24:58.801Z] error: Package 'Elastic.Apm.NetCoreAll' is incompatible with 'all' frameworks in project '/src/aspnetcore/TestAspNetCoreApp.csproj'.
```

I won't fix this now, this PR is already huge, so here is the plan:
I'll merge it and then I'll do a beta release with this feature. After that I'll remove the beta suffix, so everything will be green again, and I'll also open an issue to fix this for the next time.

Issue #829, opened 2020-04-23
Example of configuring agent in .NET Core console app


TailyFair:
**Is your feature request related to a problem? Please describe.**
It's not clear how to configure standard .NET Core app (Not ASP.NET Core). In current `HttpListenerSample` sample Agent is not configured and uses defaults. It's confusing how to setup Agent if I want to change `ServerUrls`.

**Describe the solution you'd like**
I would like to see some example on how to configure Agent in simple .NET Core console app.

Or make Agent simpler to configure, to allow just pass `IConfiguration` or set configure with builder pattern.

gregkalapos:
Hi @TailyFair,

I opened #831 to add some more docs on this.

But this is already covered [here](https://www.elastic.co/guide/en/apm/agent/dotnet/current/configuration.html) and should be simple - you can just use environment variables to configure the agent. In order to [set `ServerUrls` you can](https://www.elastic.co/guide/en/apm/agent/dotnet/current/config-reporter.html#config-server-urls) just set the environment variable `ELASTIC_APM_SERVER_URLS`

>  to allow just pass `IConfiguration` 

We also support `net461`, and `IConfiguration` does not exist there, so that's unfortunately not an option.

TailyFair:
Thank you for your response @gregkalapos. Changes to documentation in PR I believe are sufficient.

Environmental variables are partial solution to my use case, since there are multiple apps running in singe VM, so while `ServerUrls` is shared between all apps, settings like `ServiceName`, `ServiceVersion` and `Environment` are different per app.

I ended up implementing `IConfigurationReader` and `IApmLogger` on my own to have full control of configuration without environmental variables usage.

After PR is merged, I think this can be closed.

gregkalapos:
Thanks for the feedback @TailyFair.

Implementing `IConfigurationReader` is also a totally valid option - maybe 1 thing to consider: you can also set the environment variable in C# with `Environment.SetEnvironmentVariable`. If you do that before the agent initializes maybe you can set the few settings that way instead of implementing the whole `IConfigurationReader` interface. I saw some people doing this, so I though I share this.

TailyFair:
Great idea, will try that out! Thanks for documentation improvement.

Issue #824, opened 2020-04-16
Metrics reporting works, but not Transactions, on one single environment


ffMathy:
Here's a funny bug.

In my development environment, Elastic APM works just fine. It reports both transactions and metrics. However, when I deploy the same application to my production environment, ONLY metrics are reported.

What kind of condition can lead to this, and how can I fix it?

gregkalapos:
Hi @ffMathy 

> What kind of condition can lead to this, and how can I fix it?

I'd need more info for this. 

2 things: 
- Can you share more about the application? Like, ASP.NET Core, or ASP.NET (classic), or something else? 
- Can you turn on agent logs (preferable on Debug level) and add the logs here?

ffMathy:
It's an ASP .NET Core 3.1 application.

As for the logs, how do I turn it on?

gregkalapos:
> It's an ASP .NET Core 3.1 application.
> 
> As for the logs, how do I turn it on?

Ok, for ASP.NET Core it's fairly easy - with that setup the agent integrates with the ASP.NET Core logging system, so you'll see the APM Agent logs in your normal logs (e.g. either on the console on a simple setup, or if you configured a logger like Serilog, then the agent logs will end up in your sink as any other logs). 

If you pass an `IConfiguration` instance to the `UseAllElasticApm` or `UseElasticApm` method then you can configure the agent log level like [this](https://www.elastic.co/guide/en/apm/agent/dotnet/current/configuration-on-asp-net-core.html#sample-config).

```
{
  "Logging": {
    "LogLevel": {
      "Default": "Warning",
      "Elastic.Apm": "Debug"
    }
  }
// .... more configs
}
```

If you don't pass an `IConfiguration` instance just set the environment variable `ELASTIC_APM_LOG_LEVEL` to `Debug`.

ffMathy:
Alright, I have logs now. What part of the logs do you want? I can see that some of them include sensitive data, and there's a lot of them.

Do you want the first logs right after startup, or perhaps from a specific log level?

gregkalapos:
> Alright, I have logs now. What part of the logs do you want? I can see that some of them include sensitive data, and there's a lot of them.
> 
> Do you want the first logs right after startup, or perhaps from a specific log level?

[...]

PR #831, opened 2020-04-23
[Docs] Add docs explaining agent initialization

gregkalapos:
Triggered by #829 and also some learnings from #824. 

- Describes how agent initialization works (implicit vs explicit - especially with public agent api)
- Public Agent API + turning on auto instrumentation 
- New troubleshooting point about double agent initialization.

### DIFF
--- a/docs/public-api.asciidoc
+++ b/docs/public-api.asciidoc
@@ -209,7 +209,7 @@ var transaction2 = Agent.Tracer.StartTransaction("Transaction2", "TestTransactio
 //----------------------------
 [float]
 [[api-transaction]]
-=== Transaction API
+== Transaction API
 //----------------------------
 A transaction describes an event captured by an Elastic APM agent monitoring a service. Transactions help combine multiple <<api-span,Spans>> into logical groups, and they are the first <<api-span,Span>> of a service. More information on Transactions and Spans is available in the {apm-overview-ref-v}/apm-data-model.html[APM data model] documentation.
 
@@ -435,7 +435,7 @@ Agent.Tracer.CaptureTransaction("MyCustomTransaction",ApiConstants.TypeRequest, 
 //----------------------------
 [float]
 [[api-span]]
-=== Span API
+== Span API
 //----------------------------
 A span contains information about a specific code path, executed as part of a transaction.
 
--- a/docs/public-api.asciidoc
+++ b/docs/public-api.asciidoc
@@ -16,6 +16,8 @@ as well as track errors.
 
 The API does not require explicit Agent initialization - agent initialization is optional. The `Elastic.Apm.Agent.IsConfigured` property lets you check whether the agent is already initialized.
 
+[float]
+[[implicit-initialization]]
 === Implicit agent initialization
 
 If you don't explicitly initialize the agent, it will be started with a default configuration. This means the agent will read <<configuration, configuration settings>> from environment variables. In case you don't set an environment variable, the Agent will use the default value (in case of `ServerUrls` this is `http://localhost:8200`).
@@ -28,7 +30,8 @@ Another example for initialization is when you enable the agent with one of the 
 
 The default agent setup should cover most of the use cases and the primary way to configure the agent is through environment variables.
 
-
+[float]
+[[explicit-initialization]]
 === Explicit agent initialization
 
 In case you would like to replace one of the agent components, you can do so by calling the `Elastic.Apm.Agent.Setup(AgentComponents)` method.
@@ -41,6 +44,8 @@ In the AgentComponents you can pass following optional components to the agent:
 
 NOTE: In case of ASP.NET Core, when you register the agent, the `UseElasticApm` and the `UseAllElasticApm` methods both implicitly initialize the agent by calling the `Elastic.Apm.Agent.Setup` method internally. In that setup the `IConfigurationReader` implementation will read configuration from the ASP.NET Core configuration system in case you pass an `IConfiguration` instance to the method. The `IApmLogger` instance will also log through the configured logging provider by integrating into the ASP.NET Core logging system.
 
+[float]
+[[auto-instrumentation-and-agent-api]]
 === Auto instrumentation in combination with the Public Agent API
 
 With the `Elastic.ApmAgent.Subscribe(params IDiagnosticsSubscriber[] subscribers)` method you can turn on auto instrumentation for supported libraries.
--- a/docs/public-api.asciidoc
+++ b/docs/public-api.asciidoc
@@ -14,19 +14,19 @@ as well as track errors.
 [[api-initialization]]
 == Initialization
 
-The API does not require explicit Agent initialization - agent initialization is optional. The `Elastic.Apm.Agent.IsConfigured` property lets you check whether the agent is already initialized.
+The API does not require explicit Agent initialization—agent initialization is optional. The `Elastic.Apm.Agent.IsConfigured` property lets you check whether the agent is already initialized.
 
 [float]
 [[implicit-initialization]]
 === Implicit agent initialization
 
-If you don't explicitly initialize the agent, it will be started with a default configuration. This means the agent will read <<configuration, configuration settings>> from environment variables. In case you don't set an environment variable, the Agent will use the default value (in case of `ServerUrls` this is `http://localhost:8200`).
+If you don't explicitly initialize the agent, it will be started with a default configuration. This means the Agent will read <<configuration,configuration settings>> from environment variables. If you don't set an environment variable, the Agent will use the default value. For example, the `ServerUrls` default is `http://localhost:8200`.
 
 This implicit initialization of the agent happens on the first call on the `Elastic.Apm.Agent` class.
 
-NOTE: One exception is the `Elastic.Apm.Agent.IsConfigured` method. This method never initializes the agent, it only checks whether the agent is already initialized.
+NOTE: One exception is the `Elastic.Apm.Agent.IsConfigured` method. This method never initializes the agent, it only checks if the agent is already initialized.
 
-Another example for initialization is when you enable the agent with one of the technology specific methods from the <<setup>> instructions. Specifically when the `UseElasticApm` or `UseAllElasticApm` method is called in ASP.NET Core or when the IIS module is initialized in case of IIS applications.
+Another example of initialization is when you enable the Agent with one of the technology-specific methods from the <<setup>> instructions. Specifically when the `UseElasticApm` or `UseAllElasticApm` method is called in ASP.NET Core or when the IIS module is initialized in an IIS application.
 
 The default agent setup should cover most of the use cases and the primary way to configure the agent is through environment variables.
 
@@ -34,15 +34,14 @@ The default agent setup should cover most of the use cases and the primary way t
 [[explicit-initialization]]
 === Explicit agent initialization
 
-In case you would like to replace one of the agent components, you can do so by calling the `Elastic.Apm.Agent.Setup(AgentComponents)` method.
-
+If you would like to replace one of the agent components, you can do so by calling the `Elastic.Apm.Agent.Setup(AgentComponents)` method.
 In the AgentComponents you can pass following optional components to the agent:
 
-- `IApmLogger`: A logger implementation which will be used to print log messages. Default: A console logger.
-- `IPayloadSender`: A component that receives all the captured events like spans, transactions and metrics. The default implementation serializes all events and sends them to the Elastic APM Server
+- `IApmLogger`: A logger implementation that will be used to print log messages. Default: A console logger.
+- `IPayloadSender`: A component that receives all the captured events like spans, transactions, and metrics. The default implementation serializes all events and sends them to the Elastic APM Server
 - `IConfigurationReader`: A component that reads <<configuration, agent configuration settings>>. The default implementation reads configuration through environment variables.
 
-NOTE: In case of ASP.NET Core, when you register the agent, the `UseElasticApm` and the `UseAllElasticApm` methods both implicitly initialize the agent by calling the `Elastic.Apm.Agent.Setup` method internally. In that setup the `IConfigurationReader` implementation will read configuration from the ASP.NET Core configuration system in case you pass an `IConfiguration` instance to the method. The `IApmLogger` instance will also log through the configured logging provider by integrating into the ASP.NET Core logging system.
+NOTE: In the case of ASP.NET Core, when you register the agent, the `UseElasticApm` and the `UseAllElasticApm` methods both implicitly initialize the agent by calling the `Elastic.Apm.Agent.Setup` method internally. In that setup, the `IConfigurationReader` implementation will read configuration from the ASP.NET Core configuration system in case you pass an `IConfiguration` instance to the method. The `IApmLogger` instance will also log through the configured logging provider by integrating into the ASP.NET Core logging system.
 
 [float]
 [[auto-instrumentation-and-agent-api]]
@@ -50,9 +49,9 @@ NOTE: In case of ASP.NET Core, when you register the agent, the `UseElasticApm` 
 
 With the `Elastic.ApmAgent.Subscribe(params IDiagnosticsSubscriber[] subscribers)` method you can turn on auto instrumentation for supported libraries.
 
-In case of ASP.NET Core, when you turn on the agent with the `UseAllElasticApm` method, the agent will do this automatically.
+In the case of ASP.NET Core, when you turn on the agent with the `UseAllElasticApm` method, the agent will do this automatically.
 
-In case of a typical console application, you need to do this manually.
+With a typical console application, you need to do this manually.
 
 `IDiagnosticsSubscriber` implementations are offered by the agent and they subscribe to diagnostic source events or other data sources in order to capture events automatically.
 
--- a/docs/public-api.asciidoc
+++ b/docs/public-api.asciidoc
@@ -41,6 +41,22 @@ In the AgentComponents you can pass following optional components to the agent:
 
 NOTE: In case of ASP.NET Core, when you register the agent, the `UseElasticApm` and the `UseAllElasticApm` methods both implicitly initialize the agent by calling the `Elastic.Apm.Agent.Setup` method internally. In that setup the `IConfigurationReader` implementation will read configuration from the ASP.NET Core configuration system in case you pass an `IConfiguration` instance to the method. The `IApmLogger` instance will also log through the configured logging provider by integrating into the ASP.NET Core logging system.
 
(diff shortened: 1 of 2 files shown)


docs/public-api.asciidoc
```diff
@@ -10,6 +10,58 @@ customize and manually create spans and transactions,
...
+
+[float]
+[[explicit-initialization]]
+=== Explicit agent initialization
+
+In case you would like to replace one of the agent components, you can do so by calling the `Elastic.Apm.Agent.Setup(AgentComponents)` method.
+
+In the AgentComponents you can pass following optional components to the agent:
+
+- `IApmLogger`: A logger implementation which will be used to print log messages. Default: A console logger.
```
bmorelli25, line 41:
```suggestion
- `IApmLogger`: A logger implementation that will be used to print log messages. Default: A console logger.
```

docs/public-api.asciidoc
```diff
@@ -10,6 +10,58 @@ customize and manually create spans and transactions,
...
+- `IPayloadSender`: A component that receives all the captured events like spans, transactions and metrics. The default implementation serializes all events and sends them to the Elastic APM Server
```
bmorelli25, line 42:
```suggestion
- `IPayloadSender`: A component that receives all the captured events like spans, transactions, and metrics. The default implementation serializes all events and sends them to the Elastic APM Server
```

docs/public-api.asciidoc
```diff
@@ -10,6 +10,58 @@ customize and manually create spans and transactions,
...
+- `IConfigurationReader`: A component that reads <<configuration, agent configuration settings>>. The default implementation reads configuration through environment variables.
+
+NOTE: In case of ASP.NET Core, when you register the agent, the `UseElasticApm` and the `UseAllElasticApm` methods both implicitly initialize the agent by calling the `Elastic.Apm.Agent.Setup` method internally. In that setup the `IConfigurationReader` implementation will read configuration from the ASP.NET Core configuration system in case you pass an `IConfiguration` instance to the method. The `IApmLogger` instance will also log through the configured logging provider by integrating into the ASP.NET Core logging system.
```
bmorelli25, line 45:
```suggestion
NOTE: In the case of ASP.NET Core, when you register the agent, the `UseElasticApm` and the `UseAllElasticApm` methods both implicitly initialize the agent by calling the `Elastic.Apm.Agent.Setup` method internally. In that setup, the `IConfigurationReader` implementation will read configuration from the ASP.NET Core configuration system in case you pass an `IConfiguration` instance to the method. The `IApmLogger` instance will also log through the configured logging provider by integrating into the ASP.NET Core logging system.
```

docs/troubleshooting.asciidoc
```diff
@@ -90,3 +90,24 @@ For this you'll need an `IApmLogger` implementation (see above) which you need t
...
+
+Another scenario can be that you use the Public Agent API in combination of the IIS module or the ASP.NET Core NuGet package where you enable the agent with the `UseElasticApm` or `UseAllElasticApm` methods. Both the first call to the IIS module and the `UseElasticApm`/`UseAllElasticApm` methods internally call the `Elastic.Apm.Agent.Setup` method to initialize the agent.
+
+In case you use the Public Agent API with the `Elastic.Apm.Agent` class that can be potentially executed before the IIS module initializes or the `UseElasticApm`/`UseAllElasticApm` calls execute, then those will fail, since you implicitly initialized the agent before.
+
+To prevent the `InstanceAlreadyCreatedException` in those scenarios, in case you have code that can potentially run very early, before the agent is initialized, it's recommended to check with the `Elastic.Apm.Agent.IsConfigured` method whether the agent is already initialized and only then use other methods in the Public Agent API.
```
bmorelli25, line 112:
```suggestion
To prevent the `InstanceAlreadyCreatedException` in those scenarios, in case you have code that can potentially run very early before the agent is initialized, it's recommended to check with the `Elastic.Apm.Agent.IsConfigured` method whether the agent is already initialized and only then use other methods in the Public Agent API.
```

docs/public-api.asciidoc
```diff
@@ -10,6 +10,58 @@ customize and manually create spans and transactions,
 as well as track errors.
 
 
+[float]
+[[api-initialization]]
+== Initialization
+
+The API does not require explicit Agent initialization - agent initialization is optional. The `Elastic.Apm.Agent.IsConfigured` property lets you check whether the agent is already initialized.
```
bmorelli25, line 17:
```suggestion
The API does not require explicit Agent initialization—agent initialization is optional. The `Elastic.Apm.Agent.IsConfigured` property lets you check whether the agent is already initialized.
```

docs/public-api.asciidoc
```diff
@@ -10,6 +10,58 @@ customize and manually create spans and transactions,
...
+
+[float]
+[[implicit-initialization]]
+=== Implicit agent initialization
+
+If you don't explicitly initialize the agent, it will be started with a default configuration. This means the agent will read <<configuration, configuration settings>> from environment variables. In case you don't set an environment variable, the Agent will use the default value (in case of `ServerUrls` this is `http://localhost:8200`).
```
bmorelli25, line 23:
```suggestion
If you don't explicitly initialize the agent, it will be started with a default configuration. This means the Agent will read <<configuration,configuration settings>> from environment variables. If you don't set an environment variable, the Agent will use the default value. For example, the `ServerUrls` default is `http://localhost:8200`.
```

docs/public-api.asciidoc
```diff
@@ -10,6 +10,58 @@ customize and manually create spans and transactions,
...
+
+This implicit initialization of the agent happens on the first call on the `Elastic.Apm.Agent` class.
+
+NOTE: One exception is the `Elastic.Apm.Agent.IsConfigured` method. This method never initializes the agent, it only checks whether the agent is already initialized.
```
bmorelli25, line 27:
```suggestion
NOTE: One exception is the `Elastic.Apm.Agent.IsConfigured` method. This method never initializes the agent, it only checks if the agent is already initialized.
```

docs/public-api.asciidoc
```diff
@@ -10,6 +10,58 @@ customize and manually create spans and transactions,
...
+
+Another example for initialization is when you enable the agent with one of the technology specific methods from the <<setup>> instructions. Specifically when the `UseElasticApm` or `UseAllElasticApm` method is called in ASP.NET Core or when the IIS module is initialized in case of IIS applications.
```
bmorelli25, line 29:
```suggestion
Another example of initialization is when you enable the Agent with one of the technology-specific methods from the <<setup>> instructions. Specifically when the `UseElasticApm` or `UseAllElasticApm` method is called in ASP.NET Core or when the IIS module is initialized in an IIS application.
```

docs/public-api.asciidoc
```diff
@@ -10,6 +10,58 @@ customize and manually create spans and transactions,
...
+
+The default agent setup should cover most of the use cases and the primary way to configure the agent is through environment variables.
+
+[float]
+[[explicit-initialization]]
+=== Explicit agent initialization
+
+In case you would like to replace one of the agent components, you can do so by calling the `Elastic.Apm.Agent.Setup(AgentComponents)` method.
```
bmorelli25, line 37:
```suggestion
If you would like to replace one of the agent components, you can do so by calling the `Elastic.Apm.Agent.Setup(AgentComponents)` method.
```

docs/public-api.asciidoc
```diff
@@ -10,6 +10,58 @@ customize and manually create spans and transactions,
...
+
```
bmorelli25, line 38: I think we should combine these into one paragraph

```suggestion
```

docs/public-api.asciidoc
```diff
@@ -10,6 +10,58 @@ customize and manually create spans and transactions,
...
+NOTE: In case of ASP.NET Core, when you register the agent, the `UseElasticApm` and the `UseAllElasticApm` methods both implicitly initialize the agent by calling the `Elastic.Apm.Agent.Setup` method internally. In that setup the `IConfigurationReader` implementation will read configuration from the ASP.NET Core configuration system in case you pass an `IConfiguration` instance to the method. The `IApmLogger` instance will also log through the configured logging provider by integrating into the ASP.NET Core logging system.
+
+[float]
+[[auto-instrumentation-and-agent-api]]
+=== Auto instrumentation in combination with the Public Agent API
+
+With the `Elastic.ApmAgent.Subscribe(params IDiagnosticsSubscriber[] subscribers)` method you can turn on auto instrumentation for supported libraries.
+
+In case of ASP.NET Core, when you turn on the agent with the `UseAllElasticApm` method, the agent will do this automatically.
```
bmorelli25, line 53:
```suggestion
In the case of ASP.NET Core, when you turn on the agent with the `UseAllElasticApm` method, the agent will do this automatically.
```

docs/public-api.asciidoc
```diff
@@ -10,6 +10,58 @@ customize and manually create spans and transactions,
...
+
+In case of a typical console application, you need to do this manually.
```
bmorelli25, line 55:
```suggestion
With a typical console application, you need to do this manually.
```

docs/troubleshooting.asciidoc
```diff
@@ -90,3 +90,24 @@ For this you'll need an `IApmLogger` implementation (see above) which you need t
...
 Agent.Setup(new AgentComponents(logger: new ApmLoggerBridge()));
 ----
+
+[float]
+[[double-agent-initialization]]
+=== An `InstanceAlreadyCreatedException` exception is thrown
+
+It can happen that in a very early stage of the monitored process the agent throws an `InstanceAlreadyCreatedException` exception with the following message: "The singleton APM agent has already been instantiated and can no longer be configured."
+
+This happens when the agent would be initialized multiple times, which is prohibited.
```
bmorelli25, line 100: I think we should combine these into one paragraph

docs/troubleshooting.asciidoc
```diff
@@ -90,3 +90,24 @@ For this you'll need an `IApmLogger` implementation (see above) which you need t
...
+
+NOTE: The reason for forcing a single agent instance per process is that enabling multiple agents would open up problems like capturing events and metrics multiple times for each agent instance, or having multiple background threads for event serialization and transfer to the APM server.
+
+Take a look at the initialization section of the <<public-api, Public Agent API>> for more information on how agent initialization works.
+
+This issue can happen for example if you call the `Elastic.Apm.Agent.Setup` method multiple times or in case you call another method on `Elastic.Apm.Agent` which implicitly initializes the agent and then you call the `Elastic.Apm.Agent.Setup` method on an already initialized agent.
```
bmorelli25, line 106:
```suggestion
As an example, this issue can happen if you call the `Elastic.Apm.Agent.Setup` method multiple times, or if you call another method on `Elastic.Apm.Agent` that implicitly initializes the agent and then you call the `Elastic.Apm.Agent.Setup` method on the already initialized agent.
```

docs/troubleshooting.asciidoc
gregkalapos, line 100:
```suggestion
It can happen that in a very early stage of the monitored process the agent throws an `InstanceAlreadyCreatedException` exception with the following message: "The singleton APM agent has already been instantiated and can no longer be configured." This happens when the agent would be initialized multiple times, which is prohibited.
```

gregkalapos: Agreed - added it.

docs/public-api.asciidoc
```diff
@@ -10,6 +10,57 @@ customize and manually create spans and transactions,
...
+[[api-initialization]]
+== Initialization
+
+The API does not require explicit Agent initialization—agent initialization is optional. The `Elastic.Apm.Agent.IsConfigured` property lets you check whether the agent is already initialized.
+
+[float]
+[[implicit-initialization]]
+=== Implicit agent initialization
+
+If you don't explicitly initialize the agent, it will be started with a default configuration. This means the Agent will read <<configuration,configuration settings>> from environment variables. If you don't set an environment variable, the Agent will use the default value. For example, the `ServerUrls` default is `http://localhost:8200`.
```
gregkalapos, line 23:
```suggestion
If you don't explicitly initialize the agent, it will be started with a default component setup. This means the Agent will read <<configuration,configuration settings>> from environment variables. If you don't set an environment variable, the Agent will use the default value. For example, the `ServerUrls` default is `http://localhost:8200`.
```

gregkalapos:
> Nice! I have some grammar recommendations. I'd like to take another look before merging if that's ok.

Thanks for the suggestions. Sure - no rush with this.

I added a comment [here](https://github.com/elastic/apm-agent-dotnet/pull/831#discussion_r414388141) - what do you think about that?

Could you please also double check the section levels on this page? I'm not sure it's correct right now. I also see that the website splits some of the `asciidoc` files into multiple "sub pages", which I'd like to avoid here.

Here is how I'd imagine sections here - at least this was my intention:

Everything on a single page:

- Public API
  -  Initialization
      -  Implicit agent initialization
      - Explicit agent initialization
  - Auto instrumentation in combination with the Public Agent API
  - Tracer API
  - (all other section on this level)

gregkalapos:
@TailyFair @ffMathy this PR tries to cover your feedback on the agent initialization issue in the docs. Feel free to chime in, in case you feel this text could be improved.

docs/public-api.asciidoc
bmorelli25: Got it. That makes sense to me 👍

docs/public-api.asciidoc
```diff
@@ -10,6 +10,57 @@ customize and manually create spans and transactions,
...
+
+- `IApmLogger`: A logger implementation that will be used to print log messages. Default: A console logger.
+- `IPayloadSender`: A component that receives all the captured events like spans, transactions, and metrics. The default implementation serializes all events and sends them to the Elastic APM Server
+- `IConfigurationReader`: A component that reads <<configuration, agent configuration settings>>. The default implementation reads configuration through environment variables.
+
+NOTE: In the case of ASP.NET Core, when you register the agent, the `UseElasticApm` and the `UseAllElasticApm` methods both implicitly initialize the agent by calling the `Elastic.Apm.Agent.Setup` method internally. In that setup, the `IConfigurationReader` implementation will read configuration from the ASP.NET Core configuration system in case you pass an `IConfiguration` instance to the method. The `IApmLogger` instance will also log through the configured logging provider by integrating into the ASP.NET Core logging system.
+
+[float]
+[[auto-instrumentation-and-agent-api]]
+=== Auto instrumentation in combination with the Public Agent API
```
bmorelli25, line 48:
```suggestion
== Auto instrumentation in combination with the Public Agent API
```

docs/public-api.asciidoc
```diff
@@ -10,6 +10,57 @@ customize and manually create spans and transactions,
...
+
+Some examples:
+
+- `HttpDiagnosticsSubscriber`: captures HTTP calls through `HttpClient`
+- `EfCoreDiagnosticsSubscriber`: captures database calls through Entity Framework Core
+- `SqlClientDiagnosticSubscriber`: captures database calls through `SqlClient` 
+
 [float]
 [[api-tracer-api]]
 === Tracer API
```
bmorelli25, line 66:
```suggestion
== Tracer API
```

docs/troubleshooting.asciidoc
```diff
@@ -90,3 +90,21 @@ For this you'll need an `IApmLogger` implementation (see above) which you need t
...
+
+Another scenario can be that you use the Public Agent API in combination of the IIS module or the ASP.NET Core NuGet package where you enable the agent with the `UseElasticApm` or `UseAllElasticApm` methods. Both the first call to the IIS module and the `UseElasticApm`/`UseAllElasticApm` methods internally call the `Elastic.Apm.Agent.Setup` method to initialize the agent.
+
+In case you use the Public Agent API with the `Elastic.Apm.Agent` class that can be potentially executed before the IIS module initializes or the `UseElasticApm`/`UseAllElasticApm` calls execute, then those will fail, since you implicitly initialized the agent before.
+
+To prevent the `InstanceAlreadyCreatedException` in those scenarios, in case you have code that can potentially run very early before the agent is initialized, it's recommended to check with the `Elastic.Apm.Agent.IsConfigured` method whether the agent is already initialized and only then use other methods in the Public Agent API.
```
bmorelli25, line 110: I'm still struggling with this troubleshooting section. I attempted a rewrite. Feel free to take some, all, or none:

```
In the early stage of a monitored process, the Agent might throw an `InstanceAlreadyCreatedException` exception with the following message: "The singleton APM agent has already been instantiated and can no longer be configured." This happens when you attempt to initialize the Agent multiple times, which is prohibited. Allowing multiple Agent instances per process would open up problems, like capturing events and metrics multiple times for each instance, or having multiple background threads for event serialization and transfer to the APM Server.

TIP: Take a look at the initialization section of the <<public-api,Public Agent API>> for more information on how agent initialization works.

As an example, this issue can happen if you call the `Elastic.Apm.Agent.Setup` method multiple times, or if you call another method on `Elastic.Apm.Agent` that implicitly initializes the agent, and then you call the `Elastic.Apm.Agent.Setup` method on the already initialized agent.

Another example might be when you use the Public Agent API in combination with the IIS module or the ASP.NET Core NuGet package, where you enable the agent with the `UseElasticApm` or `UseAllElasticApm` methods. Both the first call to the IIS module and the `UseElasticApm`/`UseAllElasticApm` methods internally call the `Elastic.Apm.Agent.Setup` method to initialize the agent.

If you use the Public Agent API with the `Elastic.Apm.Agent` class, there's a chance it can execute before the IIS module initializes or the `UseElasticApm`/`UseAllElasticApm` calls execute. If that happens, those will fail, as the Agent has been implicitly initialized already.

To prevent the `InstanceAlreadyCreatedException` in these scenarios, first use the `Elastic.Apm.Agent.IsConfigured` method to check if the agent is already initialized. After the check, you can safely use other methods in the Public Agent API. This will prevent code from running before the Agent has initialized.
```

gregkalapos: Nice, thanks! 

I took most of it, changed 2 things:

> If you use the Public Agent API with the Elastic.Apm.Agent class, there's a chance it can execute before the IIS module initializes or the UseElasticApm/UseAllElasticApm calls execute. If that happens, those will fail, as the Agent has been implicitly initialized already.

I think the "there's a chance it can execute before..." part is a bit unfortunate so I changed that part. Both the IIS module and the `UseElasticApm/UseAllElasticApm` methods run super early, at a very well defined stage in the application startup. I wanted the text to reflect that this issue happens when users use the public API extremely early in the application process. So this does not happen "by chance" this needs active work and thinking by the user to find a point to run code before the agent initializes :)

(last sentence)
> This will prevent code from running before the Agent has initialized.

Changed to:

> This will prevent accidental implicit agent initialization.

Issue #794, opened 2020-03-27
New central config options: CaptureHeaders, LogLevel, SpanFramesMinDuration, StackTraceLimit


gregkalapos:
These 4 configs are not yet supported by .NET, but they are already [added in Kibana](https://github.com/elastic/kibana/blob/master/x-pack/plugins/apm/common/agent_configuration/setting_definitions/index.test.ts#L128-L138).

Given we already have the infrastructure, let's add these 4 and make them also runtime changeable.

```
capture_headers
log_level
span_frames_min_duration
stack_trace_limit
```

gregkalapos:
Note to future self: Don't forget to update the docs ;)

gregkalapos:
Big progress was made on this by @vhatsura in #820 👏 

Missing part to close this:
`LogLevel` does not work yet.
- We should make sure our loggers read the value that we fetch from central config
- We should also think about how this integrates with .NET Core logging (e.g. do we really want to override the log level defined in `appsettings.json` or elsewhere, etc.)

vhatsura:
@gregkalapos,
(I've tried to resolve it, at least, for console logger, but it instantiates before remote configuration and used by central config fetcher itself. Looks like we need to add a new method on `IApmLogger` interface, smth like `SetLogLevel`. WDYT?
As for
> We should also think about how this integrates with .NET Core logging

I'd say, we don't need to do it, due to log level is defined on common logging configuration.

gregkalapos:
> @gregkalapos,
> (I've tried to resolve it, at least, for console logger, but it instantiates before remote configuration and used by central config fetcher itself. Looks like we need to add a new method on `IApmLogger` interface, smth like `SetLogLevel`. WDYT?
> As for
> 
> > We should also think about how this integrates with .NET Core logging
> 
> I'd say, we don't need to do it, due to log level is defined on common logging configuration.


>  Looks like we need to add a new method on `IApmLogger` interface, smth like `SetLogLevel`. WDYT?

Well, it'd be really nice to avoid it - otherwise all implementors need to implement it - and I know for our IIS module it's common that people redirect logs by implementing `IApmLogger` - for them this would mean a build error. This would be especially less nice if all the benefit would be that our Console Logger uses it, but e.g. in ASP.NET Core we ignore this.

gregkalapos:
With https://github.com/elastic/apm-agent-dotnet/pull/1096 this is done.

PR #840, opened 2020-05-07
Prepare release 1.5

gregkalapos:
- [x] bump version
- [x] create changelog

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -27,15 +27,15 @@ endif::[]
 [float]
 ===== Features
 - Auto instrumentation for `SqlClient` (https://www.elastic.co/guide/en/apm/agent/dotnet/current/setup.html#setup-sqlclient[documentation])
-- Introducing Filter API {pull}#792[#792] (https://www.elastic.co/guide/en/apm/agent/dotnet/current/public-api.html#filter-api[documentation])
-- Auto-detect culprit for exceptions {pull}#740[#740]
+- Introducing Filter API {pull}792[#792] (https://www.elastic.co/guide/en/apm/agent/dotnet/current/public-api.html#filter-api[documentation])
+- Auto-detect culprit for exceptions {pull}740[#740]
 - New config settings: `ExcludedNamespaces`, `ApplicationNamespaces` (https://www.elastic.co/guide/en/apm/agent/dotnet/current/config-all-options-summary.html[documentation])
-- Keep `Activity.Current.TraceId` in sync with the Trace ID used by the agent {pull}#800[#800]
-- Report Kubernetes system metadata {pull}#741[#741]
+- Keep `Activity.Current.TraceId` in sync with the Trace ID used by the agent {pull}800[#800]
+- Report Kubernetes system metadata {pull}741[#741]
 
 [float]
 ===== Bug fixes
-- Database connection string parsing issue with Oracle {pull}#795[#795]
+- Database connection string parsing issue with Oracle {pull}795[#795]
 
 [[release-notes-1.4.0]]
 ==== 1.4.0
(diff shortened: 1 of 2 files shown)


gregkalapos:
I don't really [see in the logs](https://elasticsearch-ci.elastic.co/job/elastic+docs+apm-agent-dotnet+pull-request/1085/console) why this doc build fails. @bmorelli25 could you please take a look? 

I have some link that are not yet on `current`- can that be the problem? Like [here](https://github.com/elastic/apm-agent-dotnet/pull/840/files#diff-634e7b38da1fcb3d58e9db4eaa5653f2R30) - that's only on `master` right now - my plan was to move that to `current` (aka `1.x` branch) once the 1.5 release is out.

bmorelli25:
This is an interesting one. Is there a reason why you're not using attributes to link to content? That would ensure the release notes in each branch only link to content that exists in the same branch. I'll leave a few comments.

CHANGELOG.asciidoc
```diff
@@ -21,6 +21,22 @@ endif::[]
...
 
+[[release-notes-1.5.0]]
+==== 1.5.0
+
+[float]
+===== Features
+- Auto instrumentation for `SqlClient` (https://www.elastic.co/guide/en/apm/agent/dotnet/current/setup.html#setup-sqlclient[documentation])
+- Introducing Filter API {pull}792[#792] (https://www.elastic.co/guide/en/apm/agent/dotnet/current/public-api.html#filter-api[documentation])
+- Auto-detect culprit for exceptions {pull}740[#740]
+- New config settings: `ExcludedNamespaces`, `ApplicationNamespaces` (https://www.elastic.co/guide/en/apm/agent/dotnet/current/config-all-options-summary.html[documentation])
```
bmorelli25, line 32:
```suggestion
- Auto instrumentation for `SqlClient` (<<setup-sqlclient,documentation>>)
- Introducing Filter API {pull}792[#792] (<<filter-api,documentation>>)
- Auto-detect culprit for exceptions {pull}740[#740]
- New config settings: `ExcludedNamespaces`, `ApplicationNamespaces` (<<config-all-options-summary,documentation>>)
```

bmorelli25: This should fix all of the documentation errors as all of this content exists in master. When the release is pushed to 1.x, all of the content and links in the 1.x branch will be updated to point to current instead.

The only downside of this is that the links are not clickable in markdown. I think that's ok though, as neither are the PR links, and we have a header at the top that encourages readers to view RNs in the documentaiton.

gregkalapos:
Thanks @bmorelli25! 

> Is there a reason why you're not using attributes to link to content? 

That was lack of knowledge on my side - I did not know it works also from here; but now I see how I should use this 👍 

> The only downside of this is that the links are not clickable in markdown. I think that's ok though, as neither are the PR links, and we have a header at the top that encourages readers to view RNs in the documentation.

Yes, I agree. I'll also do a GitHub release here, and for that I'll turn those into markdown for [this page](https://github.com/elastic/apm-agent-dotnet/releases) - I think that's where people would look at changelogs if they are on GitHub.

vhatsura:
@gregkalapos, what about #794. Are you going to announce it after solving the issue with `LogLevel`?

gregkalapos:
> @gregkalapos, what about #794. Are you going to announce it after solving the issue with `LogLevel`?

That's a feature which also depends on Kibana, so it mainly targets 7.7+ - these additional configs should be covered in the release announcement of 7.7.

Not that I thought about this a lot, but basically 2 reasons I have not listed:
- yeah, it'd be nice to have some idea about the loglevel
- and I felt as long as 7.7 is not officially GA, listing it would be more confusing.

PR #843, opened 2020-05-11
ci: timeout tweaks and no docker CI builds

v1v:
## What
- Timeout after 2 hours.
- Timeout after 75 minutes in the `Initialising` stage.
- Docker for windows stage removed as it's not supported (will help to reduce the number of infra resources)

## Why

Overall timeout takes precedence over the the step/stage timeout, therefore we need to fine the configuration that could be applied globally and with certain configuration when there is a release on-going


Jenkinsfile
```diff
@@ -17,7 +17,7 @@ pipeline {
     BENCHMARK_SECRET  = 'secret/apm-team/ci/benchmark-cloud'
   }
   options {
-    timeout(time: 75, unit: 'MINUTES')
+    timeout(activity: true, time: 1, unit: 'HOURS')
```
v1v, line 20: `Timeout after 1 hour of no activity in logs for pipeline instead of absolute duration.`

kuisathaverat: if a command is showing an error every 30 min, the job could never end. I am not sure is a good option.

v1v:
@kuisathaverat and @gregkalapos 

>  dead lock 

Good point! https://github.com/elastic/apm-agent-dotnet/pull/843/commits/85534d06362821297d3181809febb287f6ecaeb2 should fix that particular corner case. For such, I did move the release stages from the `Initializing` as siblings , therefore I could apply a specific timeout for the `Initializing` stage and a global one for the whole pipeline.

Issue #844, opened 2020-05-11
Stacktrace with diagnosticsource - user code missing


gregkalapos:
We capture stack traces for spans that are longer than a predefined duration (`SpanFramesMinDuration `). This is also true for auto captured span where we create the span based on DiagnosticSource events.

If async is also involved then the user code is completely missing from the stack. 

E.g. in [this](https://discuss.elastic.co/t/why-is-the-stack-exception-captured-by-the-apm-agent-different-from-the-format-in-the-log) discuss entry we see DiagnosticSource for the HTTP request, lots of framework call, but it's not visible where the application does the outgoing HTTP call. Ideally there should be a method from user code on the stack, so users can know where that HTTP request in their code happened.

gregkalapos:
One option would be probably to capture the callstack on the start event.

In case of `span.duration < SpanFramesMinDuration` this'd mean we captured it unnecessarily, which is not the best for perf.

But:
- HTTP, db, and similar calls will be typically longer than `SpanFramesMinDuration` - the default is 5ms.
- Within one "average" transaction the number of HTTP, Db, and similar calls will be low, so typically there'd be a very few stacktraces that we capture unnecessarily.

gregkalapos:
Done in #845

PR #845, opened 2020-05-11
Capture stack trace on span start for HttpClient

gregkalapos:
Solves #844

Implements https://github.com/elastic/apm-agent-dotnet/issues/844#issuecomment-626821726. 

Stacktrace prior to this PR (caller code in user code is not visible - because we captured the stack in `Span.End()` called during the 2. diagnostic source call):
Screen Shot 2020-05-11 at 22 57 57

Stacktrace after this PR (user code shows up):
Screen Shot 2020-05-11 at 22 58 09

More stack frame improvements coming...

### DIFF
--- a/src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerImplBase.cs
+++ b/src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerImplBase.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
 using Elastic.Apm.Api;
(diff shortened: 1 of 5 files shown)


gregkalapos:
Also tested with EF Core async - it's not an issue there, in that case we can still capture the stack trace in `.End()` and still see the caller code in the user's code base.

wajika:
I need to use this version, how to pull its image?

gregkalapos:
@wajika this is not released yet, but it's merged to `master`. You can download the generated NuGet pacakges from our CI - [here](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-dotnet%2Fapm-agent-dotnet-mbp/detail/master/282/artifacts) are the artifacts of the last master build - those already contain the change from this PR - and some other stack trace related improvements.

Please keep in mind that those NuGet package from our CI are not official releases (our official packages are on nuget.org) - but of course you are free to use them.

Issue #227, opened 2019-05-10
Agent breakdown metrics (AKA Breakdown Graphs)


felixbarny:
Implementation of agent breakdown metrics (https://github.com/elastic/apm/issues/78)

Checklist/TL;DR version of the design:

- [ ] [Calculate self-time of transactions and spans](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.ondan294nbpt)
  - [ ] Prerequisite: Spans have a reference to their parent
  - [ ] Add a child duration reentrant timer to transactions and spans
- [ ] Add [feature flag](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.12wulrccoi9k)
- Metrics registry enhancements
  - [ ] Add support for [timers](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.qk36t41ikda3)
  - [ ] [Limit the max amount of timers](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.tzyh7loxf2is)
  - [ ] Add support for counters
  - [ ] Optional: [guard against concurrent reads/writes](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.aler9ixddp0e)
- [ ] [Track span self-times on the transaction](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.g5pdtefj3ag)
  - [ ] Prerequisite: Spans have a reference to their transaction
  - On span end:
    - [ ] track the span self-time on the corresponding transaction
  - On transaction end
    - [ ] track all self-times on the metric registry
    - [ ] track the transaction's self time with the `span.type` `app`
    - [ ] track the transaction's duration on the metrics registry
- [Reporting metrics](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.gkxkryn9w02s)
  - [ ] Add support for top-level labels for `transaction.name`, `transaction.type`, `span.type` and `span.subtype`
[...]

PR #849, opened 2020-05-13
Calculate self-time of transactions and spans (part of agent breakdown metrics)

vhatsura:
related to #227 

The PR address only the first important part of agent breakdown metrics: calculate self-time of transaction and spans.

I opened PR as early as possible to get feedback about changes under the hood. The main change is in extracting logic from `Span` and `Transaction` classes into the separate `ExecutionSegment` abstract class.

## To be done

In case it makes sense the following changes should be applied to the PR:

* Reanimate `FilterApiTests` due to commenting `JsonConstructor`
* Move methods from `ExecutionSegmentCommon` to `ExecutionSegment` class
* Go through `Span` and `Transaction` and move shared logic to `ExecutionSegment`

## Questions:

* Maybe, it makes sense to create shared `breakdownmetrics` branch for the feature and do each option from the checklist as a separate PR?

/cc @gregkalapos


gregkalapos:
Done in https://github.com/elastic/apm-agent-dotnet/pull/1271

PR #862, opened 2020-05-26
add ability to collect memory from cgroup

### LINKED ISSUE #858: APM Agent on K8s take system memory from host
**Describe the bug**
When using APM agent on Kubernetes cluster, the memory reported in **system.process.memory.size** is incorrect.

To identify this situation whe have the suport of @tanquetav

**To Reproduce**
1. In kibana, search your service metrics:
`service.name : your-service-name and processor.event : metric`
this value is all memory of host.

**Expected behavior**
The correct value of the service.

vhatsura:
closes #858 

@gregkalapos, I want to ask you about the details of `Elastic.Apm.DockerTests`. Do tests from the project run in CI? If so, is it possible to control which parameters tests are executing? I want to cover new logic with tests

### DIFF
--- a/src/Elastic.Apm/Metrics/MetricsProvider/FreeAndTotalMemoryProvider.cs
+++ b/src/Elastic.Apm/Metrics/MetricsProvider/FreeAndTotalMemoryProvider.cs
@@ -35,62 +35,109 @@ namespace Elastic.Apm.Metrics.MetricsProvider
 		{
 			if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
 			{
-				var (success, totalMemory, freeMemory) = GlobalMemoryStatus.GetTotalPhysAndAvailPhys();
-
-				if (!success || totalMemory == 0 || freeMemory == 0)
-					return null;
-
-				var retVal = new List<MetricSample>();
-
-				if (_collectFreeMemory)
-					retVal.Add(new MetricSample(FreeMemory, freeMemory));
-
-				if (_collectTotalMemory)
-					retVal.Add(new MetricSample(TotalMemory, totalMemory));
-
-				return retVal;
+				return GetSamplesForWindows();
 			}
 
 			if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
 			{
-				var retVal = new List<MetricSample>();
+				return GetSamplesForLinux();
+			}
+
+			return null;
+		}
+
+		private IEnumerable<MetricSample> GetSamplesForWindows()
+		{
+			var (success, totalMemory, freeMemory) = GlobalMemoryStatus.GetTotalPhysAndAvailPhys();
+
+			if (!success || totalMemory == 0 || freeMemory == 0)
+				return null;
+
+			var retVal = new List<MetricSample>();
+
+			if (_collectFreeMemory)
+				retVal.Add(new MetricSample(FreeMemory, freeMemory));
+
+			if (_collectTotalMemory)
+				retVal.Add(new MetricSample(TotalMemory, totalMemory));
+
+			return retVal;
+		}
+
+		private IEnumerable<MetricSample> GetSamplesForLinux()
+		{
+			var retVal = new List<MetricSample>();
+
+			var (cGroupUsageInBytes, cGroupLimitInBytes) = GetCGroupMemoryInfo();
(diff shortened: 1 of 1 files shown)


gregkalapos:
Yes, those are also executed in CI - [here](https://apm-ci.elastic.co/job/apm-agent-dotnet/job/apm-agent-dotnet-mbp/job/master/284/testReport/Elastic.Apm.DockerTests/) you can see those.

>  If so, is it possible to control which parameters tests are executing?

Not sure I get this part of the question. If you want to skip a test depending on the environment, [this is](https://github.com/elastic/apm-agent-dotnet/blob/master/test/Elastic.Apm.Tests/NetCoreFact.cs) what we usually do.

What would be the approach here? Is it enough just to make sure the test runs on Linux? If that's the case then probably a `FactAttribute` is a good way to go.

vhatsura:
> Not sure I get this part of the question

Is it possible to run a docker container with tests with additional parameters as `-m 500m` to be cover functionality with memory limits?

gregkalapos:
> > Not sure I get this part of the question
> 
> Is it possible to run a docker container with tests with additional parameters as `-m 500m` to be cover functionality with memory limits?

Oh, now I get it. That would require some changes in the `jenkinsfile`. Currently this just relies on the fact that the test itself will run within a container. [Here](https://github.com/elastic/apm-agent-dotnet/blob/master/Jenkinsfile#L481) is that part. I'm not a jenkins expert, @v1v knows more about this. I assume we could just add a parameter there. But then the next issue will be that currently we use that `def dotnet(Closure body)` part to run all our tests - so probably we should have 1 for just the docker tests which can set this parameter and run everything else as we do now.

vhatsura:
Ok, thanks for your clarification. Looking forward for answer from @v1v. From my side, I'm going to investigate other approaches to run it from the code

v1v:
The docker containers that are used in the CI are only related to the provisioning, and they can be tweaked similar to what it's executed in the command line when running `docker run <flags>`, therefore what @gregkalapos suggested is actually the way I could think about, although I see some downsides regarding how static the test could be. I wonder if the test-containers approach that you worked with in https://github.com/elastic/apm-agent-dotnet/pull/592 could be the way to programmatically assign the memory allocation for each test?

gregkalapos:
The same PR for the Java agent is here: https://github.com/elastic/apm-agent-java/pull/1211 

#### Testing
I discussed this with Java Agent devs and we tend to agree that for now it's fine to test this with sample files (just read them and assert that the parsed values are fine) - there are some already in the Java PR. So for now I'd not start adding a CI feature here to run this in containers with specific settings because that would increase the effort here. If we do it, maybe it'd be better to do it in the integration test repo and run it for all agents - that way with the same effort we'd cover all agents.

#### Key of this metric 
The other blocker here is that we currently have these metrics under `system.` and in case we send values grom cgroup under keys with `system.` we'd change the semantics of the metrics (these are ecs defined if I'm not mistaken). The goal is to find a better key for these values and use that key across all agents. Unfortunately we don't have an agreement on the new key yet.

#### Metricbeat
Also, another thing that came up if that maybe Metricbeat should read these values long term - not final decision on that yet, but we try to avoid having the same code in multiple components and this thing seems to be yet another implementation of the same thing which may lead to the decision to not to do this in APM agents - again, no final decision on this, just saying this upfront that this could be a consideration.

vhatsura:
@gregkalapos, thanks a lot for your explanation. 
Some comments based on new information.

> these are ecs defined if I'm not mistaken

I'm not sure, but as I understand and look into the right place, [ECS doesn't have the schema for `system`](https://github.com/elastic/ecs/tree/master/schemas).

> The other blocker here is that we currently have these metrics under system. and in case we send values grom cgroup under keys with system. we'd change the semantics of the metrics

It would be nice to know the exact semantics of the metrics due to no detailed description of it in the documentation: `system.memory.total` - Total memory, `system.memory.actual.free` - Actual free memory. So, in such case, it would be nice to update documentation, because (I'm not a native speaker), but `system` can mean various things:

> a group of related hardware units or programs or both, especially when dedicated to a single application.

So, k8s pod can be interpreted as a system, as well as the whole k8s cluster. However, for me, as a developer, limits of the system which have a direct influence on my application are good to know.

PR #869, opened 2020-06-04
Fix CI - add timestamp to packages we push to feedz.io

gregkalapos:
Follow up on https://github.com/elastic/apm-agent-dotnet/pull/864 - we overlooked a small thing there.

Currently master fails when we publish to feedz.io with:

```
[2020-06-03T22:35:52.167Z] + .ci/linux/deploy.sh ******** ********
[2020-06-03T22:35:52.430Z] dotnet nuget push ./src/Elastic.Apm/bin/Release/Elastic.Apm.1.5.1.nupkg
[2020-06-03T22:35:53.817Z] info : Pushing Elastic.Apm.1.5.1.nupkg to 'https://f.feedz.io/elastic/all/nuget'...
[2020-06-03T22:35:53.817Z] info :   PUT https://f.feedz.io/elastic/all/nuget/
[2020-06-03T22:35:54.391Z] info :   Conflict https://f.feedz.io/elastic/all/nuget/ 627ms
[2020-06-03T22:35:54.391Z] info : To skip already published packages, use the option --skip-duplicate
[2020-06-03T22:35:54.391Z] error: Response status code does not indicate success: 409 (Conflict).
script returned exit code 1
```

This is because the NuGet package version is static and we can't push a package with the same name and version multiple times. For pushing to nuget.org we always increment the version manually for the release. But the plan is to push every master build to feedz.io, so that approach won't work with feedz.io.

Here is my idea to solve this:
- I'd keep the manual increment approach for nuget.org, I think that's a safe way, we really don't want to accidentally push there.
- For feedz.io all packages will be suffixed with `-alpha-$timestamp`. That way we 1) make clear that those are pre-release packages and 2) we never have version collision. Example for a generated file: `Elastic.Apm.1.5.1-alpha-20200604-100605.nupkg`. 

One nit, but I think this is fine: currently for a "real release" (meaning when we increment the version number and push to nuget.org) we'd push both to feedz.io with `-alpha-timestamp` and to nuget.org with the official version number.  But I think this is fine, I'd not really stress about this.


.ci/linux/releaseWithTimeStamp.sh
```diff
@@ -0,0 +1,12 @@
...
+# This script packes the project as a release without the
+# sample projects
+#
+set -euxo pipefail
+
+# Remove sample projects - and other we don't want to pack
+dotnet sln remove test/Elastic.Apm.PerfTests/Elastic.Apm.PerfTests.csproj
+dotnet sln remove test/Elastic.Apm.AspNetFullFramework.Tests/Elastic.Apm.AspNetFullFramework.Tests.csproj
+dotnet sln remove sample/AspNetFullFrameworkSampleApp/AspNetFullFrameworkSampleApp.csproj
+dotnet pack --version-suffix alpha-`date -u +"%Y%m%d-%H%M%S"` -c Release
```
gregkalapos, line 12: This file is the same as `release.sh`, except we add the parameter 

```
--version-suffix alpha-`date -u +"%Y%m%d-%H%M%S"`
```

If there is a nicer way to do this (e.g. removing this file, and have only `release.sh` and add the `--version-suffix` based on an input parameter), that'd be very nice. So feel free to improve this.

vhatsura:
Hey, @gregkalapos.

I believe it's worth to look into [GitVersion](https://github.com/GitTools/GitVersion).
It adds auto-incrementing version on each commit and set the right Version attributes to package for manual tags also.

gregkalapos:
@vhatsura - it looks very nice. Nevertheless I'd try to keep this simple. If we go with that I feel we'd need to change a whole bunch of other thing. E.g integration tests also fetch the agent - so we need to make sure that that part is working and the integration tests can deal with the generated versions.

gregkalapos:
Failing integration test is due to [this](https://github.com/elastic/apm-integration-testing/blob/5e705162b250a9ab959978bc038f58a94b387b51/docker/opbeans/dotnet/run.sh#L43) line. I’ll open a PR for that in the integration-test repo.

.ci/linux/releaseWithTimeStamp.sh
v1v: Let me tweak the script to use a parameter and apply the same for the Jenkinsfile :)

v1v: See https://github.com/gregkalapos/apm-agent-dotnet/pull/9

gregkalapos: Nice, thanks @v1v.

Issue #883, opened 2020-06-18
Parent.id is set to a non-Elastic APM Span/Transaction


gregkalapos:
Found in Opbeans and also [here](https://discuss.elastic.co/t/nothing-to-see-in-traces/236917).

One example of such an id: `|eb88cbba-4040e4007bc3c335.`.

My quick slack analysis:

```
that id looks like an id coming from Activity https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.activity?view=netcore-3.1
Activity is a “tracing library” build into .NET (fun fact the C# OpenTelemetry SIG decided  to take this API and turn it into an  OpenTelemetry compatible API - so think it’s like a span) -  anyways, so what we do in the agent is that we try to keep our trace id in sync with  the traceid on Activity (that way if users look at Activity.Current.TraceId then it’ll be the same as our trace id in elastic). Problem is Activity pre-dates W3C TraceContext and - originally it had a hierarchical id format (that’s what you pasted), but now it also supports w3c format. We only keep the id in sync when  it’s w3c format: https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm/Model/Transaction.cs#L85
At least that’s my intention. I can see 2 options here: Either I messed up and somewhere a check is missing or it’s in W3C format, but it’s parent is not, so the check is ok, but still we set the parent id which is in the old format - not sure it’s possible.
I’ll dig deeper - I still just start my day here. But thanks for mentioning this! that id looks like an id coming from Activity https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.activity?view=netcore-3.1
Activity is a “tracing library” build into .NET (fun fact the C# OpenTelemetry SIG decided  to take this API and turn it into an  OpenTelemetry compatible API - so think it’s like a span) -  anyways, so what we do in the agent is that we try to keep our trace id in sync with  the traceid on Activity (that way if users look at Activity.Current.TraceId then it’ll be the same as our trace id in elastic). Problem is Activity pre-dates W3C TraceContext and - originally it had a hierarchical id format (that’s what you pasted), but now it also supports w3c format. We only keep the id in sync when  it’s w3c format: https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm/Model/Transaction.cs#L85
[...]

PR #888, opened 2020-06-26
Ignore Activity.Current.ParentId if no traceparent header is present

gregkalapos:
Addressing #883

### Description of the problem this PR solves

In our UI every trace must start with a transaction where the `parentid` is `null` - if that's not the case then it's not considered as a trace, therefore under "traces" in kibana, the given trace won't show up.

### Activity Parent id in the 1. service of a chain.

We keep trace id and also parentid in sync with [`System.Diagnostics.Activity`](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.activity?view=netcore-3.1), and ASP.NET Core itself already creates an activity - also in case this is the 1. service in a chain. Prior to this PR we did this when we created the transaction: `ParentId = Activity.Current.ParentId;` that's why there was never a transaction with `null` trace id. So in case the agent is in the 1. service of a chain, this PR will fix the the problem described above.

### Context propagation in ASP.NET Core

The other thing is context propagation in ASP.NET Core described [here](https://devblogs.microsoft.com/aspnet/improvements-in-net-core-3-0-for-troubleshooting-and-monitoring-distributed-apps/). Since .NET Core and ASP.NET Core 3.0 the framework itself creates traceids and propagates those across HTTP calls. By default it uses the `Request-Id` header, so our agent will ignore it. With this in case A -> B (A calls B) in B we will have an Activity with a prentId which points to an Activity in A - so far so good, no issue for us. But users can opt-in to use the W3C TraceContext standard and in that case the context propagation happens in the `traceparent` header, which our agent will read. Once users opt-in, and let's say B is monitored by us but A isn't then we'll have the issue described above, because our first transaction will have a parent id in this case pointing to an Activity from A. There is no way to know this when we start the transaction in B and this is a more general problem and I'd argue a bug in our TraceContext support. This PR does not solve this issue - This problem is discussed here: https://github.com/elastic/apm/issues/286. I'd like to emphasis that the usage of the `traceparent` header is opt-in; so this only happens if users actively set `Activity.DefaultIdFormat = ActivityIdFormat.W3C;` - I expect this to be a fairly uncommon case today.

### DIFF
--- a/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
+++ b/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
@@ -113,17 +113,14 @@ namespace Elastic.Apm.AspNetCore
 								"Incoming request with invalid {TraceParentHeaderName} header (received value: {TraceParentHeaderValue}). Starting trace with new trace id.",
 								TraceContext.TraceParentHeaderNamePrefixed, headerValue);
 
-						// In case there is no traceparent header, we suppress activity reuse, this way the transaction.parentID WON't be set
-						// to the activity parent id if the activity from ASP.NET Core.
-						transaction = _tracer.StartTransactionInternal(transactionName, ApiConstants.TypeRequest, suppressActivityReuse: true);
+						transaction = _tracer.StartTransactionInternal(transactionName, ApiConstants.TypeRequest);
 					}
 				}
 				else
 				{
 					_logger.Debug()?.Log("Incoming request. Starting Trace.");
 
-					// Same suppressActivityReuse: true as above
-					transaction = _tracer.StartTransactionInternal(transactionName, ApiConstants.TypeRequest, suppressActivityReuse: true);
+					transaction = _tracer.StartTransactionInternal(transactionName, ApiConstants.TypeRequest);
 				}
 
 				return transaction;
--- a/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
+++ b/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
@@ -119,7 +119,6 @@ namespace Elastic.Apm.AspNetCore
 				else
 				{
 					_logger.Debug()?.Log("Incoming request. Starting Trace.");
-
 					transaction = _tracer.StartTransactionInternal(transactionName, ApiConstants.TypeRequest);
 				}
 
--- a/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
+++ b/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
@@ -113,13 +113,17 @@ namespace Elastic.Apm.AspNetCore
 								"Incoming request with invalid {TraceParentHeaderName} header (received value: {TraceParentHeaderValue}). Starting trace with new trace id.",
 								TraceContext.TraceParentHeaderNamePrefixed, headerValue);
 
-						transaction = _tracer.StartTransactionInternal(transactionName, ApiConstants.TypeRequest);
+						// In case there is no traceparent header, we suppress activity reuse, this way the transaction.parentID WON't be set
+						// to the activity parent id if the activity from ASP.NET Core.
+						transaction = _tracer.StartTransactionInternal(transactionName, ApiConstants.TypeRequest, suppressActivityReuse: true);
 					}
 				}
 				else
 				{
 					_logger.Debug()?.Log("Incoming request. Starting Trace.");
-					transaction = _tracer.StartTransactionInternal(transactionName, ApiConstants.TypeRequest);
+
+					// Same suppressActivityReuse: true as above
+					transaction = _tracer.StartTransactionInternal(transactionName, ApiConstants.TypeRequest, suppressActivityReuse: true);
 				}
 
 				return transaction;
--- a/src/Elastic.Apm/Api/Tracer.cs
+++ b/src/Elastic.Apm/Api/Tracer.cs
@@ -48,11 +48,11 @@ namespace Elastic.Apm.Api
 		public ITransaction StartTransaction(string name, string type, DistributedTracingData distributedTracingData = null) =>
 			StartTransactionInternal(name, type, distributedTracingData);
 
-		internal Transaction StartTransactionInternal(string name, string type, DistributedTracingData distributedTracingData = null, bool suppressActivityReuse = false)
+		internal Transaction StartTransactionInternal(string name, string type, DistributedTracingData distributedTracingData = null)
 		{
 			var currentConfig = _configProvider.CurrentSnapshot;
 			var retVal = new Transaction(_logger, name, type, new Sampler(currentConfig.TransactionSampleRate), distributedTracingData
-				, _sender, currentConfig, CurrentExecutionSegmentsContainer, suppressActivityReuse)
+				, _sender, currentConfig, CurrentExecutionSegmentsContainer)
 			{ Service = _service };
 
 			_logger.Debug()?.Log("Starting {TransactionValue}", retVal);
--- a/src/Elastic.Apm/Api/Tracer.cs
+++ b/src/Elastic.Apm/Api/Tracer.cs
@@ -48,11 +48,11 @@ namespace Elastic.Apm.Api
 		public ITransaction StartTransaction(string name, string type, DistributedTracingData distributedTracingData = null) =>
 			StartTransactionInternal(name, type, distributedTracingData);
 
-		internal Transaction StartTransactionInternal(string name, string type, DistributedTracingData distributedTracingData = null)
+		internal Transaction StartTransactionInternal(string name, string type, DistributedTracingData distributedTracingData = null, bool suppressActivityReuse = false)
 		{
 			var currentConfig = _configProvider.CurrentSnapshot;
 			var retVal = new Transaction(_logger, name, type, new Sampler(currentConfig.TransactionSampleRate), distributedTracingData
-				, _sender, currentConfig, CurrentExecutionSegmentsContainer)
+				, _sender, currentConfig, CurrentExecutionSegmentsContainer, suppressActivityReuse)
 			{ Service = _service };
 
 			_logger.Debug()?.Log("Starting {TransactionValue}", retVal);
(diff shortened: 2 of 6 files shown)


gregkalapos:
Responding here to @axw from https://github.com/elastic/apm/issues/286 with more .NET specific details

>> Reason I open this issue is that in .NET the situation is more-common, you don't even need another tool for this to happen: In ASP.NET Core, since version 3, there is built-in trace propagation support (details here). In short: the framework itself creates a traceid and it propagates it through HTTP calls. By default it uses a header called Request-Id, which is ok, but it has opt-in support for TraceContext and users can turn on context propagation through the traceparent header. That way in our previous example if A is not monitored at all, but opted-in and B is monitored by Elastic APM, we have the same problem.

> Looking at the agent code a bit, it seems that when starting a Transaction we're creating a new Activity and setting the ASP.NET Core-created Activity as its parent. Is that right? It seems to me that the Activity created by ASP.NET Core should be used as the basis for Transactions, rather than as the parent.

Right, when we create a `Transaction` we also create an `Activity`. The reason here is the following: we want to keep our traceid and the `Activity.Current.TraceId` the same (reason: some Microsoft samples already print `Activity.Current.TraceId` - I think it's very valuable if that value is the same as our `traceid`). In case there is no activity when we start the transaction, on order to make sure we keep those 2 in sync we start an Activity and reuse its `TraceId` which will flow to all child activities created by other components - this can easily happen in a console app. But you are right, if there is already an `Activity` we could just reuse it. My thinking with always starting an activity was that it's more consistent - activities have names and users can observe them; I'd find it strange that we fire an activity for our transactions when there is none, but we don't when there is already one. Overall whether we fire a new Activity when there is already one from ASP.NET Core or not, is I think not a super key point; if we feel it'd be better to not create one in that case, we can change it. For now I'd keep the existing behaviour.

> i.e. if there's an Activity, don't create one, just use it: set the transaction ID to `Activity.Current.SpanId`, parent ID to `Activity.Current.ParentId`, trace ID to `Activity.Current.TraceId`. Also, if `Activity.Parent == null` (root activity), make a sampling decision and update the trace context flags.


- `Activity.Current.ParentId`:  `Activity` supports 2 id formats - the old hierarchical one (the ones starting with `|`) and the W3C TraceContext format. `Activity.Current.IdFormat` let's us check the format, but one thing I noticed is that `Activity.Current.SpanId` can be in W3C format, but the `Activity.Current.ParentId` could be in the old hierarchical format, which I feel we should always ignore. I managed to reproduce this, which makes me 😱  So, to be very honest I don't 100% trust just reading this ids.  Also, even if we reuse `Activity.Current.ParentId` that can still point to something outside Elastic APM - so from the **A** -> **B** example, if the agent is only in **B** then `Activity.Current.ParentId` would point to the `Activity` from **A**. And with that we have the same problem. My latest thinking on this was that we should never set `Activity.Current.ParentId`  to our transactions, because this can lead to the UI problem described in  https://github.com/elastic/apm/issues/286 (Of course we still set `Transaction.ParentId` in case the values comes from the `traceparent` header)
- Regarding setting `Activity.Current.TraceId`: I think this is a key point - so, this value is the same on all `Activity` instances - this flows through all child Activities. If we create a new Activity, this'll still be the same as the one on the `Activity` from ASP.NET Core.
- Setting `Activity.Current.SpanId`: that's a good point, we can do that. I'll adapt the PR

So, overall, for keeping the current behaviour I'd still always start a new Activity when we create a transaction and I'd not set the `prentId` to make sure we never run into the UI problem. Happy to elaborate further.

gregkalapos:
An update on reusing `Activity.Current.SpanId`:

There is a little bit of an issue with this, which is that this id is not compatible with our sampling decision logic. For sampling we rely on the fact that the transaction id has normal distribution for the 8-byte id, which does not seem to be the case for `Activity.Current.SpanId`, here is what it does:


```
public unsafe static ActivitySpanId CreateRandom()
{
   ulong id;
   ActivityTraceId.SetToRandomBytes(new Span<byte>(&id, sizeof(ulong)));
   return new ActivitySpanId(ActivityTraceId.SpanToHexString(new ReadOnlySpan<byte>(&id, sizeof(ulong))));
}

internal unsafe static void SetToRandomBytes(Span<byte> outBytes)
{
    Debug.Assert(outBytes.Length <= sizeof(Guid));     // Guid is 16 bytes, and so is TraceId 
    Guid guid = Guid.NewGuid();
    ReadOnlySpan<byte> guidBytes = new ReadOnlySpan<byte>(&guid, sizeof(Guid));
    guidBytes.Slice(0, outBytes.Length).CopyTo(outBytes);
}
```

So it generates a GUID and slices the first 8 bytes out and does some processing on it which will be the `SpanId`. This does not sounds like normal distribution and if we base our sampling decision on these ids, then our sampling tests fail.

gregkalapos:
> An update on reusing `Activity.Current.SpanId`:
> 
> There is a little bit of an issue with this, which is that this id is not compatible with our sampling decision logic. For sampling we rely on the fact that the transaction id has normal distribution for the 8-byte id, which does not seem to be the case for `Activity.Current.SpanId`, here is what it does:
> 
> ```
> public unsafe static ActivitySpanId CreateRandom()
> {
>    ulong id;
>    ActivityTraceId.SetToRandomBytes(new Span<byte>(&id, sizeof(ulong)));
>    return new ActivitySpanId(ActivityTraceId.SpanToHexString(new ReadOnlySpan<byte>(&id, sizeof(ulong))));
> }
> 
> internal unsafe static void SetToRandomBytes(Span<byte> outBytes)
> {
>     Debug.Assert(outBytes.Length <= sizeof(Guid));     // Guid is 16 bytes, and so is TraceId 
>     Guid guid = Guid.NewGuid();
>     ReadOnlySpan<byte> guidBytes = new ReadOnlySpan<byte>(&guid, sizeof(Guid));
>     guidBytes.Slice(0, outBytes.Length).CopyTo(outBytes);
> }
> ```
> 
> So it generates a GUID and slices the first 8 bytes out and does some processing on it which will be the `SpanId`. This does not sounds like normal distribution and if we base our sampling decision on these ids, then our sampling tests fail.

After slacking other team members, turns out this is actually very much aligned with W3C TraceContext. So I changed the sampling to do it similarly as the Java agent does it [here](https://github.com/elastic/apm-agent-java/blob/master/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/sampling/ProbabilitySampler.java#L74).

axw:
Re sampling decision: I think you have a good point about using an externally-provided SpanId for the sampling decision. As mentioned in https://stackoverflow.com/questions/917617/is-using-a-guid-a-valid-way-to-generate-a-random-string-of-characters-and-numbers, there's no guarantees about how uniformly-random GUIDs are. It might be safe to assume; I don't know. I'd personally only use them in this way if I knew for a fact that they were generated with a uniform distribution.

Re the rest:

I'm a bit confused, so apologies if I've led you astray...

When a `traceparent` (or `elasticapm-traceparent`) header is received we ignore `Activity.Current` (`distributedTracingData != nil` case), and just use values from the header. Is that correct?

Do we care about `Activity.Current` at all when there's no `traceparent` header (`distributedTracingData == nil` case)? I wonder if we should just _always_ ignore it when creating a transaction, including its TraceId.

gregkalapos:
1 short comment on using the Guid ids for sampling: it makes our tests pass.


> When a `traceparent` (or `elasticapm-traceparent`) header is received we ignore `Activity.Current` (`distributedTracingData != nil` case), and just use values from the header. Is that correct?

That is correct. The reason here is that `Activity.Current` does not always have ids we need - e.g. ASP.NET Core versions that are older than 3.0 or in ASP.NET Classic the framework does not create an `Activity` and even in default setup with new ASP.NET Core versions `Activity` uses an id format which is not compatible with W3C TraceContext (W3C is currently opt-in). I wanted to be on the safe side and I thought the best is to just manually read the `traceparent` ourselves and apply it - that works always and easy to test and in case `Activity` applies the same `traceparent` we are in-sync anyways (it'll have the same TraceId). 

Reading `Activity.Current` in case it applies `traceparent` and uses W3C TraceContext compatible ids could work and would be better for perf to not read `traceparent` also in the Agent, but I don't feel confident at this point to rely on `Activity` and I also feel in the majority of the cases we'd need to fall back to read the `traceparent` header anyways, because `Activity.Current` will be either null  (ASP.NET Classic and ASP.NET Core pre 3.0) or not using the same id format (ASP.NET Core post 3.0 with default setup not opting-in to use W3C).

> Do we care about `Activity.Current` at all when there's no `traceparent` header (`distributedTracingData == nil` case)? I wonder if we should just _always_ ignore it when creating a transaction, including its TraceId.

I'd say we should care about `Activity.Current` if possible, because printing `Activity.Current.TraceId` for diagnostics in .NET samples is common, so I think it's good to have a best effort approach to keep that in-sync with our traceid. Also, this whole thing started out by an issue opened by a user. If we don't care about `Activity.Current` and its TraceId then we loose this feature. I feel that'd be sad, keeping traceIds in-sync makes us better integrated with the framework.

axw:
> I'd say we should care about Activity.Current if possible, because printing Activity.Current.TraceId for diagnostics in .NET samples is common, so I think it's good to have a best effort approach to keep that in-sync with our traceid. Also, this whole thing started out by an issue opened by a user. If we don't care about Activity.Current and its TraceId then we loose this feature. I feel that'd be sad, keeping traceIds in-sync makes us better integrated with the framework.

What I was thinking was like this: when starting a transaction, call SetParentId to set its TraceId and ParentId to randomly generated values. That way _after_ calling Activity.Start, the Activity's TraceId and ParentId are in sync with the transaction. You might also set the transaction's ID based on the Activity.SpanId after calling Start, to keep SpanId in sync too.

gregkalapos:
Talked to @axw 

- Problem with calling `SetParentId` to set `TraceId` is that there can already be an `Activity.Current` with a TraceId e.g. by starting an `Activity` manually (or by some other framework) before our transaction is started and that'd mean we'd have different TraceIds in Elastic and on those Activities prior to our Transaction start. Another example that can happen is that e.g. ASP.NET Core creates an `Activity`, prints it with `TraceId` (e.g. due to some error) and if we ignore that then we'd create another `TaceId`. So we can't really ignore `Activity.Current`
- Another one that came up is that when HTTP request comes in with `elastic-apm-traceparent` then `Activity.Current.TraceId` still can diverge from our traceid. For now we accept this and not try to fix; due to RUM `traceparent` (without prefix) should be already the default and long term we hopefully get rid of the prefixed header.

Merging now.

PR #891, opened 2020-06-30
[ci] opbeans requires tags with format v<major>.<minor>.<patch>

v1v:
apm-agent-dotnet uses a different convention for the release tags compared to the opbeans one.

Besides, I added the option to run the benchmakr for the tag release in the apm-agent-dotnet since it uses some obsoleted tag convention.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -341,7 +341,7 @@ pipeline {
                 allOf {
                   anyOf {
                     branch 'master'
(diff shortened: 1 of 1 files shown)


v1v:
Unrelated test failure. I'll merge it now

gregkalapos:
> As far as I see there is an unrelated test failure

Yes, it was definitely unrelated.

Issue #650, opened 2019-12-10
Add span.context.destination.* fields used by service-map feature


SergeyKleyman:
Derived from https://github.com/elastic/apm/issues/180

gregkalapos:
Implemented in #893.

PR #893, opened 2020-07-01
Add span.context.destination.service.* 

gregkalapos:
Solves #650. This is needed for service map. Implements https://github.com/elastic/apm/issues/180

- Adds 3 new fields in `span.context.destination.service`
   - `Resource`
   - `Name`
   - `Type`
 
Also set `span.context.destination` in Elastic.Apm.Elasticsearch.

Includes tests for: 
- Outgoing HTTP
- SqlClient (which does not set `span.context.destination` as defined in https://github.com/elastic/apm/issues/180) 
- MSSQL
- Elasticsearch


gregkalapos:
@Mpdreamz could you take a quick look, especially at the changes in `Elastic.Apm.Elasticsearch`? We need to fill `destination.address` and `destination.port` - my intention is to always set those to the corresponding node.

Mpdreamz:
Late but.LGTM

Issue #688, opened 2020-01-14
Add config TRANSACTION_IGNORE_URLS


gregkalapos:
From https://github.com/elastic/apm/issues/144 

This setting does not exist in .NET yet, so we should just go with the agreement from https://github.com/elastic/apm/issues/144.

gregkalapos:
Implemented in #904.

PR #904, opened 2020-07-15
Add TransactionIgnoreUrls setting

gregkalapos:
Solves https://github.com/elastic/apm-agent-dotnet/issues/688, implements https://github.com/elastic/apm/issues/144

The `TransactionIgnoreUrls` config takes a list and matches each url path against it. If the given url path is in the list, then it'll be ignored - meaning no transaction will be captured.

It checks the path part of the url. E.g. adding `/home/index` will match all of the followings:
- `https://www.myCoolSite.com/home/index`
- `http://localhost/home/index`
- `http://whatever.com/home/index?value1=123` (so query string is ignored for matching)

By default it's case insensitive.

ToDo:
- [x] Apply this also in ASP.NET Classic
- [x] More tests 
- [x] Think about custom transactions and how to apply it there

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -672,6 +672,34 @@ the agent will capture request and response headers, including cookies.
 NOTE: Setting this to `false` reduces memory allocations, network bandwidth and disk space used by Elasticsearch.
 
 [float]
+[[config-transaction-ignore-urls]]
+==== `TransactionIgnoreUrls` (performance)
+
+[options="header"]
+|============
+| Environment variable name             | IConfiguration or Web.config key
+| `ELASTIC_APM_TRANSACTION_IGNORE_URLS` | `ElasticApm:TransactionIgnoreUrls`
+|============
+
+[options="header"]
+|============
+| Default                                                                                                                    | Type
+| `/VAADIN/*, /heartbeat*, /favicon.ico", *.js", *.css", *.jpg", *.jpeg", *.png", *.gif", *.webp", *.svg", *.woff", *.woff2` | List<string>
+|============
+
+Used to restrict requests to certain URLs from being instrumented.
+
+This property should be set to a list containing one or more strings.
+When an incoming HTTP request is detected, its request path will be tested against each element in this list.
+
+This option supports the wildcard `*`, which matches zero or more characters.
+Examples: `/foo/*/bar/*/baz*`, `*foo*`.
+Matching is case insensitive by default.
+Prepending an element with `(?-i)` makes the matching case sensitive.
+
+NOTE: All errors that are captured during a request to an ignored URL are still sent to the APM Server regardless of this setting.
+
+[float]
 [[config-use-elastic-apm-traceparent-header]]
 ==== `UseElasticTraceparentHeader` (added[1.3.0])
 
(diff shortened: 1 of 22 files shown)


docs/configuration.asciidoc
```diff
@@ -671,6 +671,34 @@ the agent will capture request and response headers, including cookies.
...
+[options="header"]
+|============
+| Default                                                                                                                    | Type
+| `/VAADIN/*, /heartbeat*, /favicon.ico", *.js", *.css", *.jpg", *.jpeg", *.png", *.gif", *.webp", *.svg", *.woff", *.woff2` | List<string>
+|============
+
+Used to restrict requests to certain URLs from being instrumented.
+
+This property should be set to a list containing one or more strings.
+When an incoming HTTP request is detected, its request path will be tested against each element in this list.
```
gregkalapos, line 693: @bmorelli25 I think this part could be improved a lot. In the issue description I write about what part of the URL is matched, but I think this text does not make it clear.

Can you come up with something better? I'd like to have something that describes very well what part of the URL is exactly matched - I think most docs just call this part 'path`, but I'm not sure its meaning is generally known.

bmorelli25: You know, I actually don't find this difficult to understand. With that being said, an example here certainly wouldn't hurt and would go a long way to making this feature easier to understand. What about adding what you have in the description to these docs?

```
Used to ...

This property should be set to a list containing one or more strings.
When an incoming HTTP request is detected, its request path will be tested against each element in this list.
For example, adding `/home/index` to this list would match and remove instrumentation from the following URLs:

[source,txt]
----
https://www.mycoolsite.com/home/index
http://localhost/home/index
http://whatever.com/home/index?value1=123
----

This option supports ...
```

gregkalapos: Great; I also think adding an example helps. I added it. Thanks!

arivasvera:
Hi @gregkalapos ,

Is this feature working?

I'm using `Elastic.Apm.NetCoreAll` version `1.6.1` and in my `appsettings.json` I have this:
```
"ElasticApm": {
    "SecretToken": "MY_TOKEN",
    "ServerUrls": "https://example.com",
    "ServiceName": "my-app",
    "TransactionIgnoreUrls": ["/api/healthcheck"]
  }
```

I have my Startup like this:
```
using Elastic.Apm.AspNetCore;
using Elastic.Apm.DiagnosticSource;
using Elastic.Apm.EntityFrameworkCore;

public class Startup
{
  public void Configure(IApplicationBuilder app, IHostingEnvironment env)
  {
    app.UseElasticApm(Configuration, new HttpDiagnosticsSubscriber(), new EfCoreDiagnosticsSubscriber());
    //…rest of the method
  }
  //…rest of the class
}
```

but I'm still getting transactions when `http://mysite.com/api/healthcheck` is requested

am I doing something wrong?

Thanks in advance

bmorelli25:
@arivasvera, I don't think this commit made it into version `1.6.1`. It looks like it was merged a day after the release.

arivasvera:
Thank you for your response @bmorelli25 

I hope this feature will release soon :) 

Best regards

arivasvera:
hello again,

do you guys have an estimated date to release this feature?

again, thanks in advance

PBarnard:
Has this been released yet? I don't see it anywhere on the releases page - https://github.com/elastic/apm-agent-dotnet/releases

gregkalapos:
Yes, this was part of the `1.7.0` release - so this setting is out and can be used.

[This doc page](https://www.elastic.co/guide/en/apm/agent/dotnet/current/config-all-options-summary.html) has all the settings from the latest release.

Regarding questions to release date: We don't publicly communicate specific release dates. Simply because we believe it'd cause more harm than it'd help people. So we'll never be able to answer questions around release date, sorry about that.

PR #911, opened 2020-07-21
Allow to load configuration from another configuration reader

kodigo35:
Gives the abilty to load APM settings from external service instead of using default appsettings keys.


kodigo35:
CLA updated

PR #912, opened 2020-07-21
Allow to load configuration from another configuration reade

kodigo35:
Gives the abilty to load APM settings from external service instead of using default appsettings keys.


kodigo35:
hi @gregkalapos 

i signed the CLA few minutes ago, can you check if it is good ?

i will add some tests asap to cover the PR.

for the doc do you want me to complete it ?

Regards

gregkalapos:
> i signed the CLA few minutes ago, can you check if it is good ?

Thank, yes, it look ok now 👍 

> i will add some tests asap to cover the PR.

Perfect, just ping me if you have any question. I think there isn't really a good place to put a test covering this - feel free to create a small project under the `test` folder for it.

> for the doc do you want me to complete it ?

Don't worry about it now, we have doc writers, I'll ping them and draft something together once we also have tests.

src/Elastic.Apm/Config/ConfigConsts.cs
```diff
@@ -132,6 +132,7 @@ public static class EnvVarNames
 			public const string ExcludedNamespaces = Prefix + "EXCLUDED_NAMESPACES";
 			public const string ApplicationNamespaces = Prefix + "APPLICATION_NAMESPACES";
 			public static string TransactionIgnoreUrls = Prefix + "TRANSACTION_IGNORE_URLS";
+      public const string ConfigurationReaderType = Prefix + "CONFIGURATION_READER_TYPE";
```
vhatsura, line 135: nit: formatting issue

src/Elastic.Apm/Config/ConfigConsts.cs
```diff
@@ -164,6 +165,7 @@ public static class KeyNames
 			public const string ExcludedNamespaces = "ElasticApm:ExcludedNamespaces";
 			public const string ApplicationNamespaces = "ElasticApm:ApplicationNamespaces";
 			public static string TransactionIgnoreUrls = "ElasticApm:TransactionIgnoreUrls";
+      public const string ConfigurationReaderType = "ElasticApm:ConfigurationReaderType";
```
vhatsura, line 168: nit: formatting issue (alignment)

src/Elastic.Apm/Helpers/ConfigHelper.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Configuration;
+using System.Text;
+using Elastic.Apm.Config;
+using Elastic.Apm.Logging;
+
+namespace Elastic.Apm.Helpers
+{
+	public class ConfigHelper
```
vhatsura, line 10: As I see, it's called only from `src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs`. Maybe it's better to store the class also in `Elastic.Apm.AspNetFullFramework` project. What do you think? It will allow not to add `System.Configuration` dependency which makes sense only for .NET Framework applications.

src/Elastic.Apm/Elastic.Apm.csproj
```diff
@@ -20,4 +20,9 @@
     <PackageReference Include="System.Threading.Tasks.Dataflow" Version="4.9.0" />
     <PackageReference Include="Ben.Demystifier" Version="0.1.6" />
   </ItemGroup>
+  <ItemGroup>
+    <Reference Include="System.Configuration">
+      <HintPath>..\..\..\..\..\..\..\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.6.1\System.Configuration.dll</HintPath>
```
vhatsura, line 25: Reference without `HintPath` should also works fine. `<Reference Include="System.Configuration" />`

src/Elastic.Apm/Helpers/ConfigHelper.cs
```diff
@@ -0,0 +1,34 @@
...
+	{
+		public static IConfigurationReader CreateReader(IApmLogger logger)
+		{
+			if (ConfigurationManager.AppSettings[ConfigConsts.KeyNames.ConfigurationReaderType] != null
```
vhatsura, line 14: It can be simplified to 

```csharp
if (!string.IsNullOrEmpty(ConfigurationManager.AppSettings[ConfigConsts.KeyNames.ConfigurationReaderType]))
```

Also, I suggest using `IsNullOrWhiteSpace`. And, maybe it's worth logging some warning message when the value is empty or whitespace.

src/Elastic.Apm/Elastic.Apm.csproj
kodigo35: done

src/Elastic.Apm/Helpers/ConfigHelper.cs
kodigo35: if i push it to Elastic.Apm.AspNetFullFramework it will not reusable by other projects and tests

src/Elastic.Apm/Config/ConfigConsts.cs
kodigo35: done

src/Elastic.Apm/Helpers/ConfigHelper.cs
kodigo35: it was a mistake to let the !=null as i added the IsNullOrEmpty after..

gregkalapos:
There is a really good point from @vhatsura [here](https://github.com/elastic/apm-agent-dotnet/pull/912#discussion_r459850628).

> As I see, it's called only from `src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs`. Maybe it's better to store the class also in `Elastic.Apm.AspNetFullFramework` project. What do you think? It will allow not to add `System.Configuration` dependency which makes sense only for .NET Framework applications.

Technically we could use `System.Configuration` also on all frameworks by pulling in the NuGet package, but: 
- we don't do it right now in the code, so in order to take advantage of it in the `Elastic.Apm` package we would need to also update our existing config readers. 
- in my opinion the benefit would be very limited (assuming most .NET Core apps load configs though `IConfiguation` and not  through `System.Configuration`) and I also feel we should not make the configuration story more complicated than it should be. On .NET Core we offer `IConfiguration` (with all possible config providers) and configuration though environment variables. I feel adding legacy `System.Configuration` would be an overkill.

So, my suggestion: let's make this .NET Framework only and let's move it into `Elastic.Apm.AspNetFullFramework` - that way we can also remove the `System.Configuration.ConfigurationManager` dependency from `Elastic.Apm` (I'd like to have very little dependency in that package)

Any thoughts on this? 

@kodigo35 there is also a small issue with the CLA: your new commits contain a [differrent email address](https://github.com/elastic/apm-agent-dotnet/commit/683522d447744fec53b7fb4c222c63ddd56e3cc1.patch) and the CLA is tied to your commit mails. Could you please sigh it again with the mail address [from the new commits]((https://github.com/elastic/apm-agent-dotnet/commit/683522d447744fec53b7fb4c222c63ddd56e3cc1.patch))?

kodigo35:
Updated my commit email to be the same. 

Updated code to follow @vhatsura proposal.

kodigo35:
Hello @gregkalapos, with which email i have to sign please ? i signed with both with julien.dacosta@kodigo.fr and julien.dacostaps@outlook.fr i dont see information about this

kodigo35:
i also didn't receive any mail after i signed the CLA is it normal or am i doing something wrong ?

gregkalapos:
@kodigo35, it could be your first commits, [see here](https://github.com/elastic/apm-agent-dotnet/commit/bc96603fd0b7351cd4e003c57e94ca0c8f128b04.patch).

That seems to use `58736349+kodigo35@users.noreply.github.com` (very strange....) - can you maybe try to sign the CLA with that?

> i also didn't receive any mail after i signed the CLA is it normal or am i doing something wrong ?

I'm not sure if you get a mail, but I looked into our CLA database and the two mail addresses you listed above are there, so those 2 signatures are fine.

kodigo35:
I signed CLA with 58736349+kodigo35@users.noreply.github.com

I made this commit directly in github web ui, not a good idea :)

gregkalapos:
> 
> 
> I signed CLA with [58736349+kodigo35@users.noreply.github.com](mailto:58736349+kodigo35@users.noreply.github.com)
> 
> I made this commit directly in github web ui, not a good idea :)

Oh, I suspected that's what happened :) 

I found another missing one [here](https://github.com/elastic/apm-agent-dotnet/commit/8c44a3709a04209fb60ad74b55ebf428c6bd9824.patch). That commit uses `julien.dacostaps@kodigo.fr` and that's not in the CLA. Hopefully this will be the last one.

Again, sorry about the trouble - I strongly suggest setting up git once and just use a single setup, so all your commits use a single mail address.

kodigo35:
> julien.dacostaps@kodigo.fr

sorry for the trouble, i set up my mail address in gi but too late :)

kodigo35:
> Great! Now CLA is also green - I'll wait for CI to finish the build and then merge.
> 
> Thanks @kodigo35 .

Thanks to you @gregkalapos

Issue #914, opened 2020-07-23
Elastic.Apm.SqlClient SqlClientDiagnosticListener ExecutionTime 0ms


cdssoftware:
In the SqlClientDiagnosticListener HandleStopCommand sometimes the ExecutionTime from the Statistics value is 0ms (even when the execution has taken much longer) in my environment (AKS running .net core microservices, talking to an Azure SQL Database).

I have fixed this locally by changing the code

```
if (propertyFetcherSet.Statistics.Fetch(payloadData) is IDictionary<object, object> statistics &&
	statistics.ContainsKey("ExecutionTime") && statistics["ExecutionTime"] is long durationInMs)
		duration = TimeSpan.FromMilliseconds(durationInMs);
```

to 

```
if (propertyFetcherSet.Statistics.Fetch(payloadData) is IDictionary<object, object> statistics &&
	statistics.ContainsKey("ExecutionTime") && statistics["ExecutionTime"] is long durationInMs &&
        durationInMs > 0)
		duration = TimeSpan.FromMilliseconds(durationInMs);
```

by doing this the transaction time is calculated as the difference between the start and stop timestamps

gregkalapos:
Thanks for reporting @cdssoftware . 

I opened #915 with the suggested fix. Don't wanna blame, but to me this sounds like a bug in the database library itself. I agree with the fix, ignoring the `0` value and just using the difference between start and stop is definitely more accurate.

gregkalapos:
Fixed in #915, it will be included in the next release.

cdssoftware:
Thanks @gregkalapos

cdssoftware:
@gregkalapos Do you know when the next release will happen (1.6.2?) as we are still waiting for this change to be published

PR #915, opened 2020-07-23
SqlClientDiagnosticListener - Fix 0ms duration

gregkalapos:
Addressing #914

According to #914 the `Statistics` property can report 0ms duration. 

If we ignore 0ms reported by diagnosticsource then the duration will be calculated as the difference between span start and end - which will be surely more accurate than 0ms.

Can't repro this, so not trying to test.


gregkalapos:
Also ran auto code cleanup.

PR #920, opened 2020-07-30
Full Framework custom config: use envvar config as fallback

gregkalapos:
Follow up from #912.

@kodigo35 feel free to chime on.

Summary:
- Added docs (Added @bmorelli25 to review)
- Made sure that similarly to other settings, this can also be configured both in `web.config` and as an environment variable
- Added a test to also cover the environment variable case
- Some nits, like removing unused `using`s and some unused references

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -142,6 +142,34 @@ Below is a sample `Web.config` configuration file for a ASP.NET application.
 </configuration>
 ----
 
+Additionally on ASP.NET you can also implement your own configuration reader. For this you need to implement the `IConfigurationReader` interface from the `Elastic.Apm.Config` namespace.
+
+Once you have your `IConfiguration` implementation you can use the <<config-full-framework-configuration-reader-type, FullFrameworkConfigurationReaderType>> setting.
+
+[float]
+[[config-full-framework-configuration-reader-type]]
+==== `FullFrameworkConfigurationReaderType`
+
+This setting is .NET Full Framework only.
+
+With this setting you can point an agent to a custom `IConfigurationReader` implementation and the agent will read configuration from your `IConfigurationReader` implementation.
+
+Use type name in  https://docs.microsoft.com/en-us/dotnet/api/system.type.assemblyqualifiedname?view=netcore-3.1#System_Type_AssemblyQualifiedName[AssemblyQualifiedName] format (e.g: `MyClass, MyNamespace`).
+
+[options="header"]
+|============
+| Environment variable name                              | Web.config key
+| `ELASTIC_APM_FULL_FRAMEWORK_CONFIGURATION_READER_TYPE` | `ElasticApm:FullFrameworkConfigurationReaderType`
+|============
+
+[options="header"]
+|============
+| Default                          | Type
+| None       | String
+|============
+
+NOTE: In case you set this setting both in the web.config file and as an environment variable then the web.config file has precedence.
+
 
 [[config-core]]
 === Core configuration options
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -167,7 +167,7 @@ Use type name in  https://docs.microsoft.com/en-us/dotnet/api/system.type.assemb
 | None       | String
 |============
 
-NOTE: In case you set this setting both in the web.config file and as an environment variable then the web.config file has precedence.
+If this setting is set in both the web.config file and as an environment variable, then the web.config file has precedence.
 
 
 [[config-core]]
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -142,9 +142,8 @@ Below is a sample `Web.config` configuration file for a ASP.NET application.
 </configuration>
 ----
 
-Additionally on ASP.NET you can also implement your own configuration reader. For this you need to implement the `IConfigurationReader` interface from the `Elastic.Apm.Config` namespace.
-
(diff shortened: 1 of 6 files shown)


docs/configuration.asciidoc
```diff
@@ -142,6 +142,34 @@ Below is a sample `Web.config` configuration file for a ASP.NET application.
 </configuration>
 ----
 
+Additionally on ASP.NET you can also implement your own configuration reader. For this you need to implement the `IConfigurationReader` interface from the `Elastic.Apm.Config` namespace.
+
+Once you have your `IConfiguration` implementation you can use the <<config-full-framework-configuration-reader-type, FullFrameworkConfigurationReaderType>> setting.
+
+[float]
+[[config-full-framework-configuration-reader-type]]
+==== `FullFrameworkConfigurationReaderType`
```
gregkalapos, line 151: Once people use this, it'll kill central config - there will be no way to pull configs in this case, since they inject their own config reader. Should we point this out? 🤔

kodigo35: May be i missed something in the current architecture. In my case i have an ecosystem of services and it needed to set configuration in each service. Instead of that i centralized configuration in an existing database configuration table loaded with a dedicacted 'ConfigurationReader' and the setup of APM configuration is set in a single place.

docs/configuration.asciidoc
```diff
@@ -142,6 +142,34 @@ Below is a sample `Web.config` configuration file for a ASP.NET application.
 </configuration>
 ----
 
+Additionally on ASP.NET you can also implement your own configuration reader. For this you need to implement the `IConfigurationReader` interface from the `Elastic.Apm.Config` namespace.
+
+Once you have your `IConfiguration` implementation you can use the <<config-full-framework-configuration-reader-type, FullFrameworkConfigurationReaderType>> setting.
```
bmorelli25, line 147: I would combine this into one paragraph. Something like this:

```suggestion
Additionally, on ASP.NET, you can implement your own configuration reader. To do this, implement the `IConfigurationReader` interface from the `Elastic.Apm.Config` namespace.
Once implemented, you can make use of the <<config-full-framework-configuration-reader-type, `FullFrameworkConfigurationReaderType`>> setting.
```

docs/configuration.asciidoc
bmorelli25: If I'm understanding correctly, it sounds like this is used as an alternate way to single-source configuration values for all .NET agents.

We should definitely note that this will break central configuration.

docs/configuration.asciidoc
```diff
@@ -142,6 +142,34 @@ Below is a sample `Web.config` configuration file for a ASP.NET application.
...
+| `ELASTIC_APM_FULL_FRAMEWORK_CONFIGURATION_READER_TYPE` | `ElasticApm:FullFrameworkConfigurationReaderType`
+|============
+
+[options="header"]
+|============
+| Default                          | Type
+| None       | String
+|============
+
+NOTE: In case you set this setting both in the web.config file and as an environment variable then the web.config file has precedence.
```
bmorelli25, line 171:
```suggestion
If this setting is set in both the web.config file and as an environment variable, then the web.config file has precedence.
```

docs/configuration.asciidoc
```diff
@@ -142,6 +142,34 @@ Below is a sample `Web.config` configuration file for a ASP.NET application.
...
 ----
 
+Additionally on ASP.NET you can also implement your own configuration reader. For this you need to implement the `IConfigurationReader` interface from the `Elastic.Apm.Config` namespace.
+
+Once you have your `IConfiguration` implementation you can use the <<config-full-framework-configuration-reader-type, FullFrameworkConfigurationReaderType>> setting.
+
+[float]
+[[config-full-framework-configuration-reader-type]]
+==== `FullFrameworkConfigurationReaderType`
+
```
bmorelli25, line 152: I would add something like this to the top

```suggestion

NOTE: Use of this setting will disable {apm-app-ref}/agent-configuration.html[Central configuration].

```

docs/configuration.asciidoc
gregkalapos: @bmorelli25 I was wrong on this one - fetching central config values happens in another component and it'll still work. "This is a feature™️"

gregkalapos: @kodigo35 I see - what you do sounds to me very much like a feature that we already work on. It's not fully implemented yet, meaning you cannot set all settings with that at the moment, but we constantly iterate on it and add new features. The basic idea is that you'll be able to configure your agents directly from Kibana. Look [here](https://www.elastic.co/guide/en/kibana/current/agent-configuration.html) for more info.

docs/configuration.asciidoc
gregkalapos: As noted [here](https://github.com/elastic/apm-agent-dotnet/pull/920#discussion_r464632697) I was originally wrong, so this'll work with central config.

gregkalapos:
Merging now - Brandon is out, but no need for more review, I basically took all the suggestions, except the one where I gave wrong info.

docs/configuration.asciidoc
kodigo35: Dedicated configuration reader is usefull for those who have externalized the whole configuration of their service and dont want to have extra load of work for ci/cd with new keys to set. 

Some configuration properties of agent can be done in kibana but applications should only knows apm server. this requires at least the url, service name and environment property to be set.  And if you have multiple services, applications it is not very suitable to have setting in web.config. envvar car be used too but in testing platform you can have several versions of applciations deployed and they would share the same settings.

Issue #922, opened 2020-08-03
Add GC time to the .Net Agent


jpenniman:
The .Net Agent already records the number collections and the heap sizes, however it is missing the pause time.

Recording GC pause times is critical for identifying memory leaks. When troubleshooting high memory utilization, we want to answer: "Is CPU high too?" and "How much of that CPU time is being used by the garbage collector?" If the garbage collector is running a lot, but collecting very little, it's an indication that there is likely a memory leak. Long GC pause times can also cause application performance issues even without a memory leak. Understanding pause times is key to understanding if application slowness is possibly do to long pause times or too frequent pauses.

The Java agent records: `jvm.gc.time`

I'm proposing a .net equivalent: `clr.gc.time` 

For full framework:
```cs
_traceEventSession.Source.AddCallbackOnProcessStart(process =>
{
	process.AddCallbackOnDotNetRuntimeLoad(runtime =>
	{
		runtime.GCEnd += (traceProcess, gc) =>
		{
			if (traceProcess.ProcessID == Process.GetCurrentProcess().Id)
				_gcPauseTime = gc.PauseDurationMSec;
		};
	});
});
```

.Net Core doens't write pause time data o the event, but it can be calculated by capturing the instant of GCStart and doing the math in GCEnd. It's not perfect, since not all collections stop the world. Labeling the metric with the generation being collected and GC settings can help the engineer determine if the gc.time is a pause or not.

```cs
sealed class GcEventListener : EventListener
{
    long _gcStartTime;

    protected override void OnEventWritten(EventWrittenEventArgs eventData)
    {
        if (eventData.EventId == GcStartEvent)
        {
            _gcStartTime = DateTime.UtcNow.Ticks;
        }
        else if (eventData.EventId == GcEndEvent)
        {
            var gcStopTime = DateTime.UtcNow.Ticks;
            var pauseTime = (double)(gcStopTime - _gcStartTime) / 10_000.0
        }
    }
}
```

Our application currently uses similar code to get the GC Pause time.  We've tested code in net461, net472, and netcoreapp3.1 on both Windows and Linux, and it successfully reports GC pause times in all targets. Ideally, we'd like to get this functionality into the Agent, so it's something we don't have to maintain separately.

I'm available to add this feature one we can agree on implementation approach.

gregkalapos:
Hi @jpenniman

[...]

PR #925, opened 2020-08-06
Implements #922: Add GC time to the .Net Agent

jpenniman:
This adds `clr.gc.time` to the .Net GC Metrics. It's a little different than my original proposal--using `DurationInMs` instead of `PauseTimeInMs` for full framework to keep it consistent with the .Net Core metric, which can only determine duration and is not aware of whether a pause occurred.

### DIFF
--- a/src/Elastic.Apm/Metrics/MetricsCollector.cs
+++ b/src/Elastic.Apm/Metrics/MetricsCollector.cs
@@ -93,10 +93,12 @@ namespace Elastic.Apm.Metrics
 				GcMetricsProvider.GcGen2SizeName);
 			var collectGen3Size = !WildcardMatcher.IsAnyMatch(currentConfigSnapshot.DisableMetrics,
 				GcMetricsProvider.GcGen3SizeName);
+			var collectGcTime = !WildcardMatcher.IsAnyMatch(currentConfigSnapshot.DisableMetrics,
+				GcMetricsProvider.GcTimeName);
 			if (collectGcCount || collectGen0Size || collectGen1Size || collectGen2Size || collectGen3Size)
 			{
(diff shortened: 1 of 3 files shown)


jpenniman:
@gregkalapos that makes sense. I agree the easiest would be to sum the time. Normally, in .Net, when looking at perfmon or other metrics collectors, `gc.time` is the time in gc for the sample period (real-time in the case of perfmon). So I think it makes sense to sum during the gcEnd event and reset it after the sample is taken.

gregkalapos:
> @gregkalapos that makes sense. I agree the easiest would be to sum the time. Normally, in .Net, when looking at perfmon or other metrics collectors, `gc.time` is the time in gc for the sample period (real-time in the case of perfmon). So I think it makes sense to sum during the gcEnd event and reset it after the sample is taken.

Perfect, we agree on this one. Could you please update the PR accordingly?

gregkalapos (approved):
This PR is ok to merge. It got a little bit outdated, so I merged master into this and resolved the conflicts.

Since the way we collect GC stats on Full Framework changed a bit in https://github.com/elastic/apm-agent-dotnet/pull/1124, I also adapted GC time collection on Full Framework - similarly to .NET Core and .NET 5 it is based on the time span between start&stop.

Another question that came up is if we use the gc time metric as a counter and we increase it forever a `long` may not be big enough to store the gc times in ticks. I checked this - we are good here:

```
var ts = new TimeSpan(long.MaxValue);
Console.WriteLine(ts.Days / 365);
Output: 29247
```

gregkalapos:
There are some CI issues, but seems to be unrelated. VS install fails, and on Windows we had an unrelated test fail. Tests are green locally. Merging this in.

PR #938, opened 2020-08-25
Improve log message in StacktraceHelper.GenerateApmStackTrace

SergeyKleyman:
1. The phrase was inverse of the intended ("extracting exception from stackTrace")
2. Added logging of exception from which we are trying to extract stack trace.


gregkalapos (approved):
LGTM

SergeyKleyman:
Included in #939

PR #939, opened 2020-08-25
Ensure exception is logged when GenerateApmStackTrace fails

SergeyKleyman:


gregkalapos (approved):
I already approved https://github.com/elastic/apm-agent-dotnet/pull/938, but this is also fine by me.

Issue #944, opened 2020-08-28
EF Core trace stacktraces don't extend into the application


kanadaj:
**Describe the bug**
There is something wrong with how stacktraces in awaited async requests get unwound for EF Core in ASP.NET Core. The result is that only system internals are part of the resulting stacktrace, but not the actual method calling the EF Core function.

Example stacktrace as follows:
```cs
Elastic.Apm.Model.Span in End
Elastic.Apm.Model.DbSpanCommon in EndSpan
Elastic.Apm.EntityFrameworkCore.EfCoreDiagnosticListener in OnNext
System.Diagnostics.DiagnosticListener in Write
Microsoft.EntityFrameworkCore.Diagnostics.RelationalLoggerExtensions in BroadcastCommandExecuted
Microsoft.EntityFrameworkCore.Diagnostics.RelationalLoggerExtensions in CommandReaderExecutedAsync
Microsoft.EntityFrameworkCore.Storage.RelationalCommand+<ExecuteReaderAsync>d__17 in ExecuteReaderAsync
System.Threading.ExecutionContext in RunInternal
System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1+AsyncStateMachineBox`1 in MoveNext
System.Threading.Tasks.AwaitTaskContinuation in RunOrScheduleAction
System.Threading.Tasks.Task in RunContinuations
System.Threading.Tasks.Task`1 in TrySetResult
MySqlConnector.Core.CommandExecutor+<ExecuteReaderAsync>d__0 in ExecuteReaderAsync at line 76
System.Threading.ExecutionContext in RunInternal
System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1+AsyncStateMachineBox`1 in MoveNext
System.Threading.Tasks.AwaitTaskContinuation in RunOrScheduleAction
System.Threading.Tasks.Task in RunContinuations
System.Threading.Tasks.Task`1 in TrySetResult
MySql.Data.MySqlClient.MySqlDataReader+<CreateAsync>d__95 in CreateAsync at line 409
System.Threading.ExecutionContext in RunInternal
System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1+AsyncStateMachineBox`1 in MoveNext
System.Threading.Tasks.AwaitTaskContinuation in RunOrScheduleAction
System.Threading.Tasks.Task in RunContinuations
System.Threading.Tasks.Task`1 in TrySetResult
[...]

PR #948, opened 2020-09-01
Use EnhancedStackTrace for spans; start the stack trace on start for EF Core

kanadaj:
Addresses https://github.com/elastic/apm-agent-dotnet/issues/944 : Entity Framework Core stacktraces are now captured at the beginning of the request to avoid async stacktrace shenanigans; I've also replaced the StackTrace from `Span` with `EnhancedStackTrace` to reduce noise in traces.

This might have a performance impact due to capturing stack traces before we can evalute  the span length, but I'm not quite sure if it's possible to get around that. It might be prudent to add a flag to disable trace capture for EF Core. @gregkalapos thoughts?

### DIFF
--- a/src/Elastic.Apm.EntityFrameworkCore/EfCoreDiagnosticListener.cs
+++ b/src/Elastic.Apm.EntityFrameworkCore/EfCoreDiagnosticListener.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using Elastic.Apm.Api;
 using Elastic.Apm.DiagnosticSource;
 using Elastic.Apm.Model;
 using Microsoft.EntityFrameworkCore.Diagnostics;
@@ -38,7 +39,8 @@ namespace Elastic.Apm.EntityFrameworkCore
 				case { } k when k == RelationalEventId.CommandExecuting.Name && _agent.Tracer.CurrentTransaction != null:
 					if (kv.Value is CommandEventData commandEventData)
 					{
-						var newSpan = _agent.TracerInternal.DbSpanCommon.StartSpan(_agent, commandEventData.Command, InstrumentationFlag.EfCore);
+						var spanName = commandEventData.Command.CommandText.Replace(Environment.NewLine, " ");
+						var newSpan = ExecutionSegmentCommon.StartSpanOnCurrentExecutionSegment(_agent, spanName, ApiConstants.TypeDb, null, InstrumentationFlag.EfCore, true);
 						_spans.TryAdd(commandEventData.CommandId, newSpan);
 					}
 					break;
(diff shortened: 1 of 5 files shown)


src/Elastic.Apm.EntityFrameworkCore/EfCoreDiagnosticListener.cs
```diff
@@ -38,7 +39,8 @@ public void OnNext(KeyValuePair<string, object> kv)
 				case { } k when k == RelationalEventId.CommandExecuting.Name && _agent.Tracer.CurrentTransaction != null:
 					if (kv.Value is CommandEventData commandEventData)
 					{
-						var newSpan = _agent.TracerInternal.DbSpanCommon.StartSpan(_agent, commandEventData.Command, InstrumentationFlag.EfCore);
```
gregkalapos, line 41: I'd prefer keeping the call to `DbSpanCommon.StartSpan` - that way all db spans are started in the same method. And then I'd just add an optional parameter to it and pass `true` so it calls `StartSpanOnCurrentExecutionSegment` with `captureStackTraceOnStart: true`

gregkalapos (commented):
Great! Thanks for the PR! 
> I've also replaced the StackTrace from Span with EnhancedStackTrace to reduce noise in traces.

Great! This was probably missing from a previous PR, the intention was to use `EnhancedStackTrace` everywhere.

> This might have a performance impact due to capturing stack traces before we can evalute the span length, but I'm not quite sure if it's possible to get around that. It might be prudent to add a flag to disable trace capture for EF Core. @gregkalapos thoughts?

We have the same with HTTP calls and I'd say we can except this. My argument for this is that within a transaction typically you'll have a few such transactions and very likely they'll be longer than the default minduration for capturing stack trace. So it's definitely not like people will have 100s of such spans in a transaction, and if they do, then the perf. impact of this won't be the biggest problem.

So far for HTTP we have no request to disable this behavior, so I'd not add a config yet. If we have request to do so I'd rethink this.

One last point: could you maybe add 1 test with EF Core with some async calls? I think the most important assert would be one asserting that the caller method is on the stack. Let me know if you can't do it - In that case I'll do it.

Other than the test and my comment from before I'd be happy to merge this.

Thanks again!

src/Elastic.Apm.EntityFrameworkCore/EfCoreDiagnosticListener.cs
kanadaj: Alright, I didn't want to mess with the internal API and HTTP calls use the same method, but I'll change it around then

kanadaj:
> One last point: could you maybe add 1 test with EF Core with some async calls? I think the most important assert would be one asserting that the caller method is on the stack. Let me know if you can't do it - In that case I'll do it.

Alright, changed the internal method to add the `captureStackTraceOnStart` parameter and added a test for capturing the stacktrace.

To note, in order to make a test that doesn't test successfully even with the old code, I had to create a table inside the DbContext, since `ExecuteSqlCommandAsync` would bypass the EF Core pipeline almost entirely.

gregkalapos (approved):
Thank you! - looks ok to me.

I guess the integration test step will fail in the PR which was fixed in https://github.com/elastic/apm-agent-dotnet/pull/949 a few hours ago - no need to rebase just for that. I'll let CI run, if all tests except the integration test step is green, then this ready to merge.

Issue #952, opened 2020-09-03
Requests that cause 404 are logged under individual actions, flooding Elastic with paths that don't exist


kanadaj:
**Is your feature request related to a problem? Please describe.**
When any request is on a path that doesn't exist (even file requests), these requests go all the way to the MVC middleware before returning a 404. Elastic APM logs each of these requests as a different action based on the path.

**Describe the solution you'd like**
404 requests should all all grouped under a path such as `GET NotFound` and `POST NotFound`, unless they were matched against an MVC route. Otherwise the result is something like this:
![image](https://user-images.githubusercontent.com/3247357/92162240-c25a7780-ee29-11ea-9cdc-cee682500306.png)

gregkalapos:
Agreed - it is also what the [spec](https://github.com/elastic/apm/blob/master/specs/agents/tracing-instrumentation-http.md) defines. There is already 1 issue related to this: https://github.com/elastic/apm-agent-dotnet/issues/930. Those 2 can be fixed together.

kanadaj:
See PR at https://github.com/elastic/apm-agent-dotnet/pull/953

gregkalapos:
Fixed in #953. Thanks @kanadaj .

Issue #930, opened 2020-08-17
Remove default `{context.Request.Method} {context.Request.Path}` transaction name in ASP.NET Core


gregkalapos:
Currently we do [this](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs#L99) as the default transaction name. Usually the transaction name will end up being something created by [this logic](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm.AspNetCore/ApmMiddleware.cs#L286) using the MVC controller/action name or the razor page name.

But if that logic fails (mainly because we don't have that info) we keep the transaction name with the path in it.

But according to [the spec](https://github.com/elastic/apm/blob/master/specs/agents/tracing-instrumentation-http.md)

> The transaction name should be aggregatable, such as the route or handler name. 

And the path should not show up in the transaction name. 

So let's default to `<METHOD> unknown route` as the spec defines.

russcam:
@gregkalapos is this something we still want to do? From a cursory look, it looks like updating

https://github.com/elastic/apm-agent-dotnet/blob/cc11d34f945425261f2ac50cc90fb6b44b55a441/src/Elastic.Apm.AspNetCore/WebRequestTransactionCreator.cs#L37

is the only change (plus tests)?

JeremyBessonElastic:
It looks like it has been fixed and doesn't have any activity for long, then closing it.

PR #953, opened 2020-09-03
Make "unknown route" the route by default

kanadaj:
Addresses https://github.com/elastic/apm-agent-dotnet/issues/952 and https://github.com/elastic/apm-agent-dotnet/issues/930.

@gregkalapos This I feel this needs further tests, including testing for paths that do not map to a controller. Sadly, with the current test setup, adding a HTTP request such as 
```cs
await httpClient.GetAsync("css/doesnotexist/somefile.css")
``` 

results in the payload sender not capturing the transaction at all. I'm not entirely sure why that is. On the bright side, all other tests seem to still pass without a problem. I've added a broken test like above, it *should* work I think, should have it fixed before this is merged

### DIFF
--- a/src/Elastic.Apm.AspNetCore/WebRequestTransactionCreator.cs
+++ b/src/Elastic.Apm.AspNetCore/WebRequestTransactionCreator.cs
@@ -32,7 +32,7 @@ namespace Elastic.Apm.AspNetCore
 				}
 
 				Transaction transaction;
-				var transactionName = $"{context.Request.Method} Unknown Route";
+				var transactionName = $"{context.Request.Method} {context.Request.Path}";
 
 				if (context.Request.Headers.ContainsKey(TraceContext.TraceParentHeaderNamePrefixed)
 					|| context.Request.Headers.ContainsKey(TraceContext.TraceParentHeaderName))
@@ -186,6 +186,10 @@ namespace Elastic.Apm.AspNetCore
 						var name = GetNameFromRouteContext(routeData);
 
 						if (!string.IsNullOrWhiteSpace(name)) transaction.Name = $"{context.Request.Method} {name}";
+					}
+					else if (context.Response.StatusCode == StatusCodes.Status404NotFound)
+					{
+						transaction.Name = $"{context.Request.Method} Unknown Route";
 					}
 				}
 
--- a/src/Elastic.Apm.AspNetCore/WebRequestTransactionCreator.cs
+++ b/src/Elastic.Apm.AspNetCore/WebRequestTransactionCreator.cs
@@ -133,7 +133,7 @@ namespace Elastic.Apm.AspNetCore
 
 				return rawPathAndQuery == null ? null : UriHelper.BuildAbsolute(httpRequest.Scheme, httpRequest.Host, rawPathAndQuery);
 			}
-			catch(Exception e)
+			catch (Exception e)
 			{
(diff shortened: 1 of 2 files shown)


gregkalapos (commented):
@kanadaj  the `.css` file is ignored because of the [`IgnoreUrls`](https://www.elastic.co/guide/en/apm/agent/dotnet/master/config-http.html#config-transaction-ignore-urls) setting. I think best would be to use use some random url without `.[extension]`.

kanadaj:
Alright, fixed the test, should be good to go :)

src/Elastic.Apm.AspNetCore/WebRequestTransactionCreator.cs
```diff
@@ -187,6 +187,10 @@ internal static void StopTransaction(Transaction transaction, HttpContext contex
 
 						if (!string.IsNullOrWhiteSpace(name)) transaction.Name = $"{context.Request.Method} {name}";
 					}
+					else if (context.Response.StatusCode == StatusCodes.Status404NotFound)
+					{
+						transaction.Name = $"{context.Request.Method} Unknown Route";
```
gregkalapos, line 192: nit: [the spec](https://github.com/elastic/apm/blob/master/specs/agents/tracing-instrumentation-http.md#http-transactions) uses non capital letters, so let's stick to that.

```suggestion
						transaction.Name = $"{context.Request.Method} unknown route";
```

gregkalapos (approved):
Great, thanks! 

Added a small change to use non capital letters in the transaction name.

I'd still keep https://github.com/elastic/apm-agent-dotnet/issues/930 open, that issue is not really solved, but https://github.com/elastic/apm-agent-dotnet/issues/952 can be definitely closed with this.

Thanks @kanadaj  👏

kanadaj:
@gregkalapos I'm not entirely sure you want to solve #930 beyond this. Middlewares might handle a path without a mapped route, but mapping them as "unknown route" is far from ideal.

gregkalapos:
> 
> 
> @gregkalapos I'm not entirely sure you want to solve #930 beyond this. Middlewares might handle a path without a mapped route, but mapping them as "unknown route" is far from ideal.

I personally agree with that, but according to the spec that's not how it should be - we had a discussion about this with @felixbarny and the outcome was to start with `[METHOD] unknown route` (similarly to what [you did originally](https://github.com/elastic/apm-agent-dotnet/pull/953/commits/2bdde2c4fd3b3cd1099e7a23ec29d4d05261aadc#diff-127f962d697b4fc6c5ca070b8241cd2dR35)) and then deduct what we can and not use the `context.Request.Path`.

The main motivation in the spec for using route is to make sure the transaction name is aggregatable. I think it'd be fair to assume that if there is no mapped route then just using the path would still be aggregatable, because typically you won't have parameters in that case. Maybe that statement is not always true, but I think it's true for most such apps.

If we still would go with `unknown route` as default then probably it'd make sense to introduce the [`use_path_as_transaction_name` config](https://www.elastic.co/guide/en/apm/agent/java/current/config-http.html#config-use-path-as-transaction-name) similar to Java.

gregkalapos:
I don't know why the new test fails in CI - locally everything is fine and I don't see anything bad in the code.

I added more logging to the test, hopefully with that I'll be able to figure this out.

gregkalapos:
Fixed, this was a concurrency issue - instead of making sure both transactions are ended, we simply do a single transaction.

PR #971, opened 2020-10-07
Avoid double initialization in HostBuilderExtensions.cs

gregkalapos:
In case the extension method `UseElasticApm()` on `IHostBuilder` runs multiple times (e.g. because there are multiple `IHostBuilder`s in a single process, then the agent throws an exception due to double initialization.

```
private static IHostBuilder CreateHostBuilder(string[] args) =>
	Host.CreateDefaultBuilder(args)
		.ConfigureServices((context, services) => { services.AddHostedService<HostedService>(); })
		.UseElasticApm(new HttpDiagnosticsSubscriber());
```

### Background of the issue

Currently the agent does not support multiple instances in a single process - I strongly thing we should keep this (multiple instances would mean multiple metrics capturing instances, multiple payloadsenders, potentially multiple configreaders - none of this is supported). 

### Proposed solution in this PR 

Let's check in `UseElasticApm` if an agent is already initialized and reuse that agent.

### DIFF
--- a/ElasticApmAgent.sln
+++ b/ElasticApmAgent.sln
@@ -94,7 +94,9 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "GrpcClientSample", "sample\
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Elastic.Apm.GrpcClient", "src\Elastic.Apm.GrpcClient\Elastic.Apm.GrpcClient.csproj", "{8F1B46E1-F1A0-4744-9D9A-AA1BB33842C8}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Elastic.Apm.Grpc.Tests", "test\Elastic.Apm.Grpc.Tests\Elastic.Apm.Grpc.Tests.csproj", "{3AD5584E-057A-4C52-84EA-32A9BBE12F03}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Elastic.Apm.Grpc.Tests", "test\Elastic.Apm.Grpc.Tests\Elastic.Apm.Grpc.Tests.csproj", "{3AD5584E-057A-4C52-84EA-32A9BBE12F03}"
+EndProject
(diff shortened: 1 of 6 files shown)


vhatsura:
Looks like it's an issue in the calling code, due to a service provider should be called at once.
The reason can be in the configuring agent in two different ways in the codebase or accessing agent before actual initialization.
In both cases, it would be better to have an error about it.

Also, the issue can be in unit tests when the service container is bootstrapped before each test class, collection, etc. In this case, it would be nice to have the `Setup` method which can be called in the `ApmAgent` instance rather than static version.

gregkalapos:
> 
> 
> Looks like it's an issue in the calling code, due to a service provider should be called at once.
> The reason can be in the configuring agent in two different ways in the codebase or accessing agent before actual initialization.
> In both cases, it would be better to have an error about it.
> 
> Also, the issue can be in unit tests when the service container is bootstrapped before each test class, collection, etc. In this case, it would be nice to have the `Setup` method which can be called in the `ApmAgent` instance rather than static version.

My only concern is: what happens when there are multiple `IHostBuilder`s in a single process?

vhatsura:
> > Looks like it's an issue in the calling code, due to a service provider should be called at once.
> > The reason can be in the configuring agent in two different ways in the codebase or accessing agent before actual initialization.
> > In both cases, it would be better to have an error about it.
> > Also, the issue can be in unit tests when the service container is bootstrapped before each test class, collection, etc. In this case, it would be nice to have the `Setup` method which can be called in the `ApmAgent` instance rather than static version.
> 
> My only concern is: what happens when there are multiple `IHostBuilder`s in a single process?

I'd say each `IHostBuilder` should work with the own version of `IApmAgent` because in such case we will have two different dependency containers. But, it cannot be achieved right now due to static Apm Agent initialization.

gregkalapos:
> ut, it cannot be achieved right now due to static Apm Agent initialization.

That'd be not easy to do and open up a whole bunch of other problems. I mean in theory I agree it sounds good, but I'm not sure it's realistic to implement it. But this made me thing and maybe I'd reverse it - just create 1 agent and reuse it in each `IHostBuilder` afterwords. Most people don't really care about separation (and we don't support it in a single process right now anyway), they just want to add the agent to their projects.

PR #973, opened 2020-10-12
Normalize transaction names for ASP.NET Full Framework

### LINKED ISSUE #201: ASP.NET - create transaction name based on route and group by optional values
The same as elastic/apm-agent-dotnet#84 but for Full Framework ASP.NET.

We will probably be able to do it automatically for ASP.NET frameworks that have routing such as MVC and WebApi. In addition we might also consider supporting manual configuration similar to the one implemented by Java agent - https://www.elastic.co/guide/en/apm/agent/java/current/config-http.html#config-url-groups

russcam:
This PR normalizes the transaction name for ASP.NET MVC Full Framework to use the convention of

```
<HTTP method> <area>/<controller>/<action> [<param>*]
```

which aligns with the convention used in ASP.NET Core transaction names. The transaction name is determined from the RouteData values, and where route values are present but the response is a 404, unknown route is used; Unknown route is not used where there are no route values because the application may contain endpoints that are not routed to by MVC controllers.

Integration tests introduce an MVC area with a HomeController and Index action, to assert that the area route datatoken is
taken into account. Without taking into account the area, same named controller actions in areas would end up aggregrated under the same transaction name.

This PR also fixes the DbTests when running with IIS ApplicationPool Identity. When AspNetFullFramework.Tests are run with an application pool configured to use an ApplicationPoolIdentity, the pool has insufficient access to execute the native SQLite.Interop.dlls used in database tests running with SQLite. This is fixed by configuring read and execute rights for the pool on the SQLite.Interop.dll files when configuring the ApplicationPool.

Closes #201


russcam:
Looks like an exception is thrown in mapping the IIS ApplicationPool Identity on CI. Will address.

src/Elastic.Apm/Api/ITransaction.cs
```diff
@@ -50,5 +50,11 @@ public interface ITransaction : IExecutionSegment
 		/// existing one.
 		/// </returns>
 		string EnsureParentId();
+
+		/// <summary>
+		/// If true, then the transaction name was modified by external code, and transaction name should not be changed
+		/// or "fixed" automatically.
+		/// </summary>
+		bool HasCustomName { get; }
```
gregkalapos, line 58: Why do we move this to the public `ITransaction`? 

Asking because I'd think that public API users would not really care about this flag, but from this point it'll show up for them as well.

russcam: I moved this to the interface because the ElasticApmModule needs to know whether it can rename a transaction name in an ASP.NET Full Framework app, similar to ASP.NET Core, and the started transaction returns the `ITransaction` interface: https://github.com/elastic/apm-agent-dotnet/blob/5f9fd02061e5eb2b83f839f51e5d47d2786587f3/src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs#L137

I did consider

1. casting the `ITransaction` to `Transaction`, which seems like it might be the wrong thing to do as users may implement their own `ITransaction` type?
2. instantiating a `Transaction` directly, but this also seems like it might be wrong because it circumvents the public interface?

Another approach here may be to encapsulate whether renaming is allowed inside of the `Transaction.Name` setter e.g.

https://github.com/elastic/apm-agent-dotnet/blob/a63b7575469659d4a83a67da5ec0a613e8ec0a0a/src/Elastic.Apm.AspNetCore/WebRequestTransactionCreator.cs#L183-L202

would be something like

```c#
internal class Transaction : ITransaction
{
        [JsonIgnore]
	internal bool HasCustomName { get; private set; }

	[JsonConverter(typeof(TrimmedStringJsonConverter))]
	public string Name
	{
		get => _name;
		set
		{
			if (!HasCustomName)
			{
				HasCustomName = true;
				_name = value;
			}
		}
	}
}
```

```c#
//fixup Transaction.Name - e.g. /user/profile/1 -> /user/profile/{id}
var routeData = context.GetRouteData()?.Values;

if (routeData != null && context.Response.StatusCode != StatusCodes.Status404NotFound)
{
	logger?.Trace()?.Log("Calculating transaction name based on route data");
	var name = GetNameFromRouteContext(routeData);

	if (!string.IsNullOrWhiteSpace(name)) transaction.Name = $"{context.Request.Method} {name}";
}
else if (context.Response.StatusCode == StatusCodes.Status404NotFound)
{
	logger?.Trace()
		?
		.Log("No route data found or status code is 404 - setting transaction name to 'unknown route");
	transaction.Name = $"{context.Request.Method} unknown route";
}
``` 

The downside is that the work to calculate a new transaction name is performed whether or not it can be updated, and if a new name is not assigned, there is no indication. The first issue could be achieved by introducing an `ITransaction.SetName(Func<string> func)`, where the delegate is run only if a new name can be applied, and the second issue could be signalled by logging.

src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
```diff
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Specialized;
+using System.Net;
```
gregkalapos, line 8:
```suggestion
```

Seems to be unused.

src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
```diff
@@ -235,11 +236,51 @@ private void ProcessEndRequest(object eventSender)
...
+			if (transaction == null) return;
 
 			SendErrorEventIfPresent(httpCtx);
 
+			if (!transaction.HasCustomName)
+			{
+				//fixup Transaction.Name - e.g. /user/profile/1 -> /user/profile/{id}
+				var values = httpApp.Request.RequestContext?.RouteData?.Values;
+
+				if (values?.Count > 0 && httpResponse.StatusCode != StatusCodes.Status404NotFound)
```
gregkalapos, line 250: One (I think) fairly common case which is not handled is when a REST API returns 404.

[Here](https://github.com/elastic/apm/blob/master/specs/agents/tracing-instrumentation-http.md) is the spec.

It says: 
>  This would normally also apply to requests to unknown endpoints, e.g. the transaction for the request GET /this/path/does/not/exist would be named GET unknown route, whereas the transaction for the request GET /users/123 would still be named GET /users/{id}

If we have `/user` endpoint which returns 404 for non-existent users then we'd skip the if block and keep the existing transaction name based on the url (so e.g. `GET /user/123` if `123` doesn't exist). And this would cause transaction name explosion.

src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
```diff
@@ -235,11 +236,51 @@ private void ProcessEndRequest(object eventSender)
...
+				{
+					if (httpResponse.StatusCode != StatusCodes.Status404NotFound)
+					{
+						// handle MVC areas. The area name will be included in the DataTokens.
```
gregkalapos, line 254: Nice 👍

gregkalapos (commented):
I looked at this and I'm not 100% sure we handle all potential cases as we should (see comment below). If you don't mind I'd give this another round before we merge.

src/Elastic.Apm/Api/ITransaction.cs
gregkalapos: Discussed over zoom - but for keeping it also here: I think best would be to check if it's `Transaction` if so, then cast, otherwise ignore. 

The idea behind having `ITransaction` is that potentially we have `NoopTransaction` when the agent is deactivated - and in that case if we ignore this field is fine, since nothing will be sent anyway.

src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
russcam: Thanks! resolved

src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
russcam: Good catch! We can differentiate here based on the presence of a 404 HttpException on the HttpContext coming from System.Web.Mvc

russcam:
rebased against master to cleanly merge changes in AspNetFullFrameworkSampleApp.csproj and force pushed.

@gregkalapos addressed the PR comments, so ready for another look when you have a chance.

src/Elastic.Apm/Model/Transaction.cs
```diff
@@ -218,10 +219,10 @@ void StartActivity()
 
 		/// <summary>
 		/// If true, then the transaction name was modified by external code, and transaction name should not be changed
-		/// or "fixed" automatically ref https://github.com/elastic/apm-agent-dotnet/pull/258.
+		/// or "fixed" automatically.
 		/// </summary>
 		[JsonIgnore]
-		internal bool HasCustomName { get; private set; }
+		public bool HasCustomName { get; private set; }
```
russcam, line 225: this should be reverted to `internal`

gregkalapos (approved):
LGTM, awesome! 

Just do this: https://github.com/elastic/apm-agent-dotnet/pull/973#discussion_r511706758 :)

Issue #967, opened 2020-09-22
Support numeric labels


devree-bvb:
**Is your feature request related to a problem? Please describe.**
While monitoring our application, we capture information regarding our sql statements, including the numbers of rows affected. Currently we add the number of rows as a label to the span measuring the statement. This data however is sent to the backend as a string, which prevents us from aggregating this data in a numeric way (average, ....)

**Describe the solution you'd like**
Support the feature of numeric labels (https://www.elastic.co/guide/en/apm/get-started/7.2/metadata.html)

**Describe alternatives you've considered**

**Additional context**

TomDeVree:
For this enhancement, would an implemenation like the Java implementation be a possibility? There they have 3 methods to add 3 different kinds of labels, string, numeric and boolean (https://github.com/elastic/apm-agent-java/blob/master/apm-agent-api/src/main/java/co/elastic/apm/api/SpanImpl.java) In C# now there is a dictionary for the labels, so the type is fixed to string.

This of course would be a breaking change in the API, so I would like some feedback before trying a PR

gregkalapos:
> This of course would be a breaking change in the API, so I would like some feedback before trying a PR

I'd try not to break it if possible.

Without looking into this too much: how about keeping the dictionary as it is and then additionally add the `AddLabel` method with different types? The way the dictionary is used can be changed, but let's try to not break the current API surface. That way you don't break current users, but you can add the overloads for new users. Then long term, probably best would be to just deprecate the dictionary.

TomDeVree:
Ok, Thanks for the feedback.

I will keep the existing API.
Problem there is that, because all labels come together in this dictionary that contain string values, and that can be modified, I don't know if the user intended to add a number in string format, or in a number format (for example). The difference is important for serialization to the backend.

I will try to find a solution for this

gregkalapos:
Implemented in #982.

Issue #788, opened 2020-03-25
Extend SDK for adding labels to allow adding numbers and bools


alex-fedotyev:
Java SDK allows adding labels which could be string, number or bool type.
https://www.elastic.co/guide/en/apm/agent/java/master/public-api.html#api-transaction-add-tag

Currently .NET SDK is limited to adding string only.
https://www.elastic.co/guide/en/apm/agent/dotnet/current/public-api.html#api-transaction-tags

It would be great to update .NET SDK to match java capabilities.
This functionality would be useful to enable exposing application KPI's programmatically.

gregkalapos:
Implemented in #982.

Issue #473, opened 2019-09-05
Add support for providing additional labels in CaptureError and CaptureException agent's public API


SergeyKleyman:
Currently [`CaptureError`](https://www.elastic.co/guide/en/apm/agent/dotnet/current/public-api.html#api-span-capture-error) and [`CaptureException`](https://www.elastic.co/guide/en/apm/agent/dotnet/current/public-api.html#api-span-capture-exception) .NET agent 's public API inherit `Context`, which in particular includes labels, from the corresponding transaction and there's no way to provide additional labels specific to the error and not to the whole transaction. For example, [Node.js](https://www.elastic.co/guide/en/apm/agent/nodejs/current/agent-api.html#apm-capture-error) and [Python](https://www.elastic.co/guide/en/apm/agent/python/current/api.html#client-api-capture-exception) agents do support providing additional labels in their capture-error public API.

Requested at https://discuss.elastic.co/t/how-to-use-distributed-tracing-in-with-elastic-apm-in-non-net-core-application/197356/5

gregkalapos:
Implemented in #982.

Issue #191, opened 2019-04-19
Captured error's context should not change after capture


SergeyKleyman:
Currently we copy a reference to transaction's context to captured error instance (AKA shallow copy) so error's context changes if transaction's context changes which is incorrect - error's context should not change after capture.

SergeyKleyman:
Add tests to verify that in particular error's context won't have `Response` property.

gregkalapos:
Decision: this is a nice to have, not needed in the next release.

gregkalapos:
Implemented in #982.

PR #982, opened 2020-10-19
 Extend Label SDK: Support numbers and bools and add label to errors

gregkalapos:
Solves #967, #788, #473, #191

Prior to this we only had
- `ITransaction.Labels` as `Dictionary<string,string>`
- `ISpan.Labels` as `Dictionary<string,string>`

This PR adds new methods to add labels to both `ISpan` and `ITransaction` (#967, #788):

```
public void SetLabel(string key, string value);
public void SetLabel(string key, bool value);
public void SetLabel(string key, double value);
public void SetLabel(string key, int value);
public void SetLabel(string key, long value);
public void SetLabel(string key, decimal value);
```

and keeps the dictionary with a backwards compatible change (the dictionary can be deprecated long term):
- `ITransaction.Labels` as ~`Dictionary<string,object>`~ `Dictionary<string,object>` (no change)
- `ISpan.Labels` as ~`Dictionary<string,object>`~ `Dictionary<string,object>` (no change)

Additionally it also adds an optional parameter to pass labels to errors both on `ITransaction` and `ISpan` #473:

```
ITransaction.CaptureError("Test Error", "TestMethod", new StackTrace().GetFrames(),
	labels: new Dictionary<string, Label> { { "stringLabel", "test" } });
```

The `labels` parameter will be added to the labels on the error, which contains all the labels on the given transaction plus the ones passed to the `CaptureError` and `CaptureException` methods

Prior to the PR every `Error.Context` was just a reference to the existing `Transaction.Context`. The problem with this was that changes on one of them would be reflected on the other - these 2 `Context` instances should be decoupled (according to spec - and that's also what other agents do). This PR implements this with adding DeepCopy to `Context` - each time an `Error` is created, it deep-copies `Transaction.Context`  #191.


src/Elastic.Apm/Api/IExecutionSegment.cs
```diff
@@ -40,7 +40,7 @@ public interface IExecutionSegment
 		/// For example <code>foo.bar</code> will be stored as <code>foo_bar</code> in Elasticsearch.
 		/// </summary>
 		/// <exception cref="ArgumentException"><c>null</c> as key is not allowed.</exception>
-		Dictionary<string, string> Labels { get; }
+		Dictionary<string, object> Labels { get; }
```
gregkalapos, line 43: This isn't very nice, but I want to keep backwards compatibility - this way things like (current API usage) this will still compile and work:

```
ISpan.Labels["foo"] = "bar"
``` 

Long term I feel it'd be better to just depreciate the `Dictionary` from the API.

russcam: Could a `Tag` (from what I see, tag is what labels really are?) type be introduced, that constrains the inputs only to 

- string
- bool
- double
- int
- long
- decimal

For example,

```c#
[JsonConverter(typeof(TagConverter))]
public class Tag
{
	internal object _object;
	
	public Tag(string value) => _object = value;
	
	public Tag(bool value) => _object = value;
	
	public Tag(double value) => _object = value;
	
	public Tag(int value) => _object = value;
	
	public Tag(long value) => _object = value;
	
	public Tag(decimal value) => _object = value;
	
	public static implicit operator Tag(string value) => new Tag(value);
	
	public static implicit operator Tag(bool value) => new Tag(value);
	
	public static implicit operator Tag(double value) => new Tag(value);
	
	public static implicit operator Tag(int value) => new Tag(value);
	
	public static implicit operator Tag(long value) => new Tag(value);
	
	public static implicit operator Tag(decimal value) => new Tag(value);
}

public class TagConverter : JsonConverter
{
	public override bool CanConvert(Type objectType) => typeof(Tag).IsAssignableFrom(objectType);

	public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
	{
		switch(reader.TokenType)
		{ 
			case JsonToken.Boolean:
				return new Tag((bool)reader.Value);
			case JsonToken.Float:
				return new Tag((double)reader.Value);
			case JsonToken.Integer:
				return new Tag((long)reader.Value);
			case JsonToken.String:
				return new Tag((string)reader.Value);
			default:
				throw new JsonException($"cannot deserialize {nameof(Tag)} from token type {reader.TokenType}");
		}
	}

	public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer) =>
		writer.WriteValue(((Tag)value)._object);
}
```

This would maintain API compatibility and also constrain tags/labels to the types supported by APM server

src/Elastic.Apm/Model/Span.cs
```diff
@@ -406,5 +408,23 @@ private Destination DeduceHttpDestination()
 				return null;
 			}
 		}
+
+		public void SetLabel(string key, string value)
+			=> Context.Labels.Add(key, value);
```
russcam, line 413: should this be assignative i.e. use the indexer?

gregkalapos: Right, fixed this - and also added a test which covers this.

src/Elastic.Apm/Api/IExecutionSegment.cs
gregkalapos: I really liked this idea and I also implemented it (as discussed on zoom I named the class `Label`, but I use the same idea -  and I pushed it to this PR.

I ran into one (I think) fairly big problem: This indeed makes it a nice API and it's backward-compatible with things like `ISpan.Labels["foo"] = "bar"`, but once you want to read the value it becomes a breaking change:

Prior to this, here is how you read the value:
```
var myValue = `ISpan.Labels["foo"] ; // here `myValue` is a string
```

But once we add the `Label` type, `myValue` won't be a `string` anymore, it'll be a `Label` and in order to read the value you need to write:

```
var myValue = `ISpan.Labels["foo"].Value
```

This is definitely not the standard use-case (I expect users just to write into the `Labels` and never really read it in the code later), nevertheless this was totally possible with the old API and we'd break that.  ...and this change also breaks our tests from the same reason, so we also need to assert with `.Value` once we introduce the new `Label` type.

So, right now the suggestion is contained in the PR, but I have slight preference to not add this class because of this problem. Sad to not have the cool API that'd promote the possible types, but I'd rather not break the reading use case.  Any idea or thoughts on this?

russcam: One could look at it from the perspective that the agent's API does not conform to the APM server spec for labels, because it only supports strings, and thus this is a bug that should be addressed. It is unfortunate that the fix is a breaking change.

I think a working around of this breaking change is probably not worth the effort and adds more complexity to the API. For example, an implicit operator of `Label` -> `string` could be added

```c#
public class Label
{
    public static implicit operator string(Label value) => value.Value.ToString();
}
```

`myValue` would still be of type `Label`, but would convert to `string` in cases where a `string` value is expected e.g.

```c#
void Main()
{
	var foo = new Foo();
	
	foo.Labels.Add("label_1", "one");
	foo.Labels.Add("label_2", 2);
	
        // conversion to string on assignment
	string string1 = foo.Labels["label_1"];
	Write(string1);
	
        // conversion to string on pass to method
	var label1 = foo.Labels["label_1"];
	Write(label1);
}

public void Write(string value) => Console.WriteLine(value);

public class Foo
{
	private Dictionary<string, Label> _labels = new Dictionary<string, Label>();	
	public Dictionary<string, Label> Labels => _labels;
}
```

What makes me uncomfortable with this approach is that any value `Label` holds would be converted to string, rather than handled as the type it is.

I personally think it would be better to introduce the breaking change, and bump the major version, if the value of supporting multiple types for label values is important; I suspect that this might be useful to have for performing certain types of queries such as range queries on numeric values.

gregkalapos: We had a chat about this over zoom. We concluded that actually all the suggestions are breaking changes (no matter if it's  `Dictionary<string,object>` or `Dictionary<string,Label>`, so we decided that I'll do another round on this and try to keep the API as `Dictionary<string,string>` - that's to only way to be sure we don't break user code.

I implemented this and pushed to this PR.

src/Elastic.Apm/Model/LabelsDictionary.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace Elastic.Apm.Model
+{
+	/// <summary>
+	/// The instance of this type holds the <see cref="Dictionary{TKey,TValue}" /> that contains all the labels on a given
+	/// event.
+	/// The public interface of this mimics a Dictionary with <see cref="string" /> as both the key and the value.
+	/// The reason for this is backwards compatibility - this type makes sure that we don't break user that rely on the old
+	/// interface.
+	/// </summary>
+	internal class LabelsDictionary : Dictionary<string, string>
```
gregkalapos, line 18: Originally I thought we can shadow the `int this[string key]` indexing operator with the `new` keyword. Unfortunately this does not help, because the API surface works with `Dictionary<string,string>`.

On `Transaction.cs` we have:

```
public Dictionary<string, string> Labels => Context.Labels;
````

and `Context.Labels` is the instance of this `LabelsDictionary`. 

But if we use the `new` keyword to shadow, then `Transaction.Labels["foo"] = "bar"` will hit the `int this[T key]` operator of `Dictionary` and not the one that we'd implement here. 

So, the only option I could come up with is to have 2 dictionaries: 1 for the items used through the `Dictionary<string,string>` API and 1 for the ones added through the `SetLabel` method - and we merge them before serialization.

**Comment by gregkalapos** [comment]
@russcam this is ready for another review.

The `IIS Tests` stage in CI does not seem to run - it's queued for a Windows machine since ~15 minutes and nothing happens. I suspect this is a CI-infra issue. All other stages are green.

File: docs/public-api.asciidoc [hunk 4]
```diff
@@ -248,9 +248,45 @@ finally
...
+Indexed means the data is searchable and aggregatable in Elasticsearch.
+Multiple labels can be defined with different key-value pairs.
+
+* Indexed: Yes
+* Elasticsearch type: {ref}/object.html[object]
+* Elasticsearch field: `labels` (previously `context.tags` in <v.7.0)
+
+Label values can be a string, boolean, or number.
+Because labels for a given key are stored in the same place in Elasticsearch, all label values of a given key must have the same data type.
+Multiple data types per key will throw an exception, e.g. `{foo: bar}` and `{foo: 42}`
```
bmorelli25, line 265: It looks like you copied this from the Java Agent, who copied it from the APM Overview. . . and they forgot the period 😅 
```suggestion
Multiple data types per key will throw an exception, e.g. `{foo: bar}` and `{foo: 42}`.
```

docs/public-api.asciidoc
```diff
@@ -248,9 +248,45 @@ finally
 ----
 
 
+[float]
+[[api-transaction-set-label]]
+==== `void SetLabel(string key, T value) added[1.7.0,Number and boolean labels require APM Server 6.7]
```
bmorelli25, line 253: Maybe add a plus here (and also the closing backtick)

```suggestion
==== `void SetLabel(string key, T value)` added[1.7.0,Number and boolean labels require APM Server 6.7+]
```

docs/public-api.asciidoc
```diff
@@ -473,10 +509,37 @@ finally
...
+----
+
+* `String key`:   The tag key
+* `String|Number|bool value`: The tag value
 
 [float]
 [[api-span-tags]]
 ==== `Dictionary<string,string> Labels`
+
+WARNING: This property is obsolete and will be be removed in a future version. Use the <<api-span-set-label>> method instead.
```
bmorelli25, line 541: This looks funky when rendered. Let's add a short title to make it cleaner
```suggestion
WARNING: This property is obsolete and will be be removed in a future version. Use the <<api-span-set-label,`void SetLabel()`>> method instead.   
```

bmorelli25 (approved):
A few recommendations, then docs LGTM.

docs/public-api.asciidoc
```diff
@@ -248,9 +248,45 @@ finally
...
+----
+
+* `String key`:   The tag key
+* `String|Number|bool value`: The tag value
+
 [float]
 [[api-transaction-tags]]
 ==== `Dictionary<string,string> Labels`
+
+WARNING: This property is obsolete and will be be removed in a future version. Use the <<api-transaction-set-label>> method instead.
```
russcam, line 288:
```suggestion
WARNING: This property is obsolete and will be be removed in a future version. Use the <<api-transaction-set-label, `void SetLabel()`>> method instead, which allows setting labels of string, boolean and number. This property remains for now in order to not break binary compatibility, and at serialization time, the values set with `.SetLabel()` are combined with `Labels` to form the set of labels sent to APM server, with values in `Labels` taking precedence. 
```

docs/public-api.asciidoc
```diff
@@ -473,10 +509,37 @@ finally
...
+----
+
+* `String key`:   The tag key
+* `String|Number|bool value`: The tag value
 
 [float]
 [[api-span-tags]]
 ==== `Dictionary<string,string> Labels`
+
+WARNING: This property is obsolete and will be be removed in a future version. Use the <<api-span-set-label,`void SetLabel()`>> method instead.
```
russcam, line 541:
```suggestion
WARNING: This property is obsolete and will be be removed in a future version. Use the <<api-span-set-label, `void SetLabel()`>> method instead, which allows setting labels of string, boolean and number. This property remains for now in order to not break binary compatibility, and at serialization time, the values set with `.SetLabel()` are combined with `Labels` to form the set of labels sent to APM server, with values in `Labels` taking precedence. 
```

test/Elastic.Apm.AspNetCore.Tests/AspNetCoreDiagnosticListenerTest.cs
```diff
@@ -53,20 +52,20 @@ public async Task TestErrorInAspNetCore(bool useOnlyDiagnosticSource)
 
 				capturedPayload.Should().NotBeNull();
 
-				capturedPayload.Transactions.Should().ContainSingle();
+				capturedPayload?.Transactions.Should().ContainSingle();
```
russcam, line 55: The null conditionals aren't needed here IMO, as a null reference will fail the test on [line 53](https://github.com/elastic/apm-agent-dotnet/pull/982/files#diff-7c522236e53c6b6553577390af5e0b16cc269c08530571ec49216b7017ebecd9R53)

test/Elastic.Apm.AspNetCore.Tests/AspNetCoreDiagnosticListenerTest.cs
```diff
@@ -90,24 +89,24 @@ public async Task TestJsonBodyRetrievalOnRequestFailureInAspNetCore(bool useOnly
 				var client = Helper.GetClient(agent, _factory, useOnlyDiagnosticSource);
 
 				var body = "{\"id\" : \"1\"}";
-				var response = await client.PostAsync("api/Home/PostError", new StringContent(body, Encoding.UTF8, "application/json"));
+				await client.PostAsync("api/Home/PostError", new StringContent(body, Encoding.UTF8, "application/json"));
 
 				capturedPayload.Should().NotBeNull();
 
-				capturedPayload.Transactions.Should().ContainSingle();
+				capturedPayload?.Transactions.Should().ContainSingle();
```
russcam, line 96: The null conditionals aren't needed here IMO, as a null reference will fail the test on [line 94](https://github.com/elastic/apm-agent-dotnet/pull/982/files#diff-7c522236e53c6b6553577390af5e0b16cc269c08530571ec49216b7017ebecd9R94)

src/Elastic.Apm/Api/Context.cs
```diff
@@ -4,25 +4,29 @@
 
 using System;
 using System.Collections.Generic;
+using Elastic.Apm.Model;
 using Elastic.Apm.Report.Serialization;
 using Newtonsoft.Json;
 
 namespace Elastic.Apm.Api
 {
 	public class Context
```
russcam, line 13: Should `Context` expose `SetLabel()` methods in order to set labels of different types, that `Transaction.SetLabel()` methods should call? In other words, are there usages where labels might be set on `Context` outside of the scope of a `Transaction`? There are some in unit tests

src/Elastic.Apm/Api/Context.cs
```diff
@@ -4,25 +4,29 @@
...
 		[JsonConverter(typeof(CustomJsonConverter))]
 		public Dictionary<string, string> Custom => _custom.Value;
 
+		internal Lazy<LabelsDictionary> InternalLabels = new Lazy<LabelsDictionary>();
+
 		/// <summary>
 		/// <seealso cref="ShouldSerializeLabels" />
 		/// </summary>
 		[JsonProperty("tags")]
 		[JsonConverter(typeof(LabelsJsonConverter))]
```
russcam, line 26: I don't think this should have converter type `LabelsJsonConverter`. The `[JsonConverter(typeof(LabelsJsonConverter))]` can be applied to `LabelsDictionary` and removed from label properties on `Context`, `Metadata`, `SpanContext` e.g.

```c#
	[JsonConverter(typeof(LabelsJsonConverter))]
	internal class LabelsDictionary : Dictionary<string, string>
    {
        // ...
    }
```

It achieves the same outcome and removes the need to attribute every labels property

src/Elastic.Apm/Api/Label.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Licensed to Elasticsearch B.V under
+// one or more agreements.
+// Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information
+
+namespace Elastic.Apm.Api
+{
+	public class Label
```
russcam, line 8: Should add some XML documentation to indicate what this represents, and highlight the implicit converion from supported values

src/Elastic.Apm/Model/Span.cs
```diff
@@ -406,5 +412,23 @@ private Destination DeduceHttpDestination()
 				return null;
 			}
 		}
+
+		public void SetLabel(string key, string value)
```
russcam, line 416: I wonder if we should have only a single `SetLabel(string key, Label value)` method, and delegate the "overloads" to the supported implicit conversions of `Label`. It would alleviate the need to maintain 6 overloads in all places that support setting labels. The negative is the somewhat opaqueness of implicit conversions, with the overloads acting to promote the implicit conversions supported. 

Both the following would work with a single `SetLabel(string key, Label value)` method

```c#
SetLabel("foo", "bar");
```

and

```c#
SetLabel("foo", new Label("bar"));
```

src/Elastic.Apm/Api/SpanContext.cs
```diff
@@ -5,14 +5,15 @@
 using System;
 using System.Collections.Generic;
 using Elastic.Apm.Helpers;
+using Elastic.Apm.Model;
 using Elastic.Apm.Report.Serialization;
 using Newtonsoft.Json;
 
 namespace Elastic.Apm.Api
 {
 	public class SpanContext
```
russcam, line 14: Should `SpanContext` expose `SetLabel()` methods in order to set labels of different types, that `Span.SetLabel()` methods should call? In other words, are there usages where labels might be set on `SpanContext` outside of the scope of a `Span`? There are some in unit tests

russcam (commented):

test/Elastic.Apm.AspNetCore.Tests/AspNetCoreDiagnosticListenerTest.cs
gregkalapos: I totally agree with you, but Rider is not smart enough and gives a warning if we don't care about the potential `null`:

image

I'd keep the `?.` that way we don't have warnings.

test/Elastic.Apm.AspNetCore.Tests/AspNetCoreDiagnosticListenerTest.cs
gregkalapos: Same as in https://github.com/elastic/apm-agent-dotnet/pull/982#discussion_r515686044

src/Elastic.Apm/Api/Context.cs
gregkalapos: > Should Context expose SetLabel() methods in order to set labels of different types, that Transaction.SetLabel() methods should call? In other words, are there usages where labels might be set on Context outside of the scope of a Transaction?

I don't think so - there are 2 use cases when you set labels for `Context`:

- with transactions and for those we have public `SetLabels()` methods
- for errors and in that case we also have a public API to pass labels: `ITransaction.CaptureError("Test Error", "TestMethod", new StackTrace().GetFrames(), labels: new Dictionary<string, Label> { { "stringLabel", "test" } });` . In case of errors the `Context` isn't exposed, because everything is filled when the error is captured on the fly - so no way to change an `Error.Context` after an error happened (expect with a filter).

I think exposing `Labels` on the `Context` wasn't a good idea because it just duplicates the API that we already have on `ITransaction` (and on the `CaptureError`/`CaptureException` methods) which is confusing: it's literally the same value with a duplicated API.

So overall, I think the cleanest would be to just keep the `.Labels` property with the obsolete attribute for backwards compatibility and then remove the whole Label thing from `Context`. 

> There are some in unit tests

I think I rewrote all of them to use the new public API (except the ones for backwards compatibility). In `SerializationTests` we have tests that new up a `Context` and serialize it and in those we use the new `internal InternalLabels` property. Those unit tests aren't real API use cases, in those cases there is no `Transaction` or `Error`, the tests only care about the serialization.

src/Elastic.Apm/Api/Context.cs
gregkalapos: Makes sense - I implemented it. Only thing is that if we do this then we'll serialize labels through the `InternalLabels` property instead of the `Labels` property and `InternalLabels` is `Lazy<LabelsDictionary>`  which again needs another converter. But overall I think it's still better that way.

src/Elastic.Apm/Api/Label.cs
gregkalapos: Agreed - added XML doc.

src/Elastic.Apm/Model/Span.cs
gregkalapos: Hmm... I personally prefer having baked in types on the API surface. If people see `SetLabel(string, bool)` and the other overloads they'll immediately know what types they can pass in. I'd keep it for now.

src/Elastic.Apm/Api/SpanContext.cs
gregkalapos: Same as https://github.com/elastic/apm-agent-dotnet/pull/982#discussion_r515903177

test/Elastic.Apm.AspNetCore.Tests/AspNetCoreDiagnosticListenerTest.cs
russcam: unfortunate to code in a specific way to satisfy tooling 😃 I had a quick look at seeing if null-checking options can be configured for this, but doesn't look so. Maybe wrapping all in a `if(capturedPayload != null) { ... }` satisfies the warning _and_ only does the check once?

src/Elastic.Apm/Api/Context.cs
russcam: got it, thanks

src/Elastic.Apm/Api/Context.cs
russcam: I think we might be talking about slightly different things. I suggested to move the `[JsonConverter(typeof(LabelsJsonConverter))]` on all label properties to the `LabelsDictionary` type. With this change, wherever `LabelsDictionary` is used, the converter will be used. I _think_ you're talking about attributing `InternalLabels`, which I don't think we should do, or need to do. The public API of the `"tags"` property is still `Dictionary<string, string>`, but because the specific implementation of the instance returned is `LabelsDictionary` that has an associated `JsonConverter`, the converter will be used. I think it's a little unituitive to be honest, but allows keeping the lazy semantics and moves the association of `LabelsDictionary` and `LabelsJsonConverter` to one place.

src/Elastic.Apm/Model/Span.cs
russcam: ok. Definitely more discoverable with overloads. I've maybe been thinking too much in Rust lately!

src/Elastic.Apm/Api/Context.cs
gregkalapos: Oh, I think now I got it. I just pushed a commit: https://github.com/elastic/apm-agent-dotnet/pull/982/commits/c1eb1680418b85f7b193cce051aebaa6e3082917 

Is that what you mean? 

I kind of agree it's a bit unituitive and the other thing is that we'll serialize labels through an obsolete property. But overall I'm fine with it. Plus I agree it's nice to have the association of `LabelsDictionary` and `LabelsJsonConverter` at one place

test/Elastic.Apm.AspNetCore.Tests/AspNetCoreDiagnosticListenerTest.cs
gregkalapos: Using `if` block would definitely work - but that's also coding in a specific way to satisfy tooling - Ultimately I would not null check here (as we agree it's not needed), but the warnings drive me crazy :)

I don't have strong feelings and I'd be happy to push a change to use `if`s - only thing is that I use `?.` for this in lots of other places.

Again, I'm fine with both `?.` and `if` blocks, but mixing the two would be maybe nice to avoid. Would you be ok with maybe changing this in a separate PR at all places in case we really want to do so?

test/Elastic.Apm.Tests/LabelTests.cs
```diff
@@ -0,0 +1,287 @@
...
+using System.Linq;
+using Elastic.Apm.Api;
+using Elastic.Apm.Tests.Mocks;
+using FluentAssertions;
+using Newtonsoft.Json;
+using Xunit;
+
+namespace Elastic.Apm.Tests
+{
+	public class LabelTests
```
russcam, line 18: The tests should use the payload serializer instead of `JsonConvert.SerializeObject()`, It better represents the agent's serialization and means the tests don't need to change when swapping to a different serializer in the future

src/Elastic.Apm/Report/Serialization/LabelsJsonConverter.cs
```diff
@@ -23,16 +23,26 @@ public override void WriteJson(JsonWriter writer, Dictionary<string, string> lab
...
 			}
 			writer.WriteEndObject();
 		}
 
-		public override Dictionary<string, string> ReadJson(JsonReader reader, Type objectType, Dictionary<string, string> existingValue,
+		public override LabelsDictionary ReadJson(JsonReader reader, Type objectType, LabelsDictionary existingValue,
 			bool hasExistingValue, JsonSerializer serializer
 		)
-			=> serializer.Deserialize<Dictionary<string, string>>(reader);
+			=> serializer.Deserialize<LabelsDictionary>(reader);
```
russcam, line 46: This causes a stackoverflow as `Deserialize<LabelsDictionary>` ends up calling the converter's `ReadJson` implementation

russcam, line 46:
```suggestion
		{
			if (reader.TokenType == JsonToken.Null)
				return null;

			if (reader.TokenType != JsonToken.StartObject)
				throw new JsonException($"expected {JsonToken.StartObject} but received {reader.TokenType}");

			var labels = new LabelsDictionary();

			while (reader.Read())
			{
				if (reader.TokenType == JsonToken.EndObject)
					break;

				var property = (string)reader.Value;
				reader.Read();

				switch (reader.TokenType)
				{
					case JsonToken.Integer:
						labels.InnerDictionary.Add(property, (long)reader.Value);
						break;
					case JsonToken.Float:
						labels.InnerDictionary.Add(property, (double)reader.Value);
						break;
					case JsonToken.String:
						labels.Add(property, (string)reader.Value);
						break;
					case JsonToken.Boolean:
						labels.InnerDictionary.Add(property, (bool)reader.Value);
						break;
					default:
						throw new JsonException(
							$"Expected {JsonToken.Integer}, {JsonToken.Float}, {JsonToken.String} or {JsonToken.Boolean} "
							+ $"but received {reader.TokenType}");
				}
			}

			return labels;
		}
```

test/Elastic.Apm.AspNetCore.Tests/AspNetCoreDiagnosticListenerTest.cs
russcam: for sure! Totally not needed for this PR

russcam (changes requested):
I had another review and left some comments.

I had a brief look and don't think we have tests for deserializing `LabelsDictionary` either directly, or via `Context` (which would test the deserialization through the `Dictionary<string,string>` property type). It might be prudent to have these tests

src/Elastic.Apm/Report/Serialization/LabelsJsonConverter.cs
russcam: One question that came to mind - do labels need to support dates? Understand that they will be serialized as a JSON string, but wondering if deserialization has to consider `JsonToken.Date`

gregkalapos: > do labels need to support dates? 

No, my understanding is that no other agent supports dates. It's only string, numbers, and bool. I guess technically we could support dates, but so far I followed other agents and those don't support it. Before we add it I'd definitely see if there is feedback and need for it at al.

>  wondering if deserialization has to consider `JsonToken.Date`

Since the de-serialization logic is not used in production and I feel date isn't needed to be supported I think it's fine to ignore it. As discussed over zoom, the reason I'm not very worried about the de-serialization logic is that the agent only uses the serialization logic to go from C# objects to JSON, but not the other way, so if de-serialization is important at al, then it's maybe in some of our internal tests scenarios (but currently that's not even a case based on what we discussed (Mock APM Server e.g. does not use this)).

test/Elastic.Apm.Tests/LabelTests.cs
gregkalapos: Makes sense. Turns out we have `SerializationTests.cs` which does exactly that - now `LabelTests.cs` does it as well.

gregkalapos:
> 
> 
> I had another review and left some comments.
> 
> I had a brief look and don't think we have tests for deserializing `LabelsDictionary` either directly, or via `Context` (which would test the deserialization through the `Dictionary<string,string>` property type). It might be prudent to have these tests

I did another round on this PR. 

- As suggested now we test with `PayloadItemSerializer` instead of `JsonConvert.SerializeObject`. 
- Kept the `LabelsJsonConverter.ReadJson` method - I took the suggested implementation. As discussed there is no production use-case for it - we never serialize in a real prod. agent. 
- Also added a full framework test which tests the labels and I also switched the `.Labels` to use `LabelsDictionary` there. With this at least in the MockApmServer we use the de-serialization logic, plus the API is also closer to the real agent with that change.

test/Elastic.Apm.Tests/LabelTests.cs
```diff
@@ -0,0 +1,295 @@
...
+			var mockPayloadSender = new MockPayloadSender();
+			using var agent = new ApmAgent(new AgentComponents(payloadSender: mockPayloadSender));
+
+			agent.Tracer.CaptureTransaction("test", "test", t =>
+			{
+				t.CaptureSpan("testSpan", "test", s =>
+				{
+					var labelName = "myLabel";
+					SetLabel(s, labelValue, labelName);
+					var jsonString = new PayloadItemSerializer(new MockConfigSnapshot()).SerializeObject(s);
```
russcam, line 67:
```suggestion
					var jsonString = SerializePayloadItem(s);
```

russcam (approved):
LGTM - great effort, @gregkalapos 👍 

I left one small suggestion to use the serialize method in one of the tests

test/Elastic.Apm.Tests/LabelTests.cs
gregkalapos: Oh, good catch - this is a leftover from our zoom discussion and replace-all of course did not find this.

PR #984, opened 2020-10-20
APM Server specification validation

russcam:
This PR introduces validation against the APM server JSON schema specification. Validation is performed in two ways:

1. Validate agent types against the schema through reflection.

   It's possible to statically validate a large part of the agent types against the JSON schema. Where it is not possible is where agent types model the APM server specification with a structure that cannot be statically mapped to the schema properties, typically because such types have a custom JSON serialization routine that transforms instances of types to a JSON structure that matches the specification.

2. Validate JSON input to mock APM server against the JSON schema.

This PR makes changes to the agent types in order for them to exactly match the schema. For example, the `sampled` field can be `"null"` or `"boolean"` within the specification, but the agent models as a `bool`. In chatting to @gregkalapos, it might make sense to model the specification as a set of POCOs, and leave the agent types as _an implementation_ of the spec as opposed to an _exact_ implementation of the spec. In modelling the spec as POCOs, these could be used outside of the agent, for example, to transform some output to an APM server input, or used by the mock APM server.


russcam:
As discussed with @gregkalapos, I've reverted the changes to the Agent types that transformed it into a complete reference implementation of the APM server schema specs. The validator still has the ability to perform this validation, with the test to perform this marked as skip for now.

This is now ready for review.

src/Elastic.Apm/Api/IError.cs
```diff
@@ -7,6 +7,7 @@ namespace Elastic.Apm.Api
 	/// <summary>
 	/// Represents an error which was captured by the agent.
 	/// </summary>
+	[Specification("docs/spec/errors/error.json")]
```
gregkalapos, line 10: In https://github.com/elastic/apm-server/pull/4427 this got change. I adapted the path locally, but then I run into another problem (comment below with the line).

src/Elastic.Apm.Specification/Validator.cs
```diff
@@ -0,0 +1,657 @@
...
+								ValidateEnum(specType, schema, schemaProperty, specTypeProperty, result);
+							else
+								ValidateString(specType, schema, schemaProperty, specTypeProperty, result);
+						}
+						else if (schemaProperty.Type.HasFlag(JsonObjectType.Object))
+							ValidateProperties(specTypeProperty.PropertyType, schemaProperty, result);
+						else if (schemaProperty.Type.HasFlag(JsonObjectType.Array))
+						{
+							if (IsEnumerableType(specTypeProperty.PropertyType, out var elementType))
+								ValidateProperties(elementType, schemaProperty.Item.ActualSchema, result);
```
gregkalapos, line 505: In case of `elementType=Elastic.Apm.Api.CapturedStackFrame` `schemaProperty.Item`  is `null`. This happens on the `Error` validation.

gregkalapos (changes requested):
https://github.com/elastic/apm-server/pull/4427 made some changes on where `json` files are. I tried to adapt this PR to the changes, but then I run into some `NullReferenceException`s.

But overall, this looks very nice! 👍

src/Elastic.Apm.Specification/Validator.cs
russcam: This was a bug in the spec and has been fixed in https://github.com/elastic/apm-server/issues/4515

src/Elastic.Apm/Api/IError.cs
russcam: I've updated to the v2 specs

src/Elastic.Apm.Specification/Validator.cs
```diff
@@ -0,0 +1,657 @@
+﻿using System;
```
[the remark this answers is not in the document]
russcam: done!

gregkalapos (approved):

PR #986, opened 2020-10-21
update DiagnosticInitializer, fix possible memory leak

WeihanLi:
The original code may cause memory leak when there're multiple `IDiagnosticListener` with the same `Name`


WeihanLi:
CLA already signed

PR #989, opened 2020-10-22
Improve metricset serialization

russcam:
This commit improves the metricset serialization by
writing directly to the JSON writer instead of
allocating instances of JObject.

Add deserialization implementation and unit tests
to assert serialization behaviour.

### DIFF
--- a/src/Elastic.Apm/Api/IMetricSet.cs
+++ b/src/Elastic.Apm/Api/IMetricSet.cs
@@ -16,6 +16,7 @@ namespace Elastic.Apm.Api
 		/// </summary>
 		IEnumerable<MetricSample> Samples { get; set; }
 
+		// TODO: Rename to Timestamp for consistency?
 		/// <summary>
 		/// Number of milliseconds in unix time
 		/// </summary>
--- a/src/Elastic.Apm/Report/Serialization/MetricSetConverter.cs
+++ b/src/Elastic.Apm/Report/Serialization/MetricSetConverter.cs
@@ -3,6 +3,9 @@
 // See the LICENSE file in the project root for more information
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
+using Elastic.Apm.Api;
 using Elastic.Apm.Metrics;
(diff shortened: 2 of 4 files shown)


src/Elastic.Apm/Report/Serialization/MetricSetConverter.cs
```diff
@@ -13,25 +16,86 @@ internal class MetricSetConverter : JsonConverter<MetricSet>
...
 			{
-				var valueObj = new JObject { { "value", item.KeyValue.Value } };
-
-				if (!samples.ContainsKey(item.KeyValue.Key))
-					samples.Add(item.KeyValue.Key, valueObj);
+				if (addedKeys.Add(item.KeyValue.Key))
+				{
+					writer.WritePropertyName(item.KeyValue.Key
+						.Replace('*', '_')
+						.Replace('"', '_'));
```
russcam, line 30: This should be sanitised according to the spec: https://github.com/elastic/apm-server/blob/7cf56b584c9ff94153018afd1758f27737628873/docs/spec/metricsets/metricset.json#L11-L22

russcam:
jenkins please run the tests

src/Elastic.Apm/Api/IMetricSet.cs
```diff
@@ -16,6 +16,7 @@ public interface IMetricSet
 		/// </summary>
 		IEnumerable<MetricSample> Samples { get; set; }
 
+		// TODO: Rename to Timestamp for consistency?
```
gregkalapos, line 19: This interface is public, but so far we don't expose anything to collect custom metrics, so if we rename, then I'd say we should do it soon and not wait until this may be exposed more.

gregkalapos (approved):
LGTM, thanks!

src/Elastic.Apm/Api/IMetricSet.cs
russcam: ++ I'll open another PR to address

PR #990, opened 2020-10-22
Introduce annotations on Agent types

russcam:
This PR introduces annotations onto the agent types to describe the constraints/properties of the APM server specification:

- Replace use of `[JsonConverter(typeof(TrimmedStringJsonConverter))]` with a `[MaxLength]` attribute, and assign the JsonConverter in `ElasticApmContractResolver`.
- Introduce `RequiredAttribute` and mark required properties as such.
- Rename `TrimmedStringJsonConverter` to `TruncateJsonConverter`

Discussed with @gregkalapos about using [Data Annotations ](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations?view=netcore-3.1) for this purpose, as it contains some of the attributes that are needed and provides types to validate instances against their attributes, something that may be useful for folks that use the agent API for manual instrumentation.


russcam:
We did also talk about replacing `JsonPropertyAttribute` with `DataMemberAttribute` as JSON.Net supports the latter, and it is serializer agnostic, with the intention that it would make it easier to move/support to System.Text.Json in future. Currently, System.Text.Json does not support  `DataMemberAttribute` (https://github.com/dotnet/runtime/issues/29975), so have held off on doing so.

src/Elastic.Apm/Api/Constraints/RequiredAttribute.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using System;
+
+namespace Elastic.Apm.Api.Constraints
+{
+	/// <summary>
+	/// Specifies that a data field value is required.
+	/// </summary>
+	[AttributeUsage(AttributeTargets.Property)]
+	public sealed class RequiredAttribute : Attribute { }
```
gregkalapos, line 14: Just out of interest - do we already have a test on the other PR that makes sure the usage of this attribute is in-sync with the APM Server schema?

gregkalapos (approved):
Nice, thanks!

src/Elastic.Apm/Api/Constraints/RequiredAttribute.cs
russcam: We do not because it's a chicken/egg scenario in that we only know what is required in the schema but not on the types. The schema required fields are validated in #984 in the mock APM server when receiving integration test data.

russcam:
Failing test

```
Elastic.Apm.Tests.ApiTests.ConvenientApiTransactionTests.ConvenientApiTransactionTests.Elastic.Apm.Tests.ApiTests.ConvenientApiTransactionTests.AsyncTask

Error Message
Expected value to be greater or equal to 1.0, but found 0.976.
```

russcam:
jenkins run the tests please

Issue #985, opened 2020-10-20
Failing tests with `Index was outside the bounds of the array` from coverlet 


gregkalapos:
One example can be seen [here](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-dotnet%2Fapm-agent-dotnet-mbp/detail/PR-983/1/pipeline).

This usually happens on Windows in the .NET Core pipeline after a test run with logs like this:

```
[2020-10-20T07:16:49.489Z] Build FAILED.

[2020-10-20T07:16:49.489Z] 

[2020-10-20T07:16:49.489Z]        "C:\Users\jenkins\workspace\tnet_apm-agent-dotnet-mbp_PR-983\apm-agent-dotnet\ElasticApmAgent.sln" (VSTest target) (1) ->

[2020-10-20T07:16:49.489Z]        "C:\Users\jenkins\workspace\tnet_apm-agent-dotnet-mbp_PR-983\apm-agent-dotnet\test\Elastic.Apm.Grpc.Tests\Elastic.Apm.Grpc.Tests.csproj" (VSTest target) (28) ->

[2020-10-20T07:16:49.489Z]        (GenerateCoverageResult target) -> 

[2020-10-20T07:16:49.489Z]          C:\Users\jenkins.PACKER-5DA68B40\.nuget\packages\coverlet.msbuild\2.5.1\build\netstandard2.0\coverlet.msbuild.targets(33,5): error : Index was outside the bounds of the array. [C:\Users\jenkins\workspace\tnet_apm-agent-dotnet-mbp_PR-983\apm-agent-dotnet\test\Elastic.Apm.Grpc.Tests\Elastic.Apm.Grpc.Tests.csproj]

```

Based on some google-ing this is very likely related to our gRPC project and the code generation based on `.proto`: https://github.com/coverlet-coverage/coverlet/issues/167 

cc @russcam

russcam:
#991 has been merged. I'll keep this open for a bit, and close if we don't see it again

russcam:
Not seen any failing CI runs for the past week so closing this.

PR #991, opened 2020-10-22
Update coverlet.msbuild to latest

russcam:
Relates: #985


gregkalapos (approved):
Thanks. 

@v1v fyi we had some failing tests caused by coverlet - see https://github.com/elastic/apm-agent-dotnet/issues/985. Hopefully this will fix it, so I guess at this point no action needed from you.

russcam:
I'm not 100% sure that this will fix the issue, but the reference coverlet issues have been fixed in a later release than the current one we've been using, so updating the version will hopefully resolve!

PR #992, opened 2020-10-27
Add HttpContextCurrentExecutionSegmentsContainer

### LINKED ISSUE #934: Agent.Tracer.CurrentTransaction is null inside IDispatchMessageInspector
**Describe the bug**
The transaction created by the HttpModule is not carried to the IDispatchMessageInspector in a WCF service.  I do see the Agent.Tracer.CurrentTransaction in the Application_BeginRequest:

		protected void Application_BeginRequest(object sender, EventArgs e)
		{
			//this is testing that we can get the current transaction
			var transaction = Agent.Tracer.CurrentTransaction;
			if (transaction != null)
			{
				HttpContext.Current.Items.Add("CurrentTransaction", transaction);
			}
		}

But when we get to the AfterReceiveRequest in a IDispatchMessageInspector the Agent.Tracer.CurrentTransaction is null and i'm forced  to grab it from the HttpContext i created in the BeginRequest:

		public object AfterReceiveRequest(ref Message request, IClientChannel channel, InstanceContext instanceContext)
		{
			if (request.IsEmpty)
				throw new ArgumentNullException(nameof(request));
			try
			{
				if (HttpContext.Current != null)
				{
					var currentTransaction = (ITransaction)HttpContext.Current.Items["CurrentTransaction"];
					//var currentTransaction1 = Agent.Tracer.CurrentTransaction;  always null!!!
					if (currentTransaction != null)
					{
						_currentSpan = currentTransaction.StartSpan(_apmName,
							"SoapRequest", serviceInstance.GetType().Name, operationName);
						_isNewSpan = true;
					}
				}
			}
			catch (Exception exception)
			{
				throw;
			}

			return null;
		}
Which does create a new span, but Agent.Tracer.CurrentTransaction is still null and Agent.Tracer.CurrentSpan has the span i just created.

I have attached a sample WCF application that demonstrates this issue.

**To Reproduce**
1. Use this config '...'
Use the sample app i supplied and debug at the break points
Global.asax.cs line: 48
WcfEndpointBehavior.cs line 40
2. Then call '....'
use WcfTestClient and call GetData(), watch the break points
3. Then do '....'
see that the transaction is not persisted
No error, just no data...

**Expected behavior**
The Agent.Tracer.CurrentTransaction should be available in the IDispatchMessageInspector, so we can add Spans.

[WcfService4.zip](https://github.com/elastic/apm-agent-dotnet/files/5113479/WcfService4.zip)

### LINKED ISSUE #972: Transaction metadata (spans, Custom) does not get created/transferred (I think) when incoming request > threshold, and when the Expect HTTP header present
I am using Elastic.Apm/Elastic.Apm.AspNetFullFramework 1.6.1 in a asp.net WebApi service.

I have noticed cases where a transaction is visible in the Kibana APM app, and does include the standard metadata (HTTP headers, etc), but not the API produced metadata such as the Spans, and the Custom items. From some testing, I think I have identified a couple of conditions that appear to effect this disappearance:
- the presence of the Expect header (which seems to occur by default in HttpClient .. I think) .. e.g. 'Expect: 100-continue'
- when the Content-Length exceeds some threshold, though not very clear on this threshold because it seemed to vary in my test between http and https. In my test, increasing the size of the request from 16K to 18K appeared to have this behavior (with no other mis-behavior in terms of web app function, nor the APM transaction being sent to the APMServer (other than the missing metadata)).

I have also seen other strangeness such as when Labels were populated that the transaction was not visible in Kibana. [Because of this issue, I am using the Custom items rather than Labels]

**To Reproduce**
This mis-behavior occured when the client was using the Expect header, and when the requests exceeded this threshold. I have subsequent reproduced this (in the context of the web app) this misbehavior using curl .. e.g

"c:\program files\curl\bin\curl" -v -H "Expect: 100-continue" -H "Content-Type: application/json" -H "Authorization : Basic xx=" --data-binary "@c:/temp/searchentity.json" http://xxx.xx.net/..

**Expected behavior**
Its very possible some aspect of my APM configuration or usage pattern is not ideal, but the fact that the transaction is registered in the APMServer, with the default metadata, and occasionally with this additional metadata (contingent on the non-presence of the conditions listed above) lead me to think this is primarily a issue with the APM agent.

Is there any logging available from the APM agent? I have not seen much at in the NT event logs.

russcam:
This commit adds an implementation of `ICurrentExecutionSegmentsContainer` that gets/sets the current transaction and span in `HttpContext.Items`, in addition to `AsyncLocal<T>`, making them available within the ExecutionContext of IIS.

`AsyncLocal<T>` are not propagated across the ASP.NET ExecutionContext running in IIS, resulting in the current transaction set in `Application_BeginRequest` returning as null from `Agent.Tracer.CurrentTransaction` in later subsequent IIS events, when such events execute asynchronously. ASP.NET does however propagate the current HttpContext through `HttpContext.Current`, allowing values to be stored in `HttpContext.Items` hashtable and shared across IIS events. The `HttpContextCurrentExecutionSegmentsContainer` uses both `AsyncLocal<T>` and `HttpContext.Items` to get/set the current transaction and span, allowing them to be available in both IIS events and async contexts. `AsyncLocal<T>` is still needed for places where `HttpContext.Current` is null, for example, DiagnosticSourceListeners running on a separate thread in an async flow; in such cases `HttpContext.Current`, which is set/get on `CallContext.HostContext`, is not propagated over async
flows.

It seems it _may_ be possible for the current transaction and current span instances stored in `AsyncLocal<T>` and `HttpContext.Items` to be different in such scenarios where one is available but not the other on beginning
the transaction/span, and where availability is reversed on end. Unsure how likely this is in practice however, and using both `AsyncLocal<T>` and `HttpContext.Items` is the common approach in resolving this issue in < .NET 4.7.1.

For .NET 4.7.1, a new `OnExecuteRequestStep` method has been added to `HttpApplication`, to allow the execution context to be restored/saved:

https://devblogs.microsoft.com/dotnet/net-framework-4-7-1-asp-net-and-configuration-features/#asp-net-execution-step-feature

Closes #934
Closes #972

### DIFF
--- a/sample/AspNetFullFrameworkSampleApp/AspNetFullFrameworkSampleApp.csproj
+++ b/sample/AspNetFullFrameworkSampleApp/AspNetFullFrameworkSampleApp.csproj
@@ -125,6 +125,13 @@
       <SubType>ASPXCodeBehind</SubType>
       <DependentUpon>Webforms.aspx</DependentUpon>
     </Compile>
+    <Compile Include="App_Start\WebApiConfig.cs" />
+    <Compile Include="Controllers\WebApiController.cs" />
+    <Compile Include="Controllers\DatabaseController.cs" />
+    <Compile Include="Models\CreateSampleDataViewModel.cs" />
+    <Compile Include="Mvc\JsonBadRequestResult.cs" />
+    <Compile Include="Mvc\JsonNetValueProviderFactory.cs" />
+    <Compile Include="Mvc\StreamResult.cs" />
   </ItemGroup>
   <ItemGroup Condition="'$(OS)' == 'WINDOWS_NT'">
     <Content Include="Content\bootstrap-grid.css" />
@@ -202,6 +209,9 @@
     <Content Include="Views\Account\ResetPassword.cshtml" />
     <Content Include="Views\Account\ResetPasswordConfirmation.cshtml" />
     <Content Include="Views\Shared\_LoginPartial.cshtml" />
+    <Content Include="Views\Shared\_LoginPartial.cshtml" />
+    <Content Include="Views\Database\Create.cshtml" />
+    <Content Include="Views\Database\Index.cshtml" />
     <Content Include="Areas\MyArea\Views\Home\Index.cshtml" />
     <Content Include="Areas\MyArea\Web.config" />
     <Content Include="Areas\MyArea\_ViewStart.cshtml" />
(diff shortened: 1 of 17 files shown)


gregkalapos (approved):
Nice!

russcam:
jenkins run the tests please

PR #993, opened 2020-10-27
Use Ellipsis

russcam:
This commit uses the ellipsis character over three individual dots, preserving more of the original string if it needs to be truncated.


gregkalapos (approved):

russcam:
jenkins run the tests please

PR #996, opened 2020-10-27
Use Request.Unvalidated to extract request details

russcam:
This commit updates the ElasticApmModule to use
Request.Unvalidated to extract details about the
request.

Using properties such as Request.Path, Request.Url,
etc. directly can trigger ASP.NET's request validation
, if enabled (recommended):

https://docs.microsoft.com/en-us/previous-versions/hh882339(v=vs.100)


gregkalapos (approved):
Nice catch!

Issue #122, opened 2019-02-20
Add ENABLED, RECORDING settings


gregkalapos:
Currently changing the app`s code is the only way to enable/disable automatic creation of transactions/spans.

Other agents have settings for this and the .NET agent should also have these settings.

The two settings:
- active
- instrument

Definition of instrument: https://github.com/elastic/apm/issues/49

SergeyKleyman:
We should consider if setting `ACTIVE` to false should disable everything agent does including connecting to APM Server (in particular that would mean that having `ACTIVE` in configuration received from APM Server AKA remote configuration would be problematic). See discussion of `ACTIVE` at https://github.com/elastic/apm/issues/76

freeranger:
> We should consider if setting `ACTIVE` to false should disable everything agent does 
Can I suggest that this would indeed be the case - per #865, in some cases consumers of our libraries simply have no interest in APM and therefore won't have a server to connect to; in other cases it may simply be that we want to disable it completely when looking at performance issues.

And locally as developers of libraries, it means we don't have to spin up an APM server just to avoid being spammed in the logs ;)

thanks

basepi:
FYI, https://github.com/elastic/apm/issues/49 was closed, so perhaps this can be closed? Or at least amended to only include `enabled` and `recording`.

gregkalapos:
I looked at other agents' docs and also asked about this in Slack. So far none of the configs mentioned here is implemented. According to current alignments across agents we'll add these 2:

- `enabled` : if `false`, the agent is totally turned off (doesn't even instrument, no background threads started, basically the agent does not do anything). This can't be changed, if the agent starts with `enabled=false` then only chance to turn on the agent is to set it to true AND restart the app.
- `recording` : This is live changeable, and even if it's `false`, the agent still listens to central config changes. If it's false  then nothing is collected (no transactions, no errors, not metrics, nothing) - but instrumentation is still in place - it's just not collecting data.

gregkalapos:
Implemented in #997.

billhong-just:
@gregkalapos  

Hi, gregkalapos  
[...]

PR #997, opened 2020-10-27
Add Enabled and Recording config and introduce NoopTransaction and NoopSpan

gregkalapos:
Implements https://github.com/elastic/apm-agent-dotnet/issues/122

Adds the `recording` and `enabled` configs. 

- `enabled` : if `false`, the agent is totally turned off (doesn't even instrument, no background threads started, basically the agent does not do anything). This can't be changed, if the agent starts with `enabled=false` then only chance to turn on the agent is to set it to true AND restart the app.
- `recording` : This is live changeable, and even if it's `false`, the agent still listens to central config changes. If it's false  then nothing is collected (no transactions, no errors, not metrics, nothing) - but instrumentation is still in place - it's just not collecting data.


src/Elastic.Apm/CurrentExecutionSegmentsContainer.cs
```diff
@@ -3,22 +3,22 @@
...
+		public ISpan CurrentSpan
 		{
 			get => _currentSpan.Value;
 			set => _currentSpan.Value = value;
 		}
 
-		public Transaction CurrentTransaction
+		public ITransaction CurrentTransaction
 		{
 			get => _currentTransaction.Value;
```
gregkalapos, line 23: Here we still return `null` if there is no current transaction (same applies to `CurrentSpan`) and this causes a bit of a pain on the current Public API.

Writing code like this isn't safe:

```
agent.Tracer.CurrentTransaction.CaptureSpan("foo", "bar", () =>
{
	//user code to run
});
```

`agent.Tracer.CurrentTransaction` can throw a nullref if the current transaction is null, even worse, accidentally people may use `?.`:

```
agent.Tracer.CurrentTransaction?.CaptureSpan("foo", "bar", () =>
{
	//user code to run
});
```

But this cause the bigger problem that the user code in the lambda won't run if the current transaction is null.

This problem could be fixed if we'd return `NoopTransaction` here. The problem with doing that is that I feel it'd be a major breaking change: right now we force people to null-check here, if we change this to never return `null` then we could cause issues in user code. Same applies to `HttpContextCurrentExecutionSegmentsContainer.cs`.

Therefore I decided that in this PR I don't change this behaviour. Everything is already in place to do this change at some point, but I think best would be to do this on a major version bump.

I'd be happy to hear if this makes sense or I should reconsider.

docs/configuration.asciidoc
```diff
@@ -173,6 +173,55 @@ If this setting is set in both the web.config file and as an environment variabl
...
 
+[float]
+[[config-recording]]
+==== `Recording` (added[1.7.0])
+
+<<dynamic-configuration, image:./images/dynamic-config.svg[] >>
+
+A boolean specifying if the agent should be recording or not.
+When recording, the agent captures HTTP requests, tracks errors and collects and sends metrics.
+When not recording, the agent works as a noop, not collecting data and not communicating with the APM sever,
```
bmorelli25, line 184:
```suggestion
When not recording, the agent works as a noop, not collecting data and not communicating with the APM server,
```

docs/configuration.asciidoc
```diff
@@ -173,6 +173,55 @@ If this setting is set in both the web.config file and as an environment variabl
...
 === Core configuration options
 
+[float]
+[[config-recording]]
+==== `Recording` (added[1.7.0])
+
+<<dynamic-configuration, image:./images/dynamic-config.svg[] >>
+
+A boolean specifying if the agent should be recording or not.
+When recording, the agent captures HTTP requests, tracks errors and collects and sends metrics.
```
bmorelli25, line 183:
```suggestion
When recording, the agent captures HTTP requests, tracks errors, and collects and sends metrics.
```

src/Elastic.Apm/CurrentExecutionSegmentsContainer.cs
russcam: > The problem with doing that is that I feel it'd be a major breaking change: right now we force people to null-check here, if we change this to never return null then we could cause issues in user code. Same applies to HttpContextCurrentExecutionSegmentsContainer.cs.

I'm guessing that folks end up writing code in a way that always runs their code, irrespective of whether there is a transaction or not i.e. if there's a transaction, run code inside it, if not, run anyway. Is there a particular pattern that you foresee, where returning a `NoopTransaction` would cause an issue? I could see one if the action taken is different depending on if there is a transaction or not.

I _think_ this behavioural change feels OK to go into a minor version. There's a chance it could cause issues in some existing user code, but it'd greatly simplify agent usage

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
```diff
@@ -34,21 +33,23 @@ internal class ApmMiddleware
...
 			_tracer = tracer;
 			_logger = agent.Logger.Scoped(nameof(ApmMiddleware));
-			_configurationReader = agent.ConfigurationReader;
 			_agent = agent;
 		}
 
 		public async Task InvokeAsync(HttpContext context)
 		{
-			var transaction = WebRequestTransactionCreator.StartTransactionAsync(context, _logger, _tracer, _agent.ConfigStore.CurrentSnapshot);
+			var newTransaction = WebRequestTransactionCreator.StartTransactionAsync(context, _logger, _tracer, _agent.ConfigStore.CurrentSnapshot);
```
russcam, line 48: Maybe handle the result of `WebRequestTransactionCreator.StartTransactionAsync` as `Transaction`? e.g.

```c#
var transaction = WebRequestTransactionCreator.StartTransactionAsync(context, _logger, _tracer, _agent.ConfigStore.CurrentSnapshot) as Transaction;
```

`newTransaction` is not used unless it's a `Transaction`, so this may simplify the flow a little

src/Elastic.Apm.AspNetCore/DiagnosticListener/AspNetCoreDiagnosticListener.cs
```diff
@@ -41,9 +41,13 @@ public void OnNext(KeyValuePair<string, object> kv)
 				case "Microsoft.AspNetCore.Hosting.HttpRequestIn.Start":
 					if (_httpContextPropertyFetcher.Fetch(kv.Value) is HttpContext httpContextStart)
 					{
-						var transaction = WebRequestTransactionCreator.StartTransactionAsync(httpContextStart, _logger, _agent.TracerInternal,
+						var newTransaction = WebRequestTransactionCreator.StartTransactionAsync(httpContextStart, _logger, _agent.TracerInternal,
```
russcam, line 44: Similar to above comment, converting the result using `as Transaction` might be a little simpler.

src/Elastic.Apm.AspNetFullFramework/HttpContextCurrentExecutionSegmentsContainer.cs
```diff
@@ -15,13 +16,13 @@ namespace Elastic.Apm.AspNetFullFramework
...
+		private readonly AsyncLocal<ISpan> _currentSpan = new AsyncLocal<ISpan>();
+		private readonly AsyncLocal<ITransaction> _currentTransaction = new AsyncLocal<ITransaction>();
 
 		private const string CurrentSpanKey = "Elastic.Apm.Agent.CurrentSpan";
 		private const string CurrentTransactionKey = "Elastic.Apm.Agent.CurrentTransaction";
 
-		public Span CurrentSpan
+		public ISpan CurrentSpan
 		{
 			get => _currentSpan.Value ?? HttpContext.Current?.Items[CurrentSpanKey] as Span;
```
russcam, line 27: should this now be 

```c#
get => _currentSpan.Value ?? HttpContext.Current?.Items[CurrentSpanKey] as ISpan;
```

src/Elastic.Apm.AspNetFullFramework/HttpContextCurrentExecutionSegmentsContainer.cs
```diff
@@ -33,7 +34,7 @@ public Span CurrentSpan
 			}
 		}
 
-		public Transaction CurrentTransaction
+		public ITransaction CurrentTransaction
 		{
 			get => _currentTransaction.Value ?? HttpContext.Current?.Items[CurrentTransactionKey] as Transaction;
```
russcam, line 39: Should this now be

```c#
get => _currentTransaction.Value ?? HttpContext.Current?.Items[CurrentTransactionKey] as ITransaction;
```

src/Elastic.Apm.Elasticsearch/ElasticsearchDiagnosticsSubscriber.cs
```diff
@@ -15,6 +15,10 @@ public class ElasticsearchDiagnosticsSubscriber : IDiagnosticsSubscriber
 		public IDisposable Subscribe(IApmAgent agentComponents)
 		{
 			var composite = new CompositeDisposable();
+
+			if (!agentComponents.ConfigurationReader.Enabled)
```
russcam, line 19: Wonder if a base class implementation should be introduced, that does this check, and then calls into an overridden method that sets up the specific subscribers for the derived diagnostics subscriber. With such a base class, the behaviour can be enforced in one place.

src/Elastic.Apm/Config/ConfigConsts.cs
```diff
@@ -129,33 +136,37 @@ public static class EnvVarNames
...
+
+			//This setting is Full Framework only:
+			public const string FullFrameworkConfigurationReaderType = Prefix + nameof(FullFrameworkConfigurationReaderType);
 			public const string GlobalLabels = Prefix + nameof(GlobalLabels);
 			public const string HostName = Prefix + nameof(HostName);
 			public const string LogLevel = Prefix + nameof(LogLevel);
 			public const string MaxBatchEventCount = Prefix + nameof(MaxBatchEventCount);
 			public const string MaxQueueEventCount = Prefix + nameof(MaxQueueEventCount);
 			public const string MetricsInterval = Prefix + nameof(MetricsInterval);
+			private const string Prefix = "ElasticApm:";
```
russcam, line 167: super minor nitpick: looks like ordering is alphabetical, but I think it should be accessibility -> alphabetical, so that `private` members appear before `public`

src/Elastic.Apm/Config/IConfigurationReader.cs
```diff
@@ -31,6 +31,12 @@ public interface IConfigurationReader
 		/// </summary>
 		IReadOnlyList<WildcardMatcher> DisableMetrics { get; }
 
+		/// <summary>
+		/// Setting to false will completely disable the agent, including instrumentation and remote config polling.
+		/// If you want to dynamically change the status of the agent, use recording instead.
+		/// </summary>
```
russcam, line 37:
```suggestion
		/// <summary>
		/// Enables the agent.
		/// When set to <c>true</c> (the default), the agent is enabled.
		/// When set to <c>false</c>, the agent is disabled, including instrumentation and remote config polling.
		/// The value of <see cref="Enabled" /> cannot be changed during the lifetime of the application.
		/// <para />
		/// To dynamically change agent operation, use <see cref="Recording" />
		/// </summary>
```

src/Elastic.Apm/Config/IConfigurationReader.cs
```diff
@@ -126,6 +127,15 @@ public interface IConfigurationReader
 
 		double MetricsIntervalInMilliseconds { get; }
 
+		/// <summary>
+		/// A boolean specifying if the agent should be recording or not. When recording, the agent instruments and capture
+		/// requests, tracks errors and collects and sends metrics. When not recording, the agent works as a noop, not collecting
+		/// data and not communicating with the APM sever, except for polling the central configuration endpoint. As this is a
+		/// reversible switch, agent threads are not being killed when inactivated, but they will be mostly idle in this state, so
+		/// the overhead should be negligible.
+		/// </summary>
```
russcam, line 136:
```suggestion
		/// <summary>
		/// Whether the agent is recording.
		/// When set to <c>true</c>. the agent instruments and capture requests, tracks errors, and 
		/// collects and sends metrics. 
		/// When set to <c>false</c>, the agent does not collect data or communicate with the APM server, except to 
		/// fetch central configuration. 
		/// Recording can be changed during the lifetime of the application.		
		/// </summary>
		/// <remarks>
		/// As this is a reversible switch, agent threads are not terminated when inactivated, but they will be mostly 
		/// idle in this state, so the overhead should be negligible.
		/// </remarks>
```

src/Elastic.Apm/Model/DbSpanCommon.cs
```diff
@@ -33,28 +33,34 @@ internal static class DefaultPorts
 				captureStackTraceOnStart);
 		}
 
-		internal void EndSpan(Span span, IDbCommand dbCommand, Outcome outcome, TimeSpan? duration = null)
+		internal void EndSpan(ISpan iSpan, IDbCommand dbCommand, Outcome outcome, TimeSpan? duration = null)
```
russcam, line 36: minor nitpick: keep this as `span` and rename the `if (iSpan is Span span)` to something else e.g. `capturedSpan`

src/Elastic.Apm/Model/NoopSpan.cs
```diff
@@ -0,0 +1,113 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading.Tasks;
+using Elastic.Apm.Api;
+
+namespace Elastic.Apm.Model
+{
+	internal class NoopSpan : ISpan
```
russcam, line 14: xml comment?

src/Elastic.Apm/Model/NoopTransaction.cs
```diff
@@ -0,0 +1,97 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading.Tasks;
+using Elastic.Apm.Api;
+
+namespace Elastic.Apm.Model
+{
+	internal class NoopTransaction : ITransaction
```
russcam, line 14: xml comment?

russcam (commented):

russcam (approved):

src/Elastic.Apm/CurrentExecutionSegmentsContainer.cs
gregkalapos: Ok - agreed. But I'd do it outside this PR. I opened https://github.com/elastic/apm-agent-dotnet/issues/1017.

src/Elastic.Apm.AspNetCore/ApmMiddleware.cs
gregkalapos: This makes sense and I also started changing it accordingly, but then I realized there was a bug in this part:

We need the `ITransaction` returned there, because we still maintain `Tracer.CurrentTransaction` when `recording=false`, so we also need to call `ITtransaction.End()` when the request is finished. So I think there is no way to immediately cast it to `Span`, because that way we won't be able to call `End()` on the created `NoopTransaction`.

src/Elastic.Apm.AspNetCore/DiagnosticListener/AspNetCoreDiagnosticListener.cs
gregkalapos: Similar issue as described above - so I ended up keeping it.

src/Elastic.Apm.AspNetFullFramework/HttpContextCurrentExecutionSegmentsContainer.cs
gregkalapos: Nice catch - fixed.

src/Elastic.Apm.AspNetFullFramework/HttpContextCurrentExecutionSegmentsContainer.cs
gregkalapos: 👍  Fixed

src/Elastic.Apm/Config/ConfigConsts.cs
gregkalapos: Oh - no idea why mi Rider did this. Fixed it.

src/Elastic.Apm/Model/NoopSpan.cs
gregkalapos: Added 👍

src/Elastic.Apm/Model/NoopTransaction.cs
gregkalapos: Added 👍

src/Elastic.Apm.Elasticsearch/ElasticsearchDiagnosticsSubscriber.cs
gregkalapos: Sounds like a good idea. Opened an issue for this: https://github.com/elastic/apm-agent-dotnet/issues/1018/.

src/Elastic.Apm/Config/AbstractConfigurationWithEnvFallbackReader.cs
```diff
@@ -73,11 +80,13 @@ internal AbstractConfigurationWithEnvFallbackReader(IApmLogger logger, string de
 		public virtual double MetricsIntervalInMilliseconds =>
 			ParseMetricsInterval(Read(KeyNames.MetricsInterval, EnvVarNames.MetricsInterval));
 
+		public bool Recording =>
+			ParseEnabled(Read(KeyNames.Enabled, EnvVarNames.Enabled));
```
billhong-just, line 84: @gregkalapos

Hi, gregkalapos
Is this line mistyped?
Should it be ParseRecording(Read(KeyNames.Recording, EnvVarNames.Recording));?

russcam: @billhong-just yes it should be, thank you for raising. I'll open a PR now to address

PR #974, opened 2020-10-12
Add ability to configure Hostname

russcam:
This commit adds the ability to configure the hostname using the environment variable `ELASTIC_APM_HOSTNAME` or
the configuration value `ElasticApm:HostName`.  A non-null or non-empty value will configure the `System.HostName`, falling back to `System.DetectedHostName`.

Closes #932

PR #998, opened 2020-10-29
Add HostName to configuration docs

russcam:
This PR adds ability to manually configure HostName to configuration docs, as added in #974


gregkalapos (approved):
LGTM, also added @bmorelli25 for the doc review.

Issue #1001, opened 2020-11-02
Support for multiple services living in the same process.


Qrious:
**Is your feature request related to a problem? Please describe.**
Currently it is no longer possible (to my knowledge, i've looked in the documentation and issues) to override the service name when starting a transaction. This is an issue in our use case. We use an actor based distributed system framework. The key issue here is that we don't have a service per process model, but that multiple services might run in the same process (controlled by hypervisor). As a result, setting an Environment variable is not a viable solution, since that would result in race conditions between multiple services setting/reading the Environment variable.

We believe this is an issue in general for actor based distributed systems (such as Microsoft Service Fabric and Akka.Net).

I do apologize if it still is actually possible to override the service name, but i've not found any way/documentation on how to do this.

**Describe the solution you'd like**
Add an option when creating a transaction to specify the service name.

As a side note, this was possible in earlier versions (I've had seen this working at version 1.0.5).

**Describe alternatives you've considered**
- Using the environment variable (leads to race condition)
- Overriding it using `Custom` context/labels. This did (obviously) not work.

**Additional context**
-

gregkalapos:
Thanks for reporting this @Qrious.

How did you solve this problem in earlier versions? 

I'd like to say that in general we have a single agent per process approach and right now the service name is tied to this single agent version. But I totally see your use case and it makes perfect sense.

One thing which is possible in the APM Server is that an agent can overwrite the service name for a given transaction. So having the single agent per process approach we could still give you an option to override the service name on a per transaction basis.

There is no API for it, but it could be something like this:

```
if(tracer.CurrentTransaction != null)
        // SetService does not exists, but it''d be something like this:
	tracer.CurrentTransaction.SetService(name: "MyServiceName", version: "2.1-beta");
```

In this case you'd need to do this for each transaction.

If there would be such an API, it'd be also possible to use it through the Filter API:

```
Agent.AddFilter(transaction =>
{
	if(IsThisACustomService())
		transaction.SetService(name: "MyServiceName", version: "2.1-beta");
	
	return transaction;
});
```
A filter runs right before the transaction is sent to the APM Server - so if you can isolate your services (that's what the `IsThisACustomService()` simulates) - then this would be a nicer way to override the service name.

If you don't overwrite the service name on a transaction then it'd be the default service name filled by our current default logic which uses the entry assembly name.


These are just my early thoughts - so nothing is final on this. How does this sound? Would that help with your use case?

Qrious:
[...]

PR #1002, opened 2020-11-02
Add ITransaction.SetService to overwrite service name and version

gregkalapos:
Solves #1001. 

Introduces an API to overwrite service info on a per transaction basis. Prior to this PR the agent auto detected the service name and the version for a given process (which is the name and version of the entry assembly - or the app pool name on ASP.NET Classic). This PR introduces a new API to overwrite the service name and version for a given transaction.

If the API is not used (in other words no custom service name/version is set) then the default service name will be used for the given transaction.

API:
```
ITransaction.SetService(string serviceName, string serviceVersion
```

Usage:
```
var transaction1 = agent.Tracer.StartTransaction("Transaction1", "test");
transaction1.SetService("Service1", "1.0-beta1");
```

Uses the Server intake API field defined [here](https://github.com/elastic/apm-server/blob/master/docs/spec/context.json#L62).
Technically we could expose the whole service, but at this point I don't really see any advantage doing so. Other fields would be things like `framework`, `runtime`, `langauge`, `agent` - I feel those are better be auto detected and exposing all would cause more confusion. If people wish to set those we could extend the API later.

### DIFF
--- a/docs/public-api.asciidoc
+++ b/docs/public-api.asciidoc
@@ -415,14 +415,14 @@ Agent.Tracer.CaptureTransaction(transactionName, transactionType, (transaction) 
 [[api-transaction-set-service]]
 ==== `void SetService(string serviceName, string serviceVersion)` (added[1.7])
 
-With this method you can overwrite the service name and version on a per transaction basis. If this is not set, the transaction will be associated with the default service.
+Overwrite the service name and version on a per transaction basis. This is useful when you host multiple services in a single process.
 
-This method is useful when you host multiple services in a single process.
+When not set, transactions are associated with the default service.
 
-The method has 2 `string` parameters:
+This method has two `string` parameters:
 
-* `serviceName`: The name of the service which the transaction will be associated with.
-* `serviceVersion`: The version of the service which the transaction will be associated with.
+* `serviceName`: The name of the service to associate with the transaction.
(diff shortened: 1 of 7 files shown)


docs/public-api.asciidoc
```diff
@@ -411,6 +411,39 @@ Agent.Tracer.CaptureTransaction(transactionName, transactionType, (transaction)
...
+==== `void SetService(string serviceName, string serviceVersion)` (added[1.7])
+
+With this method you can overwrite the service name and version on a per transaction basis. If this is not set, the transaction will be associated with the default service.
+
+This method is useful when you host multiple services in a single process.
+
+The method has 2 `string` parameters:
+
+* `serviceName`: The name of the service which the transaction will be associated with.
+* `serviceVersion`: The version of the service which the transaction will be associated with.
```
bmorelli25, line 425:
```suggestion
Overwrite the service name and version on a per transaction basis. This is useful when you host multiple services in a single process.

When not set, transactions are associated with the default service.

This method has two `string` parameters:

* `serviceName`: The name of the service to associate with the transaction.
* `serviceVersion`: The version of the service to associate with the transaction.
```

bmorelli25 (approved):
LGTM. I have one (large) wording suggestion. Feel free to cherry-pick the parts that you like if you don't want to take it all.

PR #1003, opened 2020-11-03
Add Cloud Metadata

### LINKED ISSUE #918: Collect cloud metadata
From https://github.com/elastic/apm/issues/256

russcam:
Relates: https://github.com/elastic/apm/issues/256

This commit adds the fetching of cloud metadata to
the agent, to send as part of the agent metadata
when an agent is running in a Cloud provider.

On starting the payload sender's work loop,
the first iteration will asynchronously fetch
the metadata for the cloud provider specified with
the ELASTIC_APM_CLOUD_PROVIDER env var, or the
ElasticApm:CloudProvider app setting. Values
are aws, gcp, azure and false. When false is
specified, no cloud metadata providers are
registered, so there are no attempts to fetch
metadata. The default value of null or ""
will attempt to fetch metadata using all registered
providers.

Closes #918

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -639,7 +639,7 @@ This option supports the wildcard `*`, which matches zero or more characters. Ex
 
 Allows you to specify which cloud provider should be assumed for metadata collection. By default, the agent will attempt to detect the cloud provider and, if that fails, use trial and error to collect the metadata.
 
-Valid options are `"aws"`, `"gcp"`, and `"azure"`. If this config value is set to `"false"`, no cloud metadata will be collected.
+Valid options are `"aws"`, `"gcp"`, `"azure"`, and `"false"`. If this config value is set to `"false"`, no cloud metadata will be collected.
 
 [options="header"]
 |============
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -633,6 +633,26 @@ This option supports the wildcard `*`, which matches zero or more characters. Ex
 | `ELASTIC_APM_DISABLE_METRICS` | `ElasticApm:DisableMetrics`
 |============
 
+[float]
+[[config-cloud-provider]]
+==== `CloudProvider` (added[1.7.0])
+
+Allows you to specify which cloud provider should be assumed for metadata collection. By default, the agent will attempt to detect the cloud provider and, if that fails, use trial and error to collect the metadata.
+
+Valid options are `"aws"`, `"gcp"`, and `"azure"`. If this config value is set to `"false"`, no cloud metadata will be collected.
+
+[options="header"]
+|============
+| Environment variable name      | IConfiguration or Web.config key
+| `ELASTIC_APM_CLOUD_PROVIDER`   | `ElasticApm:CloudProvider`
+|============
+
+[options="header"]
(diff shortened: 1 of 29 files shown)


russcam:
jenkins run the tests please

russcam:
Will follow up in a separate PR for retrieving metadata for Azure App Services

docs/configuration.asciidoc
```diff
@@ -633,6 +633,26 @@ This option supports the wildcard `*`, which matches zero or more characters. Ex
 | `ELASTIC_APM_DISABLE_METRICS` | `ElasticApm:DisableMetrics`
 |============
 
+[float]
+[[config-cloud-provider]]
+==== `CloudProvider` (added[1.7.0])
+
+Allows you to specify which cloud provider should be assumed for metadata collection. By default, the agent will attempt to detect the cloud provider and, if that fails, use trial and error to collect the metadata.
+
+Valid options are `"aws"`, `"gcp"`, and `"azure"`. If this config value is set to `"false"`, no cloud metadata will be collected.
```
gregkalapos, line 642: I see we have this sentence also in the python agent. To me it's a bit contradicting that it lists the 3 cloud provider names as "valid option", and then in the next sentence it says what happens if it's `false`.

But maybe this is a problem with me :)

src/Elastic.Apm/Cloud/AwsCloudMetadataProvider.cs
```diff
@@ -0,0 +1,92 @@
...
+					AvailabilityZone = metadata["availabilityZone"].Value<string>(),
+					Machine = new CloudMachine { Type = metadata["instanceType"].Value<string>() },
+					Provider = Provider,
+					Region = metadata["region"].Value<string>()
+				};
+
+			}
+			catch (Exception e)
+			{
+				_logger.Warning()?.LogException(e, "Unable to get {Provider} cloud metadata", Provider);
```
gregkalapos, line 87: Sure about the `Warning` level? Given the default for this setting is to try all `ICloudMetadataProvider` implementations, the agent will start with 3 warnings in non-cloud and with 2 if it runs in one of the cloud providers.

I feel we should use `info` here.

We have a similar situation with the central config fetcher and I see people are very confused about those warnings as well.

gregkalapos (approved):
Nice, LGTM!

I left 1 comment regarding printing warnings - definitely not a blocker, it's for consideration.

docs/configuration.asciidoc
bmorelli25: Agreed that this wording is a bit funky. Maybe something like:

```suggestion
Valid options are `"aws"`, `"gcp"`, `"azure"`, and `"false"`. If this config value is `null` or set to `"false"`, no cloud metadata will be collected.
```

bmorelli25 (approved):
Same comment as Greg, otherwise docs LG

src/Elastic.Apm/Cloud/AwsCloudMetadataProvider.cs
russcam: Will change to `info` - agree that `warning` seems way too strong

russcam:
jenkins run the tests please

PR #1012, opened 2020-11-10
Serialize payload to stream

russcam:
This commit updates PayloadSenderV2 to serialize the payload to a stream, reducing string allocations.

The stream could be backed by byte buffers rented from an array pool in future.

Add Deserialize method to PayloadItemSerializer for use in tests.


russcam:
Since APM server supports chunked transfer encoding, we could use something like [`ObjectContent<T>`](https://github.com/mono/aspnetwebstack/blob/6248bfd24c31356e75a31c1b1030d4d96f669a6a/src/System.Net.Http.Formatting/ObjectContent.cs) and serialize to the request stream directly in future.

russcam:
jenkins run the tests please

gregkalapos (approved):
LGTM

russcam:
> The build fails in CI, but it's very strange 🤔

It looks like CI is running the result of merging in master branch. I've merged in master and fixed this, so this should now be good to merge, CI checks pending.

russcam:
failing tests are unrelated, merging this in.

PR #1015, opened 2020-11-10
[Doc] Setup instructions updates: `IHostBuilder` and Zero code change setup

gregkalapos:
We introduced 2 new ways to enable the agent. I documented those in this PR.

- Zero code change. In this case you don't need NuGet, you only need a `zip` file we publish with the agent and you need to set an environment variable. I wrote about more technical details [here](https://github.com/elastic/apm-agent-dotnet/tree/master/src/ElasticApmStartupHook).
- `IHostBuilder` on .NET Core. This was needed because we had no good story to integrate into background services. The other advantage here is that we can use the same approach to turn on the agent on every .NET Core app that uses `IHostBuilder` (which is the default these days) including ASP.NET Core. Some more info on the topic is [here](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/generic-host?view=aspnetcore-3.1). Lots of other tools use this way to integrate with .NET Core, so we just followed what everyone else does.

### DIFF
--- a/docs/setup.asciidoc
+++ b/docs/setup.asciidoc
@@ -6,20 +6,30 @@ You can add the Agent to your .NET application by referencing one of these packa
 On .NET Core the agent also supports auto instrumentation without any code change and without any recompilation of your projects. See section <<zero-code-change-setup,  Zero code change setup on .NET Core>> for more details.
 
 [float]
-=== Packages
+== Get started
+
+* <<setup-dotnet-net-core>>
+* <<setup-asp-net-core>>
+* <<setup-asp-net>>
+* <<setup-ef6>>
+* <<setup-sqlclient>>
+* <<setup-general>>
+
+[float]
+== Packages
 
 The following packages are available:
 
 https://www.nuget.org/packages/Elastic.Apm.NetCoreAll[**Elastic.Apm.NetCoreAll**]::
 
-This is a meta package that references every other Elastic APM .NET agent package. If you plan to monitor a typical ASP.NET Core application that depends on the https://www.nuget.org/packages/Microsoft.AspNetCore.All[Microsoft.AspNetCore.All] package and uses Entity Framework Core then you should reference this package. 
+This is a meta package that references every other Elastic APM .NET agent package. If you plan to monitor a typical ASP.NET Core application that depends on the https://www.nuget.org/packages/Microsoft.AspNetCore.All[Microsoft.AspNetCore.All] package and uses Entity Framework Core then you should reference this package.
 In order to avoid adding unnecessary dependencies in applications that aren’t depending on the https://www.nuget.org/packages/Microsoft.AspNetCore.All[Microsoft.AspNetCore.All] package we also offer some other packages - those are all referenced by the `Elastic.Apm.NetCoreAll` package.
 
 https://www.nuget.org/packages/Elastic.Apm[**Elastic.Apm**]::
 
 This is the core of the agent, which we didn’t name “Core”, because someone already took that name :) This package also contains the <<public-api>> and it is a .NET Standard 2.0 package. We also ship every tracing component that traces classes that are part of .NET Standard 2.0 in this package, which includes the monitoring part for HttpClient. Every other Elastic APM package references this package.
 
-https://www.nuget.org/packages/Elastic.Apm.Extensions.Hosting[**Elastic.Apm.Extensions.Hosting**](Added in 1.6.0-beta)::
+https://www.nuget.org/packages/Elastic.Apm.Extensions.Hosting[**Elastic.Apm.Extensions.Hosting**](added[1.6.0-beta])::
 
 This package offers integration with `Microsoft.Extensions.Hosting.IHostBuilder` for agent registration.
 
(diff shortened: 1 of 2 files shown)


docs/setup.asciidoc
```diff
@@ -34,19 +41,104 @@ https://www.nuget.org/packages/Elastic.Apm.SqlClient[**Elastic.Apm.SqlClient**]:
...
+            .UseAllElasticApm();
+
+    public static void Main(string[] args) => CreateHostBuilder(args).Build().Run();
+  }
+}
+----
+
+With the `UseAllElasticApm()` the agent with all its components is turned on. On ASP.NET Core it'll automatically capture incoming requests, database calls through supported technologies, outgoing HTTP requests and so on.
+
+Additionally the `UseAllElasticApm` will also add an `ITracer` to the Dependency Injection system. This means you can use the <<public-api>> in your code like this:
```
gregkalapos, line 83: We are in a `Quick start` section here, and honestly I'm not sure where to stop. The code sample below already goes into manual code instrumentation.

I feel this is useful info, but reality is that the majority of the people will be happy with the code snippet above and won't do manual instrumentation.

docs/setup.asciidoc
```diff
@@ -34,19 +41,104 @@ https://www.nuget.org/packages/Elastic.Apm.SqlClient[**Elastic.Apm.SqlClient**]:
...
+            }
+            return View();
+        }
+    }
+}
+----
+
+Similarly to this ASP.NET Core controller, you can use the same approach with `IHostedService` implementations.
+
+The `Elastic.Apm.NetCoreAll` will reference every agent components. This is usually not a problem. Nevertheless in case you want to keep dependencies minimal you can also reference the `Elastic.Apm.Extensions.Hosting` and use the `UseElasticApm` method instead of `UseAllElasticApm`. With this you can control what the agent will listen for.
```
gregkalapos, line 122: Related to the comment above: here we go even further and discuss how to control specific instrumentation modules. I feel this must be documented, but not sure this is the good place for it.

docs/setup.asciidoc
```diff
@@ -34,19 +41,104 @@ https://www.nuget.org/packages/Elastic.Apm.SqlClient[**Elastic.Apm.SqlClient**]:
...
+            .UseElasticApm(new HttpDiagnosticsSubscriber());
+
+----
+
+
 [float]
 [[setup-asp-net-core]]
 ==== ASP.NET Core
 
+We suggest using the approach described in the <<setup-dotnet-net-core, .NET Core setup instructions>> section above. We keep the `IApplicationBuilder` introduced here only for backwards compatibility.
```
gregkalapos, line 140: With the approach documented above (the `IHostBuilder` thing) we basically made things that are discussed here unnecessary. Long term I think we just need to deprecate this `ASP.NET Core` section.

docs/setup.asciidoc
bmorelli25: > We are in a Quick start section here, and honestly I'm not sure where to stop.

😂 I can relate to this.

bmorelli25:
I agree with the point you're making. Perhaps the current layout of this documentation isn't best suited for how the content has grown over the past few months. What if we moved each method's set up instructions to its own page? This would allow us to highlight both the "quickstart" and other content (like manual instrumentation), without needing to group everything together. Here's a quick mock I just made of what this could look like:

apm-dotnet-setup

gregkalapos:
> I agree with the point you're making. Perhaps the current layout of this documentation isn't best suited for how the content has grown over the past few months. What if we moved each method's set up instructions to its own page? This would allow us to highlight both the "quickstart" and other content (like manual instrumentation), without needing to group everything together. Here's a quick mock I just made of what this could look like:
> 

Very nice, I think it's a great solution!

docs/setup.asciidoc
```diff
@@ -3,19 +3,36 @@
...
+
+[float]
+== Get started
+
+* <<setup-dotnet-net-core>>
+* <<setup-asp-net-core>>
+* <<setup-asp-net>>
+* <<setup-ef6>>
+* <<setup-sqlclient>>
+* <<setup-general>>
```
bmorelli25, line 16: Is this an order that makes sense in the .NET world? Would we be better of alphabetizing these?

gregkalapos: The order was intentional - it goes (based on my own judgement) from more popular to less popular. I think alphabetic ordering would be less natural.

bmorelli25:
It looks like there's a link from the release notes that isn't versioned correctly which is breaking the build:
```
12:58:19 INFO:build_docs:Bad cross-document links:
12:58:19 INFO:build_docs:  /tmp/docsbuild/target_repo/html/en/apm/agent/dotnet/1.x/release-notes-1.x.html:
12:58:19 INFO:build_docs:   - en/apm/agent/dotnet/master/setup.html#setup-asp-net
12:58:19 INFO:build_docs:  /tmp/docsbuild/target_repo/html/en/apm/agent/dotnet/current/release-notes-1.x.html:
12:58:19 INFO:build_docs:   - en/apm/agent/dotnet/master/setup.html#setup-asp-net
12:58:19 INFO:build_docs:  /tmp/docsbuild/target_repo/html/en/apm/agent/dotnet/master/release-notes-1.x.html:
12:58:19 INFO:build_docs:   - en/apm/agent/dotnet/master/setup.html#setup-asp-net
```

I'll add another commit that fixes that link and adds in a temp anchor on `setup.html`.

docs/setup.asciidoc
bmorelli25: Makes sense

bmorelli25:
Doc build is green! So. . . how does this work? Do I request you as a reviewer on your own PR? 😅

gregkalapos:
> Do I request you as a reviewer on your own PR? 

If that possible? I cannot add myself to this PR, not sure if you can, since I opened the PR.

Nevertheless: ✅ 👍  Approved, LGTM 😆

bmorelli25 (approved):
It won't let me add you either... so I'll do it! LGTM

Issue #906, opened 2020-07-16
Adding sampling weight to transactions and spans 


gregkalapos:
Implementing https://github.com/elastic/apm/issues/307

gregkalapos:
Implemented in #1020.

PR #1020, opened 2020-11-10
Add SampleRate to spans and transactions

gregkalapos:
Solves #906.

It adds a new field (`sample_rate`) to both `Span` and `Transaction`. I kept those internal.

Spec: [here](https://github.com/elastic/apm/blob/master/specs/agents/tracing-sampling.md).

Opened follow up for propagation: https://github.com/elastic/apm-agent-dotnet/issues/1021. I'd say the propagation part will be left from the current release.

### DIFF
--- a/src/Elastic.Apm/Model/Span.cs
+++ b/src/Elastic.Apm/Model/Span.cs
@@ -19,9 +19,6 @@ namespace Elastic.Apm.Model
 {
 	internal class Span : ISpan
 	{
-		[JsonProperty("sample_rate")]
-		internal readonly double SampleRate;
-
 		private readonly Lazy<SpanContext> _context = new Lazy<SpanContext>();
 		private readonly ICurrentExecutionSegmentsContainer _currentExecutionSegmentsContainer;
 		private readonly Transaction _enclosingTransaction;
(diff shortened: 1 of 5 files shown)


src/Elastic.Apm/Model/Span.cs
```diff
@@ -38,6 +38,12 @@ internal class Span : ISpan
 		/// </summary>
 		private readonly StackFrame[] _stackFrames;
 
+		/// <summary>
+		/// Captures the sample rate of the agent when this span was created.
+		/// </summary>
+		[JsonProperty("sample_rate")]
+		internal readonly double SampleRate;
```
russcam, line 45:
```suggestion
		internal readonly double SampleRate { get; set; }
```

I think this should be a property rather than a field?

src/Elastic.Apm/Sampler.cs
```diff
@@ -20,7 +20,8 @@ namespace Elastic.Apm
 	{
 		private readonly long _higherBound;
 		private readonly long _lowerBound;
-		private readonly double _rate;
+
+		public double Rate { get; }
```
russcam, line 24: With exposing the rate as a property, it'd probably be useful to update the Sampler tests that assert rounding to use this rather than the `.ToString()` representation. Doesn't need to be part of this PR though

russcam (approved):
LGTM, just left one small suggestion

src/Elastic.Apm/Model/Span.cs
gregkalapos: Agreed - added a read-only property for this.

src/Elastic.Apm/Sampler.cs
gregkalapos: Updated the `SamplerTests.Rate_Precision_Should_Be_Rounded_To_Four_Decimal_Places` to also assert on the `Rate` property. Wasn't sure if we completely want to remove the `ToString()` overload (I think it's useful), so I also kept asserting on `ToString()`.

PR #1023, opened 2020-11-11
Fix cloud metadata

russcam:
This commit fixes a few bugs with cloud
metadata:

- AWS instance metadata should be a GET
- Default cloud provider value should be `"auto"`
- Disable cloud provider value should be `"none"`
- Correct machine type parsing in GCP provider


gregkalapos (approved):

russcam:
merging this in - failed test is not related to this change

PR #1041, opened 2020-11-16
Don't subscribe IDiagnosticsSubscriber in ApmAgentExtensions.Subscribe(...) when agent is disabled

### LINKED ISSUE #1018: Introduce base DiagnosticsSubscriber
From https://github.com/elastic/apm-agent-dotnet/pull/997#discussion_r520261099.

Currently tracing components based on DiagnosticSource implement `IDiagnosticsSubscriber` and there is no logic shared between the implementations. 

Adding a common base class would enable us to share logic across implementors. For example: With the `enabled` config every implementation need to honor the setting and skip registering if `enabled=false`. 

This behaviour could be implemented in a base class.

russcam:
This commit subscribes `IDiagnosticsSubscriber`
in `ApmAgentExtensions.Subscribe(...)` **only** if the agent is
enabled. This extension method is the root for the 
subscribers subscribing to diagnostic events, and
performing the check here removes the need to perform the
check to see if the agent is enabled in each subscriber implementation.

Closes #1018

### DIFF
--- a/docs/public-api.asciidoc
+++ b/docs/public-api.asciidoc
@@ -61,6 +61,12 @@ Some examples:
 - `EfCoreDiagnosticsSubscriber`: captures database calls through Entity Framework Core
 - `SqlClientDiagnosticSubscriber`: captures database calls through `SqlClient` 
 
+[NOTE]
+--
+When the agent is configured with <<config-enabled, `Enabled` set to `false`>>, `Elastic.ApmAgent.Subscribe(params IDiagnosticsSubscriber[] subscribers)` will not subscribe the subscribers to
+diagnostic source events.
+--
+
 [float]
 [[api-tracer-api]]
 == Tracer API
--- a/src/Elastic.Apm.SqlClient/SqlClientDiagnosticSubscriber.cs
+++ b/src/Elastic.Apm.SqlClient/SqlClientDiagnosticSubscriber.cs
@@ -9,29 +9,28 @@ using Elastic.Apm.Helpers;
 
 namespace Elastic.Apm.SqlClient
 {
-	public class SqlClientDiagnosticSubscriber : IDiagnosticsSubscriber
+	/// <summary>
+	/// Subscribes to diagnostics events from System.Data.SqlClient and Microsoft.Data.SqlClient
+	/// </summary>
+	public class SqlClientDiagnosticSubscriber : DiagnosticsSubscriberBase
 	{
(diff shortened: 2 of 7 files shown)


russcam:
jenkins run the tests please

gregkalapos (commented):
The `EfCoreDiagnosticsSubscriber` and the `ElasticsearchDiagnosticsSubscriber` are not adapted if I see this correctly.

The other thing that came to my mind is that maybe the `Elastic.Apm.DiagnosticSource.DiagnosticInitializer.OnNext` method would be a better central place to check if the agent is enabled. All our DiagnosticSubscribers use that and they could pass a configreader into it. With that we'd also have a single check and we don't need to derive from a specific class.

russcam:
> The other thing that came to my mind is that maybe the Elastic.Apm.DiagnosticSource.DiagnosticInitializer.OnNext method would be a better central place to check if the agent is enabled. All our DiagnosticSubscribers use that and they could pass a configreader into it. With that we'd also have a single check and we don't need to derive from a specific class.

Sounds good to me, will update the PR with this change

russcam:
Thinking about it a bit more, might it make sense to move the `Enabled` check to `Agent.Subscribe`:

https://github.com/elastic/apm-agent-dotnet/blob/ea1801f5b39f6a8014ffdd8010bf5c92ed9839a9/src/Elastic.Apm/Agent.cs#L235

This looks like the only entrypoint to subscribing

gregkalapos:
> Thinking about it a bit more, might it make sense to move the `Enabled` check to `Agent.Subscribe`:
> 
> https://github.com/elastic/apm-agent-dotnet/blob/ea1801f5b39f6a8014ffdd8010bf5c92ed9839a9/src/Elastic.Apm/Agent.cs#L235
> 
> This looks like the only entrypoint to subscribing

There is also `ApmAgentExtensions.Subscribe` in `ApmAgentExtensions.cs`- this call also ends up there and it seems that `ApmMiddlewareExtension` and `HostBuilderExtensions` end up directly in the extension method without this method. So from what I see `ApmAgentExtensions.Subscribe` would be a better place for the check. But yeah, I also think this looks like a better solution.

russcam:
Ready for another look when you have a chance, @gregkalapos

gregkalapos (approved):
LGTM 👍 

Maybe rename the PR if you merge it in this form.

PR #1043, opened 2020-11-16
Write ConsoleLogger messages direct to writer

russcam:
This commit logs ConsoleLogger messages straight to
TextWriter, avoiding intermediate string
allocations in building the fullMessage.

Lock around multiple writes to TextWriter
and subsequent flushing.


gregkalapos (approved):
LGTM

PR #1046, opened 2020-11-18
Fetch APM server version only when not null

russcam:
This commit updates the PayloadSender workloop to only fetch APM server info on the first iteration when the version is not null.

Tests pass an APM server info with a version so the call should not be made.


russcam:
Merging this in as all tests have passed

Issue #1039, opened 2020-11-12
Elastic APM 1.7.0 has significantly increased performance impact


kanadaj:
**Describe the bug**
After updating from 1.6.1 to 1.7.0, we have observed a very significant degradation in application performance and resources used:
![image](https://user-images.githubusercontent.com/3247357/98996507-e07bdc80-252a-11eb-9ba0-1a50aa6aa369.png)
![image](https://user-images.githubusercontent.com/3247357/98996544-eec9f880-252a-11eb-8bae-9005aeadcba1.png)

Not entirely sure why or what exactly causes it but that's a _lot_ of overhead.

kanadaj:
@gregkalapos This one essentially makes it impossible to update to 1.7.0

kanadaj:
Packages used:
Elastic.Apm.AspNetCore 1.6.1 -> 1.7.0
Elastic.Apm.Elasticsearch 1.0.1-beta1 -> 1.7.0
Elastic.Apm.EntityFrameworkCore 1.6.1 -> 1.7.0

Setup:
```cs
// 1.6.1
app.UseElasticApm(Configuration, new HttpDiagnosticsSubscriber(), new AspNetCoreDiagnosticsSubscriber(), new EfCoreDiagnosticsSubscriber(), ElasticsearchDiagnosticsSubscriber()));

// 1.7.0
app.UseElasticApm(Configuration, new HttpDiagnosticsSubscriber(), new AspNetCoreErrorDiagnosticsSubscriber(), new EfCoreErrorDiagnosticsSubscriber(), ElasticsearchDiagnosticsSubscriber()));
```

cbo100:
Commenting as another datapoint to backup this observation. Only local testing so far, but with no other change beyond:

- 1.6.1 -> 1.7.0 

Seeing 50%-150% increase in response times (20ms avg -> 30-50ms avg)

cbo100:
One observation: `AspNetCoreDiagnosticsSubscriber` was renamed to `AspNetCoreErrorDiagnosticsSubscriber`. And `ApmMiddlewareExtension` was migratred to the new one:

https://github.com/elastic/apm-agent-dotnet/compare/1.6.1...1.7.0#diff-706440fb8f0d9782422e64da0ac05bbf608d2846374d384e5a77ab6e15d2dbb0L71-R77

I can see @kanadaj migrated the constructor the same way, and I did as well. 

[...]

PR #1052, opened 2020-11-18
Moving stack trace capturing away from application thread

gregkalapos:
Addresses https://github.com/elastic/apm-agent-dotnet/issues/1039

`new StackTrace(true)` still happens on the app thread, but the demystification and the conversion to our internal stack trace model happens right before serialization on the payloadsender thread.

Benchmarks (I worked with [this benchmark](https://github.com/elastic/apm-agent-dotnet/blob/2182be81af478390111061e3edee280c4cafa3b9/test/Elastic.Apm.PerfTests/AspNetCorePerf/AspNetCoreLoadTestWithAgent.cs))

Current master:

```
|                             Method |     Mean |    Error |   StdDev |   Median |
|----------------------------------- |---------:|---------:|---------:|---------:|
| WebRequestWithDbCallsAndCustomSpan | 31.96 ms | 3.569 ms | 10.52 ms | 24.93 ms |
```

With this PR:

```
|                             Method |     Mean |     Error |    StdDev |   Median |
|----------------------------------- |---------:|----------:|----------:|---------:|
| WebRequestWithDbCallsAndCustomSpan | 8.325 ms | 0.2154 ms | 0.6214 ms | 8.126 ms |
```

So from the throughput perspective of the monitored app this seems like an improvement.

Potential issues: 
- With this we moved more work to the `PayloadSenderV2` (or to be precise to the thread that handles every span after `.End()` is called on it)
- If people do `apm.Tracer.CurrentSpan` then the `StackTrace` property will be `null` on it, which wasn't the case previously (although to be fair the story there wasn't very nice, because in some cases it was filled when the span was ended (that's the default) and in other cases (e.g. HTTP and DB spans) it was filled when the span started). On the positive side, if people use filters, then at that point the property will be already filled.

### DIFF
--- a/src/Elastic.Apm/Filters/SpanStackTraceCapturingFilter.cs
+++ b/src/Elastic.Apm/Filters/SpanStackTraceCapturingFilter.cs
@@ -1,0 +1,58 @@
+// Licensed to Elasticsearch B.V under
+// one or more agreements.
+// Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information
+
+using System.Diagnostics;
+using Elastic.Apm.Api;
+using Elastic.Apm.Helpers;
+using Elastic.Apm.Logging;
+using Elastic.Apm.Model;
+using Elastic.Apm.ServerInfo;
+
+namespace Elastic.Apm.Filters
+{
+	/// <summary>
+	/// Stack trace capturing itself happens on the application thread (in order to get the real stack trace).
+	/// This filter turns <see cref="Span.RawStackTrace" /> (which is a plain .NET System.Diagnostics.StackTrace instance) into
+	/// <see cref="Span.StackTrace" /> (which is the representation of the intake API stacktrace model).
+	/// This can be done on a non-application thread right before the span gets sent to the APM Server.
+	/// </summary>
+	internal class SpanStackTraceCapturingFilter
+	{
+		private readonly IApmServerInfo _apmServerInfo;
+		private readonly IApmLogger _logger;
+
+		public SpanStackTraceCapturingFilter(IApmLogger logger, IApmServerInfo apmServerInfo) =>
+			(_logger, _apmServerInfo) = (logger, apmServerInfo);
+
+		public ISpan Filter(ISpan iSpan)
+
+		{
+			if (iSpan is Span span)
+			{
+				if (span.RawStackTrace != null)
+				{
+					StackFrame[] trace;
+					try
+					{
+						// I saw EnhancedStackTrace throwing exceptions in some environments
+						// therefore we try-catch and fall back to a non-demystified call stack.
+						trace = new EnhancedStackTrace(span.RawStackTrace).GetFrames();
+					}
+					catch
+					{
+						trace = span.RawStackTrace.GetFrames();
+					}
+
+					span.StackTrace = StacktraceHelper.GenerateApmStackTrace(trace,
+						_logger,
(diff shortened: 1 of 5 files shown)


src/Elastic.Apm/Filters/SpanStackTraceCapturingFilter.cs
```diff
@@ -0,0 +1,54 @@
...
+	/// </summary>
+	internal class SpanStackTraceCapturingFilter
+	{
+		private readonly IApmServerInfo _apmServerInfo;
+		private readonly IApmLogger _logger;
+
+		public SpanStackTraceCapturingFilter(IApmLogger logger, IApmServerInfo apmServerInfo) =>
+			(_logger, _apmServerInfo) = (logger, apmServerInfo);
+
+		public ISpan Filter(ISpan iSpan)
```
russcam, line 29: minor nitpick: use `span` for the interface argument and something else for the conversion to `Span` e.g. `concreteSpan`, `capturedSpan`, `realSpan`, etc.

src/Elastic.Apm/Model/Span.cs
```diff
@@ -186,6 +164,20 @@ private Span(double duration, string id, string name, string parentId)
 		[JsonProperty("parent_id")]
 		public string ParentId { get; set; }
 
+		/// <summary>
+		/// This holds the raw stack trace that was captured when the span either started or ended (depending on the parameter
+		/// passed to the .ctor)
+		/// This will be turned into an elastic stack trace and sent to APM Server in the <see cref="StackTrace" /> property
+		/// </summary>
+		[JsonIgnore]
+		internal StackTrace RawStackTrace;
```
russcam, line 173: Does this need the `[JsonIgnore]`, as a field?

src/Elastic.Apm/Model/Span.cs
```diff
@@ -298,38 +290,10 @@ public void End()
...
-						}
-						else
-						{
-							StackTrace = StacktraceHelper.GenerateApmStackTrace(_stackFrames,
-								_logger,
-								ConfigSnapshot, _apmServerInfo, $"Span `{Name}'");
-						}
-					}
-				}
+				if (ConfigSnapshot.StackTraceLimit != 0 && ConfigSnapshot.SpanFramesMinDurationInMilliseconds != 0 && RawStackTrace == null
```
russcam, line 293: Worth enclosing this logic in a well named method somewhere e.g. `ConfigSnapshot.CaptureStackTrace(double duration)`, or a private method on `Span`?

russcam (approved):
Change LGTM 👍 Agree that it'd be better to offload this processing to outside of the application trace flow.

Left some small comments for consideration

gregkalapos:
I did further testing:  first with `1.7.0`, then with branch of the [PR](https://github.com/elastic/apm-agent-dotnet/pull/1052). Here is the transaction duration:

image

I tested [with this sample app from the repo](https://github.com/elastic/apm-agent-dotnet/blob/master/sample/SampleAspNetCoreApp/Controllers/HomeController.cs#L47) - only change I did is that I removed the outgoing HTTP request, since that's not deterministic. This test creates 4 EF Core DB spans in each request.

Test scenario: generating full load with Netling for 10 minutes.

Avg Transaction duration:
- `1.7.0`: well above 100ms, often above 150ms
- With the [PR](https://github.com/elastic/apm-agent-dotnet/pull/1052): mostly between 40ms and 50ms.

src/Elastic.Apm/Model/Span.cs
gregkalapos: No - removed.

PR #1053, opened 2020-11-18
AllowSynchronousIO when capturing request body

### LINKED ISSUE #1044: Synchronous operations are disallowed. Call ReadAsync or set AllowSynchronousIO to true instead.
>    at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpRequestStream.Read(Byte[] buffer, Int32 offset, Int32 count)
>    at System.IO.Stream.Read(Span`1 buffer)
>    at Microsoft.AspNetCore.WebUtilities.FileBufferingReadStream.Read(Span`1 buffer)
>    at Microsoft.AspNetCore.WebUtilities.FileBufferingReadStream.Read(Byte[] buffer, Int32 offset, Int32 count)
>    at System.IO.StreamReader.ReadBuffer()
>    at System.IO.StreamReader.ReadToEnd()
>    at Elastic.Apm.AspNetCore.Extensions.HttpRequestExtensions.ExtractRequestBody(HttpRequest request, IApmLogger logger, IConfigSnapshot configSnapshot)

Latest version as of an hour ago, and .NET 5.

russcam:
This commit allows synchronous IO for a given request
when capturing the request body.

From ASP.NET Core 3.0 onwards, synchronous reading or writing
of request streams is disabled by default, throwing
an exception when attempting to do so. See

https://docs.microsoft.com/en-us/aspnet/core/migration/22-to-30?view=aspnetcore-5.0&tabs=visual-studio#allowsynchronousio-disabled

Capturing of the request body by the agent is a synchronous
operation because it may happen in the the synchronous
flow of a diagnostic listener.

Consideration was made to using GetAwaiter().GetResult()
which would effectively be sync over async, but using the
http body control feature is preferred as it is a framework feature
designed for this purpose.

Fixes #1044.

### DIFF
--- a/test/Elastic.Apm.AspNetCore.Tests/AspNetCoreBasicTests.cs
+++ b/test/Elastic.Apm.AspNetCore.Tests/AspNetCoreBasicTests.cs
@@ -244,8 +244,8 @@ namespace Elastic.Apm.AspNetCore.Tests
 
 			_agent.Service.Framework.Name.Should().Be("ASP.NET Core");
 
-			var aspNetCoreMajorMinorVersion = Assembly.Load("Microsoft.AspNetCore").GetName().Version.ToString(2);
-			_agent.Service.Framework.Version.Should().StartWith(aspNetCoreMajorMinorVersion);
+			var aspNetCoreVersion = Assembly.Load("Microsoft.AspNetCore").GetName().Version.ToString();
+			_agent.Service.Framework.Version.Should().Be(aspNetCoreVersion);
 
 			_agent.Service.Runtime.Name.Should().Be(Runtime.DotNetCoreName);
 			_agent.Service.Runtime.Version.Should().Be(Directory.GetParent(typeof(object).Assembly.Location).Name);
--- a/test/Elastic.Apm.AspNetCore.Tests/AspNetCoreBasicTests.cs
+++ b/test/Elastic.Apm.AspNetCore.Tests/AspNetCoreBasicTests.cs
@@ -244,8 +244,8 @@ namespace Elastic.Apm.AspNetCore.Tests
 
 			_agent.Service.Framework.Name.Should().Be("ASP.NET Core");
 
-			var aspNetCoreVersion = Assembly.Load("Microsoft.AspNetCore").GetName().Version.ToString();
-			_agent.Service.Framework.Version.Should().Be(aspNetCoreVersion);
+			var aspNetCoreMajorMinorVersion = Assembly.Load("Microsoft.AspNetCore").GetName().Version.ToString(2);
+			_agent.Service.Framework.Version.Should().StartWith(aspNetCoreMajorMinorVersion);
 
 			_agent.Service.Runtime.Name.Should().Be(Runtime.DotNetCoreName);
 			_agent.Service.Runtime.Version.Should().Be(Directory.GetParent(typeof(object).Assembly.Location).Name);
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -717,9 +717,24 @@ For transactions that are HTTP requests, the agent can optionally capture the re
 If the request has a body and this setting is disabled, the body will be shown as [REDACTED].
 This option is case-insensitive.
 
-WARNING: Request bodies often contain sensitive values like passwords, credit card numbers, etc.
+[IMPORTANT]
+====
+To allow capturing request bodies, the agent sets `AllowSynchronousIO` to `true` on a per
+request basis in ASP.NET Core, since capturing can occur in synchronous code paths. 
+
+https://docs.microsoft.com/en-us/aspnet/core/migration/22-to-30?#allowsynchronousio-disabled[With ASP.NET Core 3.0 onwards, `AllowSynchronousIO` is `false` by default] 
+because a large number of blocking synchronous I/O operations can lead to thread pool starvation, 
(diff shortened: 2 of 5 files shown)


test/Elastic.Apm.AspNetCore.Tests/AspNetCoreBasicTests.cs
```diff
@@ -244,8 +244,8 @@ public async Task HomeSimplePagePostTransactionTest(bool withDiagnosticSourceOnl
 
 			_agent.Service.Framework.Name.Should().Be("ASP.NET Core");
 
-			var aspNetCoreVersion = Assembly.Load("Microsoft.AspNetCore").GetName().Version.ToString();
-			_agent.Service.Framework.Version.Should().Be(aspNetCoreVersion);
+			var aspNetCoreMajorMinorVersion = Assembly.Load("Microsoft.AspNetCore").GetName().Version.ToString(2);
```
gregkalapos, line 247: Why do we change this? 🤔  We set `Service.Framework.Version` to the assembly version (without truncating it). I'd assert on the whole value still.

gregkalapos (approved):
LGTM, just added 1 thing below.

The other thing that I was thinking about: Should we maybe add a warning about this in our [docs](https://www.elastic.co/guide/en/apm/agent/dotnet/current/config-http.html#config-capture-body)? Especially the warning [here](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel?view=aspnetcore-3.0#synchronous-io) would be nice to refer to and maybe stating that we'll turn on SynchronousIO - so e.g. even if it's set to `false` on `2.1` (which is not the default), we'll still change that.

test/Elastic.Apm.AspNetCore.Tests/AspNetCoreBasicTests.cs
russcam: This test consistently fails for me locally on 2.1 and 3.1 with diagnostic source `true`. Debugging the issue, it's related to

https://github.com/elastic/apm-agent-dotnet/blob/dbad2d54ffd2bc7ba7a27e96506f7f3595765fd6/src/Elastic.Apm.Extensions.Hosting/HostBuilderExtensions.cs#L84-L98

With diagnostic source `true`, `Microsoft.AspNetCore` assembly is not found, so resorts to assemblies containing `Microsoft.AspNetCore`, which ends up picking up `Microsoft.AspNetCore.Mvc.Core` that has a different patch version for 2.1 and 3.1 (for 2.1,  `Microsoft.AspNetCore` is 2.1.7.0 and  `Microsoft.AspNetCore.Mvc.Core` is 2.1.16.0)

Assembly Version

Be curious to see which assembly your local test picks for this.

russcam:
> The other thing that I was thinking about: Should we maybe add a warning about this in our docs? Especially the warning here would be nice to refer to and maybe stating that we'll turn on SynchronousIO

Good call on the documentation, will add something

test/Elastic.Apm.AspNetCore.Tests/AspNetCoreBasicTests.cs
russcam:
```
> dotnet --list-sdks

2.1.300 [C:\Program Files\dotnet\sdk]
2.1.801 [C:\Program Files\dotnet\sdk]
2.2.103 [C:\Program Files\dotnet\sdk]
2.2.104 [C:\Program Files\dotnet\sdk]
2.2.401 [C:\Program Files\dotnet\sdk]
3.0.100 [C:\Program Files\dotnet\sdk]
3.1.100 [C:\Program Files\dotnet\sdk]
3.1.200 [C:\Program Files\dotnet\sdk]
5.0.100-preview.8.20362.3 [C:\Program Files\dotnet\sdk]
5.0.100 [C:\Program Files\dotnet\sdk]
```

gregkalapos: Hahh, I have the same problem on `2.1`.

But this would mean we may not report the correct patch version. By looking at this, there is a complete chaos on `2.1`.

I have:
```
dotnet --list-sdks
2.1.612 [C:\Program Files\dotnet\sdk]
... and more 2.2, etc...
```

Then, when I run an ASP.NET Core app, assemblies are loaded from the folder `2.1.23`: 
```
Program Files\dotnet\shared\Microsoft.AspNetCore.App\2.1.23
```

With the following assemblies and versions in it:
```
Microsoft.AspNetCore.Mvc.Core, Version=2.1.16.0,
Microsoft.AspNetCore, Version=2.1.7.0
```

These are 4 different versions and at this point I'm not even sure what we should report.

If we ignore the patch part of the version, it could be that we actually ignore a bug. This definitely should not block this PR, on the other this may be a real bug here (probably a small one, but still...)

With this we'll report `Framework{Name: ASP.NET Core, Version: 2.1.16.0}`, both on your machine and on mine, while you have sdk `2.1.801` and I have sdk `2.1.612`.

russcam: Ok, I'll revert this change so that this PR can be merged, and I'll open a new issue to discuss

russcam: Opened https://github.com/elastic/apm-agent-dotnet/issues/1061

PR #1056, opened 2020-11-19
Get the inner exception when capturing HttpUnhandledException

### LINKED ISSUE #941: Agent should use inner exception of HttpUnhandledException
In .NET Framework ASP.NET Web Forms an exception that escaped from the request handler is wrapped by the framework into `System.Web.HttpUnhandledException` and the original exception is stored as the inner exception of `System.Web.HttpUnhandledException`. The agent uses only `System.Web.HttpUnhandledException` for error event but instead it should the inner exception since this is what the application is actually thrown.

russcam:
This commit captures the inner exception
when an exception is a HttpUnhandledException.

Tidy up naming.

Closes #941

### DIFF
--- a/sample/AspNetFullFrameworkSampleApp/AspNetFullFrameworkSampleApp.csproj
+++ b/sample/AspNetFullFrameworkSampleApp/AspNetFullFrameworkSampleApp.csproj
@@ -125,6 +125,10 @@
       <SubType>ASPXCodeBehind</SubType>
       <DependentUpon>Webforms.aspx</DependentUpon>
     </Compile>
+    <Compile Include="WebformsException.aspx.cs">
+      <SubType>ASPXCodeBehind</SubType>
+      <DependentUpon>WebformsException.aspx</DependentUpon>
+    </Compile>
     <Compile Include="App_Start\WebApiConfig.cs" />
     <Compile Include="Controllers\WebApiController.cs" />
     <Compile Include="Controllers\DatabaseController.cs" />
@@ -216,6 +220,7 @@
     <Content Include="Areas\MyArea\Web.config" />
     <Content Include="Areas\MyArea\_ViewStart.cshtml" />
     <Content Include="Webforms.aspx" />
+    <Content Include="WebformsException.aspx" />
     <Content Include="Scripts\umd\popper.min.js.map" />
     <Content Include="Scripts\umd\popper.js.map" />
     <Content Include="Scripts\umd\popper-utils.min.js.map" />
@@ -292,8 +297,6 @@
     <PackageReference Include="System.Data.SQLite">
       <Version>1.0.112</Version>
     </PackageReference>
-  </ItemGroup>
-  <ItemGroup>
     <PackageReference Include="System.Runtime.CompilerServices.Unsafe">
       <Version>5.0.0</Version>
     </PackageReference>
--- a/sample/AspNetFullFrameworkSampleApp/Controllers/DatabaseController.cs
+++ b/sample/AspNetFullFrameworkSampleApp/Controllers/DatabaseController.cs
@@ -28,8 +28,10 @@ namespace AspNetFullFrameworkSampleApp.Controllers
 		/// <returns></returns>
 		public ActionResult Index()
 		{
-			using var context = new SampleDataDbContext();
-			var samples = context.Set<SampleData>().ToList();
+			List<SampleData> samples;
+			using (var context = new SampleDataDbContext())
+				samples = context.Set<SampleData>().ToList();
+
 			return View(samples);
 		}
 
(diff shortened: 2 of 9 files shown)


gregkalapos (commented):
The code to capture the inner exception and the changes look good to me 👍 

If I see it correctly, there is no test covering this, right? It'd be nice to have a test case when `if (exception is HttpUnhandledException unhandledException)` is triggered and we assert on the captured error.

russcam:
> If I see it correctly, there is no test covering this, right? It'd be nice to have a test case when if (exception is HttpUnhandledException unhandledException) is triggered and we assert on the captured error.

There's no test for this, adding one now. An interesting consequence of capturing the inner exception is that we lose the stacktrace, which is on the `HttpUnhandledException`. We _could_ 

1. Capture it with `CaptureError` and provide the `HttpUnhandledException` stacktrace, but it seems messy to construct the components to do this
2. Check for `System.Web.HttpUnhandledException` type in `ExecutionSegmentCommon.CaptureException(...)`, and if it has an inner exception, use the inner exception for the `CapturedException` with the stacktrace of `HttpUnhandledException` 
3. Allow an empty stacktrace in the ASP.NET FF testing framework (the test fails because an assertion is made that stack frames is not empty).

Thoughts?

Would it be useful to capture the base exception ([`exception.GetBaseException()`](https://docs.microsoft.com/en-us/dotnet/api/system.exception.getbaseexception?view=net-5.0)) in separate error fields in future?

gregkalapos:
> > If I see it correctly, there is no test covering this, right? It'd be nice to have a test case when if (exception is HttpUnhandledException unhandledException) is triggered and we assert on the captured error.
> 
> There's no test for this, adding one now. An interesting consequence of capturing the inner exception is that we lose the stacktrace, which is on the `HttpUnhandledException`. We _could_
> 
> 1. Capture it with `CaptureError` and provide the `HttpUnhandledException` stacktrace, but it seems messy to construct the components to do this
> 2. Check for `System.Web.HttpUnhandledException` type in `ExecutionSegmentCommon.CaptureException(...)`, and if it has an inner exception, use the inner exception for the `CapturedException` with the stacktrace of `HttpUnhandledException`
> 3. Allow an empty stacktrace in the ASP.NET FF testing framework (the test fails because an assertion is made that stack frames is not empty).
> 
> Thoughts?
> 
> Would it be useful to capture the base exception ([`exception.GetBaseException()`](https://docs.microsoft.com/en-us/dotnet/api/system.exception.getbaseexception?view=net-5.0)) in separate error fields in future?

Oh, I see. Interesting.

I think option 2 won't work, because `HttpUnhandledException` lives in `System.Web.dll`, and `ExecutionSegmentCommon.CaptureException` is in `Elastic.Apm`, which does not reference `System.Web.dll` and it shouldn't do so, otherwise it'd pull in that dll in every environment.

Is the stacktrace on `HttpUnhandledException` the real stack trace from when the original exception is thrown? So does it contain the frame of the user code if an exception is thrown there? If so, I think option 1 isn't that bad, I'd just go with that. But if that's not a real stacktrace, then I think we should not bother to capture it, and option 3 is fine.

Btw - I blame `System.Web.dll` here. Why isn't the stacktrace on the inner exception? - not very logical  🤷 Anyways, that part we won't be able to change.

russcam:
> I think option 2 won't work, because HttpUnhandledException lives in System.Web.dll, and ExecutionSegmentCommon.CaptureException is in Elastic.Apm, which does not reference System.Web.dll and it shouldn't do so, otherwise it'd pull in that dll in every environment.

We would just need to check the full type name of the exception, no reference needed. I'm not keen on this, as this logic really belongs only where `System.Web.HttpUnhandledException` can be thrown. Thought I'd put it forward for discussion nonetheless.

> Is the stacktrace on HttpUnhandledException the real stack trace from when the original exception is thrown? So does it contain the frame of the user code if an exception is thrown there? If so, I think option 1 isn't that bad, I'd just go with that. But if that's not a real stacktrace, then I think we should not bother to capture it, and option 3 is fine.

I'll see if I can come up with a test where the framework wraps in `HttpUnhandledException` rather than throwing this directly, to see. If there's no stacktrace for the inner exception, or the stacktrace of  `HttpUnhandledException` isn't that useful, I'd also lean towards option 3.

russcam:
Have added a test where the framework wraps an exception in `HttpUnhandledException`. The inner exception has the stacktrace of where the error occurred, so this is good.

In a previous test, I was throwing a `HttpUnhandledException` directly inside a controller action, but this isn't a good representation of how the framework throws `HttpUnhandledException` from inside `System.Web.UI.Page.HandlerError(Exception)`.

gregkalapos (approved):
> The inner exception has the stacktrace of where the error occurred, so this is good.

🎉 great. 

PR looks good to me.

russcam:
jenkins run the tests

russcam:
failed tests are not related to this change. Merging in

Issue #1033, opened 2020-11-12
IExecutionSegment.Labels is deprecated and there is no replacement to get a label value


tbutler-qontigo:
**Describe the bug**
In 1.7.0 the Labels dictionary has been marked as deprecated with a message to use SetLabel instead.
This is fine for setting but what about getting the label value?
We run with <WarningsAsErrors /> so this fails our builds

**To Reproduce**
1. Use version 1.7.0
2. Set Warnings as Errors in your csproj file
2. Then create a transaction: var transaction = Agent.Tracer.StartTransaction(...)
3. Then access transaction.Labels - transaction.Labels["someLabel"] or transaction.Labels.ContainsKey("someKey") 
4. See the error when you build

**Expected behavior**
I would expect replacements to be available to a) get a label and b) assert if a label exists.

gregkalapos:
Thanks for opening this @tbutler-qontigo  - this is useful feedback.

The `Labels` dictionary which is marked as obsolete should work as it did before in `1.7.0`, but we plan to remove it at some point in the future. Reason for this is that it only supports strings, but we want to offer an API to also support `bool` and numbers. Only thing to keep in mind is that if you set something with the new `SetLabel()` method, you will not get the value back from the `.Labels` dictionary even if it's a string.

> We run with so this fails our builds

These obsolete usages should by-default produce a warning. In this case as of now it's safe to ignore those warnings in your tool. E.g. you can do [this](https://stackoverflow.com/a/968311/1783306) - with that you can still keep treating warnings as errors.

> This is fine for setting but what about getting the label value?

That's actually a use case we thought no one would do. But I'm totally aware that it was possible with the old API (and that's why we kept it as it is for now). Can you share a bit more about your use case? Like, why do you read the value back? What exactly do you do with it? We thought people will just set a label, and expect the value to show up in Elasticsearch, but never really wanting to read it back in C#.

tbutler-qontigo:
Hi @gregkalapos Thanks for getting back to me.
I know I could #pragma my way out of this but using something that is obsolete today is only storing up a problem for tomorrow so we have a strong preference to avoid obsolete methods and properties - after all, the point of marking them as such is to indicate that they will go away at some point so consumers should start using the replacement methods and properties now right?

Our current use case is in fact just in unit tests of some custom middleware which creates labels under certain circumstances but not others - our tests verify this behaviour.
Going forward, we may have occasion to extract labels in actual application code.
[...]

PR #1057, opened 2020-11-19
Add GetLabel method

gregkalapos:
Addresses https://github.com/elastic/apm-agent-dotnet/issues/1033 

In release `1.7.0` we marked the `Labels` property as obsolete and added the `SetLabel()` method with multiple overloads to support `string`, `bool`, and numbers.

This PR adds a method to read back the values from the agent. Since the `Labels` property supported both writing and reading back the value, we decided to not break users who rely on the reading use case in the future - therefore we agreed we offer a way to read values also with the new API.

Here is how it works (same on span) (updated version):

```
transaction.SetLabel("intLabel", 42);
var myLabelValue = transaction.GetLabel<int>("intLabel"); //myLabelValue is an `int`
Console.WriteLine(myLabelValue);
```

TODO:
- [x] Given the API is still up for discussion, make sure the `docs/public-api.asciidoc` documents the final version.

### DIFF
--- a/docs/public-api.asciidoc
+++ b/docs/public-api.asciidoc
@@ -282,6 +282,19 @@ transaction.SetLabel("intSample", 42);
 * `String|Number|bool value`: The tag value
 
 [float]
+[[api-transaction-get-label]]
+==== `Label GetLabel(string key)` added[1.8.0,Number and boolean labels require APM Server 6.7+]
+
+Returns the label which was added in the <<api-transaction-set-label, SetLabel>> method. If the key does not exist, the method returns `null`.
+
+
+[source,csharp]
+----
+var myLabel = transaction.GetLabel("foo");
+Console.WriteLine(myLabel.Value);
+----
+
+[float]
 [[api-transaction-tags]]
 ==== `Dictionary<string,string> Labels`
 
@@ -566,6 +579,20 @@ span.SetLabel("intSample", 42);
 
 * `String key`:   The tag key
 * `String|Number|bool value`: The tag value
+
+
+[float]
+[[api-span-get-label]]
+==== `Label GetLabel(string key)` added[1.8.0,Number and boolean labels require APM Server 6.7+]
+
+Returns the label which was added in the <<api-span-set-label, SetLabel>> method. If the key does not exist, the method returns `null`.
+
+
+[source,csharp]
+----
+var myLabel = span.GetLabel("foo");
+Console.WriteLine(myLabel.Value);
+----
 
 [float]
 [[api-span-tags]]
--- a/docs/public-api.asciidoc
+++ b/docs/public-api.asciidoc
@@ -282,17 +282,17 @@ transaction.SetLabel("intSample", 42);
 * `String|Number|bool value`: The tag value
 
 [float]
-[[api-transaction-get-label]]
-==== `Label GetLabel(string key)` added[1.8.0,Number and boolean labels require APM Server 6.7+]
+[[api-transaction-try-get-label]]
+==== `T GetLabel<T>(string key, out T value)` added[1.7.1,Number and boolean labels require APM Server 6.7+]
 
-Returns the transaction's label. If the `key` does not exist, this method returns `null`.
+Returns the transaction's label in the `value` out parameter. If the `key` does not exist, this method returns false.
 Labels can be added with the <<api-transaction-set-label, SetLabel>> method.
 
 
 [source,csharp]
 ----
-var myLabel = transaction.GetLabel("foo");
-Console.WriteLine(myLabel.Value);
+if(transaction.TryGetLabel<int>("foo", our var myLabel))
+    Console.WriteLine(myLabel);
 ----
 
 [float]
@@ -583,17 +583,17 @@ span.SetLabel("intSample", 42);
 
 
 [float]
-[[api-span-get-label]]
-==== `Label GetLabel(string key)` added[1.8.0,Number and boolean labels require APM Server 6.7+]
+[[api-span-try-get-label]]
+==== `T GetLabel<T>(string key, out T value)` added[1.7.1,Number and boolean labels require APM Server 6.7+]
 
-Returns the span's label. If the `key` does not exist, this method returns `null`.
+Returns the span's label in the `value` out parameter. If the `key` does not exist, this method returns false.
 Labels can be added with the <<api-span-set-label, SetLabel>> method.
 
 
 [source,csharp]
 ----
-var myLabel = span.GetLabel("foo");
-Console.WriteLine(myLabel.Value);
+if(span.TryGetLabel<bool>("foo", out var myLabel))
+    Console.WriteLine(myLabel);
 ----
 
 [float]
--- a/docs/public-api.asciidoc
+++ b/docs/public-api.asciidoc
@@ -285,7 +285,8 @@ transaction.SetLabel("intSample", 42);
 [[api-transaction-get-label]]
 ==== `Label GetLabel(string key)` added[1.8.0,Number and boolean labels require APM Server 6.7+]
 
-Returns the label which was added in the <<api-transaction-set-label, SetLabel>> method. If the key does not exist, the method returns `null`.
+Returns the transaction's label. If the `key` does not exist, this method returns `null`.
+Labels can be added with the <<api-transaction-set-label, SetLabel>> method.
 
 
 [source,csharp]
@@ -585,7 +586,8 @@ span.SetLabel("intSample", 42);
 [[api-span-get-label]]
 ==== `Label GetLabel(string key)` added[1.8.0,Number and boolean labels require APM Server 6.7+]
 
-Returns the label which was added in the <<api-span-set-label, SetLabel>> method. If the key does not exist, the method returns `null`.
+Returns the span's label. If the `key` does not exist, this method returns `null`.
+Labels can be added with the <<api-span-set-label, SetLabel>> method.
 
 
 [source,csharp]
--- a/src/Elastic.Apm/Model/NoopSpan.cs
+++ b/src/Elastic.Apm/Model/NoopSpan.cs
@@ -112,6 +112,8 @@ namespace Elastic.Apm.Model
 
 		public void SetLabel(string key, decimal value) { }
 
+		public Label GetLabel(string key) => null;
+
 		public ISpan StartSpan(string name, string type, string subType = null, string action = null) =>
 			new NoopSpan(name, type, subType, action, _currentExecutionSegmentsContainer, Id, TraceId, this);
 	}
--- a/src/Elastic.Apm/Model/NoopSpan.cs
+++ b/src/Elastic.Apm/Model/NoopSpan.cs
@@ -112,7 +112,7 @@ namespace Elastic.Apm.Model
 
 		public void SetLabel(string key, decimal value) { }
 
-		public Label GetLabel(string key) => null;
+		public T GetLabel<T>(string key) => default;
 
 		public ISpan StartSpan(string name, string type, string subType = null, string action = null) =>
 			new NoopSpan(name, type, subType, action, _currentExecutionSegmentsContainer, Id, TraceId, this);
--- a/src/Elastic.Apm/Model/NoopSpan.cs
+++ b/src/Elastic.Apm/Model/NoopSpan.cs
@@ -112,9 +112,13 @@ namespace Elastic.Apm.Model
 
 		public void SetLabel(string key, decimal value) { }
 
-		public T GetLabel<T>(string key) => default;
-
 		public ISpan StartSpan(string name, string type, string subType = null, string action = null) =>
 			new NoopSpan(name, type, subType, action, _currentExecutionSegmentsContainer, Id, TraceId, this);
+
+		public bool TryGetLabel<T>(string key, out T value)
+		{
+			value = default;
+			return false;
+		}
 	}
 }
--- a/src/Elastic.Apm/Model/Span.cs
+++ b/src/Elastic.Apm/Model/Span.cs
@@ -223,9 +223,14 @@ namespace Elastic.Apm.Model
 			{ nameof(IsSampled), IsSampled }
 		}.ToString();
 
-		public Label GetLabel(string key) => Context.InternalLabels.Value.InnerDictionary.ContainsKey(key)
-			? Context.InternalLabels.Value.InnerDictionary[key]
-			: null;
+		public T GetLabel<T>(string key)
+		{
+			if (Context.InternalLabels.Value.InnerDictionary.TryGetValue(key, out var label))
+				if (label?.Value is T t) return t;
+
+			return default;
+		}
+
 
 		public ISpan StartSpan(string name, string type, string subType = null, string action = null)
 		{
--- a/src/Elastic.Apm/Model/Span.cs
+++ b/src/Elastic.Apm/Model/Span.cs
@@ -223,12 +223,19 @@ namespace Elastic.Apm.Model
 			{ nameof(IsSampled), IsSampled }
 		}.ToString();
 
-		public T GetLabel<T>(string key)
+		public bool TryGetLabel<T>(string key, out T value)
 		{
 			if (Context.InternalLabels.Value.InnerDictionary.TryGetValue(key, out var label))
-				if (label?.Value is T t) return t;
+			{
+				if (label?.Value is T t)
+				{
+					value = t;
+					return true;
+				}
+			}
 
-			return default;
+			value = default;
+			return false;
 		}
 
 
--- a/src/Elastic.Apm/Model/Span.cs
+++ b/src/Elastic.Apm/Model/Span.cs
@@ -223,6 +223,10 @@ namespace Elastic.Apm.Model
 			{ nameof(IsSampled), IsSampled }
 		}.ToString();
 
+		public Label GetLabel(string key) => Context.InternalLabels.Value.InnerDictionary.ContainsKey(key)
+			? Context.InternalLabels.Value.InnerDictionary[key]
+			: null;
+
 		public ISpan StartSpan(string name, string type, string subType = null, string action = null)
 		{
 			if (ConfigSnapshot.Enabled && ConfigSnapshot.Recording)
--- a/src/Elastic.Apm/Model/Transaction.cs
+++ b/src/Elastic.Apm/Model/Transaction.cs
@@ -384,7 +384,7 @@ namespace Elastic.Apm.Model
 
 		public Label GetLabel(string key) => _context.Value.InternalLabels.Value.InnerDictionary.ContainsKey(key)
 			? _context.Value.InternalLabels.Value.InnerDictionary[key]
-			: null
+			: null;
 
 		public ISpan StartSpan(string name, string type, string subType = null, string action = null)
 		{
--- a/src/Elastic.Apm/Model/Transaction.cs
+++ b/src/Elastic.Apm/Model/Transaction.cs
@@ -22,12 +22,12 @@ namespace Elastic.Apm.Model
 	internal class Transaction : ITransaction
 	{
 		private static readonly string ApmTransactionActivityName = "ElasticApm.Transaction";
+		private readonly IApmServerInfo _apmServerInfo;
 		private readonly Lazy<Context> _context = new Lazy<Context>();
 		private readonly ICurrentExecutionSegmentsContainer _currentExecutionSegmentsContainer;
 
 		private readonly IApmLogger _logger;
 		private readonly IPayloadSender _sender;
-		private readonly IApmServerInfo _apmServerInfo;
 
 		private readonly string _traceState;
 
@@ -381,6 +381,10 @@ namespace Elastic.Apm.Model
 			_sender.QueueTransaction(this);
 			_currentExecutionSegmentsContainer.CurrentTransaction = null;
 		}
+
+		public Label GetLabel(string key) => _context.Value.InternalLabels.Value.InnerDictionary.ContainsKey(key)
+			? _context.Value.InternalLabels.Value.InnerDictionary[key]
+			: null
 
 		public ISpan StartSpan(string name, string type, string subType = null, string action = null)
 		{
--- a/src/Elastic.Apm/Model/Transaction.cs
+++ b/src/Elastic.Apm/Model/Transaction.cs
@@ -382,12 +382,19 @@ namespace Elastic.Apm.Model
 			_currentExecutionSegmentsContainer.CurrentTransaction = null;
 		}
 
-		public T GetLabel<T>(string key)
+		public bool TryGetLabel<T>(string key, out T value)
 		{
 			if (Context.InternalLabels.Value.InnerDictionary.TryGetValue(key, out var label))
-				if (label?.Value is T t) return t;
+			{
+				if (label?.Value is T t)
+				{
+					value = t;
+					return true;
+				}
+			}
 
-			return default;
+			value = default;
+			return false;
 		}
 
 		public ISpan StartSpan(string name, string type, string subType = null, string action = null)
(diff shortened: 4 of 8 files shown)


src/Elastic.Apm/Model/NoopSpan.cs
```diff
@@ -112,6 +112,8 @@ public Task<T> CaptureSpan<T>(string name, string type, Func<ISpan, Task<T>> fun
 
 		public void SetLabel(string key, decimal value) { }
 
+		public Label GetLabel(string key) => null;
```
gregkalapos, line 115: This is up for discussion. If the agent is turned off (`enabled=false` or `recording=false`), then this code runs.

My thinking was that if the agent is turned off, then it doesn't do anything and the advantage here is that we cause basically no overhead. If we new up `Context` instances and keep track of labels and other things, then the overhead with the turned off agent will be more.

I feel `NoopSpan` should be really NoOp and if the agent is turned off, we should work with NoOp instances.

Same for `NoopTransaction`.

src/Elastic.Apm/Model/Span.cs
```diff
@@ -232,6 +232,10 @@ public override string ToString() => new ToStringBuilder(nameof(Span))
 			{ nameof(IsSampled), IsSampled }
 		}.ToString();
 
+		public Label GetLabel(string key) => Context.InternalLabels.Value.InnerDictionary.ContainsKey(key)
+			? Context.InternalLabels.Value.InnerDictionary[key]
+			: null;
```
russcam, line 237:
```suggestion
		public Label GetLabel(string key)
		{
		    Context.InternalLabels.Value.InnerDictionary.TryGetValue(key, out var label);
		    return label;
		}
```

russcam: With `TryGetValue` , the lookup is performed once

src/Elastic.Apm/Model/Transaction.cs
```diff
@@ -382,6 +382,10 @@ public void End()
 			_currentExecutionSegmentsContainer.CurrentTransaction = null;
 		}
 
+		public Label GetLabel(string key) => _context.Value.InternalLabels.Value.InnerDictionary.ContainsKey(key)
+			? _context.Value.InternalLabels.Value.InnerDictionary[key]
+			: null;
```
russcam, line 387:
```suggestion
		public Label GetLabel(string key)
		{
		    _context.Value.InternalLabels.Value.InnerDictionary.TryGetValue(key, out var label);
			return label;
		}
```

src/Elastic.Apm/Model/NoopSpan.cs
russcam: I think this is fine. The behaviour should be summarized in an `IMPORTANT` admonition in the enabled and recording configuration docs e.g. "When set to false. labels are not captured"

russcam (approved):
LGTM 👍 I left some suggestions for consideration

docs/public-api.asciidoc
```diff
@@ -281,6 +281,19 @@ transaction.SetLabel("intSample", 42);
 * `String key`:   The tag key
 * `String|Number|bool value`: The tag value
 
+[float]
+[[api-transaction-get-label]]
+==== `Label GetLabel(string key)` added[1.8.0,Number and boolean labels require APM Server 6.7+]
+
+Returns the label which was added in the <<api-transaction-set-label, SetLabel>> method. If the key does not exist, the method returns `null`.
```
bmorelli25, line 288:
```suggestion
Returns the transaction's label. If the `key` does not exist, the method returns `null`.

TIP: Labels are added with the <<api-transaction-set-label, SetLabel>> method.
```

docs/public-api.asciidoc
```diff
@@ -567,6 +580,20 @@ span.SetLabel("intSample", 42);
 * `String key`:   The tag key
 * `String|Number|bool value`: The tag value
 
+
+[float]
+[[api-span-get-label]]
+==== `Label GetLabel(string key)` added[1.8.0,Number and boolean labels require APM Server 6.7+]
+
+Returns the label which was added in the <<api-span-set-label, SetLabel>> method. If the key does not exist, the method returns `null`.
```
bmorelli25, line 588:
```suggestion
Returns the span's label. If the `key` does not exist, the method returns `null`.

TIP: Labels are added with the <<api-span-set-label, SetLabel>> method.
```

bmorelli25 (approved):
I think the docs are fine as-is.

> Returns the label which was added in the <<api-transaction-set-label, SetLabel>> method.

This sentence felt a little _off_ to me, so I made a recommendation, but I'm ok with you ignoring or changing it.

tbutler-qontigo:
Hi guys - I appreciate the effort you are putting into this and the speed with which you are doing so.

I have some feedback on this change:

I guess it is a consequence of Label having an `object` `Value` vs being a generic but it feels a bit odd that I can `SetLabel` with a specific type (because you provide overloads for many types) but when I `GetLabel` I get an object.
Have you considered refactoring `Label` to support a generic type for `Value`?

Alternatively (or perhaps additionally) would it make more sense for `GetLabel` to return the `Value` of the `Label` rather than an actual `Label` instance?  Again, thinking of a get being the the complement to a set method, I would either expect `SetLabel` to take a `Label` and `GetLabel` to return a `Label` or, since `SetLabel` takes a value, I would expect `GetLabel` to return the value I'd previousl set vs a `Label` with a value.
The latter could then be a generic to return the correct type for the value.

gregkalapos:
> Hi guys - I appreciate the effort you are putting into this and the speed with which you are doing so.
> 
> I have some feedback on this change:
> 
> I guess it is a consequence of Label having an `object` `Value` vs being a generic but it feels a bit odd that I can `SetLabel` with a specific type (because you provide overloads for many types) but when I `GetLabel` I get an object.
> Have you considered refactoring `Label` to support a generic type for `Value`?
> 
> Alternatively (or perhaps additionally) would it make more sense for `GetLabel` to return the `Value` of the `Label` rather than an actual `Label` instance? Again, thinking of a get being the the complement to a set method, I would either expect `SetLabel` to take a `Label` and `GetLabel` to return a `Label` or, since `SetLabel` takes a value, I would expect `GetLabel` to return the value I'd previousl set vs a `Label` with a value.
> The latter could then be a generic to return the correct type for the value.


Yeah, I personally agree with this feedback. Accepting `int`, `bool`, etc. in `SetLabel()` and returning `Label` (with `object Label.Value`) from `GetLabel()` isn't very nice.

Making it generic seems reasonable, here is how I updated the PR:

```
/// <summary>
/// Returns the value of a label.
/// </summary>
/// <param name="key">The key of the label that you would like to read</param>
/// <typeparam name="T">The type of the value for the given label. If the type does not match this method returns <code>default(T)</code></typeparam>
/// <returns>The value of the label if the key exists, <code>default(T)</code> otherwise.</returns>
T GetLabel<T>(string key);
```

Usage (same for `int`, `double`, `bool`, `string`, etc.)

```
transaction.SetLabel("intLabel", 42);
var intValue = transaction.GetLabel<int>("intLabel");
```


Advantage: 
- You put an `int` into `SetLabel()`, you get back an `int` from `GetLabel()` (some for other types).
- You don't need to know our `Label` type, specifically you don't need to know that there is `Value` property, which is an `object` that you need to cast. You only work with built-in types, which is natural to people.

Disadvantages:
1.  In `GetLabel()` you need to know the type of the given label, otherwise you don't get back the value.
2. We return `default(T)`, meaning transaction.GetLabel<int>("foo") will return `0` (`false` in case of `bool`) if the value doesn't exis, but also when the value is in fact `0` (or `false` in case of `bool`)


I think this is acceptable, specifically I think point 1 is fine. Regarding point 2 we could:
- Throw an exception if `key` doesn't exists (I personally don't think this is fine - we should never make the application crash, even if it's some invalid input into the public API.)
- Use nullable, but with that we are almost back to the problem returning `Label`: people need to read the value from `.Value` meaning they don't get back the value directly.
- Also add an `bool IsKeyPresentInLabels(string key`) method (or something with better name).This sounds like an overkill to me (3 label related methods, while most people only want `SetLabel`), but we can add it any time later if we get feedback that people really need it.

@russcam @tbutler-qontigo what do you think?

src/Elastic.Apm/Model/Span.cs
gregkalapos: Agreed - I'm using `TryGetValue` with some other changes described [here](https://github.com/elastic/apm-agent-dotnet/pull/1057#issuecomment-731802674).

src/Elastic.Apm/Model/Transaction.cs
gregkalapos: Agreed - I'm using `TryGetValue` with some other changes described [here](https://github.com/elastic/apm-agent-dotnet/pull/1057#issuecomment-731802674).

russcam:
Thanks for raising @tbutler-qontigo! 

Having the generic `GetLabel<T>(string key)` method doesn't provide any real benefit and makes the method trappy, in my opinion:

- With the generic method, you need to know the label value type to use it effectively, and using the wrong type that is a value type, returns a `default(T)` that cannot distinguish between a default value that has been set and one that hasn't. Further, this doesn't tell you if a label with the specific key has actually been set.

Conversely,

- With the non-generic method returning `Label`, it is known whether a label has been set with a specific key as an instance of `Label` is returned. To use this label type effectively, you still need to know the label value type, but at least the type can be reflected and cast to the correct type.

The incongruency of setting a specific label value type and getting an instance of `Label` back is a little unfortunate, but I think would be a breaking change to change now in 1.x. We could move to [`SetLabel()` accepting an instance of `Label` and have implicit conversions for the supported types in the next major version](https://github.com/elastic/apm-agent-dotnet/pull/982#discussion_r515689796). This would break the API surface, but would not require changes to usage.

> I guess it is a consequence of Label having an object Value vs being a generic but it feels a bit odd that I can SetLabel with a specific type (because you provide overloads for many types) but when I GetLabel I get an object.
Have you considered refactoring Label to support a generic type for Value?

It was considered to have `Label<T>` with a `T` value, but I don't think it provides any real value in this case, besides avoiding the boxing, and would be harder to work with, in my opinion. Since labels are held in a dictionary, we would need a non-generic `Label` base type to type the `TValue` in `Dictionary<TKey, TValue>` to, and type check each label at serialization time to see what closed generic type it is, or type check `T` in a generic serialization routine to understand how to serialize `T`. `T` could also be boxed at this point to `object` and passed to `WriteValue(object value)`, but then the boxing is still there 🙂 On the `GetLabel()` side, `T` would still need to be known to use effectively, I think - unless I'm missing something?

> Alternatively (or perhaps additionally) would it make more sense for GetLabel to return the Value of the Label rather than an actual Label instance?

`object GetLabel(string key)` could work, and would be a better complement to `SetLabel()` overloads. It does raise some questions

1. Are there cases where returning `Label` might be more useful? It only has a `Value` property right now, but might it have other properties in future? I don't think it will.
2. Are there cases where working with labels might afford us some flexibility? Difficult to know this ahead of time. A contrived example might be introducing methods on labels to increment numeric values. (An aside, but if `SetLabel` were to accept an instance of `Label`, consumers could derive their own label types from `Label` to add this logic).

On that last question, I think `Label GetLabel(string key)` might be a more flexible approach for now, with moving to `SetLabel(string key, Label label)` in the next major.

tbutler-qontigo:
Hi guys

I think if you are going to have `object GetLabel(string label)` then only having `SetLabel(string label, object value)` makes sense - they complement each other - to me it feels wrong to be able to add explicit types but be unable to retrieve those explicit types.

for values that have not been set you could be "inspired" by the dictionary model
 - throw a `KeyNotFoundException` if it does not exist (but unlike `Dictionary`, actually include the key in the message!
- Have a `TryGetValue` (or `TryGetLabel`?) method which does not throw if it does not exist
- Have a `ContainsKey` (or `ContainsLabel`) method to see if it exists - `GetLabel` does not necessarily have to throw if it does not exist and could return the default

> With the generic method, you need to know the label value type to use it effectively, and using the wrong type that is a value type, returns a default(T) that cannot distinguish between a default value that has been set and one that hasn't. Further, this doesn't tell you if a label with the specific key has actually been set.

One could argue that you need to know the label should exist at all - otherwise why are you trying to access it? - and that you will likely need to know the type anyway on the basis that you are retrieving it so you are likely to be doing something with it.
If you only care that it exists then you could use a suitable Contains method; if you care about the value then you are probably going to need it typed appropriately anyway.

> On that last question, I think Label GetLabel(string key) might be a more flexible approach for now, with moving to SetLabel(string key, Label label) in the next major.

Perhaps a Label has a key and a value so you `void SetLabel(Label label)` and `Label GetLabel(string key)`?
If Label is an object then why doesn't it encapsulate everything about the label - which is it's key and it's value?
Unless you can see a use case for the same instance of `Label` but with different keys?

russcam:
> I think if you are going to have object GetLabel(string label) then only having SetLabel(string label, object value) makes sense - they complement each other - to me it feels wrong to be able to add explicit types but be unable to retrieve those explicit types.

`SetLabel()` method overloads constrain label types only to those that are supported by APM and Elasticsearch. Accepting `object` would not control API invariants, allowing it to be misused and fail at runtime. This is the primary reason for having a `Label` type - it constrains the type of `Label.Value` only to those supported.

> for values that have not been set you could be "inspired" by the dictionary model
> 
> - throw a `KeyNotFoundException` if it does not exist (but unlike `Dictionary`, actually include the key in the message!
> - Have a `TryGetValue` (or `TryGetLabel`?) method which does not throw if it does not exist
> - Have a `ContainsKey` (or `ContainsLabel`) method to see if it exists - GetLabel does not necessarily have to throw if it does not exist and could return the default

With `Label GetLabel(string key)`, I don't think we need any of these - if an instance of `Label` is returned, it exists, if `null` is returned, it doesn't. I agree with your assertion that you're likely to try to get labels only if they exist and are also likely to know the types.

> Perhaps a Label has a key and a value so you void `SetLabel(Label label)` and `Label GetLabel(string key)`?
If Label is an object then why doesn't it encapsulate everything about the label - which is it's key and it's value?
Unless you can see a use case for the same instance of Label but with different keys?

Labels themselves have no associated key, they only have a value - or more accurately, they _are the value_!. The `Label` type exists to constrain the types of values that can be supplied as labels.

tbutler-qontigo:
Hi @russcam 

> Accepting object would not control API invariants,

Agreed.  This is not my preferred option - but it feels like your Sets and Gets should complement each other - so if you accept an object you should return an object and vice versa.
I would prefer that you do not accept object at all and do not return object at all.  
Likewise if you are going to return a `Label` then you should accept it in the Set

>I agree with your assertion that you're likely to try to get labels only if they exist and are also likely to know the types.

So if we assume the caller knows the type then `GetLabel<T>` would seem acceptable - and if it throws because you are casting to the wrong type then so be it.

> The Label type exists to constrain the types of values that can be supplied as labels

Don't your SetLabel overloads do that?  If all they will accept is types supported by APM and elastic search then you do not need the `Label` type at all - or at least you do not need to expose this publicly.  All you need to expose is the actual value.

gregkalapos:
image

gregkalapos:
I went through the comments and I still think what I said in [my last comment](https://github.com/elastic/apm-agent-dotnet/pull/1057#issuecomment-731802674).

I think we all agree that the main use case is setting the label. By far, most people are only interested in `SetLabel`. Therefore I still think we should keep it as simple and easy to use as possible. If we'd accept `Label` instead of the specific types then a whole bunch of question will arise (what is `Label`? Why doesn't it have the key? Can I set my own type into it? Oh it has implicit conversion... what's that? For what types does it have implicit conversion? etc.)

If we keep the current overloads then here is what users see, and that's in my opinion the cleanest and simplest way to communicate what types we accept:
image

I feel seeing that intellisense tells everything - no need to read docs or dive into the source code.

So in my opinion, using `Label` on the API surface adds more confusion for our users. The reading use case is less common and making the common API (`SetLabel`) to more confusing because of a less common API (`GetLabel`) isn't the good move.

> 1. Are there cases where returning `Label` might be more useful? It only has a `Value` property right now, but might it have other properties in future? I don't think it will.
> 2. Are there cases where working with labels might afford us some flexibility? Difficult to know this ahead of time. A contrived example might be introducing methods on labels to increment numeric values. (An aside, but if `SetLabel` were to accept an instance of `Label`, consumers could derive their own label types from `Label` to add this logic).

We are talking about a use-case here where people set a simple label on a span or on a transaction. I feel this shouldn't be more than that. So to point 1: No, because I think all we need to return is the value itself. And to point 2: I feel all those functionality would be huge overkill and confusing. Like in the increment use case: let's say you set a label on a span, and then you keep the `Label` instance and at some point you increment the number: what if the span is already sent to APM Server? Plus, do we really want to change stuff on a span through a `Label` reference? I think these are very dangerous use cases and super confusing for very little benefit - if someone wants to increment a label (which I think very very few people will do) then they should call `SetLabel(string key, int value)` on the span or transaction again with the incremented number. That's guaranteed to work as long as you don't call `.End()` on the span/transaction.

> Having the generic `GetLabel<T>(string key)` method doesn't provide any real benefit and makes the method trappy, in my opinion:

I think it does:
- You don't need to learn any new type like our `Label` (keep in mind, people won't know what that is, they'll need to read docs for that or look at the code).
- No need to know how to get out the value of the thing which we return, because we return the value itself.
- With that we basically have the same type on the `SetLabel` (`transaction.SetLabel("intLabel", 42)`) and `GetLabel` (`transaction.GetLabel<int>("intLabel")`).

> - With the generic method, you need to know the label value type to use it effectively, and using the wrong type that is a value type, returns a `default(T)` that cannot distinguish between a default value that has been set and one that hasn't. Further, this doesn't tell you if a label with the specific key has actually been set.

I totally share this concern. I don't think it's an issue that you need to know the type - but the `default(T)` problem is a valid problem indeed as I said in [my original comment](https://github.com/elastic/apm-agent-dotnet/pull/1057#issuecomment-731802674). I also offered some solutions for it (which were partly brought up here). Any thoughts on those? 

Specifically (both are partly in-sync with "being inspired by the dictionary model")
- Adding a method that returns `bool` if the key exits (this method would also care about the type of the given key)
- Throwing an exception in case the key doesn't exist

I still think both of those are be a nicer solutions than changing the common API (`SetLabel`) to something less good due to this problem.

russcam:
I'd like to elaborate on the thought process here.

Agreed that the common use case is to set labels and would expect getting labels to be much less common.

Current discussion aside for the moment, I think it's useful to look at the requirements of the agent API:

1. We want to support setting a label value of string, boolean and numeric types against a string key.
2. We don't want to allow a user to be able to set a label value type that is unsupported by APM or Elasticsearch.
3. We want to support returning the label value set against a string key.

In order to collect string keys and label values of the types we want to support, the simplest data structure is a `Dictionary<string, object>`. This data structure can't be exposed directly in the API however, as it would allow users to set values that are not supported, so the dictionary must be internal to the agent and expose methods for setting key/values and getting values by key. The simplest approach would be to expose set methods for the label value types supported. An alternative approach would be to encapsulate the label value types supported in a type _(`Label` type)_, and expose a single set method that accepts this type. With the latter approach, it would be reasonable to change the `TValue` of `Dictionary<string, object>` to be the `Label` type, to avoid boxing. This approach also has the benefit of being able to handle this type specifically when it comes to serialization.

Getting label values that have been set, in its simplest form, would be to return the `TValue` of the backing `Dictionary<string, TValue>`; `TValue` would be dictated by the chosen backing structure.

With this in mind, the current implementation is somewhat a mixture of two approaches - uses a backing `Dictionary<string, Label>` with overloaded set methods. I don't see having supported types encapsulated in a `Label` type as being inherently more complex than having set overloads when both are backed by a `Label` `TValue` type; the encapsulation of supported types is just in a different location. If labels can be set in more than one place in the code base, encapsulating supported types in a `Label` type removes the need to implement all set overloads at each location, and makes the addition of new supported types easier in the future e.g. support for `Guid`, `DateTime`, and `DateTimeOffset` as string values, etc. 

It's a valid point that encapsulating supported types in a `Label` type introduces a layer of abstraction to setting a label, because the overloads do not describe the supported types. Does this layer of abstraction make setting labels inherently more difficult? I think this is largely an opinion-based answer, but if it were to be compared to what a user must know in order to use other parts of the Agent API such as starting and ending transactions and spans, capturing exceptions, etc., setting labels with a `Label` type looks much less complex in my opinion. With the set method accepting a `Label` type, the simplest counterpart is a get method returning a `Label` type if it exists, and `null` otherwise.

Looking at what is publicly exposed currently and putting aside the internal data structure used, there are set method overloads that accept a label value type of the supported types. When a user gets a label value that was previously set, what should be returned? If the key exists, the label value should be returned, and if the key does not exist, then what? Further, in what form should a label value be returned? Given the set method overloads and the fact that C# does not support overloaded methods differing in return type, the two _simplest_ choices when a key exists appear to be

1. return the internal `TValue` of the dictionary
2. provide a generic method to return `TValue` cast as a supported label type, `T`

The approach however must also accommodate reasonable behaviour when a key does not exist. When returning the internal `TValue` per point 1, whether `TValue` is `object` or `Label`, when a key exists, returns an instance of `object` or `Label`, and when it does not exist, returns `null`. 

For point 1, there are additional considerations when `TValue` is `Label` compared to when `TValue` is `object`. Given a label value is set with one of the set overloads, returning an internal `Label` structure could be confusing, I agree. If set were to accept a `Label` in future, then it might still be considered confusing in this version, but not so in future.

When `TValue` is `object`, returning `object` for getting a label seems perfectly reasonable, given labels support different types, and allows differentiating between a value that has been set and a missing value for the given key, with a single get method.

For point 2, providing a generic method to return the value cast as `T` is problematic when 

1. the key exists but `T` is not the type that was set
2. when the key does not exist. if `default(T)` is returned, this could be misleading for value types, as the default value returned will not differentiate between a value that was set to the default value and a value that was missing.

This is what I mean by _trappy_ - values may be returned that were never set.

If a generic method were to be supported, `bool TryGetLabel<T>(string key, out T value)` overcomes these problems. Whether this should be supported _in addition_ to a `T GetLabel<T>(string key)` that throws could be discussed, but in the interests of keeping the agent API as simple as possible and not wanting to avoid throwing exceptions for misuse, `bool TryGetLabel<T>(string key, out T value)` satisfies the requirements outlined at the start. _(This approach does remove the need to have a `Label`, because the backing data structure can simply be `Dictionary<string, object>` with a custom converter. Removing the `Label` type though would need to wait, as it's public)_.

russcam:
> _(This approach does remove the need to have a Label, because the backing data structure can simply be Dictionary<string, object> with a custom converter. Removing the Label type though would need to wait, as it's public)._

On this last point, an observation to consider is that passing labels is exposed on the public `CaptureException` API

https://github.com/elastic/apm-agent-dotnet/blob/e35228127a1b207c720c6efb9455b0544a95c0c4/src/Elastic.Apm/Api/IExecutionSegment.cs#L77-L103

It follows that understanding what a `Label` represents would need to be known in order to add labels to a captured exception.

gregkalapos:
I'm just trying to quickly respond to maybe have an agreement in our short working day overlap. If needed I'll also respond to the other parts of your comment @russcam.

The idea with `bool TryGetLabel<T>(string key, out T value)` totally makes sense to me instead of the `T GetLabel<T>(string key)` method.

Removing the `Label` type completely wasn't my intention, so I think we also agree on that.

So... do we go with `bool TryGetLabel<T>(string key, out T value)`? That'd be the only method we add in the PR.

russcam:
> So... do we go with bool `TryGetLabel<T>(string key, out T value)`? That'd be the only method we add in this PR.

Designing evolving APIs is hard 😀 I'm ok with `TryGetLabel<T>(string key, out T value)`, though my preference would be to return `Label` in a get method.

I think `TryGetLabel<T>(string key, out T value)` is reasonable, if we want to align with the current set overloads and don't want to promote `Label` in this context in future. If we believe we might want to promote `Label` in the future, then I'd still prefer we return `Label` from a get method. The reason for raising this is that the former doesn't align the concept of labels across setting on transactions and spans (pass a key and the value to `SetLabel()`), and capturing exceptions (add a `Label` encapsulating the value with a key), while the latter doesn't now either, but could in future, were `SetLabel` to accept a `Label`.

PR #1060, opened 2020-11-23
HostBuilderExtensions.UseElasticApm: subscribe to `subscribers` params

### LINKED ISSUE #1059: HostBuilderExtensions.UseElasticApm doesn't register DiagnosticsSubscribers

gregkalapos:
Fixes #1059

In `Elastic.Apm.Extensions.Hosting.HostBuilderExtensions.UseElasticApm(this IHostBuilder builder, params IDiagnosticsSubscriber[] subscribers)` the `params IDiagnosticsSubscriber[] subscribers` were never activated. This was because of the following wrong check:

```
if(Agent.IsConfigured && Agent.Config.Enabled)
  if (subscribers != null && subscribers.Any() && Agent.IsConfigured) Agent.Subscribe(subscribers);
```

Problem here was that `Agent.IsConfigured` remained false until `IApmAgent` got resolved first  - but that always runs after the check.

Proposed fix: we force the resolution directly within the `UseElasticApm` method:

```
var serviceProvider = services.BuildServiceProvider();
var agent = serviceProvider.GetService<IApmAgent>();
```

And do a similar check after that.

I feel doing "lazy agent loading" here makes no sense, the agent should be immediately enabled, otherwise it would e.g. not collect metrics, plus there is no guarantee that a user forces to resolve `IApmAgent`.

### DIFF
--- a/src/Elastic.Apm.Extensions.Hosting/HostBuilderExtensions.cs
+++ b/src/Elastic.Apm.Extensions.Hosting/HostBuilderExtensions.cs
@@ -55,11 +55,8 @@ namespace Elastic.Apm.Extensions.Hosting
 				services.AddSingleton<IApmAgent, ApmAgent>(sp =>
 				{
 					if (Agent.IsConfigured) return Agent.Instance;
-
-					var components = sp.GetService<AgentComponents>();
-					var newAgentInstance = new ApmAgent(components);
-					Agent.Setup(components);
-					return newAgentInstance;
+					Agent.Setup(sp.GetService<AgentComponents>());
+					return Agent.Instance;
 				});
 
 				services.AddSingleton(sp => sp.GetRequiredService<IApmAgent>().Tracer);
--- a/src/Elastic.Apm.Extensions.Hosting/HostBuilderExtensions.cs
+++ b/src/Elastic.Apm.Extensions.Hosting/HostBuilderExtensions.cs
@@ -28,12 +28,11 @@ namespace Elastic.Apm.Extensions.Hosting
 			builder.ConfigureServices((ctx, services) =>
 			{
 				//If the static agent doesn't exist, we create one here. If there is already 1 agent created, we reuse it.
-
 				if (!Agent.IsConfigured)
 				{
 					services.AddSingleton<IApmLogger, NetCoreLogger>();
 					services.AddSingleton<IConfigurationReader>(sp =>
-					new MicrosoftExtensionsConfig(ctx.Configuration, sp.GetService<IApmLogger>(), ctx.HostingEnvironment.EnvironmentName));
+						new MicrosoftExtensionsConfig(ctx.Configuration, sp.GetService<IApmLogger>(), ctx.HostingEnvironment.EnvironmentName));
 				}
 				else
 				{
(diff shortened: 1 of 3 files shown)


src/Elastic.Apm.Extensions.Hosting/HostBuilderExtensions.cs
```diff
@@ -57,15 +56,22 @@ public static IHostBuilder UseElasticApm(this IHostBuilder builder, params IDiag
 				{
 					if (Agent.IsConfigured) return Agent.Instance;
 
-					var apmAgent = new ApmAgent(sp.GetService<AgentComponents>());
+					var newAgentInstance = new ApmAgent(sp.GetService<AgentComponents>());
 					Agent.Setup(sp.GetService<AgentComponents>());
-					return apmAgent;
+					return newAgentInstance;
```
russcam, line 61: can agent components be resolved once, and passed to both agent instance and `Agent.Setup(...)` i.e.

```c#
var components = sp.GetService<AgentComponents>();
var newAgentInstance = new ApmAgent(components );
Agent.Setup(components);					
return newAgentInstance;
```

Also, does this mean there are potentially two instances of an agent in the application - a singleton instance created by this method, and the lazy instance on `Agent` that's used internally?

test/Elastic.Apm.Extensions.Hosting.Tests/HostBuilderExtensionTests.cs
```diff
@@ -24,9 +27,49 @@ public async Task TwoHostBuildersNoException()
...
+			var fakeSubscriber = new FakeSubscriber();
+			fakeSubscriber.IsSubscribed.Should().BeFalse();
+
+			Host.CreateDefaultBuilder()
+				.ConfigureServices((context, services) => { services.AddHostedService<HostedService>(); })
+				.UseElasticApm(fakeSubscriber)
+				.Build();
+
+			fakeSubscriber.IsSubscribed.Should().BeTrue();
+		}
```
russcam, line 57: Should there also be a test to assert that subscribers are not subscribed when the agent is not enabled?

russcam (approved):
LGTM 👍

src/Elastic.Apm.Extensions.Hosting/HostBuilderExtensions.cs
gregkalapos: > Also, does this mean there are potentially two instances of an agent in the application - a singleton instance created by this method, and the lazy instance on Agent that's used internally?

Hahh - that's a good point - we should not new up an `ApmAgent` instance here. Technically they were 2 instances, but they were sharing the `AgentComponents`, but still - best is to just return `Agent.Instance`. I changed this part.

> can agent components be resolved once, and passed to both agent instance and Agent.Setup(...) i.e.

With the previous change this is not needed. Otherwise I'd agree.

test/Elastic.Apm.Extensions.Hosting.Tests/HostBuilderExtensionTests.cs
gregkalapos: Agreed - I added a test.

Issue #1051, opened 2020-11-18
Automate creating of ElasticApmAgent_[version].zip file


gregkalapos:
For the zero code change (startup hook based) agent injection we distribute a `.zip` file with all the agent components. In the future we may use other ways to distribute this.

Regardless of how we distribute it we should:
- Generate a file or folder with all the agent components. By setting `DOTNET_STARTUP_HOOKS` the agent should be injected from the generates file/folder .
- Zip the content of this for now which we can publish here on GitHub.
- Run this whole thing in CI.
- Add a test stage with a sample app, so all commits are tested with this type of setup.

russcam:
I've opened #1062 to address building the zip file. I think it might make sense to split this issue into two, one to create zip, one to test zip, as the testing requires some work to ensure startup hooks work with different major versions of System.Diagnostics.DiagnosticSource.

PR #1062, opened 2020-11-25
Build automation for performing common tasks

russcam:
Relates: #1051

This PR introduces a small F# project to automate common build tasks cross-platform:

- building all projects
- building agent zip file for use with startup hooks
- generating release notes
- building nuget packages

Tasks can be extended in future.

The root of the solution includes a `build.bat` and `build.sh` that serves as an entrypoint into invoking the project. For example

- to build all projects

    ```bat
    .\build.bat
    ```
   or
    ```bat
    ./build.sh
    ```

- to create a zip file for use with startup hooks

    ```bat
    .\build.bat agent-zip
    ```
   or
    ```bat
    ./build.sh agent-zip
    ```

etc.

The Targets.fs file in the F# scripts project defines all of the available targets/tasks and their dependencies, as well as command line arguments. Other .fs files define the functions to invoke for a given target/task.

This build automation may be of interest to the APM integration/automation team to automate tasks for the .NET agent on CI.


build.bat
```diff
@@ -0,0 +1,2 @@
+@echo Off
```
gregkalapos, line 1: nit: could we add this and the `build.sh` file to the solution under `files` - we have all other files there. 

_I'd add a suggestion, but that part of the `.sln` file is unchanged._

gregkalapos (approved):
Awesome 🚀  This is really cool.

Just pinging @v1v: fyi we have some automation here that we could reuse later - some of this we already have, but some of them are new.

v1v:
Just to be sure, this particular change is not tested in the CI at the moment, so do you want to enable it with a follow up? Or do you prefer as part of this PR?

gregkalapos:
> Just to be sure, this particular change is not tested in the CI at the moment, so do you want to enable it with a follow up? Or do you prefer as part of this PR?

My thoughts on this: 
I think this is fine to merge as it is. This is part of https://github.com/elastic/apm-agent-dotnet/issues/1051, as a follow up I think it'd make sense to run `./build.sh agent-zip` in CI and make sure the generated `.zip` file ends up in jenkins artifacts. That way we 1) always build the zip file, so if we accidentally break something, likely the build will also fail, and 2) we could just take that file and publish it the way we want. There will be also another follow up to test the generated `.zip` file - and for that we'll also need to run `./build.sh agent-zip` in CI.

For the rest (like building NuGet packages, building the whole agent, etc.) I think there is no need to change those - for that I'd keep the current CI and just keep in mind that we have this and maybe reuse this later.

But this is @russcam's PR, so let's see what he says. :)

russcam:
Agree with @gregkalapos, I think it'll make sense to run `./build.sh agent-zip` in CI, but it doesn't need to be part of this PR. I'll open an issue to track running in CI. 

Some more effort and investigation needs to go into testing the zip, so will open a separate issue for that too; it's related to what to do when different versions of `System.Diagnostics.DiagnosticSource` are used in an application versus the agent, and how this relates to subscribing to listeners in the startup hooks.

> For the rest (like building NuGet packages, building the whole agent, etc.) I think there is no need to change those - for that I'd keep the current CI and just keep in mind that we have this and maybe reuse this later.

I don't think there's an immediate need to change what currently exists in CI. I'd be happy to have a chat about using this build automation in CI in the future though; I've found it particularly useful in the past to be able to run the same commands locally as are run on CI, which this build automation helps with, without hopefully being overtly complex.

build.bat
russcam: Thanks, will add

.gitignore
```diff
@@ -1,339 +1,343 @@
-## Ignore Visual Studio temporary files, build results, and
```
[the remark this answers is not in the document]
russcam: weird, I'll change it back

russcam:
Opened https://github.com/elastic/apm-agent-dotnet/issues/1074 for the flaky test.

Merging this in.

PR #1065, opened 2020-11-30
Introduce ServerUrl and deprecate ServerUrls

### LINKED ISSUE #1035: [META 359] Consolidate server_url/server_urls option
Rename `server_urls` to `server_url` (see https://github.com/elastic/apm/issues/359) in a backwards compatible way.

Adapt documentation to only describe the `server_url` option.

russcam:
This commit consolidates ServerUrl and ServerUrls options
by

- introducing ServerUrl configuration value
- deprecating ServerUrls with ObsoleteAttribute

ServerUrl uses the ServerUrl environment or setting key value,
falling back to ServerUrls[0] when not specified, allowing
the usage of ServerUrls to continue to work.

ServerUrls uses the ServerUrls environment or setting key value
if specified, falling back to ServerUrl environment or setting key value
when not specified, allowing exitsing usage of ServerUrls outside of the
agent API to work with ServerUrl.

Closes #1035

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -484,7 +484,7 @@ NOTE: This option requires APM Server 7.2 or later. It will have no effect on ol
 
 The URL for your APM Server. The URL must be fully qualified, including protocol (`http` or `https`) and port.
 
-IMPORTANT: Use of `ServerUrls` is deprecated. Use `ServerUrl`
+IMPORTANT: Use of `ServerUrls` is deprecated. Use `ServerUrl`.
 
 [float]
 [[config-secret-token]]
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -76,7 +76,7 @@ Here is a sample `appsettings.json` configuration file for a typical ASP.NET Cor
   "AllowedHosts": "*",
   "ElasticApm": <2>
     {
-      "ServerUrls":  "http://myapmserver:8200",
+      "ServerUrl":  "http://myapmserver:8200",
       "SecretToken":  "apm-server-secret-token",
       "TransactionSampleRate": 1.0
     }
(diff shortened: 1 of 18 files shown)


src/Elastic.Apm/Config/AbstractConfigurationReader.cs
```diff
@@ -211,7 +231,7 @@ private IReadOnlyList<Uri> ParseServerUrlsImpl(ConfigurationKeyValue kv)
 			{
 				_logger?.Warning()
 					?.Log(nameof(EnvVarNames.ServerUrls)
-						+ " configuration option contains more than one URL which is not supported by the agent yet"
```
gregkalapos, line 214: Should we print something like this additionally here?

> `ServerUrls` config is obsolete, use `ServerUrl` instead.

gregkalapos (approved):
👍  LGTM

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
russcam: ++ will add

docs/configuration.asciidoc
```diff
@@ -484,7 +484,7 @@ NOTE: This option requires APM Server 7.2 or later. It will have no effect on ol
 
 The URL for your APM Server. The URL must be fully qualified, including protocol (`http` or `https`) and port.
 
-NOTE: Providing multiple URLs is not supported by the agent yet. If multiple URLs are provided only the first one will be used.
+IMPORTANT: Use of `ServerUrls` is deprecated. Use `ServerUrl`
```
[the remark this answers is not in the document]
russcam: thanks, added

russcam:
Opened #1071 for the flaky test.

Merging this in.

russcam:
ah, I've jumped the gun a bit on merging this in - intended to add this to 1.8.0 as it's a change that should go in a minor rather than a patch version. I'll update the version label to v1.7.2 for now, and raise a discussion about branching strategies for next major and minor versions.

PR #1068, opened 2020-11-30
Add git info to benchmarks

gregkalapos:
This PR adds git commit sha and branch name to our benchmarks.

Mostly inspired by recent performance issues we had.

### The broader idea

Currently we only run benchmarks when PRs are merged to `master` and we report that to observability-benchmarks cluster. That helps us to find perf issues after merge (which is not ideal).

With this PR benchmark data will always have git info (branch and last commit SHA). Once merged we could turn on running benchmarks on every PR - we could still plot only `master` on the observability-benchmarks cluster but we'll have data for every CI run (including PR branches). With that we can compare the PRs benchmark value with the last master value and put the info into the PR, which I think will help us to avoid perf. issues later.


russcam (approved):
LGTM 👍

Issue #1077, opened 2020-12-02
ELASTIC_APM_ENABLED environment variable does not have the same behavior as "Enabled" setting in appsettings.json file


jpinedanu:
**Bug**
When using appsettings.json to configure Elastic APM, if the option "Enabled" is set as false, the initial fetch to read the configuration from APM Server is ignored and no connection is attempted. This is the expected behavior. However, when the code reaches any explicit call when using the Elastic APM agent a connection to the APM server will be attempted even though ElasticApm:Enabled is set as false. 
When setting the environment variable ELASTIC_APM_ENABLED as false, instead of using appsettings.json, all connections were ignored and no attempt was made to connect against the APM server. I believe disabling Elastic APM by using appsettings.json should have the same behavior as setting the  environment variable which is to disable all connections to the APM server

**To Reproduce**
1. Set Enabled to false in appsettings.json config
```
{
  "ElasticApm": {
    "Enabled": false
  }
}
```
2. Then call Elastic.Apm.Agent.Tracer.CaptureTransaction and see an attempt is made to reach the APM server. 
3. Remove "Enabled" configuration from appsettings.json file and set environment variable "ELASTIC_APM_ENABLED" to false.
4. Run the same code calling CaptureTransaction from the Elastic APM agent and view that no connection is attempted.

**Expected behavior**
Disabling Elastic APM by using appsettings.json to set ElasticApm:Enabled to false should have the same behavior as setting the ELASTIC_APM_ENABLED environment variable to false by having them both avoid reaching out to the APM server and disabling all behavior of the agent.

gregkalapos:
Thanks for reporting this. Indeed, this is a bug. I opened https://github.com/elastic/apm-agent-dotnet/pull/1078 to fix this.

gregkalapos:
Fixed in #1078.

PR #1078, opened 2020-12-02
Fix: ASP.NET Core, enabled=false in appsetting does not disable public Agent API

gregkalapos:
Solves https://github.com/elastic/apm-agent-dotnet/issues/1077.

### The bug

The problem was that in the `ApmMiddlewareExtension.cs` when the agent was disabled, we returned very early and we did not initialize the agent properly. This was fine as long as the public `Agent.Tracer.X` API wasn't used, once it was used, an implicit initialization happened with default values where `enabled=true`, so the agent still created real `Transaction`s and sent those transactions to the APM server.

### The fix

In the `ApmMiddlewareExtension.cs` we make sure `Agent.Setup(components)` is called with the current config. If that config contains `enabled=false`, the agent will always reuse that, so the public API will also return `NoopTransaction` and `NoopSpan`.


gregkalapos:
I'll need to do another round on this one tomorrow due to the failing test.

src/Elastic.Apm.AspNetCore/ApmMiddlewareExtension.cs
```diff
@@ -52,14 +52,13 @@ public static class ApmMiddlewareExtension
...
-				return builder;
-
 			var config = new AgentComponents(configurationReader: configReader, logger: logger);
 			HostBuilderExtensions.UpdateServiceInformation(config.Service);
 
+			// Agent.Setup must be called, even if agent is disabled. This way static public API usage won't implicitly initialize an agent with default values, instead, this will be reused.
 			Agent.Setup(config);
-			return UseElasticApm(builder, Agent.Instance, logger, subscribers);
+
+			return !configReader.Enabled ? builder : UseElasticApm(builder, Agent.Instance, logger, subscribers);
```
russcam, line 61: can this just call `return UseElasticApm(builder, Agent.Instance, logger, subscribers);` and let `UseElasticApm(...)` return without setting up middleware?

russcam (approved):
LGTM 👍  Left a small comment for consideration

src/Elastic.Apm.AspNetCore/ApmMiddlewareExtension.cs
gregkalapos: Yes 👍  Makes sense.

nikhil-lad-udio:
enabled=false is app settings does not disable agent API calling in django as well.

Issue #1034, opened 2020-11-12
[META 319] Add sanitize_field_name option to Kibana


AlexanderWert:
According to [this PR](https://github.com/elastic/apm/pull/334) the `sanitize_field_names` option needs to be added to [Kibana's configuration options](https://github.com/elastic/kibana/blob/master/x-pack/plugins/apm/common/agent_configuration/setting_definitions/general_settings.ts) for the .NET agent.

gregkalapos:
Turns out this setting is not yet live changeable. I opened https://github.com/elastic/apm-agent-dotnet/pull/1082 to change this. Once PR https://github.com/elastic/apm-agent-dotnet/pull/1082 is done, I'll also update the Kibana config.

gregkalapos:
Talked to @AlexanderWert, I opened the Kibana PR now, just to make sure this setting gets into Kibana also for .NET before `7.11` feature freeze.  https://github.com/elastic/apm-agent-dotnet/pull/1082 is ready to review and we still have enough time to make sure it's merged before `7.11` is out.

PR #1082, opened 2020-12-07
Make SanitizeFieldNames config live changeable through central config

gregkalapos:
Prerequisite for https://github.com/elastic/apm-agent-dotnet/issues/1034

`SanitizeFieldNames` prior to this PR was a static setting. The `PayloadSenderV2` passed its config which it got during startup to `HeaderDictionarySanitizerConverter` which then used the values from `SanitizeFieldNames` during serialization. This was not "central config ready" - that setting is not live changeable.

This PR removes the  `HeaderDictionarySanitizerConverter` and adds a filter (`HeaderDictionarySanitizerFilter`) which has the same functionality, except it reads configs from the `ConfigSnapshot` of the given transaction. The `ConfigSnapshot` on the given transaction is updated through central config - therefore changes from Kibana are reflected there.


src/Elastic.Apm/Filters/TransactionIgnoreUrlsFilter.cs
```diff
@@ -15,8 +16,20 @@ namespace Elastic.Apm.Filters
...
 		public TransactionIgnoreUrlsFilter(IConfigSnapshot configSnapshot) => _configSnapshot = configSnapshot;
-		public ITransaction Filter(ITransaction transaction) =>
-			WildcardMatcher.IsAnyMatch(_configSnapshot.TransactionIgnoreUrls, transaction?.Context?.Request?.Url?.PathName) ? null : transaction;
+
+		public ITransaction Filter(ITransaction transaction)
+		{
+			if (transaction is Transaction realTransaction)
+			{
+				// If there is no context, there is no URL either, therefore this transaction can't be filtered based on the URL.
+				if (!realTransaction.IsContextCreated)
```
gregkalapos, line 27: This isn't strictly related, but now that we register the same filters (including this one) in the mock payload senders in our tests as we do in `PayloadSenderV2` in prod, turns out that this filter created `transaction.Context` always, even if it wasn't needed.  ...so now we check for that.

src/Elastic.Apm/Api/Request.cs
```diff
@@ -22,7 +23,9 @@ public class Request
 		[Sanitize]
```
russcam, line 23: Does this also need to be sanitized by the filter too?

Looks like this is the only usage of `SanitizeAttribute` now, so can probably be removed, since it is not used for anything (used to be used by the contract resolver to determine converter to use).

src/Elastic.Apm/Filters/TransactionIgnoreUrlsFilter.cs
russcam, line 27: can combine nested `if`s?

test/Elastic.Apm.Tests/Mocks/MockPayloadSender.cs
```diff
@@ -19,16 +19,17 @@ namespace Elastic.Apm.Tests.Mocks
...
+		private readonly List<Func<ISpan, ISpan>> _spanFilters = new List<Func<ISpan, ISpan>>();
 		private readonly List<ISpan> _spans = new List<ISpan>();
+		private readonly List<Func<ITransaction, ITransaction>> _transactionFilters = new List<Func<ITransaction, ITransaction>>();
 		private readonly List<ITransaction> _transactions = new List<ITransaction>();
 
 		public MockPayloadSender(IApmLogger logger = null)
-			=> _spanFilters.Add(
-				new SpanStackTraceCapturingFilter(logger ?? new NoopLogger(), new MockApmServerInfo(new ElasticVersion(7, 10, 0, null))).Filter);
+			=> PayloadSenderV2.SetUpFilters(_transactionFilters, _spanFilters, new MockConfigSnapshot(),
+				new MockApmServerInfo(new ElasticVersion(7, 10, 0, null)), logger ?? new NoopLogger());
```
russcam, line 32:
```suggestion
				MockApmServerInfo.Version710, logger ?? new NoopLogger());
```

test/Elastic.Apm.Tests/Mocks/SerializerMockPayloadSender.cs
```diff
@@ -22,36 +24,28 @@ namespace Elastic.Apm.Tests.Mocks
...
+		public SerializerMockPayloadSender(IConfigurationReader configurationReader)
 		{
-			get
-			{
-				_transactionTaskCompletionSource.Task.Wait();
-				return Transactions.First();
-			}
+			PayloadSenderV2.SetUpFilters(_transactionFilters, _spanFilters,
+				new ConfigSnapshotFromReader(configurationReader, nameof(SerializerMockPayloadSender)),
+				new MockApmServerInfo(new ElasticVersion(7, 10, 0, null)), new NoopLogger());
```
russcam, line 38:
```suggestion
				MockApmServerInfo.Version710, new NoopLogger());
```

russcam (changes requested):
I've left some comments for consideration

src/Elastic.Apm/Filters/TransactionIgnoreUrlsFilter.cs
gregkalapos: In https://github.com/elastic/apm-agent-dotnet/pull/1087 I added more to the outer if block: https://github.com/elastic/apm-agent-dotnet/pull/1087/files#diff-78983cac62ab3ac9a4a1ad0e2e32e6f04cb7b744979b4c8f9e3106b0fa5ef262R25 

Otherwise I'd agree :)

src/Elastic.Apm/Api/Request.cs
gregkalapos: Good point!

> Does this also need to be sanitized by the filter too?

No - body sanitization happens when we read the request body in `HttpRequestExtensions.ExtractRequestBody`. And that only happens if the body is form content type. Originally we did body sanitization during serialization, but we changed to do it during body capturing after some iterations on the problem.

Reason is that once the body is captured, we'd need to parse it again and do the serialization which is super hard (with form encoded it's easy, but if it's not form encoded we can ran into all sorts of problems), plus it's very high overhead.

> Looks like this is the only usage of SanitizeAttribute now, so can probably be removed, since it is not used for anything (used to be used by the contract resolver to determine converter to use).

I agree, I ended up just removing it. So in sum: for body sanitization nothing really changes in this PR, except that we remove this attribute, but the way we do sanitization for the body is the same as before.

gregkalapos:
@russcam this is ready for another round of review.

russcam (approved):
changes LGTM 👍

PR #1090, opened 2020-12-15
only package projects explicitly marked <IsPackable>true</IsPackable>

russcam:
This commit only packages projects
explicitly marked as <IsPackable>true</IsPackable>.
Each project in ./src is explicltly marked, and projects
in ./test and ./sample inherit <IsPackable>false</IsPackable>
from Directory.Build.props.

This allows build[.bat/.sh] pack to package just
the releasable artifacts.


gregkalapos (approved):
LGTM

PR #1091, opened 2020-12-15
Only add AspNetCoreErrorDiagnosticsSubscriber if not present

russcam:
This commit only adds the `AspNetCoreErrorDiagnosticsSubscriber` if an instance is not already present in the subscribers.

### DIFF
--- a/src/Elastic.Apm.AspNetCore/ApmMiddlewareExtension.cs
+++ b/src/Elastic.Apm.AspNetCore/ApmMiddlewareExtension.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using Elastic.Apm.AspNetCore.DiagnosticListener;
 using Elastic.Apm.Config;
 using Elastic.Apm.DiagnosticSource;
@@ -72,10 +73,11 @@ namespace Elastic.Apm.AspNetCore
 			if (!agent.ConfigurationReader.Enabled)
 				return builder;
 
-			var subs = new List<IDiagnosticsSubscriber>(subscribers ?? Array.Empty<IDiagnosticsSubscriber>())
-			{
-				new AspNetCoreErrorDiagnosticsSubscriber()
-			};
+			var subs = subscribers?.ToList() ?? new List<IDiagnosticsSubscriber>(1);
+
+			if (subs.Count == 0 || subs.All(s => s.GetType() != typeof(AspNetCoreErrorDiagnosticsSubscriber)))
+				subs.Add(new AspNetCoreErrorDiagnosticsSubscriber());
+
 			agent.Subscribe(subs.ToArray());
 			return builder.UseMiddleware<ApmMiddleware>(agent.Tracer, agent);
 		}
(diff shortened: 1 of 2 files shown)


gregkalapos (approved):
LGTM.

Maybe we could also add a test to cover this.

russcam:
> Maybe we could also add a test to cover this.

I think it's going to be somewhat difficult to test this, because the test seams are concrete types and static methods i.e. `ApmAgent` and `ApmAgentExtensions.Subscribe()`:

https://github.com/elastic/apm-agent-dotnet/blob/c01860c78a56ad72b15766d167f0407d44626a49/src/Elastic.Apm.AspNetCore/ApmMiddlewareExtension.cs#L66-L83

This could be refactored to use the `IApmAgent` interface and move `Subscribe` to a method on `IApmAgent`, but this would be a bit of work. Is there some other obvious way to test that I'm missing?

gregkalapos:
> Is there some other obvious way to test that I'm missing?

In e.g. `AspNetCoreBasicTests.cs` we could do something like this:

```
[Fact]
public void RegisterErrorListenerTwice()
{
	var builder = _factory
		.WithWebHostBuilder(n =>
		{
			n.Configure(app =>
			{
			        //this'd end up in the method that you change in this PR... so it'd trigger the code we'd like to test
				app.UseElasticApm(_agent, _agent.Logger, new AspNetCoreErrorDiagnosticsSubscriber());
			});
		});
	builder.CreateClient();
	// And then here assert that _agent.Disposables only contains a single AspNetCoreErrorDiagnosticsSubscriber
}
```

It'd be a bit of a work and additional code to iterate through `_agent.Disposables` and check on the types included there, but that should be doable.

But this made me think a bit: Is there any case where it makes sense to subscribe to any of our subscribers more than once? 🤔  I think we could generally say: A subscriber should only subscribe once. Because otherwise we'd create events multiple times (in case of `AspNetCoreErrorDiagnosticsSubscriber` we'd capture multiple errors, in case of `HttpDiagnosticsSubscriber` we'd create multiple http spans, etc.). 

So I wonder if it'd be better to have a general check - maybe in `DiagnosticInitializer.cs` or in `ApmAgentExtensions.Subscribe` - and make sure whatever `IDiagnosticsSubscriber` is passed into the agent, it'll only subscribe 1 time for each type?

_The last part about an overall check for each type is just a general thought and it's not really a problem that was ever reported, so if it's offtopic for this PR, then let's just ignore it._

russcam:
Added a unit test to assert that only one AspNetCoreErrorDiagnosticListener is registered.

> So I wonder if it'd be better to have a general check - maybe in DiagnosticInitializer.cs or in ApmAgentExtensions.Subscribe - and make sure whatever IDiagnosticsSubscriber is passed into the agent, it'll only subscribe 1 time for each type?

I can see some value in doing this for our subscribers. Perhaps we can open a separate issue to track?

gregkalapos:
> Added a unit test to assert that only one AspNetCoreErrorDiagnosticListener is registered.

Nice, looks good to me 👍 

> 
> > So I wonder if it'd be better to have a general check - maybe in DiagnosticInitializer.cs or in ApmAgentExtensions.Subscribe - and make sure whatever IDiagnosticsSubscriber is passed into the agent, it'll only subscribe 1 time for each type?
> 
> I can see some value in doing this for our subscribers. Perhaps we can open a separate issue to track?

Opened https://github.com/elastic/apm-agent-dotnet/issues/1119. 

I think this is ok to merge.

PR #1092, opened 2020-12-15
Remove IConfigSnapshot parameter from CollectRequestBody

gregkalapos:
Just a nit from when I looked at the code during [this discussion](https://github.com/elastic/apm-agent-dotnet/pull/1082#discussion_r543376989).

We had this method:

```
internal static void CollectRequestBody(this Transaction transaction, bool isForError, HttpRequest httpRequest, IApmLogger logger, IConfigSnapshot configSnapshot)
```

The 1. parameter is a `Transaction`, the last one is an `IConfigSnapshot`. The last parameter is unnecessary, we can just use the `Transaction.ConfigSnapshot` property.

This PR removed the `IConfigSnapshot` parameter and adapts caller code.


russcam (approved):
LGTM 👍

PR #1093, opened 2020-12-16
Add diff build target

### LINKED ISSUE #1079: Diff built assemblies in PRs and Release against latest nuget packages
Relates: #1067

In striving to follow semantic versioning principles, we want to avoid introducing binary breaking changes within releases of a major version of the APM agent.

### Possible implementation

To aid with this, there is [assembly-differ](https://github.com/nullean/assembly-differ), a dotnet tool that can diff assemblies from file, directories, nuget packages and git repositories. This tooling has been successfully used with the Elasticsearch .NET client in CI for over a year (https://github.com/elastic/elasticsearch-net/pull/4223), and used manually for much longer. It has proven its value in catching changes before releasing.

[A target can be added to the scripts project](https://github.com/elastic/apm-agent-dotnet/blob/7a3896c6598d9bf959eccc43bad9a7fb64a1607e/build/scripts/Targets.fs#L73-L106) that performs a diff on PRs and on releasing, to notify of breaking changes to the public API, optionally failing if a breaking change it introduced.

russcam:
This commit adds the ability to diff built assemblies against
the previously released version on Nuget. If there are changes,
an XML file containing the diffs is emitted in ./build/output
with a name matching the assembly name.

To run the diff command for all assemblies

```
build.bat diff
```

This will use the netstandard2.0 TFM by default, but a TFM
can be supplied with `--framework`.

To run for only certain assemblies

```
build.bat diff --packageids Elastic.Apm --packageids Elastic.Apm.SqlClient
```

Closes #1079

### DIFF
--- a/ElasticApmAgent.sln
+++ b/ElasticApmAgent.sln
@@ -50,6 +50,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "files", "files", "{B406113B
 		RELEASING.md = RELEASING.md
 		build.bat = build.bat
 		build.sh = build.sh
+		dotnet-tools.json = dotnet-tools.json
 	EndProjectSection
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Elastic.Apm.PerfTests", "test\Elastic.Apm.PerfTests\Elastic.Apm.PerfTests.csproj", "{F069CE99-F418-4BC2-9E44-8F03497D8DA8}"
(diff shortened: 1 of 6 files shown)


gregkalapos:
Very cool!

Can we maybe add a readme about this? ...just copy the PR description, it's already good. It can be under `/build/scripts/`, or whatever place feels good for this. 

https://github.com/elastic/apm-agent-dotnet/pull/1062#issue-527033078 also had useful info in the PR description - I think it'd be useful to also move that info into the same readme file.

gregkalapos (approved):
👍 LGTM

PR #1095, opened 2020-12-17
Add private readonly IConfiguration variable

insukcho:
Add below private variable definition line in Startup.cs source snippet.

> private readonly IConfiguration Configuration;

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -33,7 +33,10 @@ using Elastic.Apm.AspNetCore;
 
 public class Startup
 {
-    public Startup(IConfiguration configuration)
+
+    private readonly IConfiguration Configuration;
+    
+	public Startup(IConfiguration configuration)
     {
         Configuration = configuration;
     }
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -34,11 +34,11 @@ using Elastic.Apm.AspNetCore;
 public class Startup
 {
 
-    private readonly IConfiguration Configuration;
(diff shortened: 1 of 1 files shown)


docs/configuration.asciidoc
```diff
@@ -33,7 +33,10 @@ using Elastic.Apm.AspNetCore;
 
 public class Startup
 {
-    public Startup(IConfiguration configuration)
+
+    private readonly IConfiguration Configuration;
```
russcam, line 37: we should rename this from `Configuration` to `_configuration`, in line with private field naming conventions. Line 41 would also need to change to `_configuration = configuration;`

russcam (changes requested):
Thanks for the submission, @insukcho - I left a comment about naming conventions.

docs/configuration.asciidoc
insukcho: agree. I made a new commit.

russcam (approved):
LGTM 👍

docs/configuration.asciidoc
```diff
@@ -33,9 +33,12 @@ using Elastic.Apm.AspNetCore;
 
 public class Startup
 {
-    public Startup(IConfiguration configuration)
+
+    private readonly IConfiguration _configuration;
+    
+	public Startup(IConfiguration configuration)
```
russcam, line 39:
```suggestion
    public Startup(IConfiguration configuration)
```

russcam: minor nitpick: spacing

russcam (approved):

PR #1097, opened 2020-12-18
Use ConfigureAwait(false)

russcam:
This commit uses `ConfigureAwait(false)` in Elastic.Apm to not capture the context when awaiting async Tasks.

The exception is `AgentTimer`, which needs some further investigation. Marking awaited tasks as `.ConfigureAwait(false)` in `AgentTimer` breaks unit tests. My suspicion is that the tests assume a captured context somewhere.


russcam:
run tests

gregkalapos (approved):
LGTM

PR #1102, opened 2020-12-21
Check port is available when creating LocalServer

russcam:
This commit adds static methods to create a LocalServer,
checking that a HttpListener can be started to listen on a
specific port as part of creation, and optionally looping through
ports until the listener can be started.


gregkalapos (approved):
LGTM

PR #1104, opened 2020-12-23
Add Microsoft.NET.Sdk.Razor package reference for ASP.NET Core 2.2

russcam:
This commit adds a reference to Microsoft.NET.Sdk.Razor package reference for ASP.NET Core 2.2. This is to fix a known bug in the .NET 5.0 SDK where the build target RazorTagHelper fails when DOTNET_HOST_PATH is not set.


russcam:
Merging this in - test failures are not related to this change

PR #1106, opened 2020-12-30
Run test assemblies in parallel

russcam:
This PR 

1. separates out the common test components from Elastic.Apm.Tests into a new assembly, Elastic.Apm.Tests.Utilities. This allows assemblies containing tests to be run in parallel, which is not possible when test assemblies reference another test assembly that may potentially be running at the same time.
2. adds coverlet and JunitXml.TestLogger packages to all test projects using Directory.Build.Props. This removes the need to add them in CI scripts or to each project individually.
3. updates linux CI scripts to run tests in the solution file. In conjunction with the xunit configuration, this allows test assemblies to run in parallel.
4. Renames Elastic.Apm.DockerTests and Elastic.ApmPerfTests to Elastic.Apm.Docker.Tests and Elastic.Apm.Perf.Tests, respectively, to align the test assembly naming convention.
5. Excludes all test assemblies from being included in code coverage output


russcam:
jenkins run tests

russcam:
run tests

russcam:
This is now ready for review @gregkalapos. 

The benchmark test fails currently as the the host does not have the net50 SDK installed, and the log switchable test looks like a timing condition, which I'll see if I can address.

.ci/linux/benchmark.sh
```diff
@@ -4,6 +4,6 @@
 #
 set -euxo pipefail
 
-cd ./test/Elastic.Apm.PerfTests
+cd ./test/Elastic.Apm.Perf.Tests
```
gregkalapos, line 7: Since we rename this: I'd vote for calling it `Elastic.Apm.Benchmarks`.

russcam: Agreed

gregkalapos (approved):
LGTM

PR #1109, opened 2021-01-04
Fix nullref in MetricsCollector with custom IConfigurationReader

gregkalapos:
From [discuss](https://discuss.elastic.co/t/elastic-apm-object-reference-not-set-to-an-instance-of-an-object).

With a custom `IConfigurationReader` where the `IConfigurationReader.DisabledMetrics` property returns `null`, the `MetricsCollector` `.ctor` throws an exception.

This PR fixes that.

### DIFF
--- a/src/Elastic.Apm/Helpers/WildcardMatcher.cs
+++ b/src/Elastic.Apm/Helpers/WildcardMatcher.cs
@@ -68,6 +68,7 @@ namespace Elastic.Apm.Helpers
 			{
 				if (!matcher.StartsWith(Wildcard) && !matcher.EndsWith(Wildcard))
 					return new VerbatimMatcher(split[0], ignoreCase);
+
 				return new SimpleWildcardMatcher(split[0], matcher.StartsWith(Wildcard), matcher.EndsWith(Wildcard), ignoreCase);
 			}
 
@@ -110,7 +111,9 @@ namespace Elastic.Apm.Helpers
 		/// <returns>The first matching <see cref="WildcardMatcher" />, or <code>null</code> if none match.</returns>
 		internal static WildcardMatcher AnyMatch(IReadOnlyCollection<WildcardMatcher> matchers, string firstPart, string secondPart)
 		{
-			for (var i = 0; i < matchers?.Count; i++)
+			if (matchers is null) return null;
+
+			for (var i = 0; i < matchers.Count; i++)
 			{
 				if (matchers.ElementAt(i).Matches(firstPart, secondPart))
 					return matchers.ElementAt(i);
--- a/src/Elastic.Apm/Helpers/WildcardMatcher.cs
+++ b/src/Elastic.Apm/Helpers/WildcardMatcher.cs
@@ -110,7 +110,7 @@ namespace Elastic.Apm.Helpers
 		/// <returns>The first matching <see cref="WildcardMatcher" />, or <code>null</code> if none match.</returns>
 		internal static WildcardMatcher AnyMatch(IReadOnlyCollection<WildcardMatcher> matchers, string firstPart, string secondPart)
 		{
-			for (var i = 0; i < matchers.Count; i++)
+			for (var i = 0; i < matchers?.Count; i++)
 			{
 				if (matchers.ElementAt(i).Matches(firstPart, secondPart))
 					return matchers.ElementAt(i);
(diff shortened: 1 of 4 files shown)


src/Elastic.Apm/Metrics/MetricsCollector.cs
```diff
@@ -46,73 +46,83 @@ internal class MetricsCollector : IDisposable, IMetricsCollector
 
 		public MetricsCollector(IApmLogger logger, IPayloadSender payloadSender, IConfigSnapshotProvider configSnapshotProvider)
 		{
-			_logger = logger.Scoped(nameof(MetricsCollector));
-			_payloadSender = payloadSender;
-			_configSnapshotProvider = configSnapshotProvider;
+			try
```
gregkalapos, line 49: This `try-catch` is up for discussion. The real fix is the null-check in `WildcardMatcher.cs ` and that should be enough.

Having said that, this `.ctor` runs in `Agent.Setup` which is a fully unprotected code so this can crash the user application (that's what happened [here](https://discuss.elastic.co/t/elastic-apm-object-reference-not-set-to-an-instance-of-an-object)). I feel it's better to never throw an exception here, if something happens, we just log, the agent won't collect metrics, but we let the application run and not crash it. That's a better behaviour for an APM Agent than crashing the app. And the `try-catch` makes sure that's really the case.

src/Elastic.Apm/Helpers/WildcardMatcher.cs
```diff
@@ -110,7 +110,7 @@ public static WildcardMatcher ValueOf(string wildcardString)
 		/// <returns>The first matching <see cref="WildcardMatcher" />, or <code>null</code> if none match.</returns>
 		internal static WildcardMatcher AnyMatch(IReadOnlyCollection<WildcardMatcher> matchers, string firstPart, string secondPart)
 		{
-			for (var i = 0; i < matchers.Count; i++)
+			for (var i = 0; i < matchers?.Count; i++)
```
russcam, line 113: `if (matchers is null) return null;` on the first line might be easier to read. 

Had to have a look at the IL to see what this results in, and it instantiates an `int?` and checks if it has a value initialized from a non-null matcher count on each loop, so the initial null check looks like it'd be better.

src/Elastic.Apm/Metrics/MetricsCollector.cs
russcam: It feels like a bug if the metrics collector initialization fails but the user has opted to collect metrics. A couple of questions come to mind: 

1. has there been a reported issue with metrics collector initialization?
2. is there a concerning component of initialization that warrants a try/catch?

If a user did happen to have an issue with metric collector initialization that threw an exception, hopefully it'd be reported to be fixed, and in the meantime, the user can disable metric collection through configuration, so wouldn't be blocked from using the agent whilst waiting for a fix. Conversely, if metric collection initialization failure is handled in a try/catch and the rest of the agent is allowed to proceed, the lack of metrics and reason why may be less obvious to the user, and may be a higher barrier to reporting it.

russcam (changes requested):
I left a couple of comments for consideration

src/Elastic.Apm/Helpers/WildcardMatcher.cs
gregkalapos: Ok - changed this to the suggested null check.

src/Elastic.Apm/Metrics/MetricsCollector.cs
gregkalapos: I think this is a fundamental discussion where in my opinion APM is drastically different from other type of software. I think that an APM Agent should in general not make the monitored application crash. So with that in mind I think it's preferable to have some a "buggy behaviour" in APM compared to crashing the app - again, I totally agree this is not true in other type of software, but I argue that APM is different in this sense. In this specific scenario: for a production system with the agent it's still better to have an agent that does not collect metrics than crashing it and make it unusable for every user.

To the questions:

> 1. has there been a reported issue with metrics collector initialization?

Yes - that's exactly the linked [discuss](https://discuss.elastic.co/t/elastic-apm-object-reference-not-set-to-an-instance-of-an-object) entry which triggered me to open this PR.

> 2. is there a concerning component of initialization that warrants a try/catch?

The problem here is that we don't know this upfront. I remember looking at one PR where we changed something in the initialization and I was thinking: "Is there anything that could throw here?" I concluded "no", but as it turned out I was wrong. So now that `WildcardMatcher` can deal with `null` likely there isn't, but maybe we overlooked something.

gregkalapos: As discussed over zoom, I removed the try-catch for now to unblock the PR.

russcam (approved):
LGTM 👍

PR #1112, opened 2021-01-06
Update Specification attributes for MockApmServer

russcam:
This commit updates the paths used in Specification attributes for DTOs used by MockApmServer.


russcam:
Merging this in as IIS tests will fail without it

PR #1122, opened 2021-01-13
Expose creating agent components for ASP.NET Full Framework

### LINKED ISSUE #1045: Elastic.Apm.Agent.AddFilter doesn't work in Global.asax's Application_Start()
**Describe the bug**
In my MVC project, I wanted to add some global filters. To do that, I've added my code in the `Global.asax` `Application_Start()` method. Here is my code:

```
protected void Application_Start()
{
  //my other code
  Utils.ElasticUtilities.AddAPMGlobalFilter();
}
```

```
public class ElasticUtilities
    {
        public static void AddAPMGlobalFilter()
        {
            AgentDependencies.Logger = new ApmLoggerBridge();

            try
            {
                bool filterAdded = Elastic.Apm.Agent.AddFilter((Elastic.Apm.Api.ITransaction transaction) =>
                {
                    try
                    {
                        LogMessage("APM Transaction filter" + transaction.Id);
                        transaction.Labels.Add("UserID", MyCustomCode.UserID.ToString());
                        transaction.Labels.Add("EmailAddress", MyCustomCode.EmailAddress);
                    }
                    catch (Exception exc)
                    {
                        transaction.CaptureException(exc);
                        LogException(exc);
                    }
                    return transaction;
                });
                LogMessage("Filter added:" + filterAdded.ToString());
            }
            catch (Exception ex)
            {
                LogException(ex);
            }
        }

        internal class ApmLoggerBridge : IApmLogger
        {
            public bool IsEnabled(LogLevel level)
            {
                return true;
            }

            public void Log<TState>(LogLevel level, TState state, Exception e, Func<TState, Exception, string> formatter)
            {
                string message = formatter(state, e);
                if (e == null)
                {
                    LogMessage(message);
                }
                else
                {
                    LogException(e);
                }
            }
        }

    }
```

But transactions aren't sent to the APM... The only exceptions that I received are:
```
Exception type:
System.UnauthorizedAccessException

Message:
Access to the registry key 'Global' is denied.
```

and

```
Source:
Microsoft.Diagnostics.Tracing.TraceEvent

Exception type:
System.UnauthorizedAccessException

Message:
Error Starting ETW:  Access Denied (Administrator rights required to start ETW)
```

The configuration is valid because the code above everything is logged to the APM correctly.
If I move my code from `Application_Start()` to `Session_Start()` it works, but I don't want to add a filter for each session.

Thank you!

### LINKED ISSUE #1098: AspNetFullFramework - Implicit initialization inconsistent IConfigurationReader
When using AspNetFullFramework and accessing Agent before it has been initialized via a request, (e.g. by accessing Agent.AddFilter()), the Agent will be initialized with a EnvironmentConfigurationReader instead of a FullFrameworkConfigReader.
This means that accessing Agent.something before it has been initialized, initializes the Agent incorrectly.
It is also not possible to manually call Agent.Setup and have it use the FullFrameworkConfigReader.

I would expect the same result when calling Agent.CaptureTransaction before the first request and after the first request

Our current solution is to always check Agent.IsConfigured before doing anything, but that means we are dropping som manually captured transactions, and not able to set up our filters beforehand.
A secondary solution would be to duplicate our config values, having the same values in both web.config and Envirionemt, so both configReaders read the same values.

I would like if accessing Agent before it has been setup gives the same result as after, or giving a way of manually setting it up to read from web.config, without having to write a custom ConfigReader, that does the same as FullFrameworkConfigReader

russcam:
This commit exposes a CreateAgentComponents method on
ElasticApmModule that allows a user to create agent components
for ASP.NET Full Framework and set up the APM agent, before
the ElasticApmModule Init() runs. This is useful in scenarios
where a user wishes to capture transactions/spans or setup filters at application
start.

The CreateAgentComponents returns a derived FullFrameworkAgentComponents
type that SafeAgentSetup method uses to determine whether an already initialized
agent is likely to lead to unexpected behaviour e.g. agent component implementations
not designed to work with ASP.NET Full Framework.

An example usage of CreateAgentComponents is

    public class MvcApplication : HttpApplication
    {
        protected void Application_Start()
        {
            // other application startup e.g. RouteConfig, etc.

            // configure agent filter on startup
            var agentComponents = ElasticApmModule.CreateAgentComponents();
            Agent.Setup(agentComponents);
            Agent.AddFilter((ITransaction t) =>
            {
                t.SetLabel("foo", "bar");
                return t;
            });
        }
    }

Fixes #1098
Fixes #1045

### DIFF
--- a/src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
+++ b/src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
@@ -409,8 +409,8 @@ namespace Elastic.Apm.AspNetFullFramework
 				{
 					BuildLogger()
 						.Scoped(dbgInstanceName)
-						.Error()
-						?.LogException(ex, "The Elastic APM agent was already initialized before call to"
+						.Info()
+						?.Log("The Elastic APM agent was already initialized before call to"
 							+ $" {nameof(ElasticApmModule)}.{nameof(Init)} - {nameof(ElasticApmModule)} will use existing instance");
 				}
 				else
--- a/src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
+++ b/src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
@@ -6,7 +6,6 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.Linq;
-using System.Reflection;
 using System.Security.Claims;
 using System.Web;
 using Elastic.Apm.Api;
@@ -19,6 +18,9 @@ using Elastic.Apm.Model;
 
 namespace Elastic.Apm.AspNetFullFramework
 {
+	/// <summary>
+	/// Captures each request in an APM transaction
+	/// </summary>
 	public class ElasticApmModule : IHttpModule
 	{
 		private static bool _isCaptureHeadersEnabled;
@@ -29,9 +31,13 @@ namespace Elastic.Apm.AspNetFullFramework
 		private HttpApplication _application;
 		private IApmLogger _logger;
 
-		// ReSharper disable once ImpureMethodCallOnReadonlyValueField
-		public ElasticApmModule() => _dbgInstanceName = DbgInstanceNameGenerator.Generate($"{nameof(ElasticApmModule)}.#");
+		public ElasticApmModule() =>
+			// ReSharper disable once ImpureMethodCallOnReadonlyValueField
+			_dbgInstanceName = DbgInstanceNameGenerator.Generate($"{nameof(ElasticApmModule)}.#");
 
+		public void Dispose() => _application = null;
+
+		/// <inheritdoc />
 		public void Init(HttpApplication application)
 		{
 			try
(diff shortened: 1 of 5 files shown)


src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
```diff
@@ -455,12 +405,24 @@ private static void SafeAgentSetup(string dbgInstanceName)
...
+						.Scoped(dbgInstanceName)
+						.Error()
+						?.LogException(ex, "The Elastic APM agent was already initialized before call to"
+							+ $" {nameof(ElasticApmModule)}.{nameof(Init)} - {nameof(ElasticApmModule)} will use existing instance");
+				}
+				else
+				{
+					BuildLogger()
+						.Scoped(dbgInstanceName)
+						.Error()
```
gregkalapos, line 420: As discussed over zoom, let's not consider this as an error - an info log would be better.

gregkalapos (approved):
Nice, LGTM. 

As discussed on zoom, a bit of a documentation would be helpful to this so people can use it easier.

Issue #1136, opened 2021-01-15
Implicit agent initialization + Serilog log correlation does not work


gregkalapos:
Repro:

```
static void Main(string[] args)
{
    Log.Logger = new LoggerConfiguration()
      .Enrich.WithElasticApmCorrelationInfo()
      .WriteTo.Console(new EcsTextFormatter()).CreateLogger();
    
    
    Agent.Tracer.CaptureTransaction("foo", "bar", () =>
    {
        // This log line won't have `trace_id` field - therefore log correlation won't work either
        Log.Information("Bla");
    });
}
```

Workaround: Call `Agent.Setup(new AgentComponents());`. 

But that should not be needed - the 1. call to the agent API should initialize the agent, which does not seem to happen.

gregkalapos:
[This check](https://github.com/elastic/ecs-dotnet/blob/master/src/Elastic.Apm.SerilogEnricher/ElasticApmEnricher.cs#L18) makes it `return` and skip adding the traceid because we don't set the `Agent.IsConfigured` flag, which we should. 

PR opened.

gregkalapos:
Fixed in https://github.com/elastic/apm-agent-dotnet/pull/1137

PR #1137, opened 2021-01-15
Set _isConfigured flag also with implicit agent initialization

gregkalapos:
Solves https://github.com/elastic/apm-agent-dotnet/issues/1136.

The bug was that if someone only uses the public API then the initialized itself as designed, except the `Agent.IsConfigured` flag remained `false`.

I also added 1 test which tests `Agent.IsConfigured`. 

We already had 1 test for that, and with this we now have 2 tests that test a static property - nothing else depends on anything static. I moved the two static tests into their own projects, otherwise we'd have a problem that the static instance will be shared between the tests.


russcam (approved):
LGTM 👍

Issue #1073, opened 2020-12-01
Test ElasticApmAgent_<Version>.zip on CI


russcam:
Relates: #1051, #1072

Once the ElasticApmAgent_<Version>.zip build is running on CI (#1072), the zip file should be tested to ensure that the generated artifact can auto instrument an ASP.NET Core web application without code changes.

### Possible implementation

1. A barebones sample ASP.NET Core application is introduced, as the target application for startup hood testing
2. A build target is added to the scripts F# project with a dependency on the `agent-zip` target that
    1. unzips the  ElasticApmAgent_<Version>.zip to a known location
    2. uses docker to run the sample ASP.NET Core with the `DOTNET_STARTUP_HOOKS` environment variable pointing to the unzipped ElasticApmAgent_<Version> location
3. Collects agent logs from the sample application, or indexed data from Elasticsearch to verify that auto instrumentation is operating correctly.

### Investigation

From current investigation, the major version of `System.Diagnostics.DiagnosticSource` that the agent references must match the version that the web application uses. 

The agent subscribes to diagnostic events using `DiagnosticListener.AllListeners.Subscribe(...)`. If the ASP.NET Core application is using `System.Diagnostics.DiagnosticSource` 4.x, but the APM agent is using `System.Diagnostics.DiagnosticSource` 5.0.0, the agent will not be able to load the 5.0.0 version of the assembly into `AssemblyLoadContext.Default` - a `FileLoadException` is thrown.

An approach to load assemblies with conflicting versions is to create a different `AssembyLoadContext` and load them into this context. Using a different `AssemblyLoadContext` to load a 5.0.0 version and a 4.x version of `System.Diagnostics.DiagnosticSource` works in getting around the version conflict, but since the agent depends on the 5.0.0 version, it will try to subscribe to the `DiagnosticListener.AllListeners` property of the 5.0.0 version, not the 4.x version that the application is publishing events to. Further, the static `DiagnosticListener.AllListeners` property is isolated to the context in which it is loaded, so the agent assemblies that subscribe need to be loaded into the same context and be binary compatible, which doesn't seem to be the case if they're compiled against a different version. One way to get around this issue is to reference `System.Diagnostics.DiagnosticSource` 5.0.0 as a dependency of the application, which works, but we might want to take an alternative approach that doesn't potentially require a user to make changes to their application.

PR #1138, opened 2021-01-16
Allow ElasticApmStartupHook to work with System.Diagnostics.DiagnosticSource 4 and 5

### LINKED ISSUE #1072: Build ElasticApmAgent_<Version>.zip on CI
#1062 introduced build automation to build a ElasticApmAgent_<Version>.zip file that uses startup hooks to auto instrument an application without code changes. The command for building the zip file is 

```sh
./build.sh agent-zip
```

which outputs a zip file at `/build/output/ElasticApmAgent_<Version>.zip`, using the version encoded in `/src/Directory.Build.Props`.

The agent-zip command should be run in CI, capturing the output zip file as a build artifact.

### LINKED ISSUE #1134: Loading agent with DOTNET_STARTUP_HOOK on .NET 5
The DiagnosticSource dll can't be loaded - this is a version incompatibility - we try to load version 4.x on every platform, but on .NET 5 we should try loading version 5 of the `.dll`.

```
Description: The process was terminated due to an unhandled exception.
Exception Info: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.IO.FileNotFoundException: Could not load file or assembly 'System.Diagnostics.DiagnosticSource, Version=4.0.5.0, Culture=neutral,
```

russcam:
This commit updates the startup hook implementation to allow it to work with applications that are compiled against either
`System.Diagnostics.DiagnosticSource` 4 or 5.

Introduce a `UseDiagnosticSourceFour` MSBuild property to Elastic.Apm to conditionally reference `System.Diagnostics.DiagnosticSource` 4.x.x or 5.x.x. The build agent-zip task uses the property to compile Elastic.Apm against both System.Diagnostics.DiagnosticSource 4.x.x and 5.x.x, and package both into the startup hook zip file.

Split the startup hook agent assembly and dependency loading into its own assembly, `Elastic.Apm.StartupHook.Loader`, compiled against Elastic.Apm, allowing different versions to be compiled for different referenced `System.Diagnostics.DiagnosticSource` versions. `ElasticApmStartupHook` now determines if a version of `System.Diagnostics.DiagnosticSource` is already loaded, using this to determine which version of `Elastic.Apm.StartupHook.Loader` to load and invoke.

Update the build project `pack` target to build nuget packages **and** the zip file for startup hooks, and update CI linux scripts to use the build project `pack` target. This change requires the CI scripts to look in one directory for nuget packages, and allows artifacting the zip file.

Closes #1072
Closes #1134


russcam:
run tests

russcam:
run tests

russcam:
The build target `agent-docker` builds a docker image containing all of the assemblies to auto instrument using startup hooks. This can be tested by

1. Building the docker image with

   ```
   ./build.bat agent-docker
   ```

   _(use `./build.sh` on macOS or Linux)_

2. Build a docker image that uses the built docker image. An example Dockerfile

   ```
   FROM observability/apm-agent-dotnet:1.7.1 AS agent

   FROM mcr.microsoft.com/dotnet/core/sdk:3.1 AS build-env
   WORKDIR /sampleapp
   
   RUN dotnet new mvc
   RUN dotnet publish -c Release -o out
   
   # Build runtime image
   FROM mcr.microsoft.com/dotnet/core/aspnet:3.1
   WORKDIR /app
   ENV DOTNET_STARTUP_HOOKS=/usr/agent/ElasticApmAgentStartupHook.dll
   COPY --from=agent /usr/agent /usr/agent
   COPY --from=build-env /sampleapp/out .
   ENTRYPOINT ["dotnet", "sampleapp.dll"]
   ```
 
   To build the image. In the directory container the above Dockerfile

   ```
   docker build --tag startuphook-sample .
   ```

3. Run the startuphook-sample image

   ```
   docker run --publish 5000:80 -e ELASTIC_APM_LOG_LEVEL=Trace -e ELASTIC_APM_STARTUP_HOOKS_LOGGING=1 --rm startuphook-sample
   ```

A similar approach would help with #1073

.ci/linux/build.sh
```diff
@@ -4,7 +4,4 @@
 #
 set -euxo pipefail
 
-# Remove Full Framework projects
-.ci/linux/remove-projects.sh
-
-dotnet build
+./build.sh
```
gregkalapos, line 7: nit: I think we can also remove the comment from line 3.

.ci/windows/msbuild.bat
```diff
@@ -5,4 +5,4 @@ echo "Prepare context for VsDevCmd.bat"
 call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\Common7\Tools\VsDevCmd.bat"
 nuget restore -verbosity detailed -NonInteractive
 
-msbuild /p:Configuration=Release
```
gregkalapos, line 8: Do we go back to building debug here? If so, why? I think you changed this not long ago.

src/Elastic.Apm.StartupHook.Loader/Loader.cs
```diff
@@ -0,0 +1,66 @@
+﻿using System;
```
gregkalapos, line 1: nit: the licence header is missing.

src/Elastic.Apm.StartupHook.Loader/Loader.cs
```diff
@@ -0,0 +1,66 @@
...
+		[MethodImpl(MethodImplOptions.NoInlining)]
+		private static void StartAgent()
+		{
+			Agent.Setup(new AgentComponents());
+			Agent.Subscribe(new HttpDiagnosticsSubscriber());
+
+			if (AppDomain.CurrentDomain.GetAssemblies().Any(n => n.GetName().Name.Contains("Microsoft.AspNetCore.")))
+			{
+				Agent.Subscribe(
+					new AspNetCoreErrorDiagnosticsSubscriber(),
```
gregkalapos, line 57: `AspNetCoreErrorDiagnosticsSubscriber` is not needed. I see it was there before, likely you just moved it here. It should not have been there in the original code. `AspNetCoreDiagnosticSubscriber` alone captures everything including errors. `AspNetCoreErrorDiagnosticsSubscriber` is used in combination with the middleware to capture errors that don't bubble up to the middleware. `AspNetCoreErrorDiagnosticsSubscriber` has this functionality, so that alone can capture everything we need.

src/Elastic.Apm/Elastic.Apm.csproj
```diff
@@ -2,6 +2,7 @@
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
     <IsPackable>true</IsPackable>
+    <UseDiagnosticSourceVersionFour>false</UseDiagnosticSourceVersionFour>
```
gregkalapos, line 5: As discussed a small comment would be nice here - or go with a flag which adds the version.

src/ElasticApmAgentStartupHook/StartupHook.cs
```diff
@@ -0,0 +1,172 @@
+﻿using System;
```
gregkalapos, line 1: Nit: licence header missing.

gregkalapos (approved):
Nice 👍 

Added a few minor things I noticed.

.ci/windows/msbuild.bat
russcam: Changed this back to the previously successful test run state, as tests were hanging in CI for Windows .NET Framework when built in Release configuration. Having changed this back, the tests did still hang on a run when compiling in Debug configuration, so building in Release configuration may not be an issue. I'll open a separate issue to track because tests hanging is a more general problem to investigate.

src/Elastic.Apm.StartupHook.Loader/Loader.cs
russcam: good spot!

src/Elastic.Apm.StartupHook.Loader/Loader.cs
russcam: Thanks, removed

src/Elastic.Apm/Elastic.Apm.csproj
russcam: Added a comment

src/ElasticApmAgentStartupHook/StartupHook.cs
russcam: nice spot! added

.ci/linux/build.sh
russcam: done!

russcam:
Will merge this in and follow up with a PR with automated tests. Manually tested with ASP.NET Core 3.1 and ASP.NET 5

PR #1139, opened 2021-01-20
fix doc markup

trentm:
Just noticed this while browsing the docs:
https://www.elastic.co/guide/en/apm/agent/dotnet/current/public-api.html#api-span-set-label

I think this was a typo from commit 2c433d31c45444e026081721e9c4190dadc383b8


gregkalapos (approved):
Thanks!

PR #1140, opened 2021-01-21
Remove coverlet.msbuild

russcam:
This commit removes coverlet.msbuild from
windows CI. The Windows .NET Framework
tests are intermittently failing, so this is a test to see if
this improves runs.

Code coverage is collected on Linux CI
so I think we'd miss code coverage for net461 with this change.


russcam:
Looking into this further, [coverlet recommends to use the `Xplat Code Coverage` coverlet.collector with dotnet test](https://github.com/coverlet-coverage/coverlet/issues/859#issuecomment-636308749) as it uses an in-process collector. If I'm understanding the test CI setup correctly, coverlet,msbuild is only currently used for

- test.bat: `dotnet test` on Windows that **does not** specify using `--collect:"XPlat Code Coverage"` but **does** specify `/p:CollectCoverage=true`

which I think we could switch over to using coverlet.collector instead.

russcam:
jenkins run tests

russcam:
jenkins run tests

russcam:
jenkins run tests

gregkalapos (approved):
Changes LGTM, but it still seems to timeout :(

image

russcam:
jenkins run tests

russcam:
All green build. Re-running

russcam:
Jenkins run tests

PR #1147, opened 2021-01-29
Propagate sample rate through tracestate

### LINKED ISSUE #1021: Propagate sample rate in tracestate
Follow up to https://github.com/elastic/apm-agent-dotnet/issues/906.

With https://github.com/elastic/apm-agent-dotnet/pull/1020 we already mark sample rate on transactions and spans. The next step is to propagate it in the `tracestate`.

Spec [here](https://github.com/elastic/apm/blob/master/specs/agents/tracing-sampling.md#propagation).

russcam:
This commit propagates the sample rate of the APM agent through the tracestate header, and uses the sample rate received in incoming tracestate as the sample rate for non-root transactions and spans.

Introduce `TraceState` type to model tracestate, allowing incoming headers to be added and sample rate to be set. Move TraceState validation from `TraceContext` onto `TraceState`.

Obsolete `TryDeserializeFromString` and `SerializeToString` on `DistributedTracingData` in favour of a method that accepts both traceparent and tracestate, and properties that get the traceparent and tracestate text headers, respectively.

Change Transaction and Span SampleRate to `double?` to allow a value to be omitted when incoming Trace context does not contain a tracestate header or the header does not contain the `es` vendor `s` attribute.

Closes #1021


gregkalapos (commented):
I have a bit of a mixed feeling about adding `[Obsolete()]` to the `DistributedTracingData` public APIs.

For HTTP we propagate the `tracestate` out of the box and I'm not sure how relevant it is to propagate this value manually for other use cases. I usually tend to prefer not to change public API. I'd totally agree with this PR if we'd start out new, but I'm not sure the benefit of adding the ability to propagate `tracestate` manually outweighs technically breaking user code (sure, with the attribute we don't break directly, but long term users need to change user code).

I'll ask around on what other agents do.

russcam:
In its current form, `DistributedTracingData` only exposes traceparent and **not** tracestate. The Elastic APM agents may be used with other systems that require manual propagation of traceparent and tracestate, but they would only be able to get the former and not the latter, which on the surface seems limiting; for example, [`System.Diagnostics.ActivityContext`](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.activitycontext.-ctor?view=net-5.0#System_Diagnostics_ActivityContext__ctor_System_Diagnostics_ActivityTraceId_System_Diagnostics_ActivitySpanId_System_Diagnostics_ActivityTraceFlags_System_String_System_Boolean_) exposes `TraceState`.

Let's sync to discuss.

russcam:
Would you mind taking another look at this, @gregkalapos? I think it's good to go now, with the changes in 9c6d5f3252e588722e5c358561fb94e4e7f7ce68, following our discussion.

src/Elastic.Apm/Model/Transaction.cs
```diff
@@ -120,54 +122,100 @@ internal Transaction(ApmAgent agent, string name, string type)
...
-				// PrentId could be also set here, but currently in the UI each trace must start with a transaction where the ParentId is null,
+				// ParentId could be also set here, but currently in the UI each trace must start with a transaction where the ParentId is null,
 				// so to avoid https://github.com/elastic/apm-agent-dotnet/issues/883 we don't set it yet.
 			}
 			else
 			{
 				if (_activity != null)
+				{
 					Id = _activity.SpanId.ToHexString();
+					// TODO: should _activity's trace context be populated with values from distributedTracingData?
```
gregkalapos, line 182: Good question - we have never done that and so far there was no feedback on this. Having said that, if we don't do it, then in this case the data between Activity and our transaction will be different. So I'm leaning toward syncing the data here as well.

src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerImplBase.cs
```diff
@@ -153,12 +155,15 @@ private void ProcessStartEvent(TRequest request, Uri requestUrl)
...
-			if (!RequestHeadersContain(request, TraceContext.TraceStateHeaderName) && transaction.OutgoingDistributedTracingData != null
-				&& transaction.OutgoingDistributedTracingData.HasTraceState)
-			{
-				RequestHeadersAdd(request, TraceContext.TraceStateHeaderName,
-					TraceContext.BuildTraceState(transaction.OutgoingDistributedTracingData));
-			}
+			if (!RequestHeadersContain(request, TraceContext.TraceStateHeaderName) && span.OutgoingDistributedTracingData != null && span.OutgoingDistributedTracingData.HasTraceState)
+				RequestHeadersAdd(request, TraceContext.TraceStateHeaderName, span.OutgoingDistributedTracingData.TraceState.ToTextHeader());
+
+			//if (!RequestHeadersContain(request, TraceContext.TraceStateHeaderName) && transaction.OutgoingDistributedTracingData != null
```
gregkalapos, line 161: nit: I think this is just some leftover, right? Could you please remove it?

gregkalapos (approved):
LGTM 👍 

Just to document here one thing I was interested in and played with:
`Activity` also propagates the `traceparent` and `tracestate` in HTTP requests - so it writes the exact same HTTP headers as we do. I did some testing and it seems that always our values end up being in those headers. We specifically check in `HttpDiagnosticListenerImplBase.cs` if those hears are already present and only write those if they aren't. As it seems this is all fine and works as intended.

src/Elastic.Apm/Model/Transaction.cs
russcam: This has now been done as part of #1228, so I'll remove the comment 

https://github.com/elastic/apm-agent-dotnet/blob/eb4f85b79b2b7b942d60a54c7a7b1256ca3d4343/src/Elastic.Apm/Model/Transaction.cs#L148-L155

src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerImplBase.cs
russcam: done!

PR #1148, opened 2021-01-29
Fix scoped logger naming in GcMetricsProvider

russcam:


russcam:
small change, merging this in.

PR #1161, opened 2021-02-01
Handling exceptions in the agent

gregkalapos:
Adapting code according to https://github.com/elastic/apm-agent-dotnet/pull/1162.

### Remove InstanceAlreadyCreatedException, instead we log now

In cases of multiple initializations, the agent previously fired an `InstanceAlreadyCreatedException`. This always happened when an already initialized agent was present and a 2. initialization was triggered. E.g. calling `Agent.Setup(AgentComponents)` twice, or enabling the agent with `UseAllElasticApm` and then calling `Agent.Setup()` manually, etc. With this PR, instead of throwing an exception and potentially crashing the app, we log an error and keep the original agent setup.

### Introduce DiagnosticListenerBase class which handles errors in OnNext

To make sure all DiagnosticSource listeners handle errors, this PR introduces `DiagnosticListenerBase` with a try-catch. With this, code in sub classes is automatically guarded.

_We almost already introduced such a base class in one of the PRs earlier... Now I think we have enough reason for it._

### DIFF
--- a/src/Elastic.Apm/Agent.cs
+++ b/src/Elastic.Apm/Agent.cs
@@ -252,7 +252,8 @@ namespace Elastic.Apm
 		public static void Setup(AgentComponents agentComponents)
 		{
 			if (LazyApmAgent.IsValueCreated)
-				agentComponents?.Logger?.Error()?.Log("The singleton APM agent has already been instantiated and can no longer be configured.");
+				agentComponents?.Logger?.Error()?.Log("The singleton APM agent has" +
+					" already been instantiated and can no longer be configured. Reusing existing instance");
 
 			Components = agentComponents;
 			_isConfigured = true;
--- a/src/Elastic.Apm/Agent.cs
+++ b/src/Elastic.Apm/Agent.cs
@@ -252,7 +252,7 @@ namespace Elastic.Apm
 		public static void Setup(AgentComponents agentComponents)
 		{
 			if (LazyApmAgent.IsValueCreated)
-				throw new InstanceAlreadyCreatedException("The singleton APM agent has already been instantiated and can no longer be configured.");
+				agentComponents?.Logger?.Error()?.Log("The singleton APM agent has already been instantiated and can no longer be configured.");
 
 			Components = agentComponents;
 			_isConfigured = true;
@@ -262,11 +262,6 @@ namespace Elastic.Apm
 		{
 			if (!LazyApmAgent.IsValueCreated)
 				Setup(apmAgent.Components);
-		}
-
-		internal class InstanceAlreadyCreatedException : Exception
-		{
-			internal InstanceAlreadyCreatedException(string message) : base(message) { }
 		}
 	}
 }
--- a/src/Elastic.Apm/DiagnosticListeners/DiagnosticListenerBase.cs
+++ b/src/Elastic.Apm/DiagnosticListeners/DiagnosticListenerBase.cs
@@ -1,0 +1,59 @@
+// Licensed to Elasticsearch B.V under
+// one or more agreements.
+// Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information
+
+using System;
+using System.Collections.Generic;
+using Elastic.Apm.DiagnosticSource;
+using Elastic.Apm.Logging;
+
+namespace Elastic.Apm.DiagnosticListeners
+{
+	/// <summary>
+	/// A base class for DiagnosticSource listeners which encapsulates common functionality - e.g. exception handling.
+	/// </summary>
+	public abstract class DiagnosticListenerBase : IDiagnosticListener
+	{
+		/// <summary>
+		/// A logger scoped to the child class.
+		/// </summary>
+		protected readonly IApmLogger Logger;
+
+		/// <summary>
+		/// Current Agent instance.
+		/// </summary>
+		protected readonly IApmAgent ApmAgent;
+
+		protected DiagnosticListenerBase(IApmAgent apmAgent)
+		{
+			ApmAgent = apmAgent;
+			Logger = apmAgent.Logger.Scoped(GetType().Name);
+		}
+
+		/// <summary>
+		/// Fires each time when <see cref="OnNext"/> is called.
+		/// Code within this method is guarded by a try-catch.
+		/// </summary>
+		/// <param name="kv"></param>
+		protected abstract void HandleOnNext(KeyValuePair<string, object> kv);
+
+		public abstract string Name { get; }
+
+		public void OnCompleted() { }
+
+		public virtual void OnError(Exception error) { }
+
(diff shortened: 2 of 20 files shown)


bmorelli25 (commented):
Troubleshooting docs look fine 👍

src/Elastic.Apm/Agent.cs
```diff
@@ -252,7 +252,7 @@ private static bool CheckAndAddFilter(Action<PayloadSenderV2> action)
 		public static void Setup(AgentComponents agentComponents)
 		{
 			if (LazyApmAgent.IsValueCreated)
-				throw new InstanceAlreadyCreatedException("The singleton APM agent has already been instantiated and can no longer be configured.");
+				agentComponents?.Logger?.Error()?.Log("The singleton APM agent has already been instantiated and can no longer be configured.");
```
russcam, line 255:
```suggestion
				agentComponents?.Logger?.Error()?.Log("The singleton APM agent has already been instantiated and can no longer be configured. Using existing instance");
```

src/Elastic.Apm/DiagnosticListeners/DiagnosticListenerBase.cs
```diff
@@ -0,0 +1,59 @@
...
+{
+	/// <summary>
+	/// A base class for DiagnosticSource listeners which encapsulates common functionality - e.g. exception handling.
+	/// </summary>
+	public abstract class DiagnosticListenerBase : IDiagnosticListener
+	{
+		/// <summary>
+		/// A logger scoped to the child class.
+		/// </summary>
+		protected readonly IApmLogger Logger;
```
russcam, line 21:
```suggestion
		protected IApmLogger Logger { get; };
```

src/Elastic.Apm/DiagnosticListeners/DiagnosticListenerBase.cs
```diff
@@ -0,0 +1,59 @@
...
+
+		/// <summary>
+		/// Current Agent instance.
+		/// </summary>
+		protected readonly IApmAgent ApmAgent;
```
russcam, line 26:
```suggestion
		protected IApmAgent ApmAgent { get; }
```

src/Elastic.Apm/DiagnosticListeners/DiagnosticListenerBase.cs
```diff
@@ -0,0 +1,59 @@
...
+		/// <summary>
+		/// Fires each time when <see cref="OnNext"/> is called.
+		/// Code within this method is guarded by a try-catch.
+		/// </summary>
+		/// <param name="kv"></param>
+		protected abstract void HandleOnNext(KeyValuePair<string, object> kv);
+
+		public abstract string Name { get; }
+
+		public void OnCompleted() { }
```
russcam, line 43:
```suggestion
		public virtual void OnCompleted() { }
```

russcam (changes requested):
On the whole, the changes look good.

The removal of `InstanceAlreadyCreatedException` is a behavioural change. In theory, a thrown `InstanceAlreadyCreatedException` would bubble up to the application which I would suspect would be noticed and fixed by a user early, as the agent would not be functioning otherwise. The removal of `InstanceAlreadyCreatedException` could be viewed as a positive behavioural change, as it's removal leaves an application in no less of a worse state. On the other hand, where an exception would have been thrown, an error is now logged, which is less visible and could leave a user in a state where they believe a change has taken effect (e.g. changed the setup of the agent) when it has not. This looks like an acceptable minor version change.

The introduction of `DiagnosticListenerBase` _looks_ binary compatible - do the differ results flag anything?

gregkalapos:
Ready for another round.


> The introduction of `DiagnosticListenerBase` _looks_ binary compatible - do the differ results flag anything?

It flags some unrelated things, which I think don't came from this PR. The `DiagnosticListenerBase` shows up as a new type:

```
<Type Name="Elastic.Apm.DiagnosticListeners.DiagnosticListenerBase" DiffType="New" />
```

Nothing is flagged on types where this is now the parent type except `GrpcClientDiagnosticListener` - I'd expect the same for other derived types. The methods marked as `Deleted` are still there through the parent class, plus they were empty, and on top of that these were there for the framework to call (which it never does), so I think this is fine. 

```
<Module Name="Elastic.Apm.GrpcClient.dll" DiffType="Modified">
    <Type Name="Elastic.Apm.GrpcClient.GrpcClientDiagnosticListener" DiffType="Modified">
      <Method Name="HandleOnNext(KeyValuePair&lt;String, Object&gt;)" DiffType="New" />
      <Method Name="OnCompleted()" DiffType="Deleted" />
      <Method Name="OnError(Exception)" DiffType="Deleted" />
      <Method Name="OnNext(KeyValuePair&lt;String, Object&gt;)" DiffType="Deleted" />
    </Type>
```

src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerImplBase.cs
```diff
@@ -51,85 +43,89 @@ protected HttpDiagnosticListenerImplBase(IApmAgent agent)
...
-			else
-				Logger.Trace()?.Log("Unrecognized key `{DiagnosticEventKey}'", kv.Key);
 		}
 
 		private void ProcessStartEvent(TRequest request, Uri requestUrl)
 		{
 			Logger.Trace()?.Log("Processing start event... Request URL: {RequestUrl}", Http.Sanitize(requestUrl));
 
-			var transaction = _agent.Tracer.CurrentTransaction;
+			var transaction = Agent.Tracer.CurrentTransaction;
```
russcam, line 121:
```suggestion
			var transaction = ApmAgent.Tracer.CurrentTransaction;
```

russcam (approved):
Changes LGTM 👍 

One thing to note is that now in some uses of the base `IApmAgent ApmAgent` property, a cast to `ApmAgent` happens in the handling of each diagnostic event. For example, in `HandleOnNext()` in `AspNetCoreDiagnosticListener`. Wondering if in these cases it is worth avoiding the cast each time by assigning the `ApmAgent` in the ctor to a `private ApmAgent _agent` field?

gregkalapos:
> One thing to note is that now in some uses of the base `IApmAgent ApmAgent` property, a cast to `ApmAgent` happens in the handling of each diagnostic event. For example, in `HandleOnNext()` in `AspNetCoreDiagnosticListener`. Wondering if in these cases it is worth avoiding the cast each time by assigning the `ApmAgent` in the ctor to a `private ApmAgent _agent` field?

It makes sense - I don't think it'll make a huge difference, but why not... Specifically in `AspNetCoreDiagnosticListener` I don't think is super relevant, since we only cast once there and the reason for that cast is to get the current config snapshot. Since people already asked about making that public maybe we can get rid of that single cast later. In `EfCoreDiagnosticListener` and `SqlClientDiagnosticListener` we had multiple casts, I think your suggestion would be especially useful there, so I adapted those 2.

PR #1162, opened 2021-02-01
Document "Never crash" approach

gregkalapos:
Following our discussion from one of the planning meetings, documenting the "never crash" approach.

fyi: @AlexanderWert.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -125,6 +125,14 @@ If this is the case then a perf. test should be added to the `test\Elastic.Apm.B
 
 We care both about memory and CPU overhead and both should be measured. The `test\Elastic.Apm.Benchmarks` is configured to measure both.
 
+#### Do not crash the monitored process
+
+The agent must never crash the monitored application.
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -125,6 +125,14 @@ If this is the case then a perf. test should be added to the `test\Elastic.Apm.B
 
 We care both about memory and CPU overhead and both should be measured. The `test\Elastic.Apm.Benchmarks` is configured to measure both.
 
+#### Do not crash the monitored process
+
+The agent must never crash the monitored application.
+
+One aspect of this is that the agent code itself never throws exceptions. Instead, the agent uses the logger to log errors and it either skips specific events or completely stops collecting events. Skipping events must be preferred over unhandled exceptions in agent code.
```
russcam, line 132:
```suggestion
One aspect of this is that the agent code itself should never throws exceptions. Instead, the agent uses the logger to log errors and it either skips specific events or completely stops collecting events. Skipping events must be preferred over unhandled exceptions in agent code.
```

russcam (approved):
LGTM, left a small comment.

CONTRIBUTING.md
russcam: **must** never throw exceptions is more correct here, I think.

PR #1163, opened 2021-02-01
Send destination info for embedded databases

gregkalapos:
Prior to this PR we had a check to not send destination info for local databases. Reason for this was the part `When to set destination.service.* fields?` in the [APM issue](https://github.com/elastic/apm/issues/180) that introduced this feature.

This causes things like SQLite never show up on service map. It seems that no other agent had this check - at least I asked and I got no specific acknowledgements that other agents would do the same.

Therefore this PR removes this check.

With this, local databases (like SQLite) will also show up on the service map.

image

@alex-fedotyev fyi.


russcam (approved):
LGTM 👍

PR #1165, opened 2021-02-04
StartupHook fix: remove check on Microsoft.AspNetCore assemblies

gregkalapos:
### The bug

In the startup hook logic we had this check:

```
if (AppDomain.CurrentDomain.GetAssemblies().Any(n => n.GetName().Name.Contains("Microsoft.AspNetCore.")))
{
	Agent.Subscribe(
		new AspNetCoreErrorDiagnosticsSubscriber(),
		new AspNetCoreDiagnosticSubscriber(),
		new EfCoreDiagnosticsSubscriber(),
		new SqlClientDiagnosticSubscriber(),
		new ElasticsearchDiagnosticsSubscriber()
	);
}
```

In other words we only turned on most of our diagnostic subscribers when at least one `Microsoft.AspNetCore.[X].dll` was loaded. The thinking behind this check was that e.g. `AspNetCoreDiagnosticSubscriber`  (which starts the transactions for incoming HTTP calls) only makes sense to load when ASP.NET Core itself is also loaded. Problem is that this check is not reliable and does not always work. Specifically this check returns false on an ASP.NET Core Web API project on 3.1 which leads to not capturing any transaction. `Microsoft.AspNetCore.[X]` dlls get loaded in that case, but those are NOT loaded yet when the startup hook code gets called. Fun fact: on ASP.NET Core MVC the check works perfectly. Bottom line is, this check is not good.

### The fix 

I removed the check and try to load all our `IDiagnosticsSubscriber` implementations without any check - I feel we can't really be smart here, best is to try loading everything. Now some of those due to their dependencies can fail, so I also added exception handling and logging around this.

### Next steps

@russcam already works on automated testing for the startuphook feature. In order to incorporate the learning from here we should also test against non-MVC (specifically ASP.NET Core WebAPI) projects and also against a console app with minimal dependencies.


src/Elastic.Apm.StartupHook.Loader/Loader.cs
```diff
@@ -57,15 +57,24 @@ private static void StartAgent()
...
+			void LoadDiagnosticSubscriber(IDiagnosticsSubscriber diagnosticsSubscriber)
 			{
-				Agent.Subscribe(
-					new AspNetCoreDiagnosticSubscriber(),
-					new EfCoreDiagnosticsSubscriber(),
-					new SqlClientDiagnosticSubscriber(),
-					new ElasticsearchDiagnosticsSubscriber(),
-					new GrpcClientDiagnosticSubscriber()
-				);
+				var logger = StartupHookLogger.CreateLogger();
```
russcam, line 68: could create once outside the local function and reuse

src/Elastic.Apm.StartupHook.Loader/Loader.cs
```diff
@@ -57,15 +57,24 @@ private static void StartAgent()
 			Agent.Setup(new AgentComponents());
 			Agent.Subscribe(new HttpDiagnosticsSubscriber());
```
russcam, line 58: Should also be done inside `LoadDiagnosticSubscriber(new HttpDiagnosticsSubscriber())`?

src/Elastic.Apm.StartupHook.Common/StartupHookLogger.cs
```diff
@@ -0,0 +1,58 @@
...
+
+using System;
+using System.IO;
+
+namespace Elastic.Apm.StartupHook.Common
+{
+	/// <summary>
+	/// Logs startup hook process, useful for debugging purposes.
+	/// </summary>
+	public class StartupHookLogger
```
russcam, line 14: If it's just the `StartupHookLogger` that's shared between `ElasticApmAgentStartupHook` and `Elastic.Apm.StartupHook.Loader`, a single linked file (or even copy/paste!) would be less involved than a shared project :)

src/Elastic.Apm.StartupHook.Common/StartupHookLogger.cs
```diff
@@ -0,0 +1,58 @@
...
+		{
+			_logPath = logPath;
+			_enabled = enabled;
+		}
+
+		/// <summary>
+		/// Returns a logger and initializes it based on environment variables
+		/// </summary>
+		/// <returns></returns>
+		public static StartupHookLogger CreateLogger()
```
russcam, line 29:
```suggestion
		public static StartupHookLogger Create()
```

russcam (commented):
I've left some comments. Overall, I think it's reasonable to instantiate all the diagnostic listeners as the overhead in instantiating but not using is probably small, and overcomes assembly load ordering.

src/Elastic.Apm.StartupHook.Loader/Loader.cs
gregkalapos: Yeah, it definitely doesn't hurt. 

`HttpDiagnosticsSubscriber` lives within `Elastic.Apm` and only depends on things that are always loaded, so I don't expect that to fail. Nevertheless I agree it's best to also guard this one. Moved it into `LoadDiagnosticSubscriber()`.

src/Elastic.Apm.StartupHook.Common/StartupHookLogger.cs
gregkalapos: Ok, I changed this to share a file from one project and link it into the other.

src/Elastic.Apm.StartupHook.Common/StartupHookLogger.cs
gregkalapos: Done

src/Elastic.Apm.StartupHook.Loader/Loader.cs
gregkalapos: I'm not sure I follow this one. This is the only place currently where we use the logger in `Loader`. Sure, if we'd log later at other places we can move the logger.

gregkalapos:
@russcam ready for another round.

src/Elastic.Apm.StartupHook.Loader/Loader.cs
russcam: A new logger is created on every local function instantiation of `LoadDiagnosticSubscriber(...)`, for each DiagnosticSubscriber. Whilst the overhead in creating a new logger each time is small, a single instance could be reused for loading all DiagnosticSubscribers.

ElasticApmAgent.sln
```diff
@@ -101,31 +101,34 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Elastic.Apm.GrpcClient", "s
...
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Elastic.Apm.StaticImplicitInitialization.Tests", "test\Elastic.Apm.StaticImplicitInitialization.Tests\Elastic.Apm.StaticImplicitInitialization.Tests.csproj", "{CE0120A5-A094-40A3-905E-8E14526800D3}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Elastic.Apm.StaticExplicitInitialization.Tests", "test\Elastic.Apm.StaticExplicitInitialization.Tests\Elastic.Apm.StaticExplicitInitialization.Tests.csproj", "{DC879845-353B-4C62-9640-6EA3008326F7}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Elastic.Apm.StaticExplicitInitialization.Tests", "test\Elastic.Apm.StaticExplicitInitialization.Tests\Elastic.Apm.StaticExplicitInitialization.Tests.csproj", "{DC879845-353B-4C62-9640-6EA3008326F7}"
 EndProject
 Global
 	GlobalSection(SharedMSBuildProjectFiles) = preSolution
+		src\Elastic.Apm.StartupHook.Common\Elastic.Apm.StartupHook.Common.projitems*{5c5c19e2-a35b-4f5f-809c-4ca797c886d9}*SharedItemsImports = 13
+		src\Elastic.Apm.StartupHook.Common\Elastic.Apm.StartupHook.Common.projitems*{7b18c4c5-099b-41d1-8cc2-2d2c412b80cd}*SharedItemsImports = 5
```
russcam, line 129: can be removed?

ElasticApmAgent.sln
```diff
@@ -101,31 +101,34 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Elastic.Apm.GrpcClient", "s
...
 		test\Elastic.Apm.DatabaseTests.Common\Elastic.Apm.DatabaseTests.Common.projitems*{968e1e85-e996-42de-9845-d20dae16165a}*SharedItemsImports = 5
+		src\Elastic.Apm.StartupHook.Common\Elastic.Apm.StartupHook.Common.projitems*{b5200dc5-5e48-4f72-ada9-96d0380a00cb}*SharedItemsImports = 5
```
russcam, line 131: can be removed?

src/Elastic.Apm.StartupHook.Common/Elastic.Apm.StartupHook.Common.projitems
```diff
@@ -0,0 +1,14 @@
...
+    <HasSharedItems>true</HasSharedItems>
+    <SharedGUID>5c5c19e2-a35b-4f5f-809c-4ca797c886d9</SharedGUID>
+  </PropertyGroup>
+  <PropertyGroup Label="Configuration">
+    <Import_RootNamespace>Elastic.Apm.StartupHook.Common</Import_RootNamespace>
+  </PropertyGroup>
+  <ItemGroup>
+    <Compile Include="$(MSBuildThisFileDirectory)StartupHookLogger.cs" />
+  </ItemGroup>
+</Project>
```
russcam, line 14: can be removed?

src/Elastic.Apm.StartupHook.Common/Elastic.Apm.StartupHook.Common.shproj
```diff
@@ -0,0 +1,13 @@
...
+    <ProjectGuid>5c5c19e2-a35b-4f5f-809c-4ca797c886d9</ProjectGuid>
+    <MinimumVisualStudioVersion>14.0</MinimumVisualStudioVersion>
+  </PropertyGroup>
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.Common.Default.props" />
+  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.Common.props" />
+  <PropertyGroup />
+  <Import Project="Elastic.Apm.StartupHook.Common.projitems" Label="Shared" />
+  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.CSharp.targets" />
+</Project>
```
russcam, line 13: can be removed?

src/Elastic.Apm.StartupHook.Common/StartupHookLogger.cs
```diff
@@ -0,0 +1,12 @@
...
+// Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information
+
+using System;
+using System.IO;
+
+namespace Elastic.Apm.StartupHook.Common
+{
+
+}
```
russcam, line 12: can be removed?

src/Elastic.Apm.StartupHook.Loader/Elastic.Apm.StartupHook.Loader.csproj
```diff
@@ -14,4 +14,12 @@
       <ProjectReference Include="..\Elastic.Apm\Elastic.Apm.csproj" />
     </ItemGroup>
 
+    <ItemGroup>
+      <Compile Include="..\ElasticApmAgentStartupHook\StartupHookLogger.cs">
+        <Link>StartupHookLogger.cs</Link>
+      </Compile>
+    </ItemGroup>
+
+    <Import Project="..\Elastic.Apm.StartupHook.Common\Elastic.Apm.StartupHook.Common.projitems" Label="Shared" />
```
russcam, line 23: can be removed?

src/ElasticApmAgentStartupHook/ElasticApmAgentStartupHook.csproj
```diff
@@ -5,4 +5,5 @@
     <IsPackable>false</IsPackable>
     <RootNamespace>ElasticApmStartupHook</RootNamespace>
   </PropertyGroup>
+  <Import Project="..\Elastic.Apm.StartupHook.Common\Elastic.Apm.StartupHook.Common.projitems" Label="Shared" />
```
russcam, line 8: can be removed?

russcam (commented):
Looks like there are some leftover files from the shared project, other LGTM 👍 

I'll remove these and merge, as this is needed for #1167

PR #1169, opened 2021-02-08
StartupHook: Introduce AssemblyLoadContext to load agent assemblies

gregkalapos:
Prior to this PR with the startuphook feature we loaded everything into `AssemblyLoadContext.Default`. This PR introduces a separate AssemblyLoadContext  to load agent assemblies and dependencies of agent assemblies.

Why we need this?
- The agent depends on `Newtonsoft.Json`  `Version="11.0.2"`
- If the application depends on a different version (which is not unlikely), we cause a conflict. What happens in those cases is that when we try to load `Newtonsoft.Json` with `AssemblyLoadContext.Default`, it throws an exception and the agent won't work.
- With an extra load context we can load the two versions at the same time, the app uses its own version and the agent also uses its own version, all good.

The problem this PR solves can be reproduced by creating an empty project, adding `<PackageReference Include="Newtonsoft.Json" Version="9.0.2" />` (or any version) to its `csproj` file and use the startuphook feature. Same happens when there is a dependency in the app which depends on `Newtonsoft.Json`.

### DIFF
--- a/src/Elastic.Apm.StartupHook.Loader/Elastic.Apm.StartupHook.Loader.csproj
+++ b/src/Elastic.Apm.StartupHook.Loader/Elastic.Apm.StartupHook.Loader.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
     <PropertyGroup>
-        <TargetFramework>netcoreapp2.1</TargetFramework>
+        <TargetFramework>netcoreapp2.2</TargetFramework>
         <IsPackable>false</IsPackable>
     </PropertyGroup>
 
--- a/src/Elastic.Apm.StartupHook.Loader/Elastic.Apm.StartupHook.Loader.csproj
+++ b/src/Elastic.Apm.StartupHook.Loader/Elastic.Apm.StartupHook.Loader.csproj
@@ -1,18 +1,9 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
     <PropertyGroup>
-        <TargetFramework>netcoreapp3.1</TargetFramework>
+        <TargetFramework>netcoreapp2.1</TargetFramework>
         <IsPackable>false</IsPackable>
     </PropertyGroup>
-
-    <ItemGroup>
-      <ProjectReference Include="..\Elastic.Apm.AspNetCore\Elastic.Apm.AspNetCore.csproj" />
-      <ProjectReference Include="..\Elastic.Apm.Elasticsearch\Elastic.Apm.Elasticsearch.csproj" />
-      <ProjectReference Include="..\Elastic.Apm.EntityFrameworkCore\Elastic.Apm.EntityFrameworkCore.csproj" />
-      <ProjectReference Include="..\Elastic.Apm.SqlClient\Elastic.Apm.SqlClient.csproj" />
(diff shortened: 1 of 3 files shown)


russcam:
@gregkalapos did you want to rebase this against master, to pick up the startup hook tests merged in https://github.com/elastic/apm-agent-dotnet/pull/1167?

src/Elastic.Apm.StartupHook.Loader/Loader.cs
```diff
@@ -40,40 +44,72 @@ private static string AssemblyDirectory
...
+
+			foreach (var libToLoad in AgentLoadContext.AgentDependencyLibsToLoad)
+				LoadAssembly(libToLoad);
+			foreach (var libToLoad in AgentLoadContext.AgentLibsToLoad)
+				LoadAssembly(libToLoad);
+
+			AssemblyLoadContext.Default.Resolving += (context, assemblyName) =>
+			{
+				if (AgentLoadContext.AgentDependencyLibsToLoad.Contains(assemblyName.Name)
+					|| AgentLoadContext.AgentLibsToLoad.Contains(assemblyName.Name))
```
russcam, line 57: Should we make both `AgentDependencyLibsToLoad` and `AgentLibsToLoad` to be `HashSet<string>` for this op?

src/Elastic.Apm.StartupHook.Loader/Loader.cs
```diff
@@ -40,40 +44,72 @@ private static string AssemblyDirectory
...
+			};
 
 			StartAgent();
 		}
 
+		public static void LoadAssembly(string assemblyName)
+		{
+			if (AppDomain.CurrentDomain.GetAssemblies().Any(n => n.FullName.Equals(assemblyName, StringComparison.Ordinal)))
+			{
+				Logger.WriteLine($"{assemblyName} is alrady loaded - we don't try to load it");
```
russcam, line 73:
```suggestion
				Logger.WriteLine($"{assemblyName} is already loaded");

```

src/Elastic.Apm.StartupHook.Loader/Elastic.Apm.StartupHook.Loader.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
     <PropertyGroup>
-        <TargetFramework>netcoreapp2.2</TargetFramework>
+        <TargetFramework>netcoreapp3.1</TargetFramework>
```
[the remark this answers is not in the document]
gregkalapos: Originally I used an API which was only present in `3.1`, but it turned out to be not needed. I can revert this. So, given there is no change needed here, I'd say let's keep it as it is. The next push will change it back to `2.2.`.

gregkalapos:
This is still work in progress. I'm still working on some cases where the fix does not help.

gregkalapos:
I gave up on this - loading the agent into a separate AssemblyLoadContext has issues with diagnostic source, since it depends on a static instance.

I opened https://github.com/elastic/apm-agent-dotnet/pull/1227 to solve the Newtonsoft.Json issue by using `assembly-rewriter` instead of relying on AssemblyLoadContext.

PR #1177, opened 2021-02-15
DotnetProject.cs  - handle path with space in it

gregkalapos:
Ran into an issue on Windows where the path includes an empty space. Adding `"` to handle it.


russcam (approved):
LGTM, thanks @gregkalapos 👍

Issue #1080, opened 2020-12-04
Transaction.Custom is null if Apm is disabled


tbutler-qontigo:
**Describe the bug**
Version 1.7.1

When Enabled = false you get a NoopTransaction when you call StartTransaction.
The Custom dictionary is null so any attempt to add to that will throw a NullReferenceException

**To Reproduce**
1. Ensure Apm is disabled (Enabled = false)
2. Then call `var transaction = Agent.Tracer.StartTransaction(...)` and note that you get a NoopTransaction
3. Then do `transaction.Custom.Add(....)`
4. Observe that you will get a `NullReferenceException` because Custom is null

**Expected behavior**
1transaction.Custom.Add1 should not fail whether Apm is enabled or not.

gregkalapos:
Thanks for reporting @tbutler-qontigo.

I opened https://github.com/elastic/apm-agent-dotnet/pull/1081 to fix it.

tbutler-qontigo:
BTW just discovered this is also a problem with `OutgoingDistributedTracingData` - we call SerializeToString on that and again that is null for a `NoopTransaction`

gregkalapos:
> BTW just discovered this is also a problem with `OutgoingDistributedTracingData` - we call SerializeToString on that and again that is null for a `NoopTransaction`

That one is known to me and I think it's best to not fake that one. If the agent is disabled, you won't get `OutgoingDistributedTracingData` out of the agent. If we'd fake it, we'd either need to generate all the ids (`traceid`, `transactionid`, and `spanid`) which would mean the agent still does a lot of things, or return something fake, but that'd mean we give you invalid data.

gregkalapos:
On the other hand, I think we should warn people about this. I'll add some docs to make this clear.

tbutler-qontigo:
Hi - I think the problem (or "a" problem) is that to the developer, they don't know that they are getting a `NoopTransaction` rather than a real one and so don't know to check the properties have values.
Also it does mean that 1.7.x should really be 2.x (if you follow semver?) because it is essentially a breaking change - code that used to always work now doesn't if Apm is not enabled because properties which were always non null before are now potentially null in some circumstances.

Putting foo?.bar?.baz everywhere is a bit of a smell tbh...but if that's what we have to do then that's what we have to do....
If there are more props that might be null on the NoopTransaction can you add them to the docs too please?

[...]

PR #1179, opened 2021-02-15
[Doc] Adding some notes about disabled agent

gregkalapos:
Inspired by https://github.com/elastic/apm-agent-dotnet/issues/1080. The goal here is to make clear that if you disable the agent, the Public API will behave differently.


russcam (approved):
LGTM

Issue #1195, opened 2021-02-19
Getting no GC metrics on .NET5.0


stefannikolei:
Using .NET5.0 and Elastic.Apm.NetCoreAll v1.8.0

I used created a new WebApplication from Visual Studio and just added app.UseAllElasticApm(configuration).

I am not getting any gc metrics.
The metric logs look like this:

```
{MetricsCollector} Metrics collected: MetricSample{system.process.cpu.total.norm.pct: 0,0005209173955098381}, MetricSample{system.cpu.total.norm.pct: 0,16761072158813475}, MetricSample{system.process.memory.size: 2225303339008}, MetricSample{system.process.memory.rss.bytes: 99045376}, MetricSample{system.memory.actual.free: 10745376768}, MetricSample{system.memory.total: 34278375424}
```
Configuration looks like this 
```
"ElasticApm":{
"active":true,
"LogLevel": Debug,
"ServerUrl": xxx,
"SecretToken": xxx,
"Environment": "dev",
"ServiceName": "GcTest",
"VerifyServerCert": "false",
"TransactionSampleRate": 1.0
}
```

Do i need to configure something else to get gc information

stefannikolei:
I only get the entry of the GcEventListener that it was initialized. After this i never see anything from him again (LogLevel: trace)

gregkalapos:
I merged #1196 and opened https://github.com/elastic/apm-agent-dotnet/pull/1199 to make sure this is also covered by our tests. 

I'll keep this open until https://github.com/elastic/apm-agent-dotnet/pull/1199 is merged.

gregkalapos:
Test which covers this is merged now in #1199. Thanks for opening this @stefannikolei  and also thanks for the fix.

PR #1196, opened 2021-02-19
Add GCHeapStats_V2

stefannikolei:
When using .NET5 we do not get GCHeapStats_V1 only GCHeapStats_V2

#1195


gregkalapos (approved):
Thanks @stefannikolei.

LGTM

I'll follow up and look at our tests. This should be covered in `MetricsTests.CollectGcMetrics`.

PR #1197, opened 2021-02-19
Adding Redis and background hosted service to SampleAspNetCoreApp to show more of a real world end to end sample.

ejsmith:
@russcam as we talked about last night, I've added some features to the AspNetSampleCoreApp to represent more of a real world application and show the Elastic APM client fully working.

Currently, it does not appear to be capturing Redis spans or other spans that are happening in the background queue hosted service that I've added.

### DIFF
--- a/sample/SampleAspNetCoreApp/Controllers/HomeController.cs
+++ b/sample/SampleAspNetCoreApp/Controllers/HomeController.cs
@@ -19,6 +19,7 @@ using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using SampleAspNetCoreApp.Data;
 using SampleAspNetCoreApp.Models;
+using StackExchange.Redis;
 
 namespace SampleAspNetCoreApp.Controllers
 {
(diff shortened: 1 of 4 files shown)


russcam:
The SampleAspNetCoreApp is used for integration tests, with a SQLite db as the only external dependency,so that tests can use Microsoft.AspNetCore.Mvc.Testing to spin up the application from code.

I'm thinking we might want to have a separate project for a kitchen sink sample, with a docker-compose file or similar, for spinning up the third party services it uses. What do you think, @gregkalapos?

ejsmith:
Yeah, that makes sense. I wasn't sure if you were running Elasticsearch to report the data to or not.

gregkalapos:
> I'm thinking we might want to have a separate project for a kitchen sink sample, with a docker-compose file or similar, for spinning up the third party services it uses. What do you think, @gregkalapos?

Why don't we just extend the `StackExchangeRedisSample` with the new sample from this PR? That already uses `DotNet.Testcontainers` so it spins up everything we need.

ejsmith:
@gregkalapos how are you going about actually looking at the APM data in Elastic and not just running tests with fake exporters? I'm not seeing any sort of end to end example that appears to be reporting the data to elastic. That is what I am interested in. I want to see all these things working together across multiple services.

gregkalapos:
In this repo we mostly mock the APM Server.

We have another repo where we have end-to-end tests (for all agents, not only .NET) [here](https://github.com/elastic/apm-integration-testing/).

ejsmith:
@gregkalapos ok, just checked that out and it appears that there is an extremely simple sample app there. Am I wrong in thinking that there should be much more real world sample app that spans multiple services so you can see how everything is working together?

gregkalapos:
> @gregkalapos ok, just checked that out and it appears that there is an extremely simple sample app there. Am I wrong in thinking that there should be much more real world sample app that spans multiple services so you can see how everything is working together?

From one of the dockerfiles we have in that repo we fetch [this ASP.NET Core app](https://github.com/elastic/opbeans-dotnet), we have the same app for other agents, we start all of them, they talk to each other and we assert based on data from elasticsearch. That's our end-to-end test where we don't mock anything.

russcam:
Hey @ejsmith, we discussed in our weekly sync the idea of having more of a "kitchen sink" sample application that instruments multiple different services. We agreed it would be a good idea to have one, and that it should be separate from SampleAspNetCoreApp, which is used in integration tests for ASP.NET Core integration. At the moment, we've got a few things in progress

- MongoDB integration
- Azure Service Bus integration
- Azure Storage integration
- Azure CosmosDB integration

which we're working on getting in the next release or release after that. In light of this, it might be a little bit of time before we can put one together. I've opened #1246 to track and discuss it, and will close this PR for now.

Issue #1194, opened 2021-02-19
Detected pacakge downgrade on blank ASP.NET Core WebApi NET5.0


stefannikolei:
I created a fresh WebApi project in VS2019 targeting .NET5.0.

When trying to install the package Elastic.Apm.NetCoreAll 1.8.0 version i get following Nuget error:

```
Detected package downgrade Microsoft.Extensions.Logging.Abstractions from 5.0.0 to 3.0.0
Detected package downgrade Microsoft.Extensions.Hosting.Abstractions from 5.0.0 to 3.0.0
```

gregkalapos:
Thanks for reporting this. Opened https://github.com/elastic/apm-agent-dotnet/pull/1198 to fix it.

PR #1198, opened 2021-02-21
Add TargetFramework NET5.0 to Elastic.Apm.AspNetCore & co

gregkalapos:
Fixes https://github.com/elastic/apm-agent-dotnet/issues/1194. 

Also turning on all ASP.NET Core tests for .NET 5 as well.

### DIFF
--- a/test/Elastic.Apm.AspNetCore.Tests/DistributedTracingAspNetCoreTests.cs
+++ b/test/Elastic.Apm.AspNetCore.Tests/DistributedTracingAspNetCoreTests.cs
@@ -2,6 +2,7 @@
 // Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information
 
+using System.Diagnostics;
 using System.Linq;
 using System.Net.Http;
 using System.Reflection;
@@ -144,7 +145,7 @@ namespace Elastic.Apm.AspNetCore.Tests
 		{
 			_agent1.ConfigStore.CurrentSnapshot = new MockConfigSnapshot(transactionSampleRate: "0");
 
-			await ExecuteAndCheckDistributedCall();
+			await ExecuteAndCheckDistributedCall(true);
 
 			// Since the trace is non-sampled both transactions should me marked as not sampled
 			_payloadSender1.FirstTransaction.IsSampled.Should().BeFalse();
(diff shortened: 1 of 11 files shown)


stefannikolei:
No Sunday for you?

test/Elastic.Apm.AspNetCore.Tests/DistributedTracingAspNetCoreTests.cs
```diff
@@ -191,8 +192,25 @@ public async Task DisposeAsync()
 			_agent2?.Dispose();
 		}
 
-		private async Task ExecuteAndCheckDistributedCall()
+		/// <summary>
+		/// Executes the distributed tracing call between the 2 services.
+		/// </summary>
+		/// <param name="PreventStartingActivity">If true, we'll create an actvity on .NET 5</param>
+		/// <returns></returns>
+		private async Task ExecuteAndCheckDistributedCall(bool PreventStartingActivity = false)
```
russcam, line 200: Would calling `PreventStartingActivity` something like `startActivityBeforeHttpCall` (and have boolean semantics that are the inverse of current impl) be a better description for what the parameter controls?

russcam (approved):
LGTM. Left a comment about parameter naming

test/Elastic.Apm.AspNetCore.Tests/DistributedTracingAspNetCoreTests.cs
gregkalapos: Sounds ok to me - I changed it. In all tests we create the activity, except in 1, that's why I called it `PreventStartingActivity`. But this can change later... I renamed it.

michaelbeckersgit:
Hi Greg,

I would need to have the version to support .Net 5. When will this change be released please?

stefannikolei:
> Hi Greg,
> 
> I would need to have the version to support .Net 5. When will this change be released please?

It works in a net5 app. You just need to pin your dependencies. Microsoft.extensions.abstractions.hosting and I think configuration to 5.x.x 

the only downside is that there are no GC metrics. But there is already a fix in the main branch

gregkalapos:
We plan to push out a patch soon containing both this fix and the other fix about GC metrics.

michaelbeckersgit:
Many thanks, works now fine with 1.8.1!

PR #1201, opened 2021-02-24
Move InternalsVisibleTo into project files

russcam:
This commit moves the usage of the InternalsVisibleTo into project files.

The exception is AspNetFullFrameworkSampleApp, which uses the older project file format.


gregkalapos (approved):
LGTM - love this ❤️

PR #1202, opened 2021-02-24
Update editorconfig

russcam:
This commit updates the rules in editorconfig; It removes warnings for syntax errors and includes configuration to underscore prefix
private fields.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -29,15 +29,22 @@ indent_size = 2
 # ---
 
 # ---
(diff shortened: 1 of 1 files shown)


gregkalapos (approved):
LGTM, 

we should somehow automate to keep this in-sync with the other .NET repositories (like [elasticsearch-net](https://github.com/elastic/elasticsearch-net/blob/master/.editorconfig))

PR #1203, opened 2021-02-24
output Activity id format before and after HTTP call

russcam:


gregkalapos (approved):
I'd vote for merging this. I'm doing some other work and it's not nice that this test causes red builds. Overall I feel it'd be better for us to just disable it for now.

russcam:
the `CollectGcMetrics` test is hanging, which is being investigated in #1218. Merging this in.

Issue #1204, opened 2021-02-24
Performance issue : allowing to capture stacktrace on exception even if it's disabled for regular process


NicolasReyDotNet:
**Is your feature request related to a problem? Please describe.**
I have an API route with a lot of business logic (with serveral Elasticsearch calls) that is monitored by APM and which is supposed to reply in 200ms or less (acceptance criteria).

With APM disabled (`"TransactionSampleRate" : 0`), my mean response time (based on 50 calls) is approx **100ms**
As soon as I monitor with APM (`"TransactionSampleRate" : 1`) with all parameters as default, the mean response time jumps to **430ms** (with exactly the same data)
I tried many parameters combinations accordind to the performance tagged settings : [All options settings](https://www.elastic.co/guide/en/apm/agent/dotnet/current/config-all-options-summary.html)

What I concluded, is, as I already read it in another thread, that the "problem" is linked to captured stacktraces.
Indeed I achieved the best performances disabling it (`"StackTraceLimit" : 0`) which is anyway useless for me for a regular call that process with no error.
![image](https://user-images.githubusercontent.com/37657851/109029955-a7274100-76c3-11eb-8095-2b278e8e1c79.png)
Now my mean response time is **180ms,** almost twice than without APM, but in accordance criteria. Great!

**BUT** this parameter seems also responsible for capturing stacktrace of uncaught throw exceptions, so when it occurs, I don't get the stacktrace either.
![image](https://user-images.githubusercontent.com/37657851/109029652-5879a700-76c3-11eb-96fa-49df7fb6c835.png)

**Describe the solution you'd like**
1) The solution I expect is using a dedicated parameters for capturing stacktraces of normal process and another one for uncaught exceptions (or the same one with all possibilities as enum do the job as well)

2) Or increase overall performance / speed so that I don't have to loose functionnalities because I need response time performance in my own application

**Describe alternatives you've considered**
1) I considered sampling request, but this has been rejected because of the loss of information + for the sampled transactions, the response time is beyond the limits (> 200ms as response time)

2) I also considered disabling stacktrace (`"StackTraceLimit" : 0`) but for uncaught exception it's a real problem

**Additional context**
Versions of Lib :
```
    <PackageReference Include="Elastic.Apm.NetCoreAll" Version="1.8.0" />
    <PackageReference Include="NEST" Version="7.11.1" />
```

[...]

Issue #1145, opened 2021-01-26
Allow Stack Trace configuration for Errors / Exceptions individually of Spans


dasMulli:
Currently it is possible to disable stack trace capture for spans by setting `StackTraceLimit` to `0`.
This helped us a lot before recent perf improvements in demystifying. However we still decided to leave it off b/c we trace with a high sample rate and capture lots of spans (mostly DB transactions but also custom ones).

This has the unfortunate effect that exceptions also no longer contain stack traces and investigating errors is a bit harder (still possibly by correlating logs with APM but more cumbersome).

It would be great if I could disable span stack trace collection independent of error / exception stack trace collection.

gregkalapos:
Hey @dasMulli,

there is also another setting called [SpanFramesMinDuration](https://www.elastic.co/guide/en/apm/agent/dotnet/current/config-stacktrace.html#config-span-frames-min-duration). If you set that to `0`  then that will only apply to spans and with `0` you'll disable stack trace capturing for every span. With that you should still have stack traces on errors.

Does that help?

gregkalapos:
I close this for now - I think we have a fairly ok workaround here. If not feel free to reopen and we can continue the discussion.

dasMulli:
Argh forgot to close it - working great in production for weeks

PR #1205, opened 2021-02-24
[Doc] Improve docs on stack trace capturing

gregkalapos:
Added a note on how to turn off stack trace capturing for spans, but keep it for errors. Also added a troubleshooting item which explains stack trace capturing overhead in a but more detail.

Inspired by https://github.com/elastic/apm-agent-dotnet/issues/1204 and https://github.com/elastic/apm-agent-dotnet/issues/1145.

### DIFF
--- a/docs/troubleshooting.asciidoc
+++ b/docs/troubleshooting.asciidoc
@@ -134,12 +134,10 @@ writing to standard output.
 [[agent-overhead]]
 === The agent causes too much overhead
 
-A good place to start is our <<config-all-options-summary>>. There are multiple settings there with the `Performance` keyword which help you to tweak the agent for your needs.
+A good place to start is [config-all-options-summary]. There are multiple settings with the `Performance` keyword which can help you tweak the agent for your needs.
 
-The most expensive operation in the agent is typically stack trace capturing. The agent, by default only captures stack traces for spans with a duration of 5ms or more, and with a limit of 50 stack frames.
+The most expensive operation in the agent is typically stack trace capturing. The agent, by default, only captures stack traces for spans with a duration of 5ms or more, and with a limit of 50 stack frames.
+If this is too much in your environment, consider disabling stack trace capturing either partially or entirely:
 
-If this is too much in your environment, the <<config-span-frames-min-duration>> and the <<config-stack-trace-limit>> are two configuration worth looking at.
-
-By setting <<config-stack-trace-limit>> to `0` you can disable stack trace capturing completely which in most applications reduces the agent overhead dramatically.
-
-If you would like to capture stack traces for errors but skip stack trace capturing for spans, you can set the <<config-span-frames-min-duration>> to `0` and leave the <<config-stack-trace-limit>> on default.
+- To disable stack trace capturing for spans, but continue to capture stack traces for errors, set the [config-span-frames-min-duration] to `0` and leave the [config-stack-trace-limit] on its default.
+- To disable stack trace capturing entirely –which in most applications reduces the agent overhead dramatically– set [config-stack-trace-limit] to `0`.
--- a/docs/troubleshooting.asciidoc
+++ b/docs/troubleshooting.asciidoc
@@ -129,3 +129,17 @@ set ELASTIC_APM_STARTUP_HOOKS_LOGGING=1
 and then running the application in a context where the environment variable will be visible. In setting this value,
 an `ElasticApmAgentStartupHook.log` file is written to in the directory containing the startup hook assembly, in addition to
 writing to standard output.
+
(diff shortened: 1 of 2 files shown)


docs/troubleshooting.asciidoc
```diff
@@ -128,4 +128,18 @@ set ELASTIC_APM_STARTUP_HOOKS_LOGGING=1
...
 an `ElasticApmAgentStartupHook.log` file is written to in the directory containing the startup hook assembly, in addition to
-writing to standard output.
\ No newline at end of file
+writing to standard output.
+
+[float]
+[[agent-overhead]]
+=== The agent causes too much overhead
+
+A good place to start is our <<config-all-options-summary>>. There are multiple settings there with the `Performance` keyword which help you to tweak the agent for your needs.
```
russcam, line 137:
```suggestion
A good place to start is <<config-all-options-summary>>. There are multiple settings with the `Performance` keyword which can help to tweak the agent for your needs.
```

docs/troubleshooting.asciidoc
```diff
@@ -128,4 +128,18 @@ set ELASTIC_APM_STARTUP_HOOKS_LOGGING=1
...
+
+The most expensive operation in the agent is typically stack trace capturing. The agent, by default only captures stack traces for spans with a duration of 5ms or more, and with a limit of 50 stack frames.
+
+If this is too much in your environment, the <<config-span-frames-min-duration>> and the <<config-stack-trace-limit>> are two configuration worth looking at.
```
russcam, line 141:
```suggestion
If this is too much in your environment, the <<config-span-frames-min-duration>> and the <<config-stack-trace-limit>> are two configuration settings worth looking at.
```

docs/troubleshooting.asciidoc
```diff
@@ -128,4 +128,18 @@ set ELASTIC_APM_STARTUP_HOOKS_LOGGING=1
...
+
+By setting <<config-stack-trace-limit>> to `0` you can disable stack trace capturing completely which in most applications reduces the agent overhead dramatically.
```
russcam, line 143:
```suggestion
By setting <<config-stack-trace-limit>> to `0`, stack trace capturing is completely disabled, which reduces the agent overhead dramatically for most applications.
```

docs/troubleshooting.asciidoc
```diff
@@ -128,4 +128,18 @@ set ELASTIC_APM_STARTUP_HOOKS_LOGGING=1
...
+
+If you would like to capture stack traces for errors but skip stack trace capturing for spans, you can set the <<config-span-frames-min-duration>> to `0` and leave the <<config-stack-trace-limit>> on default.
```
russcam, line 145:
```suggestion
To **capture stack traces only for errors** and not for spans, set the <<config-span-frames-min-duration>> to `0` and leave the <<config-stack-trace-limit>> as the default.
```

russcam (approved):
LGTM 👍 Left some small wording suggestions

docs/troubleshooting.asciidoc
```diff
@@ -128,4 +128,18 @@ set ELASTIC_APM_STARTUP_HOOKS_LOGGING=1
...
+writing to standard output.
+
+[float]
+[[agent-overhead]]
+=== The agent causes too much overhead
+
+A good place to start is our <<config-all-options-summary>>. There are multiple settings there with the `Performance` keyword which help you to tweak the agent for your needs.
+
+The most expensive operation in the agent is typically stack trace capturing. The agent, by default only captures stack traces for spans with a duration of 5ms or more, and with a limit of 50 stack frames.
+
```
bmorelli25, line 140:
```suggestion
```

The sentences before and after this line rely on each other. I'd remove this break. You also might consider bulleting out your suggestions and leading with the outcome instead of the configuration variable. In other words, something like this:

> **The agent causes too much overhead**
> A good place to start is [config-all-options-summary]. There are multiple settings there with the `Performance` keyword which can help you tweak the agent for your needs.
> 
> The most expensive operation in the agent is typically stack trace capturing. The agent, by default, only captures stack traces for spans with a duration of 5ms or more, and with a limit of 50 stack frames.
> If this is too much in your environment, consider disabling stack trace capturing either partially or entirely:
> 
> * To disable stack trace capturing for spans, but continue to capture stack traces for errors, set the [config-span-frames-min-duration] to `0` and leave the [config-stack-trace-limit] on its default.
> * To disable stack trace capturing entirely–which in most applications reduces the agent overhead dramatically–set [config-stack-trace-limit] to `0`.

bmorelli25 (approved):
A suggestion for your consideration 👍

Issue #1158, opened 2021-01-29
Instrumentation for MongoDB


AlexanderWert:
Include MongoDB instrumentation in the agent distribution.

Community package: https://github.com/vhatsura/elastic-apm-mongo 

Another package which generates diagnostic source events on top of its own diagnostic: https://github.com/jbogard/MongoDB.Driver.Core.Extensions.DiagnosticSources

russcam:
@vhatsura, would you be open to contributing https://github.com/vhatsura/elastic-apm-mongo for instrumenting MongoDB?

RascoApps:
We have this need as well. a generic DiagnosticSource listener by Name would be fantastic!

PR #1215, opened 2021-03-08
Add MongoDb support

gregkalapos:
Closes https://github.com/elastic/apm-agent-dotnet/issues/1158 

This is a port of https://github.com/vhatsura/elastic-apm-mongo

@vhatsura we are asked a lot about MongoDb, so we decided to make it officially supported. This PR is based on your work - that [repo](https://github.com/vhatsura/elastic-apm-mongo) is nice, so we'd like to move it into this repo and maintain it here. If you want, feel free to open a PR and we can take it from there, otherwise I'll just continue with this one.


src/Elastic.Apm.MongoDb/MongoDbDiagnosticsSubscriber.cs
```diff
@@ -0,0 +1,37 @@
...
+using Elastic.Apm;
+
+// ReSharper disable UnusedMember.Global
+
+namespace Elastic.Apm.Mongo
+{
+	/// <summary>
+	///     A subscriber to events from mongoDB driver diagnostic source.
+	/// </summary>
+	public class MongoDbDiagnosticsSubscriber : IDiagnosticsSubscriber
```
gregkalapos, line 14: We don't handle `enabled` and `recording` here. https://github.com/elastic/apm-agent-dotnet/pull/1161 is about to move that logic into a base class - I'll adapt this once https://github.com/elastic/apm-agent-dotnet/pull/1161 is ready to go.

docs/supported-technologies.asciidoc
```diff
@@ -94,8 +94,13 @@ We support automatic instrumentation for the following data access technologies.
 
 | Redis (StackExchange.Redis)
 | 2.0.495
-| A DB span is automatically created for each profiled redis command peformed by StackExchange.Redis 
+| A DB span is automatically created for each profiled redis command performed by StackExchange.Redis 
 | 1.8.0
+
+| MondoDb (MondoDB.Driver)
+| 2.4.4
+| A DB span is automatically created for each profiled call to MondoDB.
```
bmorelli25, line 102:
```suggestion
| A DB span is automatically created for each profiled call to MongoDB.
```

docs/setup.asciidoc
```diff
@@ -361,6 +366,25 @@ connection.UseElasticApm();
 A callback is registered with the `IConnectionMultiplexer` to provide a profiling session for each transaction and span that captures redis commands
 sent with `IConnectionMultiplexer`.
 
+[[setup-mongo-db]]
+=== MongoDB.Driver
+
+A prerequisite for auto instrumentation with [`MondoDb.Driver`] is to configure the `MongoClient` with `MongoDbEventSubscriber`:
+
+```
```
bmorelli25, line 374:
```suggestion
[source,csharp]
----
```

gregkalapos:
File headers are missing - I'll follow up with those.

src/Elastic.Apm.MongoDb/Elastic.Apm.MongoDb.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <AssemblyName>Elastic.Apm.MongoDb</AssemblyName>
+    <PackageId>Elastic.Apm.MongoDb</PackageId>
+    <Description>
+      Elastic APM integration with the MongoDB.Driver package.
+    </Description>
+    <IsPackable>true</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="MongoDB.Driver.Core" Version="2.4.4" />
```
russcam, line 15: Should this be set to a Version range, up to 3.0.0 initially? e.g. 

```
<PackageReference Include="MongoDB.Driver.Core" Version="[2.4.4,3.0)" />
```

test/Elastic.Apm.MongoDb.Tests/MongoApmTests.cs
```diff
@@ -0,0 +1,159 @@
...
+			configurationReaderMock.Setup(x => x.TransactionSampleRate)
+				.Returns(() => 1.0);
+			configurationReaderMock.Setup(x => x.TransactionMaxSpans)
+				.Returns(() => 50);
+			configurationReaderMock.Setup(x => x.Enabled)
+				.Returns(() => true);
+			configurationReaderMock.Setup(x => x.Recording)
+				.Returns(() => true);
+
+			var config = new AgentComponents(configurationReader: configurationReaderMock.Object,
```
russcam, line 34: Can be replaced with `TestAgentComponents` and a `MockConfigSnapshot`

russcam (changes requested):
I've had a first pass through and this looks good. I've left some small comments for consideration.

docs/setup.asciidoc
```diff
@@ -361,6 +366,25 @@ connection.UseElasticApm();
...
+settings.ClusterConfigurator = builder => builder.Subscribe(new MongoEventSubscriber());
+var mongoClient = new MongoClient(settings);
+```
+
+If the agent is activated with the `Elastic.Apm.NetCoreAll`, the agent will automatically capture calls to MongoDb on every active transaction.
+
+Otherwise you can manually activate auto instrumentation from the `Elastic.Apm.MongoDb` package by calling `Agent.Subscribe(new MongoDbDiagnosticsSubscriber());`.
+
+[float]
+==== Quick start
```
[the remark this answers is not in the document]
gregkalapos: Oh, this should be at the top of this section. I moved it.

src/Elastic.Apm.MongoDb/DiagnosticSource/MongoDbDiagnosticInitializer.cs
```diff
@@ -0,0 +1,35 @@
+using System;
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Elastic.Apm.MongoDb.DiagnosticSource
+{
+	internal sealed class MongoDbDiagnosticInitializer : IObserver<DiagnosticListener>, IDisposable
```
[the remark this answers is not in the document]
gregkalapos: Yes, agreed. Done.

src/Elastic.Apm.MongoDb/DiagnosticSource/MongoDiagnosticListener.cs
```diff
@@ -0,0 +1,135 @@
...
+		{
+			_apmAgent = apmAgent;
+			_logger = _apmAgent.Logger.Scoped(nameof(MongoDiagnosticListener));
+		}
+
+		public void OnNext(KeyValuePair<string, object> value)
+		{
+			switch (value.Key)
+			{
+				case Constants.Events.CommandStart when value.Value is EventPayload<CommandStartedEvent> payload &&
```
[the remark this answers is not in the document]
gregkalapos: Yes, good point. I added more logging. You had some comments below to log at other places as well - added logs to all those places.

src/Elastic.Apm.MongoDb/DiagnosticSource/MongoDiagnosticListener.cs
```diff
@@ -0,0 +1,135 @@
...
+		{
+			// do nothing because it's not necessary to handle such event from provider
+		}
+
+		private void HandleCommandStartEvent(CommandStartedEvent @event)
+		{
+			try
+			{
+				var transaction = _apmAgent.Tracer.CurrentTransaction;
+				var currentExecutionSegment = _apmAgent.Tracer.CurrentSpan ?? (IExecutionSegment)transaction;
```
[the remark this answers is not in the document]
gregkalapos: 👍 done

src/Elastic.Apm.MongoDb/Elastic.Apm.MongoDb.csproj
```diff
@@ -0,0 +1,22 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net461;netstandard2.0</TargetFrameworks>
```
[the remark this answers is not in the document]
gregkalapos: Hmm, `net461` was included in the original code, that's why it's here. I think we are good to start with `netstandard2.0`. That'll cover also newer Full Framework versions. `net461` is usually added specifically, because `netstandard2.0` doesn't work well on that version. Let's go with `netstandard2.0` only, and if someone needs `net461`, we can add it later.

src/Elastic.Apm.MongoDb/Elastic.Apm.MongoDb.csproj
gregkalapos: Why would we do that? That way we'd not work with newer versions even the code would work otherwise. This PR only depends on `MongoDB.Driver.Core.Events.IEventSubscriber`. I'd be optimistic and think that it's more likely that this will not change in later versions. So I'd take a more optimistic approach here, otherwise we'll always need to update these when new versions get released.

test/Elastic.Apm.MongoDb.Tests/MongoApmTests.cs
gregkalapos: 👍  done.

gregkalapos:
Thanks for the feedback - pushed to this, ready for another round.

src/Elastic.Apm.MongoDb/Elastic.Apm.MongoDb.csproj
russcam: It is not known at this point in time whether this will work for _all_ future versions of MongoDB.Driver.Core, so this is just a cautious suggestion. It would only need to be updated when a new major version is released, which is not very frequent.

docs/setup.asciidoc
```diff
@@ -361,6 +366,25 @@ connection.UseElasticApm();
...
+
+[source,csharp]
+----
+var settings = MongoClientSettings.FromConnectionString(mongoConnectionString);
+
+settings.ClusterConfigurator = builder => builder.Subscribe(new MongoEventSubscriber());
+var mongoClient = new MongoClient(settings);
+----
+
+If the agent is included by referencing the `Elastic.Apm.NetCoreAll` package, it will automatically capture calls to MongoDB on every active transaction.
```
russcam, line 385:
```suggestion
With the above configuration in place, If the agent is included by referencing the `Elastic.Apm.NetCoreAll` package, it will automatically capture calls to MongoDB on every active transaction.
```

I think this paragraph should be connected with the code sample, so that it can't be misinterpreted in isolation

src/Elastic.Apm.MongoDb/DiagnosticSource/MongoDiagnosticListener.cs
```diff
@@ -0,0 +1,142 @@
...
+				Logger.Trace()?.Log(nameof(HandleCommandStartEvent));
+				var currentExecutionSegment = ExecutionSegmentCommon.GetCurrentExecutionSegment(ApmAgent);
+				var span = currentExecutionSegment.StartSpan(
+					@event.CommandName,
+					ApiConstants.TypeDb,
+					"mongo");
+
+				if (!_processingQueries.TryAdd(@event.RequestId, span))
+				{
+					Logger.Debug()?.Log("Failed adding item to _processingQueries with RequestId: {RequestId}",
```
russcam, line 61: maybe `Trace()` level might be better suited?

src/Elastic.Apm.MongoDb/DiagnosticSource/MongoDiagnosticListener.cs
```diff
@@ -0,0 +1,142 @@
...
+
+		private void HandleCommandSucceededEvent(CommandSucceededEvent @event)
+		{
+			try
+			{
+				Logger.Trace()?.Log(nameof(HandleCommandSucceededEvent));
+
+				if (!_processingQueries.TryRemove(@event.RequestId, out var span))
+				{
+					Logger.Debug()?.Log("Failed removing item from _processingQueries for RequestId: {RequestId}", @event.RequestId);
```
russcam, line 104: maybe `Trace()` level might be better suited?

src/Elastic.Apm.MongoDb/DiagnosticSource/MongoDiagnosticListener.cs
```diff
@@ -0,0 +1,142 @@
...
+
+		private void HandleCommandFailedEvent(CommandFailedEvent @event)
+		{
+			try
+			{
+				Logger.Trace()?.Log(nameof(HandleCommandFailedEvent));
+
+				if (!_processingQueries.TryRemove(@event.RequestId, out var span))
+				{
+					Logger.Debug()?.Log("Failed removing item from _processingQueries for RequestId: {RequestId}", @event.RequestId);
```
russcam, line 126: maybe `Trace()` level might be better suited?

docs/setup.asciidoc
bmorelli25: Smart. What do you think about changing the wording just a hair?

```suggestion
Once the above configuration is in place, and if the agent is included by referencing the `Elastic.Apm.NetCoreAll` package, it will automatically capture calls to MongoDB on every active transaction.
```

russcam: LGTM, thanks @bmorelli25 😃

src/Elastic.Apm.MongoDb/Elastic.Apm.MongoDb.csproj
gregkalapos: Currently we don't do this with any of the supported libraries - so far the approach is to depend on the lowest possible version and not limit support. The main thinking behind this is that we usually only depend on a few basic APIs - in this case it's a single interface.

I feel we should be consistent here - either we purposely limit support for current versions and proactively watch new releases and check compatibility for all our supported libraries, or we keep the current approach. But mixing them does not feel good. If we limit it now, we should have some strategy to make sure once `3.0` is out we know about it and we release immediately. It'd be bad to wait for users to report.

If we move to this approach then we'll need to
- actively check for new versions
- when a new version comes out we should test
- update the version of the dependency and release a new version

In my opinion we'd take on a huge maintenance work and so far we haven't seen incompatibility issues in newer versions, so I feel the benefit would be limited.

I agree with your statement
> It is not known at this point in time whether this will work for all future versions 

But if we want to tackle this issue (which holds true for all our supported libraries) then first we need to have a system to proactively release for each new version of the supported library, otherwise we'd always be behind and users will ping us to update in order to support newer versions. Also, based on the experience I think in most of the cases we'll only need to update the version range, because changes that break our code are fairly rare.

docs/setup.asciidoc
gregkalapos: :) updated.

src/Elastic.Apm.MongoDb/DiagnosticSource/MongoDiagnosticListener.cs
gregkalapos: I was even thinking about `warning` :) In this case there won't be a span for the given call. But I don't have super strong opinion on this, I update it to `Trace()`. Same for the same suggestions to the same lines.

src/Elastic.Apm.MongoDb/Elastic.Apm.MongoDb.csproj
gregkalapos: We discussed this in the last sync-meeting: for now we keep the support open-ended and deal with this once this really causes a problem.

gregkalapos:
Rebased on master and added license headers.  @russcam I think this is ready to merge.

NOTICE.txt
```diff
@@ -0,0 +1,7 @@
+==========
+Notice for: elastic-apm-mongo
```
russcam, line 2:
```suggestion
Notice for: Elastic.Apm.MongoDb
```

NOTICE.txt
```diff
@@ -0,0 +1,7 @@
...
+----------
+Based on the elastic-apm-mongo project by Vadim Hatsura (@vhatsura),
+licensed under the Apache 2.0 License. https://github.com/vhatsura/elastic-apm-mongo
+
+<text of the Apache 2.0 License>
```
russcam, line 7: The NOTICE.txt should be added to the Elastic.Apm.MongoDb .csproj in order to pack it into the Nuget package distribution, similar to how the shared LICENSE file is packed in src/Directory.Build.Props

https://github.com/elastic/apm-agent-dotnet/blob/26ebf44b49446f1b5b2e79a6724bc8502d1ca5d2/src/Directory.Build.props#L33-L36

What's not clear to me is whether NOTICE.txt should be included in _all_ Nuget packages - @gtback? 

I _think_ we can also remove `<text of the Apache 2.0 License>`, based on [Apache's example NOTICE file](https://infra.apache.org/licensing-howto.html#example-notice). @gtback?

docs/setup.asciidoc
```diff
@@ -15,6 +15,7 @@ On .NET Core the agent also supports auto instrumentation without any code chang
 * <<setup-sqlclient>>
 * <<setup-stackexchange-redis>>
 * <<setup-azure-servicebus>>
+* <<setup-mongo-db>>
```
russcam, line 18: minor nit: we'll be adding more cloud integrations over time, so it might be better to put MongoDB integration before Azure ServiceBus, so that new cloud integrations are added after and grouped together.

docs/supported-technologies.asciidoc
russcam: I think this is good to commit

test/Elastic.Apm.Extensions.Hosting.Tests/Elastic - Backup.Apm.Extensions.Hosting.Tests.csproj
```diff
@@ -0,0 +1,35 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
russcam, line 1: Looking at the name of this file, I think it can be removed?

test/Elastic.Apm.MongoDb.Tests/LICENSE.txt
```diff
@@ -0,0 +1,2 @@
+The code in this directory is based on the elastic-apm-mongo project by Vadim Hatsura (@vhatsura),
```
russcam, line 1: I think this should be concatenated with the contents of LICENSE. An MSBuild target might be a good approach to handle this.

src/Elastic.Apm.MongoDb/LICENSE.txt
```diff
@@ -0,0 +1,2 @@
+The code in this directory is based on the elastic-apm-mongo project by Vadim Hatsura (@vhatsura),
+licensed under the Apache 2.0 License. https://github.com/vhatsura/elastic-apm-mongo
```
russcam, line 2: Might be good to use the file name LICENSE, for consistency with the existing LICENSE in solution root.

I think this content should be concatenated with the contents of `$(SolutionRoot)/LICENSE` and included in the nuget package. An MSBuild target might be a good approach to handle this.

Then, src/Directory.Build.Props needs to be changed to **not** include `$(SolutionRoot)/LICENSE` in the nuget package when a LICENSE file exists in the current project root.

NOTICE.txt
russcam: minor nitpick: Might be good to use the file name NOTICE, for consistency with the existing LICENSE in solution root.

russcam (commented):
Thanks for your effort on this, @gregkalapos! I think it's super close- I left some comments mainly around license and notice

test/Elastic.Apm.MongoDb.Tests/LICENSE.txt
gtback: I agree. This would improve clarity around which terms apply to which source files!

NOTICE.txt
gtback: > What's not clear to me is whether NOTICE.txt should be included in _all_ Nuget packages - @gtback?

I'm not sure exactly how Nuget works, but I think you mean all of the various packages [here](https://www.nuget.org/packages?q=elastic.apm) (minus the ones we don't publish, of course). Any package which includes the MongoDB code should definitely have it, but if there are any packages that only cause the MongoDB code to be separately downloaded, I don't think we need it (but it wouldn't hurt). Does that make sense?
 
> I _think_ we can also remove `<text of the Apache 2.0 License>`, based on [Apache's example NOTICE file](https://infra.apache.org/licensing-howto.html#example-notice). @gtback?

Our policy is to include the full text of any licenses of any dependencies listed in NOTICE.txt. I meant to put the actual full text of the license here, not the literal text `<text of the Apache 2.0 License>`; sorry for any confusion.

That said, given that the entire project is already Apache-2.0 licensed, I think we could get rid of this without causing any problems (we don't need the same text in both `LICENSE.txt` and `NOTICE.txt`

> minor nitpick: Might be good to use the file name NOTICE, for consistency with the existing LICENSE in solution root.

Either is fine. I think consistency within a repo is more important than consistency between repos.

.ci/windows/dotnet.bat
```diff
@@ -6,6 +6,7 @@ dotnet sln remove src/Elastic.Apm.AspNetFullFramework/Elastic.Apm.AspNetFullFram
 dotnet sln remove test/Elastic.Apm.AspNetFullFramework.Tests/Elastic.Apm.AspNetFullFramework.Tests.csproj
 dotnet sln remove test/Elastic.Apm.SqlClient.Tests/Elastic.Apm.SqlClient.Tests.csproj
 dotnet sln remove test/Elastic.Apm.EntityFramework6.Tests/Elastic.Apm.EntityFramework6.Tests.csproj
+dotnet sln remove test/Elastic.Apm.MongoDb.Tests/Elastic.Apm.MongoDb.Tests.csproj
```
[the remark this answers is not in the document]
gregkalapos: Because it needs to start a container and our CI is not prepared for that on Windows.  Similar to `Elastic.Apm.SqlClient.Tests`.

docs/setup.asciidoc
```diff
@@ -403,6 +409,26 @@ A new span is created when there is a current transaction, and when
...
+[source,csharp]
+----
+var settings = MongoClientSettings.FromConnectionString(mongoConnectionString);
+
+settings.ClusterConfigurator = builder => builder.Subscribe(new MongoEventSubscriber());
+var mongoClient = new MongoClient(settings);
+----
+
+Once the above configuration is in place, and if the agent is included by referencing the `Elastic.Apm.NetCoreAll` package, it will automatically capture calls to MongoDB on every active transaction.
+Otherwise, you can manually activate auto instrumentation from the `Elastic.Apm.MongoDb` package by calling `Agent.Subscribe(new MongoDbDiagnosticsSubscriber());`.
```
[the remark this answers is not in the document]
gregkalapos: Done

src/Elastic.Apm.MongoDb/Constants.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+	internal static class Constants
+	{
+		internal const string MongoDiagnosticName = "MongoDB.Driver";
+
+		public static class Events
+		{
+			internal const string CommandStart = "CommandStart";
+			internal const string CommandEnd = "CommandEnd";
+			internal const string CommandFail = "CommandFail";
```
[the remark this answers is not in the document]
gregkalapos: Done

NOTICE.txt
gregkalapos: Done

NOTICE.txt
gregkalapos: Thanks for the input on this.

1. > Might be good to use the file name NOTICE

Agreed - renamed it and merged it with what we already had.

2. Regarding the `<text of the Apache 2.0 License>`: 

I removed it and I did not add the Apache 2.0 text itself.

3. > What's not clear to me is whether NOTICE.txt should be included in all Nuget packages 

We already have one notice for `Elastic.Apm` and now we have one for `Elastic.Apm.MongoDb`. I think the best and easiest would be to include this in all our packages. 2 advantages I see here: 1) no matter which package people use, they'll be able to easily see our notices - otherwise they'd need to collect the info from multiple NuGet packages following dependencies, 2) it's easier for us to maintain (single file, we include the same in all packages).

So what I did is that I added the NOTICE file to `apm-agent-dotnet/src/Directory.Build.props` (same as our LICENSE file), so it'll be included in all our packages.

Any thoughts on this?

docs/setup.asciidoc
gregkalapos: I merged in the other one and fixed conflicts, so this should be ok.

test/Elastic.Apm.Extensions.Hosting.Tests/Elastic - Backup.Apm.Extensions.Hosting.Tests.csproj
gregkalapos: Right, removed it.

src/Elastic.Apm.MongoDb/LICENSE.txt
gregkalapos: Regarding file name: 👍 done.

Regarding the other part: I'm a bit confused here. We have our NOTICE file which says what 3rd party things we use and that's already included in the package. This file talks about `this directory` (which in the context of a NuGet package doesn't have a meaning), so the text sounds like this applies to the source code and not to the binary we generate.

I don't see why we need to have both NOTICE and LICENSE with basically telling the same, and furthermore if we need both then I think we need to adapt the text here.

My understanding was that we keep the LICENSE file here in the folder to communicate that this code is based on vhatsura/elastic-apm-mongo and we added the NOTICE flag to communicate that our binaries are inherited from code from vhatsura/elastic-apm-mongo.

@gtback @russcam what am I missing here? I'm really not a license person, so maybe I just misunderstand something.

test/Elastic.Apm.MongoDb.Tests/LICENSE.txt
gregkalapos: I think the comment below covers this topic, let's this discuss it in https://github.com/elastic/apm-agent-dotnet/pull/1215#discussion_r609254103

NOTICE.txt
gtback: > So what I did is that I added the NOTICE file to `apm-agent-dotnet/src/Directory.Build.props` (same as our LICENSE file), so it'll be included in all our packages.

I think this is fine.

.ci/windows/dotnet.bat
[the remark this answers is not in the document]
russcam: thanks for clarification 👍

NOTICE.txt
russcam: > We already have one notice for Elastic.Apm and now we have one for Elastic.Apm.MongoDb. I think the best and easiest would be to include this in all our packages. 2 advantages I see here: 1) no matter which package people use, they'll be able to easily see our notices - otherwise they'd need to collect the info from multiple NuGet packages following dependencies, 2) it's easier for us to maintain (single file, we include the same in all packages).

> So what I did is that I added the NOTICE file to apm-agent-dotnet/src/Directory.Build.props (same as our LICENSE file), so it'll be included in all our packages.

I am not a lawyer, so would cede to our policies here, but I don't believe that the NOTICE file distributed with a nuget package should include information pertaining to the licenses of derived works that are not part of the nuget package in question. For example,

- The NOTICE file distributed with the Elastic.Apm package should not contain information pertaining to the license of derivative works in Elastic.Apm.MongoDB package

Doing so seems like this would make it much harder for an end consumer of a package to understand what is and isn't relevant, from a compliance perspective. From this point of view, the NOTICE file distributed with a package should only contain information pertaining to the licenses of derived works that are part of the nuget package, whilst the NOTICE file in the root of the source should readily contain information pertaining to the licenses of derivative works in the repository in source form. It would be easier for us to include the same root NOTICE file in every package, and harder to pull out only the relevant bits for each respective package, but in doing the latter, the NOTICE aligns with the actual package contents.

russcam (commented):
Left some further comments re: licenses and notices

NOTICE.txt
gtback: Thinking about this more, I agree with @russcam . We generally try to follow the Apache Software Foundation's [practices](https://infra.apache.org/licensing-howto.html) in most cases, and they provide pretty good guidance here.

> The `LICENSE` and `NOTICE` files must exactly represent the contents of the distribution they reside in. Only components and resources that are actually included in a distribution have any bearing on the content of that distribution's `NOTICE` and `LICENSE`.

If it's easier, we can include `NOTICE` files in subdirectories of this repo, if those match up with binary distributions which are published on NuGet.

src/Elastic.Apm.MongoDb/LICENSE.txt
gtback: Yep, I meant to reply to this yesterday, but @russcam got it right. Per our policies,

> All product builds ... must export the full original license and copyright text of any dependencies where required by the dependency’s license in the NOTICE.txt file

Broadly speaking, `LICENSE` contains copyright and license information for the code in the repo (including code we've "vendored" into the repo), and `NOTICE` contains information about sources built into a distribution. This is a bit of a special case, in that the code **is** in the repo (so is included in `LICENSE`) **and** is part of the distribution, so we include the same information in both. Note that the ASF guidelines do not require (and in fact recommend against) putting the text of dependencies' licenses in NOTICE, but our practice is to do just that. Since we include the full copyright and license text in LICENSE as well, it ends up looking the same. 

---

> This file talks about `this directory` (which in the context of a NuGet package doesn't really have a meaning), so the text sounds like this applies to the source code and not to the binary we generate.

Good catch! In the case of binary distributions (NuGet packages), it probably makes more sense to say "The code for `Elastic.Apm.MongoDB` and `Elastic.Apm.MongoDb.Tests`..."

NOTICE.txt
gregkalapos: Ok - I moved the two `NOTICE` files into the corresponding projects (aka subdirectories)

We have 2 `NOTICE` files now:
- `Elastic.Apm` has one about `Newtonsoft.json` (`Ben.Demystifier` will be likely added later as well)
- `Elastic.Apm.MondoDb` this is the new one about `vhatsura/elastic-apm-mongo`

Both will be included in the corresponding NuGet package we generate and those packages will only contain notices about the relevant dependencies.

src/Elastic.Apm.MongoDb/LICENSE.txt
gregkalapos: Thanks for the input on this.

So, I implemented @russcam's suggestion with concatenating `LICENSE` files with an msbuild target. [Here](https://github.com/elastic/apm-agent-dotnet/pull/1215/commits/d4e5d6368289faa2af360b7c8926714b27e9268b) is the commit.

In summary:

- We have the main `LICENSE` file which is untouched and shipped with all NuGet packages (except next point)
- If there is a `LICENSE_LOCAL` file. (happy to rename if someone comes up with something better) in a specific project, then the main `LICENSE` file won't be added to the project and msbuild target will concatenate the main `LICENSE` file with the `LICENSE_LOCAL` and create the `LICENSE` file for the given project, which will be included in the generated package.
- Originally I wanted to only have a `LICENSE` file in the specific folders as well, but this would be harder to concatenate, because 1) we'd need to know if it was already concatenated 2) we'd need to detect if the main `LICENSE` file changed 3) we'd need to know if the local `LICENSE` file changed.

src/Elastic.Apm.MongoDb/LICENSE
```diff
@@ -1,2 +1,207 @@
-The code in this directory is based on the elastic-apm-mongo project by Vadim Hatsura (@vhatsura),
```
gregkalapos: This file is generated on every build.

src/Elastic.Apm.MongoDb/LICENSE.txt
russcam: This looks good; wonder if we can introduce a newline between each concatenated section, as I think it'd combine better? The license separator could also be part of the content between concatenated sections. Not a big deal though, and could be addressed later.

I was thinking that we could have a `LICENSE` file in the project and concatenate with the repo license into another file e.g. `COMBINED_LICENSE`, then pack this into the nuget package as `LICENSE`, but it seems that nuget's behaviour is buggy with extensionless files and renaming as part of packing (https://github.com/NuGet/Home/issues/6351#issuecomment-478903940), so `LICENSE_LOCAL` is pragmatic.

docs/supported-technologies.asciidoc
```diff
@@ -94,8 +94,13 @@ We support automatic instrumentation for the following data access technologies.
 
 | Redis (StackExchange.Redis)
 | 2.0.495+
-| A DB span is automatically created for each profiled redis command peformed by StackExchange.Redis 
+| A DB span is automatically created for each profiled redis command performed by StackExchange.Redis 
 | 1.8
+
+| MondoDb (MondoDB.Driver)
```
russcam, line 100:
```suggestion
| MongoDb (MongoDB.Driver)
```

gregkalapos: Fixed

src/Elastic.Apm.MongoDb/LICENSE.txt
gregkalapos: > wonder if we can introduce a newline between each concatenated section, as I think it'd combine better?

Good idea, I implemented this, now we have a newline between the sections.

src/Elastic.Apm.MongoDb/LICENSE
russcam: I think we should add src/Elastic.Apm.MongoDb/LICENSE to .gitignore as it flags as a change when running `build.bat pack`

russcam (approved):
I think there's just this one small change to address (https://github.com/elastic/apm-agent-dotnet/pull/1215#discussion_r591791553), otherwise LGTM 👍

src/Elastic.Apm.MongoDb/LICENSE
gregkalapos: I thought about it as well, but we do `WriteOnlyWhenDifferent="true"` and locally for me it does not flag it. Is this maybe a line ending issue again? 🤔 

Anyways, I leave it out now, I think if `LICENSE` and `LICENSE_LOCAL` doesn't change then this should not generate a git change. If we see that it happens then I agree it makes sense to ignore it - will do in a follow-up if needed.

docs/supported-technologies.asciidoc
gregkalapos: Done ✅

gtback:
Great to see this merged! 🎉

PR #1216, opened 2021-03-09
[Doc] Document clr.gc.time

gregkalapos:
Follow up from https://github.com/elastic/apm-agent-dotnet/pull/925.

PR #1219, opened 2021-03-10
[Doc] Fix GC time doc

gregkalapos:
In https://github.com/elastic/apm-agent-dotnet/pull/1216 yesterday I made a mistake - by looking at the code I realize we send gc times in ms and not in ticks.


bmorelli25 (approved):
🥈 (a milliSECOND place medal)

PR #1225, opened 2021-03-16
Add instrumentation for Azure Service Bus

### LINKED ISSUE #1157: [META 410] Instrumentation for Azure ServiceBus
Provide instrumentation for Azure ServiceBus.

Meta issue: elastic/apm#410
SPEC: elastic/apm#411

russcam:
This PR adds instrumentation for Azure Service Bus when an application is using `Microsoft.Azure.ServiceBus` 3.0.0+ or `Azure.Messaging.ServiceBus` 7.0.0+ nuget packages.

Two `IDiagnosticListener` implementations, one for `Microsoft.Azure.ServiceBus` and another for `Azure.Messaging.ServiceBus`,  create transactions and spans for received and sent messages:

A new transaction is created when

* one or more messages are received from a queue or topic subscription.
* a message is receive deferred from a queue or topic subscription.

A new span is created when there is a current transaction, and when

* one or more messages are sent to a queue or topic.
* one or more messages are scheduled to a queue or a topic.

A terrform template file is used to create a resource group,  Azure Service Bus namespace resource in the resource group, and set RBAC rules to allow the Service Principal that issues the creation access to the resources. The Service Principal credentials can are sourced from a `.credentials.json` file in the root of the repository for CI, and from an account authenticated with `az` for local development. A default location is set within the template, but all variables can be passed using [standard Terraform input variable conventions](https://www.terraform.io/docs/language/values/variables.html#using-input-variable-values).

Closes #1157


russcam:
An outstanding item is to set the distributed tracing data for receive/receive deferred messages. Messages are typically retrieved and sent in batches however, where each message may represent its own distributed trace. https://github.com/elastic/apm/issues/122 proposes introducing a `links` property to capture each. Both `Microsoft.Azure.ServiceBus` and `Azure.Messaging.ServiceBus` expose the linked traces.

russcam:
Going to rebase this against master, to pick up #1240 to run tests on CI

src/Elastic.Apm.Azure.ServiceBus/MicrosoftAzureServiceBusDiagnosticListener.cs
```diff
@@ -0,0 +1,240 @@
...
+			// transaction creation will create an activity, so use this as the key.
+			var activityId = Activity.Current.Id;
+
+			if (!_processingSegments.TryAdd(activityId, transaction))
+			{
+				Logger.Trace()?.Log(
+					"Could not add {Action} transaction {TransactionId} for activity {ActivityId} to tracked segments",
+					action,
+					transaction.Id,
+					activityId);
```
gregkalapos, line 115: If this happens we should also make sure we don't leak the transaction. My best idea so far is to just call `End()` on it - better than leaking, but far from perfect. I just realize we have the same e.g. in `HttpDiagnosticListenerImplBase`. 

```suggestion
					activityId);
					transaction.End();
```

src/Elastic.Apm.Azure.ServiceBus/MicrosoftAzureServiceBusDiagnosticListener.cs
```diff
@@ -0,0 +1,240 @@
...
+				{
+					Name = ServiceBus.SubType,
+					Resource = queueName is null ? ServiceBus.SubType : $"{ServiceBus.SubType}/{queueName}",
+					Type = ServiceBus.Type
+				}
+			};
+
+			if (!_processingSegments.TryAdd(activity.Id, span))
+			{
+				Logger.Trace()?.Log(
```
gregkalapos, line 178: Similarly to above, here we should do something with the span we just created.

gregkalapos (commented):
I had a first look, very nice 👍 

I added some comments below. On the central config part: I'm not sure it's worth the effort to make the new config runtime changeable right now in this PR - if it turns out to be more work, I think it's totally fine to do it in a follow up - just update the doc in that case.

src/Elastic.Apm.Azure.ServiceBus/ServiceBus.cs
```diff
@@ -0,0 +1,14 @@
+// Licensed to Elasticsearch B.V under
+// one or more agreements.
+// Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information
+
+namespace Elastic.Apm.Azure.ServiceBus
+{
+	internal static class ServiceBus
```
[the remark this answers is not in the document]
russcam: I agree with moving the `messaging` type to `ApiConstants` as it's a general type that will be needed for other messaging systems. The other constants are only relevant to Azure Service Bus though, so I think should stay in the only project in which they are used

src/Elastic.Apm.Azure.ServiceBus/AzureMessagingServiceBusDiagnosticListener.cs
```diff
@@ -0,0 +1,246 @@
...
+					transaction.Id,
+					activityId);
+			}
+		}
+
+		private bool MatchesIgnoreMessageQueues(string name)
+		{
+			if (name != null && _realAgent != null)
+			{
+				var matcher = WildcardMatcher.AnyMatch(_realAgent.ConfigStore.CurrentSnapshot.IgnoreMessageQueues, name);
```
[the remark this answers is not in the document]
russcam: I didn't try with central configuration. I was assuming that since this is part of the merged messaging spec, it should be part of the implementation. I'll add a test for it.

src/Elastic.Apm.Azure.ServiceBus/MicrosoftAzureServiceBusDiagnosticListener.cs
russcam: Hmm, I'm not sure ending the transaction is the right thing to do, as it'll end up being sent to APM server and Elasticsearch with details that aren't a true reflection of the operation that it tried to instrument. 

This seems like an awkward issue to resolve for what is hopefully an edge case. Should we consider introducing something like an `Abandon()` method, that unsets it as the current transaction, and does't send to APM? Could this result in spans being sent to APM server though that are linked to this transaction? This seems unlikely because the service bus method that results in the diagnostic event is either synchronous or likely awaited.

My inclination is to leave this as is in the PR and to discuss more generally how we want to tackle this kind of issue. Sound good?

src/Elastic.Apm.Azure.ServiceBus/MicrosoftAzureServiceBusDiagnosticListener.cs
```diff
@@ -0,0 +1,240 @@
...
+		private readonly PropertyFetcherCollection _receiveDeferredProperties = new PropertyFetcherCollection { "Entity", "Endpoint", "Status" };
+		private readonly PropertyFetcher _exceptionProperty = new PropertyFetcher("Exception");
+		private readonly Service _service;
+
+		public override string Name { get; } = "Microsoft.Azure.ServiceBus";
+
+		public MicrosoftAzureServiceBusDiagnosticListener(IApmAgent agent) : base(agent)
+		{
+			_realAgent = agent as ApmAgent;
+			_service = Service.GetDefaultService(agent.ConfigurationReader, agent.Logger);
```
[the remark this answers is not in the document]
russcam: A new `Service` needs to be created to assign the `Framework` to. What should the values for `Service.Name` and `Service.Version` be? Should they simply be `null`?

src/Elastic.Apm.Azure.ServiceBus/MicrosoftAzureServiceBusDiagnosticListener.cs
russcam: See https://github.com/elastic/apm-agent-dotnet/pull/1225#discussion_r603722336

src/Elastic.Apm.Azure.ServiceBus/MicrosoftAzureServiceBusDiagnosticListener.cs
gregkalapos: > for what is hopefully an edge case.

Definitely an edge case - we have similar logging in other parts of the code, I have never seen those in any apps so far.

>  Should we consider introducing something like an `Abandon()` method

This sounds good to me, yes.

> My inclination is to leave this as is in the PR and to discuss more generally how we want to tackle this kind of issue. Sound good?

Agreed 👍

PR #1227, opened 2021-03-17
agent-zip: use assembly-rewriter during agent-zip creation to not depend on Newtonsoft.Json.dll

gregkalapos:
With this PR we get rid of the `Newtonsoft.Json.dll` dependency in the agent-zip for zero code attachment.

Some comments: I wanted to merge `Newtonsoft.Json.dll` instead of just renaming it, but I ran into this error:

```
[2021-03-17T18:53:36.74+01:00][Repack][Info   ]Adding assembly for merge: ...\build\output\ElasticApmAgent\5.0.0\Elastic.Apm.dll
[2021-03-17T18:53:36.88+01:00][Repack][Info   ]Adding assembly for merge:  ...\build\output\ElasticApmAgent\5.0.0\ApmNewtonsoft.Json.dll
[2021-03-17T18:53:36.92+01:00][Repack][Info   ]Processing references
Mono.Cecil.AssemblyResolutionException: Failed to resolve assembly: 'IKVM.Runtime, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'
   at Mono.Cecil.BaseAssemblyResolver.Resolve(AssemblyNameReference name, ReaderParameters parameters)
   at Mono.Cecil.BaseAssemblyResolver.Resolve(String fullName, ReaderParameters parameters)
   at Mono.Cecil.BaseAssemblyResolver.Resolve(String fullName)
   at ILRepacking.IKVMLineIndexer.PostRepackReferences()
   at ILRepacking.Steps.ReferencesRepackStep.Perform()
   at ILRepacking.ILRepack.Repack()
   at AssemblyRewriter.Program.Run(Options options) in /home/runner/work/assembly-rewriter/assembly-rewriter/src/assembly-rewriter/Program.cs:line 69
:: agent-zip: Failed! Process exited with '2' "assembly-rewriter -i  ...\output\ElasticApmAgent\5.0.0\Elastic.Apm.dll -o  ...\build\output\ElasticApmAgent\5.0.0\Elastic.Apm.dll -i  ...\build\output\ElasticApmAgent\5.0.0\Newtonsoft.Json.dll -o  ...\build\output\ElasticApmAgent\5.0.0\ApmNewtonsoft.Json.dll -m" (24.75 s)
```

Both our F# scripts and just running something like this ends up with the same issue: `assembly-rewriter -i Elastic.Apm.dll -o NElastic.Apm.dll -i Newtonsoft.Json.dll -o NNewtonsoft.Json.dll -m`

Merging into a single assembly is not strictly needed, but if there are ideas on how to fix it I can follow up on this.

### DIFF
--- a/build/scripts/Build.fs
+++ b/build/scripts/Build.fs
@@ -15,9 +15,7 @@ open Buildalyzer
 open Fake.Core
 open Fake.DotNet
 open Fake.IO
-open Fake.IO
 open Fake.IO.Globbing.Operators
-open Fake.SystemHelper
 open Fake.SystemHelper
 open Tooling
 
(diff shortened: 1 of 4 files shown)


gregkalapos:
We decided to go with https://github.com/elastic/apm-agent-dotnet/pull/1241

The advantage of https://github.com/elastic/apm-agent-dotnet/pull/1241 is that it applies to every deployment model, not only to startup hook, plus we don't depend on additional tools to internalize the dependency (turns out we ran into sum issues e.g. on .NET 5) - therefore we just internalize the source code.

PR #1228, opened 2021-03-18
Use current activity if there is one

russcam:
This commit fixes a bug whereby `Activity.Current` should be used
to set the TraceId if

1. it's not null
2. it's using W3C Id format
3. activity are not being ignored

### DIFF
--- a/src/Elastic.Apm/Model/Transaction.cs
+++ b/src/Elastic.Apm/Model/Transaction.cs
@@ -29,6 +29,8 @@ namespace Elastic.Apm.Model
 		private readonly IPayloadSender _sender;
 		private readonly string _traceState;
 
+		// Flags whether the agent created a new Activity for this transaction.
+		// If there is an existing Activity using W3C format, no Activity will be created and this remains false.
 		private bool _createdActivity;
 
 		// This constructor is meant for serialization
@@ -171,7 +173,7 @@ namespace Elastic.Apm.Model
 			}
 
 			// Also mark the sampling decision on the Activity
-			if (IsSampled && _activity != null && !ignoreActivity && _createdActivity)
+			if (IsSampled && _activity != null && !ignoreActivity)
 				_activity.ActivityTraceFlags |= ActivityTraceFlags.Recorded;
 
 			SampleRate = IsSampled ? sampler.Rate : 0;
--- a/src/Elastic.Apm/Model/Transaction.cs
+++ b/src/Elastic.Apm/Model/Transaction.cs
@@ -29,10 +29,6 @@ namespace Elastic.Apm.Model
 		private readonly IPayloadSender _sender;
 		private readonly string _traceState;
 
-		// Flags whether the agent created a new Activity for this transaction.
-		// If there is an existing Activity using W3C format, no Activity will be created and this remains false.
-		private bool _createdActivity;
-
 		// This constructor is meant for serialization
 		[JsonConstructor]
 		private Transaction(Context context, string name, string type, double duration, long timestamp, string id, string traceId, string parentId,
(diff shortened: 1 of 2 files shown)


russcam:
This is needed for #1225

src/Elastic.Apm/Model/Transaction.cs
```diff
@@ -25,12 +25,12 @@ internal class Transaction : ITransaction
 		private readonly IApmServerInfo _apmServerInfo;
 		private readonly Lazy<Context> _context = new Lazy<Context>();
 		private readonly ICurrentExecutionSegmentsContainer _currentExecutionSegmentsContainer;
-
 		private readonly IApmLogger _logger;
 		private readonly IPayloadSender _sender;
-
 		private readonly string _traceState;
 
+		private bool _createdActivity;
```
gregkalapos, line 32: Could we add a short comment on this one? Something like:
```
Flags whether the agent fired a new `Activity` for this transaction or not.
If there is an already existing transaction, no new transaction will be created and this bool remains false.
```

I'm saying this because [this part in the `ctor`](https://github.com/elastic/apm-agent-dotnet/pull/1228/files#diff-16bfd2f9d441e32ed34c366af81982b949e3857e24ef2abc53e89384b1538344R123-R131) took me some time to understand. With such comment hopefully it'll be easier to understand that part of the code.

src/Elastic.Apm/Model/Transaction.cs
```diff
@@ -153,7 +171,7 @@ internal Transaction(ApmAgent agent, string name, string type)
 			}
 
 			// Also mark the sampling decision on the Activity
-			if (IsSampled && _activity != null && !ignoreActivity)
+			if (IsSampled && _activity != null && !ignoreActivity && _createdActivity)
```
gregkalapos, line 174: One scenario which can be relevant here is when we reuse the existing Activity (so `_createdActivity` is `false`), but the recorded flag on that activity is not set. That flag is propagated for downstream services - technically it's not an issue, because we set the `traceparent` flag for the outgoing HTTP based on our `IsSampled` flag, nevertheless I think in that case we still should set the `Recorded` flag for the existing activity.

For doing so, I'm not sure it's enough to just remove the `_createdActivity` check, we should also nullcheck the activity and check if it's id format is `w3c`.

gregkalapos (commented):
Overall it looks good to me. I left 1 minor thing to add a comment and another one (that's the more important) about setting the recorded flag on the already existing `Activity`.

Jenkins also showed failing gRPC tests, I looked into that - I found the problem within `WebRequestTransactionCreator.cs`, I opened a PR against your branch: https://github.com/russcam/apm-agent-dotnet/pull/1 - I think that'll fix the gRPC problem.

src/Elastic.Apm/Model/Transaction.cs
russcam: > I think in that case we still should set the Recorded flag for the existing activity.

Sounds good to me, I'll update with this change.

> For doing so, I'm not sure it's enough to just remove the _createdActivity check, we should also nullcheck the activity and check if it's id format is w3c.

Removing the check is good enough, as `_activity` is only set to `Activity.Current` when not null and using W3C id format  (and there's a null check for `_activity` already)

src/Elastic.Apm/Model/Transaction.cs
russcam: done 👍

src/Elastic.Apm/Model/Transaction.cs
russcam: done 👍

russcam:
#1235 has made me think about whether we should create a new activity or not, based on there being an existing activity that meets the criteria, vs. always creating a new activity, but reusing the traceid and tracestate of the existing activity, if it meets criteria. I'm thinking about this from a user's usability perspective - I don't have a good handle on what a user might want to do with the current Activity inside of an APM transaction, and whether they might need to know if a new activity has been started or not. What are your thoughts, @gregkalapos?

gregkalapos:
> 
> #1235 has made me think about whether we should create a new activity or not, based on there being an existing activity that meets the criteria, vs. always creating a new activity, but reusing the traceid and tracestate of the existing activity, if it meets criteria. I'm thinking about this from a user's usability perspective - I don't have a good handle on what a user might want to do with the current Activity inside of an APM transaction, and whether they might need to know if a new activity has been started or not. What are your thoughts, @gregkalapos?

That's a good point. I feel the most important thing is that `Activity.Current.TraceId` should be in-sync with the traceid we use, that was the main motivation behind messing with `Activity` during transaction creation. If we'd always create an activity, that'll still be true an it'd be consistent - so the agent would always fires an activity on transaction creation.

> always creating a new activity, but reusing the traceid and tracestate of the existing activity

If there is already an activity `A` and we create a new activity `B`, then traceid and tracestate will be the same on both I think, because traceid and tracestate flow to child activities.

A related thing that comes to my mind:
The other thing that happens is that the `Id` of our transaction is the same as the `spanId` of the `Activity` the agent fires. We were already asked why our transaction id is different from the `Activity.SpanId` when there is an already existing `Activity`. Here I think if we reuse `Activity.SpanId` we should only do that for the `Activity` that we fire in the agent. Reason is that the transaction is different from any other `Activity`, specifically its duration may be completely different.

So, in sum: I think always creating an activity won't change a lot here, and it's be more consistent.

Regardless of this, I'd keep the code from #1235 where we walk the `Activity` chain to find the right one for gRPC info.

gregkalapos (approved):
LGTM 👍

russcam:
> So, in sum: I think always creating an activity won't change a lot here, and it's be more consistent.

I've just committed eb4f85b79b2b7b942d60a54c7a7b1256ca3d4343 to always create and start a new Activity, if we're not ignoring them. 

This commit also tries to set the ParentId of the created Activity to the `DistributedTracingData`, if passed; this will overwrite the `ParentSpanId` and `TraceId` of the created activity, that may have been set from `Activity.Current` if it was not null when the created activity was started. This _seems_ like the right thing to do, in order to keep explicitly passed tracing data in sync for the Transaction and Activity.

russcam:
Would you mind taking this another look, @gregkalapos?

PR #1230, opened 2021-03-18
Allow access to current configuration snapshot

glucaci:
This is proposal how to access the current configuration snapshot from `IApmAgent`

### DIFF
--- a/src/Elastic.Apm/Agent.cs
+++ b/src/Elastic.Apm/Agent.cs
@@ -14,6 +14,8 @@ namespace Elastic.Apm
 {
 	public interface IApmAgent
 	{
+		IConfigSnapshotProvider ConfigurationSnapshot { get; }
+
 		IConfigurationReader ConfigurationReader { get; }
 
 		IApmLogger Logger { get; }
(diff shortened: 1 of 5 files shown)


glucaci:
Maybe will be a good idea to mark `IApmAgent.ConfigurationReader` as `Obsolete` and to be removed in `2.x`. It's confusing to have 2 properties for accessing the configuration.

Something like:
```csharp
[Obsolete("Use IApmAgent.ConfigurationSnapshot, this will be removed in 2.x")]
IConfigurationReader ConfigurationReader { get; }
```

src/Elastic.Apm/Agent.cs
```diff
@@ -14,6 +14,8 @@ namespace Elastic.Apm
 {
 	public interface IApmAgent
 	{
+		IConfigSnapshotProvider ConfigurationSnapshot { get; }
```
glucaci, line 17: Maybe ConfigurationProvider will be better name

russcam:
Hi @glucaci, 

Thanks for your interest in improving the agent! Exposing the _current_ configuration snapshot is something that we're looking to do but we're not yet decided on exactly what this should look like. I've opened https://github.com/elastic/apm-agent-dotnet/issues/1244 to discuss first, so that we have an agreed path forward before opening a PR to add the feature. I'm going to close this PR for now.

PR #1235, opened 2021-03-21
[Temp] gRPC fix with Activity reuse

gregkalapos:
Test to run https://github.com/russcam/apm-agent-dotnet/pull/1 in CI.

No need to merge.


russcam:
Thanks @gregkalapos, I've incorporated this into #1228

PR #1236, opened 2021-03-22
Allow access to MetricSet and improve memory allocation on MetricSample

glucaci:
For users to send custom metrics, must have access to `MetricSet` as in #322 mentioned.
In the same time I did an optimization on `MetricSample`.

I'm working o a prototype to integrate `IPayloadSender.QueueMetrics` with `Metrics.NET` library

### DIFF
--- a/src/Elastic.Apm/Api/MetricSample.cs
+++ b/src/Elastic.Apm/Api/MetricSample.cs
@@ -10,13 +10,17 @@ namespace Elastic.Apm.Api
 	/// <summary>
 	/// A single metric sample.
 	/// </summary>
-	public class MetricSample
+	public readonly struct MetricSample
 	{
 		public MetricSample(string key, double value)
-			=> KeyValue = new KeyValuePair<string, double>(key, value);
+		{
+			Key = key;
+			Value = value;
+		}
 
-		internal KeyValuePair<string, double> KeyValue { get; }
+		internal string Key { get; }
+		internal double Value { get; }
 
-		public override string ToString() => new ToStringBuilder(nameof(MetricSample)) { { KeyValue.Key, KeyValue.Value } }.ToString();
+		public override string ToString() => new ToStringBuilder(nameof(MetricSample)) { { Key, Value } }.ToString();
 	}
 }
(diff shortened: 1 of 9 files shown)


glucaci:
If this is ok to merge I would like to start with `tags` support,
I saw that is supported on the server schema https://github.com/elastic/apm-server/blob/7cf56b584c9ff94153018afd1758f27737628873/docs/spec/metricsets/metricset.json#L47

glucaci:
@russcam can we merge this?
Thanks!

gregkalapos:
@glucaci let me try to give an overview on what we had in mind.

We'd like to have something similar across the Elastic APM Agents. One metrics library we wanted to have support for across agents is Prometheus: https://github.com/elastic/apm-agent-dotnet/issues/1127 - at the moment we don't actively work on that in .NET.

The current metrics API was internal on purpose: we were not sure we'd like to have it public in its current form. If we make anything public, likely we'd prefer to have an internal sync with other Elastic APM agent devs to make sure that if we make this public then the API is somewhat similar across agents and we don't end up breaking it later.

So what I want to say with this: we may need to think about this a bit longer.

glucaci:
Thanks and I understand your point of view. I thought that this https://github.com/elastic/apm-agent-dotnet/issues/322#issuecomment-507646456 was the final decision.

On the other hand is like a buggy API on `IPayloadSender` because any custom implementation of `IMetricsSet` will not work and this change would be a quick win for `1.x` version.

In my opinion this will be a fix for anyone how wants to use custom metrics with `1.x` and could be improved with a braking change on a `2.x` version.

russcam:
@glucaci thank you for your interest in improving the agent regarding metric collection! 

The `MetricSample` change from `class` to `readonly struct` looks like a good one to make, but we would need to make it in a 2.x version since it would be a breaking change. With the change to `MetricSet`, I think this is something worth discussing more on #322 before committing to a PR, so I'm going to close this for now.

glucaci:
Hi @russcam, I'm not sure that `readonly struct` was a problem, I could revert it because indeed it's a braking change.

The main point of this PR is to make `MetricSet` public which is a must that `IPayloadSender.QueueMetrics` works. (at least in 1.x)

PR #1242, opened 2021-03-24
Support GraphQL Activity

### LINKED ISSUE #1184: Flexibility for replacing the AspNetCoreDiagnosticListener
Hi,

We are working on a prototype to integrate into APM a GraphQL request using `HotChocolate.AspNetCore`. 

For this to look more like GraphQL and not REST we would like to replace the `AspNetCoreDiagnosticListener` completely because we want to control the `Result` of the main transaction. (e.g. to don't have as result `HTTP2xx` or `HTTP5xx` but `Success` or `Failed`).

At the moment this is not possible because the `Result` is set on `HttpRequestIn.Stop` event.

I would suggest:

1. Split `AspNetCoreDiagnosticListener` into:
- `AspNetCoreHttpDiagnosticListener` which hast the `Http` stack events `HttpRequestIn.Start` and `HttpRequestIn.Stop`
- `AspNetCoreHostingDiagnosticListener` which has the rest of events which are useful in any case,
2. Make `TraceContext` or maybe `WebRequestTransactionCreator` public, to can use the useful implementation of parsing and creating/filling transaction. 

... or to have a way to hook in `StopTransaction` and provide some custom `Result` and `Outcome`

I will be glad to make a PR for this.

Thanks!

glucaci:
As we discussed [here](https://github.com/elastic/apm-agent-dotnet/issues/1184) I added a posibility for any GraphQL library to can overwrite the `Name`, `Result` and `Outcome` of the `AspNetCore` `Transaction` in the same way as `Grpc` is implemented.

Because there is no common way of doing it, I've defined that:
- `Activity.OperationName` should be `GraphQL`
- `Activity.Tag` `graphql.operation` would be `Transaction.Name`
- `Activity.Tag` `graphql.result` would be `Transaction.Result`
- `Activity.Tag` `graphql.outcome` would be `Transaction.Outcome`

This will resolve #1184

### DIFF
--- a/src/Elastic.Apm.AspNetCore/WebRequestTransactionCreator.cs
+++ b/src/Elastic.Apm.AspNetCore/WebRequestTransactionCreator.cs
@@ -176,8 +176,6 @@ namespace Elastic.Apm.AspNetCore
 		{
 			if (transaction == null) return;
 
-			var grpcCallInfo = CollectGrpcInfo();
-
 			try
 			{
 				if (!transaction.HasCustomName)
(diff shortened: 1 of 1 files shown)


glucaci:
Using `Activity` it's not possible.

PR #1243, opened 2021-03-25
Change NetCoreFact to NetCoreAndNet5Fact

gregkalapos:
Some of our HTTP tests were .NET Core only. I just realized that it makes sense to also run those on .NET 5. Therefore I adapted the `FactAttribute` so those run on .NET Core and on .NET 5 as well.


russcam (approved):
LGTM 👍

PR #1247, opened 2021-03-26
Azure storage integration

### LINKED ISSUE #1155: [META 410] Instrumentation for Azure Storage
Provide instrumentation for Azure Storage.

Meta issue: elastic/apm#410
SPEC: elastic/apm#411

### LINKED ISSUE #1156: [META 410] Instrumentation for Azure Queue
Provide instrumentation for Azure Queue.

Meta issue: elastic/apm#410
SPEC: elastic/apm#411

russcam:
This PR provides integrations for Azure storage, using the newer version 12 libraries.

The version 12 libraries is still relatively new, and I suspect that many folks are still using the previous library versions. The new libraries have emit Diagnostic source events for every operation, which is how this integration creates transactions and spans. 

The previous libraries do not appear to emit diagnostic events, so would require inspecting HTTP calls to ascertain if the request is to a known service. This could be done as part of this PR, or could be done in a follow up PR, if we agree to support the previous major library versions (different nuget packages).

Closes #1156 
Closes #1155


russcam:
Opened as a draft PR for the moment, as this PR branch is branched from #1225 in order to use the terraform components.

russcam:
This is ready for review. It instruments the new Azure storage packages, with #1251 as a proposal to instrument the previous Azure storage packages.

src/Elastic.Apm.Azure.Storage/AzureBlobStorageDiagnosticListener.cs
```diff
@@ -0,0 +1,205 @@
...
+// See the LICENSE file in the project root for more information
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using Elastic.Apm.Api;
+using Elastic.Apm.DiagnosticListeners;
+using Elastic.Apm.Helpers;
```
gregkalapos, line 13:
```suggestion

```

src/Elastic.Apm.Azure.Storage/AzureFileShareStorageDiagnosticListener.cs
```diff
@@ -0,0 +1,186 @@
...
+		private readonly ConcurrentDictionary<string, IExecutionSegment> _processingSegments =
+			new ConcurrentDictionary<string, IExecutionSegment>();
+
+		public AzureFileShareStorageDiagnosticListener(IApmAgent agent) : base(agent) { }
+
+		public override string Name { get; } = "Azure.Storage.Files.Shares";
+
+		protected override void HandleOnNext(KeyValuePair<string, object> kv)
+		{
+			Logger.Trace()?.Log("Called with key: `{DiagnosticEventKey}'", kv.Key);
```
gregkalapos, line 37: A bit off-topic comment, but maybe we could move this (and the same logs from the other new listeners) into `DiagnosticListenerBase` - we do this logging in lots of types that derive from `DiagnosticListenerBase` - we could do this once in the base class.

Doesn't have to be in this PR - this was just a thought when I looked at this.

src/Elastic.Apm.Azure.Storage/AzureBlobStorageDiagnosticListener.cs
```diff
@@ -0,0 +1,205 @@
...
+		private void OnStart(KeyValuePair<string, object> kv, string action)
+		{
+			var currentSegment = ApmAgent.GetCurrentExecutionSegment();
+			if (currentSegment is null)
+			{
+				Logger.Trace()?.Log("No current transaction or span - exiting");
+				return;
+			}
+
+			if (!(kv.Value is Activity activity))
```
gregkalapos, line 112: `Activity.Id` according to the API could be `null` - it's very hard for me to imagine that, but to be on the safe side I think null-checking for `activity.Id` would be better, something like this:

```suggestion
			if (!(kv.Value is Activity activity) || string.IsNullOrEmpty(activity.Id))

```

We have the same in the `Elastic.Apm.Azure.ServiceBus` and at other places here (also in the `OnStop` and `OnException` methods).

Other nice thing about null-checking these is that it'll make Rider code analysis happy (currently it shows this as warning for me).

src/Elastic.Apm.Azure.Storage/AzureBlobStorageDiagnosticListener.cs
```diff
@@ -0,0 +1,205 @@
...
+			{
+				Logger.Trace()?.Log($"Key is {(kv.Key == null ? "null" : "an empty string")} - exiting");
+				return;
+			}
+
+			switch (kv.Key)
+			{
+				case "BlobContainerClient.Create.Start":
+				case "PageBlobClient.Create.Start":
+					OnStart(kv, "Create");
```
gregkalapos, line 50: Any reason we use capital letter here (`"Create"`)? This will end up being the `action` for the span and I think we use lower case at other places. Definitely won't break anything, just wondering if this is intentional.

There are some other instances in the PR as well.

test/Elastic.Apm.Azure.Storage.Tests/AzureBlobStorageDiagnosticListenerTests.cs
```diff
@@ -0,0 +1,248 @@
+﻿using System;
```
gregkalapos, line 1: There are some unused things in this file - could you please run code cleanup on this?

Update: also in `AzureFileShareStorageDiagnosticListenerTests.cs` and `AzureQueueStorageDiagnosticListenerTests,cs`, `BlobContainerScope.cs`.

gregkalapos (approved):
Nice 👍  I found some small things - other than those I think it's ready to merge.

src/Elastic.Apm.Azure.Storage/AzureFileShareStorageDiagnosticListener.cs
russcam: agreed, be good to move to the base class. Will open a follow up PR to do so.

src/Elastic.Apm.Azure.Storage/AzureBlobStorageDiagnosticListener.cs
russcam: [An `Activity` is assigned an id when it's started](https://github.com/dotnet/runtime/blob/6578f257e3be2e2144a65769706e981961f0130c/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs#L623-L670) and `Activity.Current` validates that the value passed to the setter is either a `null` value or a started `Activity` that hasn't finished [[1]](https://github.com/dotnet/runtime/blob/6578f257e3be2e2144a65769706e981961f0130c/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs#L1233-L1242) [[2]](https://github.com/dotnet/runtime/blob/79ae74f5ca5c8a6fe3a48935e85bd7374959c570/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.Current.net46.cs#L16-L26). 

Where the `Activity` is retrieved from the diagnostic event, I think we're safe to assume that the library has started the activity, and where the `Activity` is retrieved from `Activity.Current`, I think we can also assume that this is OK because it will either be the activity from the diagnostic event, or an activity started by the agent in case of a transaction.

src/Elastic.Apm.Azure.Storage/AzureBlobStorageDiagnosticListener.cs
russcam: Following [the convention used for AWS S3 span names](https://github.com/elastic/apm/blob/master/specs/agents/tracing-instrumentation-aws.md#s3-simple-storage-service) (the spec says operation name is camel case, but I think this should read pascal case based on the examples for S3, DynamoDB and the agent weekly conversations). Note that the span action is lowercase.

I don't really have an opinion about which might be best to use for this. Happy to follow up afterwards

test/Elastic.Apm.Azure.Storage.Tests/AzureBlobStorageDiagnosticListenerTests.cs
russcam: done!

docs/setup.asciidoc
```diff
@@ -58,10 +59,14 @@ https://www.nuget.org/packages/Elastic.Apm.StackExchange.Redis[**Elastic.Apm.Sta
...
 This packages contains instrumentation to capture spans for commands sent to redis with https://www.nuget.org/packages/StackExchange.Redis/[StackExchange.Redis] package.
 
-https://www.nuget.org/packages/Elastic.Apm.StackExchange.Redis[**Elastic.Apm.Azure.ServiceBus**]::
+https://www.nuget.org/packages/Elastic.Apm.Azure.ServiceBus[**Elastic.Apm.Azure.ServiceBus**]::
 
 This packages contains instrumentation to capture transactions and spans for messages sent and received from Azure Service Bus with https://www.nuget.org/packages/Microsoft.Azure.ServiceBus/[Microsoft.Azure.ServiceBus] and https://www.nuget.org/packages/Azure.Messaging.ServiceBus/[Azure.Messaging.ServiceBus] packages.
 
+https://www.nuget.org/packages/Elastic.Apm.Azure.Storage[**Elastic.Apm.Azure.Storage**]::
+
+This packages contains instrumentation to capture spans for interation with Azure Storage with https://www.nuget.org/packages/azure.storage.queues/[Azure.Storage.Queues], https://www.nuget.org/packages/azure.storage.blobs/[Azure.Storage.Blobs] and https://www.nuget.org/packages/azure.storage.files.shares/[Azure.Storage.Files.Shares] packages.
```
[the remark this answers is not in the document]
russcam: done

src/Elastic.Apm.Azure.Storage/AzureBlobStorageDiagnosticListener.cs
russcam: done

russcam:
The one failed test doesn't look related to these changes so merging in.

PR #1248, opened 2021-03-26
clarify the globallabels format

robb3rt:
With this I'm targetting to update [the documentation here](https://www.elastic.co/guide/en/apm/agent/dotnet/master/config-core.html), specifically the configuration section for "Globallabels" and the displayed format. The format should be key and values in separate strings like "GlobalLabels": "key=value,key=value".

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -455,7 +455,7 @@ See https://www.owasp.org/index.php/Information_exposure_through_query_strings_i
 [[config-global-labels]]
 ==== `GlobalLabels` (added[1.2])
 
-Labels added to all events, with the format `key=value[,key=value[,...]]`.
+Labels added to all events, with the format `"GlobalLabels": "key=value,key=value,..."`.
 Any labels set by the application via the agent's public API will override global labels with the same keys.
 
 [options="header"]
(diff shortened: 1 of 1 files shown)


docs/configuration.asciidoc
```diff
@@ -455,7 +455,7 @@ See https://www.owasp.org/index.php/Information_exposure_through_query_strings_i
 [[config-global-labels]]
 ==== `GlobalLabels` (added[1.2])
 
-Labels added to all events, with the format `key=value[,key=value[,...]]`.
+Labels added to all events, with the format `"GlobalLabels": "key=value,key=value,..."`.
```
russcam, line 458: I don't think the format should include `"GlobalLabels"`, as it's not correct for a value coming from environment variables, or certain `IConfiguration` structures [such as from ASP.NET Full Framework's appSettings](https://www.elastic.co/guide/en/apm/agent/dotnet/master/configuration-on-asp-net.html#asp-net-sample-config).

russcam, line 458: Something like

```suggestion
Labels added to all events, where each label is specified as `key=value` pair in which the `key` is the label name and `value` is the label value, and where multiple labels are each separated by a comma, for example, `label1=value1,label2=value2`.
```

@bmorelli25 might be able to come up with something more succinct.

russcam (changes requested):
Thanks for opening @robb3rt, I've left some suggestions.

docs/configuration.asciidoc
bmorelli25: I think that's fine. I would probably separate the thought into a few different sentences to make it easier to consume:

> Labels added to all events. Labels are `key=value` pairs, where `key` is the label name and `value` is the label value. Separate multiple labels with a comma. For example, `label1=value1,label2=value2`.

PR #1251, opened 2021-04-06
Enrich HTTP spans with additional information for known services

russcam:
This a draft PR to demonstrate an approach of enriching HTTP request spans with additional data when the request is to a known service. The demonstration instruments the older Azure Blob storage client package.

PR #1258, opened 2021-04-12
Create HTTP spans for known services

russcam:
Alternative approach to #1251.

HTTP span tracers can start spans for HTTP calls to known services. The demonstration instruments the older Azure Blob storage client package.

HTTP span tracers are collected on `AgentComponents`, and retrieved in the HttpDiagnosticListener in a `ReaderWriterLockSlim`


russcam:
This is now ready to review @gregkalapos

src/Elastic.Apm/DiagnosticSource/HttpDiagnosticsSubscriber.cs
```diff
@@ -15,20 +14,39 @@ namespace Elastic.Apm.DiagnosticSource
...
+
 		/// <summary>
 		/// Start listening for HttpClient diagnostic source events.
 		/// </summary>
 		public IDisposable Subscribe(IApmAgent agent)
 		{
 			var retVal = new CompositeDisposable();
+			var realAgent = agent as ApmAgent;
+
+			if (realAgent != null && realAgent.HttpDiagnosticListener != null)
```
gregkalapos, line 33:
```suggestion
			if (realAgent?.HttpDiagnosticListener != null)
```

src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerImplBase.cs
```diff
@@ -127,8 +136,32 @@ private void ProcessStartEvent(TRequest request, Uri requestUrl)
...
+						{
+							span = httpSpanTracer.StartSpan(ApmAgent, method, requestUrl, HeaderGetter);
+							if (span != null)
+								break;
+						}
+					}
+				}
+			}
+
+			if (span is null && StartHttpSpan)
```
gregkalapos, line 160: With this `span` can remain `null`, below we have some usages of the `span` variable and I think we should null-check those.

src/Elastic.Apm/DiagnosticListeners/TraceableHttpDiagnosticListener.cs
```diff
@@ -0,0 +1,27 @@
+// Licensed to Elasticsearch B.V under
+// one or more agreements.
+// Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information
+
+namespace Elastic.Apm.DiagnosticListeners
+{
+	internal abstract class TraceableHttpDiagnosticListener : DiagnosticListenerBase
```
gregkalapos, line 8: To be honest I'm not sure if we really need this.

The code around HTTP capturing gets fairly complicated and explaining this to users wouldn't be trivial either. Maybe I'd jut leave this out and say users need to subscribe to `HttpDiagnosticsSubscriber` which captures all HTTP requests either in form of HTTP spans or in case of known services different spans capturing the corresponding service.

I'd keep it simple first and if needed we could add this later.

gregkalapos (commented):
The `IHttpSpanTracer` and the way we add it looks good to me.

We should think about the part where we start a new HTTP span even if a `IHttpSpanTracer` implementation already started a span.

There is [this new spec](https://github.com/elastic/apm/pull/433) which would conflict with that.

src/Elastic.Apm/DiagnosticListeners/TraceableHttpDiagnosticListener.cs
russcam: This is added because the subscription of a `HttpDiagnosticListener` is tracked on `ApmAgent` so that the addition of `IHttpSpanTracer`s can determine whether an instance of `HttpDiagnosticListener` has subscribed already, or whether one should be subscribed to allow the Http span tracer to work. Since there are two different implementations of `HttpDiagnosticListener`, one for HttpClient and one for HttpWebRequest. the base class, `HttpDiagnosticListenerImplBase<TRequest, TResponse>` is generic, but we really want a non-generic class to work with and attach the `StartHttpSpan` property to, to avoid casting on each HttpDiagnosticListener subscription check.

> The code around HTTP capturing gets fairly complicated and explaining this to users wouldn't be trivial either. Maybe I'd jut leave this out and say users need to subscribe to HttpDiagnosticsSubscriber which captures all HTTP requests either in form of HTTP spans or in case of known services different spans capturing the corresponding service.

The implementation details are not publicly exposed and I think the proposed approach makes it easier for users. For example. consider the following three different scenarios:

1. User wishes to trace **all** HTTP requests

   ```c#
   Agent.Subscribe(new HttpDiagnosticsSubscriber());
   ```

   This will create a HTTP span for any HTTP request

2. User wishes to trace **all** HTTP requests **and** Azure Blob storage (when they're using the older library, Microsoft.Azure.Storage.Blob)

   ```c#
   Agent.Subscribe(new HttpDiagnosticsSubscriber(), new AzureBlobStorageDiagnosticsSubscriber());
   ```
   This will create specific spans to Azure Blob storage and a HTTP span for any other HTTP request

3. User wishes to trace only Azure Blob storage (when they're using the older library, Microsoft.Azure.Storage.Blob)

   ```c#
   Agent.Subscribe(new AzureBlobStorageDiagnosticsSubscriber());
   ```
   This will create specific spans to Azure Blob storage only

With this approach, a user does not need to know the implementation details of Azure Blob storage integration (when they're using the older library, Microsoft.Azure.Storage.Blob) and its dependency on a subscribed HttpDiagnosticsSubscriber, and the agent does not create HTTP spans for HTTP requests when a user has not explicitly subscribed with HttpDiagnosticsSubscriber.

src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerImplBase.cs
russcam: Good point, will check

src/Elastic.Apm/DiagnosticSource/HttpDiagnosticsSubscriber.cs
russcam: 👍

src/Elastic.Apm/DiagnosticListeners/TraceableHttpDiagnosticListener.cs
gregkalapos: > This is added because the subscription of a HttpDiagnosticListener is tracked on ApmAgent so that the addition of IHttpSpanTracers can determine whether an instance of HttpDiagnosticListener has subscribed already, or whether one should be subscribed to allow the Http span tracer to work. Since there are two different implementations of HttpDiagnosticListener, one for HttpClient and one for HttpWebRequest. the base class, HttpDiagnosticListenerImplBase<TRequest, TResponse> is generic, but we really want a non-generic class to work with and attach the StartHttpSpan property to, to avoid casting on each HttpDiagnosticListener subscription check.


Oh, ok I see the challenge here now. This HTTP capturing code started out with a non-generic base class and 2 implementations, but over time we ended up with `HttpDiagnosticListenerImplBase<TRequest, TResponse>`, which now seems to backfire :( 

Nevertheless, basically all `TraceableHttpDiagnosticListener` does is that it offers the `StartHttpSpan` `bool` property - regarding that one, I think ([as mentioned earlier](https://github.com/elastic/apm-agent-dotnet/pull/1258#pullrequestreview-641430540)) that will conflict with [this spec](https://github.com/elastic/apm/pull/433). According to the new spec we should never start child spans for exit spans - which would mean whenever an `IHttpSpanTracer` captures a `Span`, no further HTTP span should be captured. The main motivation in the spec is that this way we'll have more meaningful breakdown graph (otherwise basically everything here will be considered as time spent in HTTP call - and not in e.g. CosmosDB call).

So, I see why we need a non-generic class, but if we follow the new spec, I guess we'll basically remove the only property from the class.

russcam: The `StartHttpSpan` property doesn't conflict with the spec- it's responsibility is to create a HTTP span when none of the registered `IHttpSpanTracer`s create a span for a HTTP request, **and** the agent is configured to trace HTTP calls. It can be thought of as the "catch-all" control for HTTP tracing. 

For example, if a user has configured the agent to trace all HTTP and Azure blob storage, when a HTTP request is inferred to be a call to Azure blob storage by a specific `IHttpSpanTracer` implementation, then an exit span will be created that represents the call to Azure blob storage; It won't contain a child HTTP span. Similarly, if a call to Azure blob storage is traced in some other way, e.g. a DiagnosticListener for activities coming from an Azure blob storage diagnostic source, an exit span will be created that represents the call to Azure blob storage. Even if the call to Azure blob storage is ultimately made over HTTP, a child HTTP span will not be created either, because the HttpDiagnosticListener exits early on a `.Start` activity [when a current span exists for a known instrumentation](https://github.com/elastic/apm-agent-dotnet/pull/1258/files#diff-560772ea12a6aab2de9de8eb0d64815b5308bf1776a4925e5c07c129dcfb7307R122-R129) (_the `InstrumentationFlag` usage might be better represented in a form that indicates something like "suppress this HTTP span" - perhaps setting the `InstrumentationFlag.HttpClient` in Azure diagnostic listeners and checking that this flag is set in HttpDiagnosticListener  would be a better signal,_)

src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerImplBase.cs
```diff
@@ -118,6 +119,14 @@ protected override void HandleOnNext(KeyValuePair<string, object> kv)
 
 		private void ProcessStartEvent(TRequest request, Uri requestUrl)
 		{
+			if (_realAgent?.TracerInternal.CurrentSpan is Span currentSpan)
+			{
+				// if there's a current span that has been instrumented for Azure, don't create a span for
```
gregkalapos, line 124: 👍 nice, this is definitely needed to avoid double span creation in combination with the Azure diagnostic source listeners.

One consequence here: In `ProcessStopEvent` we do `ProcessingRequests.TryRemove(request, out var span)` and print a warning if it returns `false`. That warning needs to be adapted because with this it can return `false` and that's ok.

src/Elastic.Apm/DiagnosticListeners/TraceableHttpDiagnosticListener.cs
gregkalapos: We discussed this via zoom. tldr: we agreed to merge it as it is.

A bit more summery:
- We agreed to keep the feature that users can capture 1) only HTTP spans 2) Calls to well known services via `IHttpSpanTracer` AND HTTP spans (to not well known servicces) and 3) only calls to well known services via `IHttpSpanTracer` and no HTTP spans
- We weren't super happy about the fact that we need the `TraceableHttpDiagnosticListener` type - ideally the bool property from this type would live in `HttpDiagnosticListenerImplBase<TRequest, TResponse>`, but that won't work because we need a reference of the base class and the generic `HttpDiagnosticListenerImplBase<TRequest, TResponse>` is not ok for that
- We considered maybe working with `object` and casting, but that's worse 
- So at the end we accepted the PR in its current form.

gregkalapos (approved):
LGTM 👍 

Only this minor thing should be changed before merge: https://github.com/elastic/apm-agent-dotnet/pull/1258#discussion_r621958377

russcam:
@gregkalapos I've changed the implementation slightly since our sync yesterday. 

- Removed `TraceableHttpDiagnosticListener` and moved the bool property for capturing of spans and whether a Http diagnostic listener has been subscribed to a new `HttpTraceConfiguration` type, an instance of which is created inside `AgentComponents` ctor.

- Removed the `AddHttpSpanTracer` static method from Agent as it is not needed.

I think this is slightly cleaner.

src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerImplBase.cs
```diff
@@ -140,10 +145,9 @@ private void ProcessStartEvent(TRequest request, Uri requestUrl)
 			string HeaderGetter(string header) => RequestTryGetHeader(request, header, out var value) ? value : null;
 
 			ISpan span = null;
-
-			if (_realAgent != null)
+			if (_configuration != null)
 			{
-				using (var httpTracers = _realAgent.Components.GetHttpTracers())
+				using (var httpTracers = _configuration.GetTracers())
```
gregkalapos, line 150: This is not new in this commit, but now that I see this again I was wondering: this happens for every outgoing HTTP request and this allocates. Not a blocker, but it'd be nice to somehow avoid allocation here.

gregkalapos (approved):
I went through it - looks ok to me 👍 

I added 1 comment, but I think it's fine.

src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerImplBase.cs
russcam: There could be a check to see if there are any http tracers before allocating

russcam: Added check

PR #1259, opened 2021-04-12
Move error capturing to Error event handler

russcam:
This PR moves the capturing of exceptions to an Error event handler. This handler runs before an `Application_Error` event handler that may be defined on the `HttpApplication` derived type in Global.asax.cs, which may perform some action with errors that affects collecting them in the EndRequest handler e.g. calling `Server.ClearError()`;


gregkalapos (approved):
LGTM

PR #1271, opened 2021-04-21
Breakdown metrics

gregkalapos:
Solves https://github.com/elastic/apm-agent-dotnet/issues/227

image

TODOs:
- [x] Add docs
- [x] ~Test with non-sampled transactions~ update: not needed - non-sampled transactions and spans also generate these metrics - according to the spec that's preferred.
- [x] Disable breakdown metric collection


bluetentacle:
What's the status of this PR?

docs/metrics.asciidoc
```diff
@@ -174,4 +174,70 @@ format: ms
 Platform: all.
 
 The approximate accumulated collection elapsed time in milliseconds.
+
+[float]
+[[metrics-application]]
+=== Built-in application metrics
```
gregkalapos, line 180: 100% stolen from the java docs.

gregkalapos:
> What's the status of this PR?

Ready for review.

gregkalapos:
Pushed a few small changes:
- After asking around I learned that the duration of spans with `type: "app"` needs to be handled as part of the transaction self-time duration, I adapted the code according to this and added tests.
- Rebased on master.

src/Elastic.Apm/Model/Transaction.cs
```diff
@@ -477,6 +504,13 @@ public void End()
 			handler?.Invoke(this, EventArgs.Empty);
 			Ended = null;
 
+			if (SpanTimings.ContainsKey(("app", null)))
+				SpanTimings[("app", null)].IncrementTimer(SelfDuration);
+			else
+				SpanTimings.TryAdd(("app", null), new SpanTimer(SelfDuration));
+
+			_breakdownMetricsProvider?.CaptureTransaction(this);
```
russcam, line 512: Could breakdown metrics be captured by subscribing to the `Transaction.End` event in `Tracer`? It would avoid the need to pass the breakdown metrics provider to each transaction to call in the `End()` method.

russcam (commented):
Left some initial comments to discuss

docs/metrics.asciidoc
```diff
@@ -174,4 +174,70 @@ format: ms
...
+*`transaction.duration`*::
++
+--
+type: simple timer
+
+This timer tracks the duration of transactions and allows for the creation of graphs displaying a weighted average.
+
+Fields:
+
+* `sum.us`: The sum of all transaction durations in ms since the last report (the delta)
```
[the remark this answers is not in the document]
gregkalapos: Hahh - I looked into this and indeed there was an issue here. 

>  is the `u` in `.us` intended to be `μ` (mu)?

Yes! 

> but the description says ms, milliseconds. Is that correct?

No, it wasn't. This was a copy from the java agent docs, and in fact we sent the data in milliseconds in the PR originally. I looked into more docs and this clearly should be microseconds (e.g. the Go agent docs has this correctly). I fixed the code, so now we send it in microseconds and I also fixed this part of the doc as well - all microseconds now.

Side-note: since the breakdown metrics are used to calculate the time spent in different span types by percentage, I think the unit will not break the feature as long as all the data is in the same unit (which was the case), so likely that's why it was not an issue before - it'd be nice to see if all agents send it in the right unit :)

docs/metrics.asciidoc
```diff
@@ -174,4 +174,70 @@ format: ms
...
+*`span.self_time`*::
++
+--
+type: simple timer
+
+This timer tracks the span self-times and is the basis of the transaction breakdown visualization.
+
+Fields:
+
+* `sum.us`: The sum of all span self-times in ms since the last report (the delta)
```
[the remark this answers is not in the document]
gregkalapos: Same as above.

src/Elastic.Apm/Model/Transaction.cs
gregkalapos: Yes, we can do that. In addition to that we also don't need to pass `_breakdownMetricsProvider` to spans - so I also removed it from the span `.ctor`.

gregkalapos:
Rebased, ready to another review round.

src/Elastic.Apm/Metrics/MetricsCollector.cs
```diff
@@ -64,6 +66,19 @@ public MetricsCollector(IApmLogger logger, IPayloadSender payloadSender, IConfig
 
 			MetricsProviders = new List<IMetricsProvider>();
 
+			if (metricsProvider != null)
+			{
+				foreach (var item in metricsProvider)
+				{
+					if (item != null)
+					{
+						if (!WildcardMatcher.IsAnyMatch(currentConfigSnapshot.DisableMetrics, BreakdownMetricsProvider.SpanSelfTime))
```
russcam, line 75: This makes the assumption that any added metrics providers should be controlled by the span self-time metric. Whilst this holds true now as breakdown provider is the only one added in the ctor, maybe the string(s) to match should be a property on an `IMetricsProvider` to allow each individual provider determine if it should be added.

russcam (commented):
I've gone through in some more detail.

The gathering of breakdown metrics looks like it could be performed in `PayloadSenderV2`, when the transaction is queued. `PayloadSenderV2` would take `BreakdownMetricsProvider` in its ctor. I know I previously suggested hooking into the `.Ended` event as has been done here, but in retrospect, performing in `PayloadSenderV2` would actually be cleaner.

An aside- 

`PayloadSenderV2` is a little bit of a misnomer as its performing a few different functions: 

1. holding a queue of items to send to APM server
2. filtering items that should be sent
3. serializing items to be sent
4. actually sending items to APM server

Might be useful to refactor this in the next major version.

src/Elastic.Apm/Model/Transaction.cs
```diff
@@ -13,25 +15,35 @@
...
 
 namespace Elastic.Apm.Model
 {
 	internal class Transaction : ITransaction
 	{
 		private static readonly string ApmTransactionActivityName = "ElasticApm.Transaction";
+
+		internal readonly TraceState _traceState;
+
+		internal readonly ConcurrentDictionary<(string, string), SpanTimer> SpanTimings = new();
```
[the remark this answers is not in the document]
gregkalapos: Done

src/Elastic.Apm/Metrics/MetricsCollector.cs
gregkalapos: Yeah, I agree that this isn't ok.

> maybe the string(s) to match should be a property on an IMetricsProvider to allow each individual provider determine if it should be added.

Problem is we'd need to address 2 scenarios with this:
- The `BreakdownMetricsProvider` and everything passed to the `ctor` is already created - so yes, in that case we could just query the string(s) from `IMetricsProvider` and it'd indeed help.
- Every other `IMetricsProvider` is not instantiated at this point, so we can't query the string(s) list for those. In that cases what we do is that we check the `DisabledMetrics` list and if one of those is disabled we don't even instantiate those.

What I did is that I moved the check to `AgentCompontents` - that's where we instantiate the `BreakdownMetricsProvider`. 

Long term we could come up with an infrastructure to better handle this - I think it'd be fine to introduce the string list you suggest, but that'll also mean we'd need to instantiate all the `IMetricsProvider` and if they are not enabled based on the string(s) list, we'd just not add them to the list and let them GCed. Nevertheless that'd be a longer discussion, so maybe we could do that in a separate PR.

gregkalapos: Update: I decided to move on and implement it nevertheless. I added `IMetricsProvider.IsEnabled` - so we instantiate each `IMetricsProvider` implementation, but only add them if they are enabled. This changed a few things around metrics collection, but in my opinion this is now much cleaner.

gregkalapos:
> I've gone through in some more detail.
> 
> The gathering of breakdown metrics looks like it could be performed in `PayloadSenderV2`, when the transaction is queued. `PayloadSenderV2` would take `BreakdownMetricsProvider` in its ctor. I know I previously suggested hooking into the `.Ended` event as has been done here, but in retrospect, performing in `PayloadSenderV2` would actually be cleaner.
> 

There are multiple places where we can capture the metrics - basically anything works after the transaction ends and before it goes out to APM server. 

In what sense do you think it would be cleaner? 

I see a few issues with that:
- `PayloadSenderV2` is not a component to create metrics, it's to send the payload.
- If someone replaces it (which is super rare, but I saw some people doing it), they won't have breakdown metrics anymore - which to me makes putting it into `PayloadSenderV2` not clean. If I replace the payload sender, I'd not expect that affecting breakdown metric collection. Same applies for us: let's say we implement a new payload sender - why does it need to change breakdown metrics collection as well?
- The breakdown metrics are emitted by transactions - it's defined to be emitted when the transaction finishes, so I think the closer it is to `Transaction.End()`, the better - yes, `Transaction.EndI()` directly calls `PayloadSenderV2`, but that is a different component for different purposes.
- If transactions emit the metrics, testing is also very natural - no additional config is needed. If we move it into `PayloadSenderV2`, we'll need to inject a `BreakdownMetricsProvider` into our test payload sender. This is not the end of the world, but again raises the question: why does the payload sender influence breakdown metrics collection, which is a metric purely based on transactions (and spans)?

I opened https://github.com/elastic/apm-agent-dotnet/pull/1341 which moves `BreakdownMetricsProvider` to `PayloadSenderV2`. Overall I think the PR in its current form is better. If it's ok for you I'd leave it out, if not I'll cherry pick the commit that moved `BreakdownMetricsProvider` to `PaylaadSenderV2`.

Otherwise it's ready for another round.

russcam:
> In what sense do you think it would be cleaner?

1. It avoids the creation and invocation of an event delegate for each transaction, to capture breakdown metrics
2. Breakdown metrics _could_ be captured on the dedicated payload processing thread

The `transaction.Ended` event has a valid use case in scenarios where some integration outside of Elastic.Apm needs to be notified when a transaction ends, such as in StackExchange.Redis, but doesn't have to be used for breakdown metrics, which are calculated in Elastic.Apm.

> * `PayloadSenderV2` is not a component to create metrics, it's to send the payload.

It's doing a lot more than sending payloads; it's more of a processing pipeline for payloads, which could also include calculating breakdown metrics for a particular payload, transactions.

> * If someone replaces it (which is super rare, but I saw some people doing it), they won't have breakdown metrics anymore - which to me makes putting it into `PayloadSenderV2` not clean. If I replace the payload sender, I'd not expect that affecting breakdown metric collection. Same applies for us: let's say we implement a new payload sender - why does it need to change breakdown metrics collection as well?

If someone replaces `IPayloadSender`, the same holds true for 

1. transaction, span and error filtering
2. enqueuing payloads and processing on a configurable interval on a dedicated thread
3. sending payloads to APM server

but the public interface doesn't define a contract for this behaviour, only that payloads are queued

```c#
public interface IPayloadSender
{
	void QueueError(IError error);
	void QueueMetrics(IMetricSet metrics);
	void QueueSpan(ISpan span);
	void QueueTransaction(ITransaction transaction);
}
```

I'm trying to think about the architecture of the agent in terms of how it is evolving and where we might want to go with it. Happy to discuss here, though it might be easier to chat over zoom 😃

src/Elastic.Apm/Metrics/MetricsCollector.cs
russcam: ++ it'd be possible to build on this for the custom metrics scenarios.

src/Elastic.Apm/Metrics/MetricsProvider/BreakdownMetricsProvider.cs
```diff
@@ -0,0 +1,87 @@
...
+							new("transaction.breakdown.count", _transactionCount),
+						}) { Transaction = new TransactionInfo { Name = transaction.Name, Type = transaction.Type } };
+
+				_itemsToSend.Add(transactionMetric);
+			}
+		}
+
+		public IEnumerable<MetricSet> GetSamples()
+		{
+			var retVal = new List<MetricSet>();
```
russcam, line 75: could avoid resizing the internal array by allocating a capacity based on count of `_itemsToSend` (or 1000)

src/Elastic.Apm/Metrics/MetricsProvider/BreakdownMetricsProvider.cs
```diff
@@ -0,0 +1,87 @@
...
+
+			lock (_lock)
+			{
+				retVal.AddRange(_itemsToSend.Take(1000));
+				_itemsToSend.Clear();
```
russcam, line 80: should `_itemsToSend` be cleared? What if it still contains items to send after taking 1000?

src/Elastic.Apm/Model/SpanTypeAndSubtype.cs
```diff
@@ -0,0 +1,19 @@
...
+// one or more agreements.
+// Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information
+
+namespace Elastic.Apm.Model
+{
+	/// <summary>
+	/// Encapsulates type and subtype
+	/// </summary>
+	internal struct SpanTypeAndSubtype
```
russcam, line 11: minor nitpick: naming

```suggestion
	internal struct SpanTimerKey
```

gregkalapos:
> > In what sense do you think it would be cleaner?
> 
> 1. It avoids the creation and invocation of an event delegate for each transaction, to capture breakdown metrics
> 2. Breakdown metrics _could_ be captured on the dedicated payload processing thread
> 
> The `transaction.Ended` event has a valid use case in scenarios where some integration outside of Elastic.Apm needs to be notified when a transaction ends, such as in StackExchange.Redis, but doesn't have to be used for breakdown metrics, which are calculated in Elastic.Apm.

1. I agree the delegate is not optimal for that. So far we had 3 options: 1) doing it in transaction 2) doing it in tracer 3) doing it in payload sender. Option 1 (the original code) would not have this issue and I think it's more natural.
2. That's true for creating the `MetricSet`, but not true for all the calculation. Also, this'd only apply to this single metric, but not for the other. If this is a goal (which I think in this PR should not be) then we should make a dedicated thread for all metrics collection.

> 
> > * `PayloadSenderV2` is not a component to create metrics, it's to send the payload.
> 
> It's doing a lot more than sending payloads; it's more of a processing pipeline for payloads, which could also include calculating breakdown metrics for a particular payload, transactions.
> 

That's true and everyone (rightfully) criticises it, yet we end up putting more stuff into it. So this is again a reason why I'd not put it into there - but the main reason I would not put it there is that I think breakdown metrics creation is really not the job for the payload sender.

> > * If someone replaces it (which is super rare, but I saw some people doing it), they won't have breakdown metrics anymore - which to me makes putting it into `PayloadSenderV2` not clean. If I replace the payload sender, I'd not expect that affecting breakdown metric collection. Same applies for us: let's say we implement a new payload sender - why does it need to change breakdown metrics collection as well?
> 
> If someone replaces `IPayloadSender`, the same holds true for
> 
> 1. transaction, span and error filtering
> 2. enqueuing payloads and processing on a configurable interval on a dedicated thread
> 3. sending payloads to APM server
> 
> but the public interface doesn't define a contract for this behaviour, only that payloads are queued
> 
> ```cs
> public interface IPayloadSender
> {
> 	void QueueError(IError error);
> 	void QueueMetrics(IMetricSet metrics);
> 	void QueueSpan(ISpan span);
> 	void QueueTransaction(ITransaction transaction);
> }
> ```
 
That's true and unfortunate, but why do we need to make yet another component part of this list? I still don't see how it's related to payload sender. Breakdown is a metric created by spans and transactions.

> I'm trying to think about the architecture of the agent in terms of how it is evolving and where we might want to go with it. Happy to discuss here, though it might be easier to chat over zoom 😃

Ok, let's zoom - maybe I just don't get the core of your motivation. I think the cleanest and simplest was to just call `BreakdownMetricsProvider?.CaptureTransaction(Transaction)` from `Transaction.End()` right after we create the last breakdown metric for the transaction. And I feel that holds true regardless of how future payloadsenders may look. But let's zoom...

gregkalapos:
> > > In what sense do you think it would be cleaner?
> > 
> > 
> > 
> > 1. It avoids the creation and invocation of an event delegate for each transaction, to capture breakdown metrics
> > 2. Breakdown metrics _could_ be captured on the dedicated payload processing thread
> > 
> > The `transaction.Ended` event has a valid use case in scenarios where some integration outside of Elastic.Apm needs to be notified when a transaction ends, such as in StackExchange.Redis, but doesn't have to be used for breakdown metrics, which are calculated in Elastic.Apm.
> 
> 1. I agree the delegate is not optimal for that. So far we had 3 options: 1) doing it in transaction 2) doing it in tracer 3) doing it in payload sender. Option 1 (the original code) would not have this issue and I think it's more natural.
> 2. That's true for creating the `MetricSet`, but not true for all the calculation. Also, this'd only apply to this single metric, but not for the other. If this is a goal (which I think in this PR should not be) then we should make a dedicated thread for all metrics collection.
> 
> > > * `PayloadSenderV2` is not a component to create metrics, it's to send the payload.
> > 
> > 
> > It's doing a lot more than sending payloads; it's more of a processing pipeline for payloads, which could also include calculating breakdown metrics for a particular payload, transactions.
> 
> That's true and everyone (rightfully) criticises it, yet we end up putting more stuff into it. So this is again a reason why I'd not put it into there - but the main reason I would not put it there is that I think breakdown metrics creation is really not the job for the payload sender.
> 
> > > * If someone replaces it (which is super rare, but I saw some people doing it), they won't have breakdown metrics anymore - which to me makes putting it into `PayloadSenderV2` not clean. If I replace the payload sender, I'd not expect that affecting breakdown metric collection. Same applies for us: let's say we implement a new payload sender - why does it need to change breakdown metrics collection as well?
> > 
> > 
> > If someone replaces `IPayloadSender`, the same holds true for
> > 
> > 1. transaction, span and error filtering
> > 2. enqueuing payloads and processing on a configurable interval on a dedicated thread
> > 3. sending payloads to APM server
> > 
> > but the public interface doesn't define a contract for this behaviour, only that payloads are queued
> > ```cs
> > public interface IPayloadSender
> > {
> > 	void QueueError(IError error);
> > 	void QueueMetrics(IMetricSet metrics);
> > 	void QueueSpan(ISpan span);
> > 	void QueueTransaction(ITransaction transaction);
> > }
> > ```
> 
> That's true and unfortunate, but why do we need to make yet another component part of this list? I still don't see how it's related to payload sender. Breakdown is a metric created by spans and transactions.
> 
> > I'm trying to think about the architecture of the agent in terms of how it is evolving and where we might want to go with it. Happy to discuss here, though it might be easier to chat over zoom 😃
> 
> Ok, let's zoom - maybe I just don't get the core of your motivation. I think the cleanest and simplest was to just call `BreakdownMetricsProvider?.CaptureTransaction(Transaction)` from `Transaction.End()` right after we create the last breakdown metric for the transaction. And I feel that holds true regardless of how future payloadsenders may look. But let's zoom...

Zoom summary: we agreed to move `BreakdownMetricsProvider.CaptureTransaction()` to `Transaction.cs` - I implemented that.

src/Elastic.Apm/Api/Tracer.cs
```diff
@@ -61,16 +65,35 @@ IApmServerInfo apmServerInfo
...
+			{
+				Service = _service
+			};
+
+			retVal.Ended += RetValOnEnded;
 
 			_logger.Debug()?.Log("Starting {TransactionValue}", retVal);
 			return retVal;
+
+			void RetValOnEnded(object sender, EventArgs e)
```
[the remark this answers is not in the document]
gregkalapos: This part could got removed.

src/Elastic.Apm/Metrics/MetricsProvider/BreakdownMetricsProvider.cs
gregkalapos: Yes, makes sense - done.

src/Elastic.Apm/Metrics/MetricsProvider/BreakdownMetricsProvider.cs
gregkalapos: > should _itemsToSend be cleared? What if it still contains items to send after taking 1000?

The spec says: "...agents must cap the total number of metricsets. A reasonable default seems to be  [1000]. When the limit is reached, agents should log sla warning...". So my understanding is that everything above 1k should be thrown away, so it's fine to clear those. Also worth noting that in this case we'd try to show time spent in percentage across 1K groups, so at this point I'd say the visualization can be also challenging.

On the other hand by thinking about this: we should not even collect those in `_itemsToSend`, plus the warning was also missing. So I moved the handling of this into `CaptureTransaction(Transaction)`: we only collect up to 1K items and if 1K is reached we log.

I also included a test to cover this.

src/Elastic.Apm/Model/SpanTypeAndSubtype.cs
gregkalapos: Done.

docs/metrics.asciidoc
```diff
@@ -174,4 +174,70 @@ format: ms
 Platform: all.
 
 The approximate accumulated collection elapsed time in milliseconds.
+
```
bmorelli25, line 177: This will fix the doc build:

```suggestion
--

```

test/Elastic.Apm.Tests.Utilities/MockPayloadSender.cs
```diff
@@ -20,7 +20,10 @@ internal class MockPayloadSender : IPayloadSender
 	{
 		private readonly List<IError> _errors = new List<IError>();
 		private readonly List<Func<IError, IError>> _errorFilters = new List<Func<IError, IError>>();
-		private readonly object _lock = new object();
+		private readonly object _spanLock = new object();
```
gregkalapos, line 23: I saw tests failing with such logs - this specific test was a new one which I introduced in the PR. I wasn't able to repro this locally:

```
System.ArgumentException : Destination array was not long enough. Check the destination index, length, and the array's lower bounds.
Parameter name: destinationArray
Stacktrace
System.ArgumentException : Destination array was not long enough. Check the destination index, length, and the array's lower bounds.
Parameter name: destinationArray
Stack Trace:
   at System.Array.Copy(Array sourceArray, Int32 sourceIndex, Array destinationArray, Int32 destinationIndex, Int32 length, Boolean reliable)
   at System.Collections.Generic.List`1.CopyTo(T[] array, Int32 arrayIndex)
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at Elastic.Apm.Tests.BreakdownTests.TransactionWithSpans() in C:\Users\jenkins\workspace\net_apm-agent-dotnet-mbp_PR-1271\apm-agent-dotnet\test\Elastic.Apm.Tests\BreakdownTests.cs:line 62
   ```
   
  This exception isn't new - I already saw it on failing tests. I think the root cause of those failing tests is that the agent still puts an event into the `MockPayloadSender` while it makes the copy of the list. To address this I introduced locks in the `MockPayloadSender`.

File: test/Elastic.Apm.Tests.Utilities/MockPayloadSender.cs [hunk 13]
```diff
@@ -72,16 +69,22 @@ public bool WaitForTransactions(TimeSpan? timeout = null, int? count = null)
...
+					{
+						while (_transactions.Count < count && signalled)
+							signalled = _transactionWaitHandle.WaitOne(DefaultTimeout);
+					}
 				}
 				else
 				{
 					var stopWatch = Stopwatch.StartNew();
-					while (_transactions.Count < count && signalled)
+					lock (_transactionLock)
```
russcam, line 81: If I understand this correctly, the lock around waiting for `x` count of transactions can't succeed because adding transactions is also within a lock, meaning no transactions can be added whilst waiting for the count to be reached.

I _think_ the locks around `WaitForX()` count can be removed, and that locking is needed only on adding items or copying to an immutable collection.

russcam (commented):
Left a comment around locking, rest looks OK.

test/Elastic.Apm.Tests.Utilities/MockPayloadSender.cs
gregkalapos: Oh, there is an `AutoResetEvent.WaitOne()` within the lock 🤦 . I really wonder why no test failed, everything was green both locally and in CI. Yeah, I also think this should fail if first thread calls `WaitForX()` and then another one tries to add to the queue.

I think locking on the reading of `_transaction.Count` and `spans.Count` is needed - otherwise the number which the thread calling `WaitForX()` sees can be incorrect if another thread adds to the queue, but the lock should be limited to the call to `.Count`, and that finishes immediately.

I updated the PR accordingly.

russcam (approved):

PR #1274, opened 2021-04-27
fix: force to pass the BUILD_ID

kuisathaverat:
The BUILD_ID environment variable is not populated in the don't Docker container making the IP to connect wrong.


mdelapenya (approved):

cachedout (approved):

kuisathaverat:
it is not the cause

Issue #1232, opened 2021-03-18
AssemblyLoadException for `Ben.Demystifier` with startup hook setup


gregkalapos:
During startup and basic transaction capturing, this does not happen, since code using `Ben.Demystifier` doesn't run until we capture a stack trace.

- Tests covering startup hook should be also extended to cover such scenario.

gregkalapos:
Solved in https://github.com/elastic/apm-agent-dotnet/pull/1275

PR #1275, opened 2021-04-27
Internalize Ben.Demystifier

gregkalapos:
Solves https://github.com/elastic/apm-agent-dotnet/issues/1232


src/Elastic.Apm/Libraries/Ben.Demystifier/Ben.Demystifier.csproj
```diff
@@ -0,0 +1,56 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
russcam, line 1: I think this can be removed

NOTICE
```diff
@@ -13,3 +13,6 @@ Permission is hereby granted, free of charge, to any person obtaining a copy of
 The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
 
 THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+The source code in src/Elastic.Apm/Libraries/Ben.Demystifier is copied from Ben.Demystifier by @benaadams,
```
russcam, line 17: Should we make a note of the commit hash from which this is copied?

src/Elastic.Apm/Libraries/Ben.Demystifier/EnhancedStackFrame.cs
```diff
@@ -0,0 +1,92 @@
+﻿// Copyright (c) Ben A Adams. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Reflection;
+
+#nullable enable
+namespace System.Diagnostics
```
russcam, line 7: We should change the namespace here to `Elastic.Apm...`, to avoid a conflict where someone also references Ben.Demystifier as a package

src/Elastic.Apm/Libraries/Ben.Demystifier/EnhancedStackFrame.cs
```diff
@@ -0,0 +1,92 @@
...
+{
+	public class EnhancedStackFrame : StackFrame
```
russcam, line 9: Does this need to be public? Ditto for any other public types

src/Elastic.Apm/Libraries/Ben.Demystifier/Enumerable/EnumerableIList.cs
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) Ben A Adams. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+#nullable enable
+namespace System.Collections.Generic.Enumerable
```
russcam, line 5: change namespace?

russcam (changes requested):
This is taking shape- I've added some comments of things that I think should be addressed.

Now that we're source including both Newtonsoft.Json and Ben.Demystifier, a tool to be able to update both would be super useful. Doesn't need to be part of this PR, but would make keeping in sync easier down the road

src/Elastic.Apm/Libraries/Ben.Demystifier/Ben.Demystifier.csproj
gregkalapos: Yes, removed.

NOTICE
gregkalapos: Ok, added it.

src/Elastic.Apm/Libraries/Ben.Demystifier/EnhancedStackFrame.cs
gregkalapos: Agreed - But I ended up with `Elastic.Apm.Ben.Demystifier` otherwise we'd end up with `Elastic.Apm.System...` which would cause some trouble for types in the `Elastic.Apm` namespace that use types from `System`.

src/Elastic.Apm/Libraries/Ben.Demystifier/EnhancedStackFrame.cs
gregkalapos: No - I made everything internal.

src/Elastic.Apm/Libraries/Ben.Demystifier/Enumerable/EnumerableIList.cs
gregkalapos: Done.

gregkalapos:
I went through it  - I made everything private and renamed the namespaces.

Also opened https://github.com/elastic/apm-agent-dotnet/issues/1280 to deal with updating these later.

src/Elastic.Apm/Libraries/Ben.Demystifier/EnhancedStackFrame.cs
```diff
@@ -0,0 +1,93 @@
+﻿// Copyright (c) Ben A Adams. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Diagnostics;
+using System.Reflection;
+
+#nullable enable
+namespace Elastic.Apm.Ben.Demystifier.System.Diagnostics
```
russcam, line 8: I think the namespace for all of these should start `Elastic.Apm.Libraries.Ben.Demystifier`, in line with the convention used for Newtonsoft.Json. Further, for those types in Ben.Demystifier that are placed in `System.*` namespaces, we can remove that part as it won't serve any purpose for our usage.

russcam (changes requested):
Left a comment about the namespaces

src/Elastic.Apm/Libraries/Ben.Demystifier/EnhancedStackFrame.cs
gregkalapos: The original `Ben.Demystifier` doesn't have it's own namespace, it's all under `System.Diagnostics` and creates subclasses like `EnhancedStackTrace` derived from `System.Diagnostics.StackTrace`, but leaves it in the same namespace. So it's a bit different compared to Newtonsoft.Json which has its own namespace and we just prefixed it.

Nevertheless I agree with you, best would be to just have this under `Elastic.Apm.Libraries.Ben.Demystifier` - so I moved everything there.

russcam (approved):
LGTM 👍 

Wondering if we should have a unit test that verifies all types in Elastic.Apm.Libraries.Ben.Demystifier.* namespace are non-public

gregkalapos:
> Wondering if we should have a unit test that verifies all types in Elastic.Apm.Libraries.Ben.Demystifier.* namespace are non-public

Good idea - I opened https://github.com/elastic/apm-agent-dotnet/pull/1287 - it also covers Newtonsoft.Json.

PR #1284, opened 2021-04-30
Use storage account in destination.service.resource

russcam:
Relates: elastic/apm#426


russcam:
Merging this in as Azure storage tests pass

PR #1287, opened 2021-05-03
Add test to assert that types under Elastic.Apm.Libraries.* are non-public

gregkalapos:
Tests that our included 3rd party libs are all non-public.

Idea from https://github.com/elastic/apm-agent-dotnet/pull/1275#pullrequestreview-648794299


russcam (approved):
LGTM 👍

PR #1290, opened 2021-05-04
Add configured hostname

### LINKED ISSUE #1289: Send configured_hostname to APM server
When a user explicitly [configures a hostname](https://www.elastic.co/guide/en/apm/agent/dotnet/current/config-core.html#config-hostname), the value should be sent to APM server with the the `configured_hostname` key. Currently, the agent sends the value with the `hostname` key, however `hostname` was deprecated and replaced with `detected_hostname` and `configured_hostname`.

russcam:
This commit adds the configured hostname to the System properties sent
to APM server.

Fixes #1289

### DIFF
--- a/src/Elastic.Apm/Api/System.cs
+++ b/src/Elastic.Apm/Api/System.cs
@@ -2,6 +2,7 @@
 // Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information
 
+using System;
 using Elastic.Apm.Api.Constraints;
 using Elastic.Apm.Api.Kubernetes;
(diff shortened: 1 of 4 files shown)


gregkalapos (approved):
Oh, nice catch. This was introduced in https://github.com/elastic/apm-server/pull/2540. 

The changes look good to me. One thing which'd be nice to have is some test coverage. `ConfigTests.Set_HostName` was my 1. candidate to extend and cover this, but unfortunately it seems we use these values on `PayloadSenderV2`, and that test only reads the value coming from the config, but doesn't test where we send the value.

russcam:
Add a couple of unit tests to assert that serialized system container configured_hostname and detected_hostname

PR #1298, opened 2021-05-06
Capture errors with startup hook auto instrumentation

### LINKED ISSUE #1233: Startup hook setup + ASP.NET Core: HTTP request with unhandled exception is not captured
With above setup, nothing is captured. The transaction is missing and the error for the unhandled exception isn't captured.

I looked at logs during a call, and it seems that the transaction is started, but after the exception, none of the events we expect [here](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm.AspNetCore/DiagnosticListener/AspNetCoreDiagnosticListener.cs#L38) are received. We listen to `Microsoft.AspNetCore.Diagnostics.UnhandledException`, `Microsoft.AspNetCore.Diagnostics.HandledException`, and `Microsoft.AspNetCore.Hosting.UnhandledException` - none of those seemed to be triggered.

russcam:
This commit updates the startup hook auto instrumentation feature to capture thrown exceptions through the subscribed
`AspNetCoreDiagnosticListener`.

When an unhandled exception is thrown, ASP.NET Core's `DeveloperExceptionPageMiddleware` will raise a diagnostic
event with key "Microsoft.AspNetCore.Diagnostics.UnhandledException" if there is a listener listening to the event. If no listener is subscribed, the event is not captured.

Update how assemblies are loaded in the startup hook auto instrumentation. Register an event handler to the `AssemblyLoadContext.Default.Resolving` event to load assemblies that exist in the loader directory with matching version and public key. For assemblies starting with Elastic.Apm, that may contain DiagnosticListeners, load into `AssemblyLoadContext.Default` to allow listeners to register via `DiagnosticListener.AllListeners.Subscribe()`. For any other assemblies, load them into a separate AssemblyLoadContext to avoid version conflicts with assemblies that the application may reference. An example of the problem this avoids is System.Reflection.Metadata; The APM agent's source included version of Ben.Demystifier depends on System.Reflection.Metadata 5.0.0, but an application may reference a different version, such as ASP.NET Core 3.0, which references System.Reflection.Metadata 1.4.4.0 by default.

Add integration tests for capturing errors from startup hooks for netcoreapp3.0, netcoreapp3.1 and net5.0.

Update documentation for startup hooks to indicate that this feature is supported in .NET Core 3.0 and newer. .NET Core 2.2 is End of Life (EOL) and no longer supported by Microsoft. In testing startup hooks with .NET Core 2.2, Diagnostic Listeners do not subscribe to ASP.NET Core diagnostic events. On initial investigation, the System.Diagnostics.DiagnosticSource assembly loaded in the startup hooks is version 4.0.3.1. When listeners come to subscribe, System.Diagnostics.DiagnosticSource 4.0.4.0 that listeners are compiled against is loaded into the separate AssemblyLoadContext, which might be why listeners do not end up subscribing. Considering .NET Core 2.2 is EOL however, no further effort has been put into making this work.

Fixes #1233

### DIFF
--- a/src/Elastic.Apm.StartupHook.Loader/Loader.cs
+++ b/src/Elastic.Apm.StartupHook.Loader/Loader.cs
@@ -20,7 +20,7 @@ using ElasticApmStartupHook;
 namespace Elastic.Apm.StartupHook.Loader
 {
 	/// <summary>
-	/// Loads the agent assemblies, its dependent assemblies and starts it
+	/// Starts the agent
 	/// </summary>
 	internal class Loader
 	{
@@ -37,23 +37,9 @@ namespace Elastic.Apm.StartupHook.Loader
 		}
 
 		/// <summary>
-		/// Initializes assemblies and starts the agent
+		/// Initializes and starts the agent
 		/// </summary>
 		public static void Initialize()
-		{
-			var agentLibsToLoad =  new[]{ "Elastic.Apm", "Elastic.Apm.Extensions.Hosting", "Elastic.Apm.AspNetCore", "Elastic.Apm.EntityFrameworkCore", "Elastic.Apm.SqlClient", "Elastic.Apm.GrpcClient", "Elastic.Apm.Elasticsearch" };
-			var agentDependencyLibsToLoad = new[] { "System.Diagnostics.PerformanceCounter", "Microsoft.Diagnostics.Tracing.TraceEvent", "Elasticsearch.Net" };
-
-			foreach (var libToLoad in agentDependencyLibsToLoad)
-				AssemblyLoadContext.Default.LoadFromAssemblyPath(Path.Combine(AssemblyDirectory, libToLoad + ".dll"));
-			foreach (var libToLoad in agentLibsToLoad)
-				AssemblyLoadContext.Default.LoadFromAssemblyPath(Path.Combine(AssemblyDirectory, libToLoad + ".dll"));
-
-			StartAgent();
-		}
-
-		[MethodImpl(MethodImplOptions.NoInlining)]
-		private static void StartAgent()
 		{
 			Agent.Setup(new AgentComponents());
 
--- a/src/ElasticApmAgentStartupHook/StartupHook.cs
+++ b/src/ElasticApmAgentStartupHook/StartupHook.cs
@@ -14,7 +14,9 @@ using ElasticApmStartupHook;
 // ReSharper disable once CheckNamespace - per doc. this must be called StartupHook without a namespace with an Initialize method.
 internal class StartupHook
 {
-	private const string ElasticApmStartuphookLoaderDll = "Elastic.Apm.StartupHook.Loader.dll";
+	private const string LoaderDll = "Elastic.Apm.StartupHook.Loader.dll";
+	private const string LoaderTypeName = "Elastic.Apm.StartupHook.Loader.Loader";
+	private const string LoaderTypeMethod = "Initialize";
 	private const string SystemDiagnosticsDiagnosticsource = "System.Diagnostics.DiagnosticSource";
 
 	private static readonly byte[] SystemDiagnosticsDiagnosticSourcePublicKeyToken = { 204, 123, 19, 255, 205, 45, 221, 81 };
@@ -40,7 +42,7 @@ internal class StartupHook
 
 		var assemblies = AppDomain.CurrentDomain.GetAssemblies();
 
-		_logger.WriteLine($"Assemblies loaded:{Environment.NewLine}{string.Join(",", assemblies.Select(a => a.GetName()))}");
+		_logger.WriteLine($"Assemblies loaded:{Environment.NewLine}{string.Join(Environment.NewLine, assemblies.Select(a => a.GetName()))}");
 
 		var diagnosticSourceAssemblies = assemblies
 			.Where(a => a.GetName().Name.Equals(SystemDiagnosticsDiagnosticsource, StringComparison.Ordinal))
@@ -63,7 +65,9 @@ internal class StartupHook
 				break;
 		}
 
-		Assembly loader = null;
+		Assembly loaderAssembly = null;
+		string loaderDirectory;
+
 		if (diagnosticSourceAssembly is null)
 		{
 			// use agent compiled against the highest version of System.Diagnostics.DiagnosticSource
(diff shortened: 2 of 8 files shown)


src/Elastic.Apm.StartupHook.Loader/Loader.cs
```diff
@@ -20,7 +20,7 @@
 namespace Elastic.Apm.StartupHook.Loader
 {
 	/// <summary>
-	/// Loads the agent assemblies, its dependent assemblies and starts it
+	/// Starts the agent
 	/// </summary>
 	internal class Loader
```
gregkalapos, line 25: With this change the name of the type and the whole project isn't very accurate - it's not really loading things anymore, it only subscribes. It's not a big deal, and I don't have super strong opinion, but wanted to note this. Maybe we could consider renaming.

src/ElasticApmAgentStartupHook/StartupHook.cs
```diff
@@ -104,7 +108,53 @@ public static void Initialize()
...
+					var assemblyName = AssemblyName.GetAssemblyName(assemblyPath);
+					if (name.Version == assemblyName.Version)
+					{
+						var keyToken = name.GetPublicKeyToken();
+						var assemblyKeyToken = assemblyName.GetPublicKeyToken();
+						if (keyToken.SequenceEqual(assemblyKeyToken))
+						{
+							// load Elastic.Apm assemblies with the default assembly load context, to allow DiagnosticListeners to subscribe.
+							// For all other dependencies, load with a separate load context to not conflict with application dependencies.
+							return name.Name.StartsWith("Elastic.Apm", StringComparison.Ordinal)
```
gregkalapos, line 144: Are we totally sure about this? 🤔 

This means if we have let's say `Elstic.Apm.X` which depends on assembly `Y`, we load `Elastic.Apm.X`  into the default context, but load `Y` into the custom context - in this case `Elastic.Apm.X` would have missing typerefs in the default load context.

This seems to run and with our current references it does not seem to have any problems, but this looks a bit strange to me.

src/ElasticApmAgentStartupHook/StartupHook.cs
```diff
@@ -104,7 +108,53 @@ public static void Initialize()
...
+	}
+
+	/// <summary>
+	/// Loads assemblies from the loader directory if they exist
+	/// </summary>
+	/// <param name="loaderDirectory"></param>
+	private static void LoadAssembliesFromLoaderDirectory(string loaderDirectory)
+	{
+		var context = new ElasticApmAssemblyLoadContext();
+		AssemblyLoadContext.Default.Resolving += (_, name) =>
```
gregkalapos, line 128: Previously when we had Newtonsoft.Json and Ben.Demystifier as normal dependencies, often I ran into stackoverflow exceptions with this. Seems working now, just noting it here.

gregkalapos (commented):
I'm not really sure this fixes https://github.com/elastic/apm-agent-dotnet/issues/1233 - or changes anything around that.

I did some tests to reproduce the original problem (no success so far) and documented my findings here: https://github.com/elastic/apm-agent-dotnet/issues/1233#issuecomment-847893988.

If you add the new test `Auto_Instrument_With_StartupHook_Should_Capture_Error` (with `throw new Exception("...");` to `Elastic.Apm.StartupHook.Sample`) on `master`, it'll be green. With my testing today and this fact I think the original problem is more like an edge case.

 I think the main change here is that we load dependencies into an extra `AssemblyLoadContext` - I think that does not change how the ASP.NET Core diagnostic source events are sent once we already subscribed (and subscribing worked before as well).

I added some comments regarding this change below.

src/ElasticApmAgentStartupHook/StartupHook.cs
russcam: There are two problems that this is attempting to address:

1. The Elastic.Apm assemblies that contain Diagnostic Listeners must be loaded into the default load context to be able to subscribe to `DiagnosticListeners.AllListeners` and observe diagnostic events - subscription will not occur if loaded into a different load context.
2. Dependencies of Elastic.Apm assemblies cannot version conflict with dependencies of the application. In testing and debugging, the dependencies of the internalized Ben.Demystifier, namely, System.Reflection.Metadata, can very likely be a different version to the one referenced by the application, which was true for ASP.NET Core 3.0 if I recall.

Are there additional tests that can be written to provide more assurance in this approach?

src/Elastic.Apm.StartupHook.Loader/Loader.cs
russcam: Loading the loader assembly and invoking the `Initialize()` method causes the Elastic.Apm.* assemblies it references to be loaded, though this is implicit now rather than being explicitly loaded. Thoughts on a better name?

russcam:
> I'm not really sure this fixes #1233 - or changes anything around that.

When I first started looking at this, there were a couple of runs with ASP.NET Core 3.0 where an error was not captured. I don't recall being able to replicate, but I did find an issue with a `TypeLoadException` thrown when attempting to load the agent referenced version of System.Reflection.Metadata, to demystify a stack trace, which led to the loading changes in this PR.

src/ElasticApmAgentStartupHook/StartupHook.cs
gregkalapos: > Are there additional tests that can be written to provide more assurance in this approach?

I did some manual testing and so far everything worked. Maybe later we could add more tests where:
- We make sure we also trigger stack trace capturing (e.g. making an outgoing HTTP call in the sample app)
- Where we have agent assemblies with additional dependencies (e.g. using elastisearch-net and make the app depend on a different version than the agent)

I think those will trigger code paths that are critical. I did those tests manually and those worked for me.

src/Elastic.Apm.StartupHook.Loader/Loader.cs
gregkalapos: > Thoughts on a better name?

Not really - and you have a good point, it still triggers the agent dlls to be loaded. Let's leave it for now.

gregkalapos:
We discussed this in a meeting - we agreed to merge this as it is.

The `TypeLoadException` related to `Ben.Demystifier` is a real issue which we should address. We (or rather I :) ) still don't fully see if and how this changes #1233 - but we can follow up on that. 

Merging this in now.

PR #1299, opened 2021-05-10
Fix SanitizeFieldNamesTests

gregkalapos:
One of the tests seemed to fail - relaxing one less-relevant assert.

Also found another issue where we changed the dictionary during enumeration.

### DIFF
--- a/src/Elastic.Apm/Filters/ErrorContextSanitizerFilter.cs
+++ b/src/Elastic.Apm/Filters/ErrorContextSanitizerFilter.cs
@@ -22,7 +22,7 @@ namespace Elastic.Apm.Filters
 			{
 				if (realError.Context.Request?.Headers != null && realError.ConfigSnapshot != null)
 				{
-					foreach (var key in realError.Context?.Request?.Headers?.Keys)
+					foreach (var key in realError.Context?.Request?.Headers?.Keys.ToList())
 					{
 						if (WildcardMatcher.IsAnyMatch(realError.ConfigSnapshot.SanitizeFieldNames, key))
 							realError.Context.Request.Headers[key] = Consts.Redacted;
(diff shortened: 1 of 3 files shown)


test/Elastic.Apm.AspNetCore.Tests/SanitizeFieldNamesTests.cs
```diff
@@ -320,7 +320,7 @@ public async Task SanitizeHeadersOnError(string headerName, bool useOnlyDiagnost
 			_capturedPayload.FirstTransaction.Context.Request.Headers[headerName].Should().Be("[REDACTED]");
 
 			_capturedPayload.WaitForErrors();
-			_capturedPayload.Errors.Should().ContainSingle();
+			_capturedPayload.Errors.Should().NotBeEmpty();
```
russcam, line 323: Is capturing more than one error now expected?

gregkalapos: > Is capturing more than one error now expected?

I debugged it and I saw a failing outgoing HTTP request into some resource which looked like part of the sample app. Below we specifically assert that the first error must contain the header and the value must be `[REDACTED]`, so I thought the easiest is to just relax this part and make sure there is at least 1 error.

src/Elastic.Apm/Filters/ErrorContextSanitizerFilter.cs
```diff
@@ -22,7 +22,7 @@ public IError Filter(IError error)
 			{
 				if (realError.Context.Request?.Headers != null && realError.ConfigSnapshot != null)
 				{
-					foreach (var key in realError.Context?.Request?.Headers?.Keys)
+					foreach (var key in realError.Context?.Request?.Headers?.Keys.ToList())
```
gregkalapos, line 25: We need `.ToList()` here, otherwise we can't change the dictionary within the `foreach`. Same in `HeaderDictionarySanitizerFilter.cs `

russcam (approved):

gregkalapos:
I see some failing tests, they seem to be related to the docker issue we have, e.g.:

```
System.TypeInitializationException : The type initializer for 'DotNet.Testcontainers.Client.DockerApiClient' threw an exception.
---- System.TimeoutException : The operation has timed out.
```

PR #1302, opened 2021-05-19
Prefer W3C traceparent over elastic-apm-traceparent

gregkalapos:
Prefer `traceparent` if present over `elastic-apm-traceparent`.

This PR does it for ASP.NET Core, fun fact: in the IIS Module for ASP.NET Classic we already have this precedence.


felixbarny:
Relates to https://github.com/elastic/apm/pull/442

russcam (approved):

PR #1305, opened 2021-05-25
Use Logger to log exception in AgentComponents initialization

### LINKED ISSUE #1254: Object Reference after updating 1.8.1 -> 1.9 (Elastic.Apm.Agent.Tracer is null)
### APM Agent version
Elastic.Apm 1.9.

### Environment
Windows
.Net Framework 4.7.2

### Describe the bug
It looks like Elastic.Apm.Agent.Tracer behavior changed in 1.9 (previously was 1.8.1)
In DEBUG mode (local execution) we don't Initialize APM.
After updating the assembly Elastic.Apm.Agent.Tracer became null

For example, we have an Object reference
1. Windows Service 
` var transaction = Elastic.Apm.Agent.Tracer.StartTransaction(FullName, ApiConstants.TypeRequest);`
2. WebSite in APM SQL Interceptor
image

It is not critical, I've fixed it all on our side. This Issue FYI only, feel free to close it.

russcam:
This commit fixes a bug in logging an exception that may occur during agent initialization; the Logger property is used instead of the logger argument, the former will always be initialized with a value.

Fixes #1254

### DIFF
--- a/src/Elastic.Apm/AgentComponents.cs
+++ b/src/Elastic.Apm/AgentComponents.cs
@@ -53,22 +53,21 @@ namespace Elastic.Apm
 
 				HttpTraceConfiguration = new HttpTraceConfiguration();
 
+				TracerInternal = new Tracer(Logger, Service, PayloadSender, ConfigStore,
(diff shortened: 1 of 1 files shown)


gregkalapos (approved):
LGTM

PR #1308, opened 2021-05-26
Create receive messaging span when inside transaction

russcam:
This commit updates the Azure Service Bus integrations to create messaging spans for receiving messages from
a queue or subscription when inside a traced transaction.


gregkalapos (approved):
LGTM

PR #1310, opened 2021-05-26
Implement config to ignore the sampled flag in traceparent

gregkalapos:
### The problem 

We started seeing environments where the full trace is not within elastic and the sample rate is effectively 0% regardless of the `TransactionSampleRate` setting. Very likely this is because unmonitored .NET5+ services send a `traceparent` header with a sampled flag set to `0` ([doc on the change from Microsoft](https://docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/5.0/default-activityidformat-changed#change-description))

### Proposed solution in this PR

We add a new setting called `SuppressTraceContextHeaders`. If this is `true` we ignore any distributed tracing data during transaction start and we make a new sampling decision regardless of the `traceparent` value.

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -1045,6 +1045,31 @@ you must instead set the `LogLevel` for the internal APM logger under the `Loggi
 | `Error`                 | String
 |============
 
+[float]
+[[config-suppress-tracecontext-headers]]
+==== `SuppressTraceContextHeaders`
+
+By setting this to `true` the agent will ignore the W3C TraceContext headers.
+In practice this means that in case a caller service calls another service where this value is `true`, 
+the agent will ignore the known W3C headers, and it will start a new trace.
+
+This can be useful when a caller service always sets the sampled flag to false and the agent would have no chance to create any sampled transaction.
+
+NOTE: .NET 5 by default sets the W3C TraceContext headers, but without an active agent it sets the sampled flag to `false` leading to 0% sampling. If your service is called by an unmonitored .NET 5 application, setting this flag will instruct the agent to start a new trace and make a new sampling decision based on <<config-transaction-sample-rate>>.
+
+[options="header"]
+|============
+| Environment variable name | IConfiguration or Web.config key
+| `SUPPRESS_TRACECONTEXT_HEADERS`   | `ElasticApm:SuppressTraceContextHeader`
+|============
+
+[options="header"]
+|============
+| Default                 | Type
+| `false`                 | Boolean
+|============
+
+
 [[config-all-options-summary]]
 === All options summary
 
@@ -1080,6 +1105,7 @@ you must instead set the `LogLevel` for the internal APM logger under the `Loggi
 | <<config-service-version,`ServiceVersion`>> | No | Core
 | <<config-span-frames-min-duration,`SpanFramesMinDuration`>> | No | Stacktrace, Performance
 | <<config-stack-trace-limit,`StackTraceLimit`>> | No | Stacktrace, Performance
+| <<config-suppress-tracecontext-headers,`SuppressTraceContextHeaders`>> | No | Core
 | <<config-transaction-ignore-urls,`TransactionIgnoreUrls`>>  | Yes | HTTP, Performance
 | <<config-transaction-max-spans,`TransactionMaxSpans`>>  | Yes | Core, Performance
 | <<config-transaction-sample-rate,`TransactionSampleRate`>> | Yes | Core, Performance
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -1046,21 +1046,22 @@ you must instead set the `LogLevel` for the internal APM logger under the `Loggi
 |============
 
 [float]
-[[config-suppress-tracecontext-headers]]
-==== `SuppressTraceContextHeaders`
+[[config-trace-context-ignore-sampled-false]]
+==== `TraceContextIgnoreSampledFalse`
 
-By setting this to `true` the agent will ignore the W3C TraceContext headers.
-In practice this means that in case a caller service calls another service where this value is `true`, 
-the agent will ignore the known W3C headers, and it will start a new trace.
+By setting this to `true` the agent will ignore the sampled part of the W3C TraceContext traceparent header when it's false and when the upstream transaction is not coming from a non-Elastic APM agent.
+In practice this means that in case a caller service calls another service where this value is `true`,
+the agent will ignore the sampling decision of the upstream service, and it will make a new sampling decision.
 
-This can be useful when a caller service always sets the sampled flag to false and the agent would have no chance to create any sampled transaction.
+This can be useful when a caller service always sets the sampled flag to false and the agent would have no chance to
+create any sampled transaction.
 
 NOTE: .NET 5 by default sets the W3C TraceContext headers, but without an active agent it sets the sampled flag to `false` leading to 0% sampling. If your service is called by an unmonitored .NET 5 application, setting this flag will instruct the agent to start a new trace and make a new sampling decision based on <<config-transaction-sample-rate>>.
 
 [options="header"]
 |============
-| Environment variable name | IConfiguration or Web.config key
-| `SUPPRESS_TRACECONTEXT_HEADERS`   | `ElasticApm:SuppressTraceContextHeader`
+| Environment variable name            | IConfiguration or Web.config key
+| `TRACE_CONTEXT_IGNORE_SAMPLED_FALSE` | `ElasticApm:TraceContextIgnoreSampledFalse`
 |============
 
 [options="header"]
@@ -1105,7 +1106,7 @@ NOTE: .NET 5 by default sets the W3C TraceContext headers, but without an active
 | <<config-service-version,`ServiceVersion`>> | No | Core
 | <<config-span-frames-min-duration,`SpanFramesMinDuration`>> | No | Stacktrace, Performance
 | <<config-stack-trace-limit,`StackTraceLimit`>> | No | Stacktrace, Performance
-| <<config-suppress-tracecontext-headers,`SuppressTraceContextHeaders`>> | No | Core
+| <<config-trace-context-ignore-sampled-false,`TraceContextIgnoreSampledFalse`>> | No | Core
 | <<config-transaction-ignore-urls,`TransactionIgnoreUrls`>>  | Yes | HTTP, Performance
 | <<config-transaction-max-spans,`TransactionMaxSpans`>>  | Yes | Core, Performance
 | <<config-transaction-sample-rate,`TransactionSampleRate`>> | Yes | Core, Performance
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -1049,21 +1049,18 @@ you must instead set the `LogLevel` for the internal APM logger under the `Loggi
 [[config-trace-context-ignore-sampled-false]]
 ==== `TraceContextIgnoreSampledFalse`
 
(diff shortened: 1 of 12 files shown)


gregkalapos:
Alternatives:

- make the setting into a URL pattern (set URLs where you ignore the headers and apply it for the rest)
- only apply when `tracestate=es:...` is NOT present (aka the request is not from our agent)


Maybe make it `TraceContextHeadersIgnore`.

russcam:
@gregkalapos, @AlexanderWert  Another alternative, which is along similar lines to ones proposed:

New configuration value: `TraceContextIgnoreSampledFalse`: `true|false`, with a default value of `false`.

If `TraceContextIgnoreSampledFalse` is set to `true`, when a transaction is started, if there is Trace Context:

- If the [sampled flag](https://www.w3.org/TR/trace-context/#sampled-flag) is 1, proceed as per current implementation
- If the [sampled flag](https://www.w3.org/TR/trace-context/#sampled-flag) is 0, check the tracestate
  - If tracestate contains `es` vendor and `s` value, proceed as per current implementation
  - If tracestate does not contain `es` vendor and `s` value
    1. determine whether the transaction should be sampled based on the transaction id and set `IsSampled` 
    2. set the sample rate from the sampler
    3. use the trace id, parent id of Trace Context
    4. [mutate tracestate](https://www.w3.org/TR/trace-context/#mutating-the-tracestate-field) of Trace Context to add `es` vendor and `s` value to and use

The end result is that sampled flag value of 0 is ignored when tracestate does not contain `es` vendor and `s` value

gregkalapos:
> @gregkalapos, @AlexanderWert Another alternative, which is along similar lines to ones proposed:
> 
> New configuration value: `TraceContextIgnoreSampledFalse`: `true|false`, with a default value of `false`.
> 
> If `TraceContextIgnoreSampledFalse` is set to `true`, when a transaction is started, if there is Trace Context:
> 
> * If the [sampled flag](https://www.w3.org/TR/trace-context/#sampled-flag) is 1, proceed as per current implementation
> * If the [sampled flag](https://www.w3.org/TR/trace-context/#sampled-flag) is 0, check the tracestate
>   
>   * If tracestate contains `es` vendor and `s` value, proceed as per current implementation
>   * If tracestate does not contain `es` vendor and `s` value
>     
>     1. determine whether the transaction should be sampled based on the transaction id and set `IsSampled`
>     2. set the sample rate from the sampler
>     3. use the trace id, parent id of Trace Context
>     4. [mutate tracestate](https://www.w3.org/TR/trace-context/#mutating-the-tracestate-field) of Trace Context to add `es` vendor and `s` value to and use
> 
> The end result is that sampled flag value of 0 is ignored when tracestate does not contain `es` vendor and `s` value

@russcam the suggestion sounds good to me. 

The advantage is clearly that this would solve the issues we have with .NET5+ unmonitored services while it causes the least harm in other environments.

The only disadvantage I see is that people will have a super hard time to understand this - I don't expect users to understand the W3C TraceContext spec, so it'll be super hard to explain in the docs what this config exactly does.

Overall I think this is a good way to proceed and worth accepting the more complicated setting - I'll update the PR.

gregkalapos:
Pushed the changes discussed above - ready for another review round.

docs/configuration.asciidoc
```diff
@@ -1046,6 +1046,32 @@ you must instead set the `LogLevel` for the internal APM logger under the `Loggi
 | `Error`                 | String
 |============
 
+[float]
+[[config-trace-context-ignore-sampled-false]]
+==== `TraceContextIgnoreSampledFalse`
+
+By setting this to `true` the agent will ignore the sampled part of the W3C TraceContext traceparent header when it's false and when the upstream transaction is not coming from a non-Elastic APM agent.
```
russcam, line 1053: It might be useful to explain the "normal" flow when Trace Context is received, and then lead into describing what this setting does. In doing so, I think it would provide more context around why it is needed and if it might be relevant to a specific user.

docs/configuration.asciidoc
```diff
@@ -1046,6 +1046,32 @@ you must instead set the `LogLevel` for the internal APM logger under the `Loggi
 | `Error`                 | String
 |============
 
+[float]
+[[config-trace-context-ignore-sampled-false]]
+==== `TraceContextIgnoreSampledFalse`
```
russcam, line 1051: Wondering if there might be a better name for this setting? It's a little awkward with the sampled flag value to which it applies included in the name, but it's hard coming up with something clear 😃 

I know it's more verbose, but maybe `IgnoreTraceContextSampledFalseNoAgentTracestate` better describes (and aligns with `IgnoreMessageQueues` naming)?

docs/configuration.asciidoc
```diff
@@ -1046,6 +1046,32 @@ you must instead set the `LogLevel` for the internal APM logger under the `Loggi
...
+
+By setting this to `true` the agent will ignore the sampled part of the W3C TraceContext traceparent header when it's false and when the upstream transaction is not coming from a non-Elastic APM agent.
+In practice this means that in case a caller service calls another service where this value is `true`,
+the agent will ignore the sampling decision of the upstream service, and it will make a new sampling decision.
+
+This can be useful when a caller service always sets the sampled flag to false and the agent would have no chance to
+create any sampled transaction.
+
+NOTE: .NET 5 by default sets the W3C TraceContext headers, but without an active agent it sets the sampled flag to `false` leading to 0% sampling. If your service is called by an unmonitored .NET 5 application, setting this flag will instruct the agent to start a new trace and make a new sampling decision based on <<config-transaction-sample-rate>>.
```
russcam, line 1060:
```suggestion
[IMPORTANT]
--
.NET 5 applications set the https://www.w3.org/TR/trace-context/[W3C Trace Context] headers for outgoing HTTP requests by default, but with the traceparent sampled flag set to `false`. If the application has an active agent, the agent will ensure that the sampled flag is propagated appropriately. If the application does not have an active agent however, and the application calls another application (the callee) that does have an active agent, the propagation of a sampled flag of `false` results in no sampled traces in the callee application.

If your application is called by an unmonitored .NET 5 application, setting `TraceContextIgnoreSampledFalse` to `true` will instruct the agent to start a new trace and make a new sampling decision based on <<config-transaction-sample-rate>> when the traceparent sampled flag is `false` and there is no agent specific tracestate present.
--
```

Maybe the section should lead with this?

russcam (commented):
I've had an initial look and left some comments

src/Elastic.Apm/Config/ConfigConsts.cs
```diff
@@ -32,6 +32,7 @@ public static class DefaultValues
 			public const string SpanFramesMinDuration = "5ms";
 			public const double SpanFramesMinDurationInMilliseconds = 5;
 			public const int StackTraceLimit = 50;
+			public const bool SuppressTraceContextHeaders = false;
```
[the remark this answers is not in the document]
gregkalapos: 👍  Fixed this and the usage of it as well.

docs/configuration.asciidoc
gregkalapos: Done

docs/configuration.asciidoc
gregkalapos: Sounds good, moved this to the top of the section.

docs/configuration.asciidoc
gregkalapos: That sounds super verbose to me. I think this is complicated enough that we won't find a simple name - I think the docs explains it well and the shorter name is easier.

docs/configuration.asciidoc
```diff
@@ -1045,6 +1045,39 @@ you must instead set the `LogLevel` for the internal APM logger under the `Loggi
 | `Error`                 | String
 |============
 
+[float]
+[[config-trace-context-ignore-sampled-false]]
+==== `TraceContextIgnoreSampledFalse`
+
+[IMPORTANT]
+--
+..NET 5 applications set the https://www.w3.org/TR/trace-context/[W3C Trace Context] headers for outgoing HTTP requests by default, but with the traceparent sampled flag set to `false`. If the application has an active agent, the agent will ensure that the sampled flag is propagated appropriately. If the application does not have an active agent however, and the application calls another application (the callee) that does have an active agent, the propagation of a sampled flag of `false` results in no sampled traces in the callee application.
```
gregkalapos, line 1054: It seems we need an additional `.` to render `.NET` in asciidoc when it's the 1. word in the section.

russcam: This looks like it renders the paragraph in red italics in asciidoctor.js

image

An attribute entry can be used

```suggestion
:dot: .

{dot}NET 5 applications set the https://www.w3.org/TR/trace-context/[W3C Trace Context] headers for outgoing HTTP requests by default, but with the traceparent sampled flag set to `false`. If the application has an active agent, the agent will ensure that the sampled flag is propagated appropriately. If the application does not have an active agent however, and the application calls another application (the callee) that does have an active agent, the propagation of a sampled flag of `false` results in no sampled traces in the callee application.
```

which renders as

image

docs/configuration.asciidoc
```diff
@@ -1045,6 +1045,39 @@ you must instead set the `LogLevel` for the internal APM logger under the `Loggi
...
+--
+
+The agent uses the W3C TraceContext standard for distributed tracing. The traceparent header from the W3C TraceContext standard defines a flag called `sampled`, which stored whether a trace is sampled or not and by default the agent accepts this flag and behaves accordingly.
+
+By setting this config to `true`, the agent will ignore the sampled part of the W3C TraceContext traceparent header when it's false and when the upstream transaction is not coming from a non-Elastic APM agent.
+In practice this means that in case a caller service calls another service where this value is `true`,
+the agent will ignore the sampling decision of the upstream service, and it will make a new sampling decision.
+
+This can be useful when a caller service always sets the sampled flag to false and the agent would have no chance to
+create any sampled transaction.
```
russcam, line 1066: I've reworded slightly to use consistent phrasing and clarify the subject

```suggestion
The agent uses the https://www.w3.org/TR/trace-context/[W3C Trace Context] specification and standards for distributed tracing. The traceparent header from the W3C Trace Context specification defines a https://www.w3.org/TR/trace-context/#sampled-flag[sampled flag] which is propagated from a caller service to a callee service, and determines whether a trace is sampled in the callee service. The default behavior of the agent honors the sampled flag value and behaves accordingly.

There may be cases where you wish to change the default behavior of the agent with respect to the sampled flag. By setting the `TraceContextIgnoreSampled` configuration value to `true`, the agent ignores the sampled flag of the W3C Trace Context traceparent header when it has a value of `false` **and** and there is no agent specific tracestate header value present. In ignoring the sampled flag, the agent makes a sampling decision based on the <<config-transaction-sample-rate, sample rate>>. This can be useful when a caller service always sets a sampled flag value of `false`, that results in the agent never sampling any transactions.

[IMPORTANT]
--
:dotnet5: .NET 5

{dotnet5} applications set the W3C Trace Context for outgoing HTTP requests by default, but with the traceparent header sampled flag set to `false`. If a {dotnet5} application has an active agent, the agent ensures that the sampled flag is propagated with the agent's sampling decision. If a {dotnet5} application does not have an active agent however, and the application calls another service that does have an active agent, the propagation of a sampled flag value of `false` results in no sampled transactions in the callee service.

If your application is called by an {dotnet5} application that does not have an active agent, setting the `TraceContextIgnoreSampledFalse` configuration value to `true` instructs the agent to start a new transaction and make a sampling decision based on the <<config-transaction-sample-rate, sample rate>>, when the traceparent header sampled flag has a value of `false` **and** there is no agent specific tracestate header value present.
--
```

docs/configuration.asciidoc
russcam: Can ignore the suggestion here as it's included in the next suggestion

russcam (approved):
LGTM. I've left a suggestion for the docs

Issue #1309, opened 2021-05-26
NullReferenceException ApmErrorLogger/GenerateApmStackTrace


aheubusch:
### APM Agent version

Elastic.Apm 1.9.0
Elastic.Apm.Extensions.Hosting 1.9.0
Elastic.Apm.Extensions.Logging 1.9.0

### Environment

.NET Core 3.1

### Describe the bug

`ApmErrorLogger` passes a null logger to `StacktraceHelper.GenerateApmStackTrace` (second arg)
https://github.com/elastic/apm-agent-dotnet/blob/4591f02a53de6104c80fd3c56bce640e14c80602/src/Elastic.Apm.Extensions.Logging/ApmErrorLogger.cs#L45-L46
which then causes a NullReferenceException. Several try/catch keep the exception from reaching user code though.
https://github.com/elastic/apm-agent-dotnet/blob/4591f02a53de6104c80fd3c56bce640e14c80602/src/Elastic.Apm/Helpers/StacktraceHelper.cs#L51

### To Reproduce
```csharp
public static void Main()
{
    var host = Host
        .CreateDefaultBuilder()
        .ConfigureLogging(logging => logging.ClearProviders())
        .UseElasticApm()
        .Build();

    var logger = (ILogger)host.Services.GetService(typeof(ILogger<object>));
    logger.LogError(new Exception(), "error log with exception");
}
```

gregkalapos:
😱 

Thanks for reporting it - super useful summary 👍 

Opened https://github.com/elastic/apm-agent-dotnet/pull/1311 to fix it.

PR #1311, opened 2021-05-26
Fix nullref in Elastic.Apm.Extensions.Logging

gregkalapos:
Fixes https://github.com/elastic/apm-agent-dotnet/issues/1309 

Also extended the test to cover the case where the nullref happened.

### DIFF
--- a/src/Elastic.Apm.Extensions.Logging/ApmErrorLogger.cs
+++ b/src/Elastic.Apm.Extensions.Logging/ApmErrorLogger.cs
@@ -42,7 +42,7 @@ namespace Elastic.Apm.Extensions.Logging
 
 			if (_agent is ApmAgent apmAgent && exception != null)
 			{
-				errorLog.StackTrace = StacktraceHelper.GenerateApmStackTrace(exception, null, "CaptureErrorLogsAsApmError",
+				errorLog.StackTrace = StacktraceHelper.GenerateApmStackTrace(exception, _agent.Logger, "CaptureErrorLogsAsApmError",
 					apmAgent.ConfigurationReader, apmAgent.Components.ApmServerInfo);
 			}
 
--- a/src/Elastic.Apm/Helpers/StacktraceHelper.cs
+++ b/src/Elastic.Apm/Helpers/StacktraceHelper.cs
@@ -48,7 +48,7 @@ namespace Elastic.Apm.Helpers
 			var len = stackTraceLimit == -1 ? frames.Length : Math.Min(frames.Length, stackTraceLimit);
 			var retVal = new List<CapturedStackFrame>(len);
 
-			logger?.Trace()?.Log("transform stack frames");
+			logger.Trace()?.Log("transform stack frames");
 
 			try
 			{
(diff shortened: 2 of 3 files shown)


src/Elastic.Apm.Extensions.Logging/ApmErrorLogger.cs
```diff
@@ -42,7 +42,7 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
 
 			if (_agent is ApmAgent apmAgent && exception != null)
 			{
-				errorLog.StackTrace = StacktraceHelper.GenerateApmStackTrace(exception, null, "CaptureErrorLogsAsApmError",
```
gregkalapos, line 45: This was the problem 🤦

src/Elastic.Apm/Helpers/StacktraceHelper.cs
```diff
@@ -48,7 +48,7 @@ internal static class StacktraceHelper
 			var len = stackTraceLimit == -1 ? frames.Length : Math.Min(frames.Length, stackTraceLimit);
 			var retVal = new List<CapturedStackFrame>(len);
 
-			logger.Trace()?.Log("transform stack frames");
+			logger?.Trace()?.Log("transform stack frames");
```
russcam, line 51: The null conditional operator on `logger` looks unneccessary. `StacktraceHelper` is internal and `GenerateApmStackTrace` is called only by agent code where the logger should never be null (if it is null, it's a bug).

src/Elastic.Apm/Helpers/StacktraceHelper.cs
```diff
@@ -64,7 +64,7 @@ internal static class StacktraceHelper
 
 					var fileName = frame?.GetFileName();
 
-					logger.Trace()?.Log("{MethodName}, {lineNo}", functionName, frame?.GetFileLineNumber());
+					logger?.Trace()?.Log("{MethodName}, {lineNo}", functionName, frame?.GetFileLineNumber());
```
russcam, line 67: The null conditional operator on `logger` looks unneccessary. `StacktraceHelper` is internal and `GenerateApmStackTrace` is called only by agent code where the logger should never be null (if it is null, it's a bug).

src/Elastic.Apm.Extensions.Logging/ApmErrorLogger.cs
```diff
@@ -42,7 +42,7 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
...
+				errorLog.StackTrace = StacktraceHelper.GenerateApmStackTrace(exception, _agent.Logger, "CaptureErrorLogsAsApmError",
```
russcam, line 45: minor nitpick: 

```suggestion
				errorLog.StackTrace = StacktraceHelper.GenerateApmStackTrace(exception, apmAgent.Logger, "CaptureErrorLogsAsApmError",
```

It makes no difference either way, but reads more intuitively to me, since the other passed values come from `apmAgent`

russcam (approved):
LGTM.

The null conditional operator on `logger` inside `GenerateApmStackTrace` look unnecessary. Removing them would avoid performing a null check on each logger invocation.

src/Elastic.Apm/Helpers/StacktraceHelper.cs
gregkalapos: Ok, there were some calls on `logger` with `.?` and some without it - I removed it from all of them.

PR #1312, opened 2021-05-27
Skip running Elasticsearch docker test when docker not available

russcam:
This commit updates the Elasticsearch integration test
that runs against a docker instance to be skipped
when docker is not on the machine e.g. Windows
.NET Core tests in CI


russcam:
Fixes the issue on Windows .NET Core. Merging in.

PR #1315, opened 2021-05-28
Fix the variable name

vitorm04:
Fix the variable name

image


russcam (changes requested):
Thanks for opening @vitorm04.

I think it would be better to update 

```
app.UseElasticApm(Configuration);
```

to

```
app.UseElasticApm(_configuration);
```

If that change is made and the CLA is signed, I can pull the change in. Thanks!

russcam:
Superseded by https://github.com/elastic/apm-agent-dotnet/pull/1327

PR #1316, opened 2021-05-28
Don't package Elastic.Apm.Specification

russcam:
This commit sets Elastic.Apm.Specification as
a `<IsPackable>false</IsPackable>` project so
that it is not packaged into a nuget package for
release.

Releasing from CI uses a constrained list
in .ci/linux/deploy.sh to determine what can
be released, but if there is a need to release from
local, this constraint will not be adhered to.


gregkalapos (approved):
👍 

I also unlisted the unintentionally published package on NuGet.

Issue #1244, opened 2021-03-26
Expose configuration snapshot on IApmAgent


russcam:
`IApmAgent` exposes the starting configuration of the agent through `IConfigurationReader`:

https://github.com/elastic/apm-agent-dotnet/blob/c6b7cb5f974f07301153ccf64389c30003511f00/src/Elastic.Apm/Agent.cs#L17

The agent's configuration can be updated dynamically through central configuration, with updated configuration captured in `IConfigSnapshot` in `IConfigStore` on the concrete `ApmAgent` implementation:

https://github.com/elastic/apm-agent-dotnet/blob/c6b7cb5f974f07301153ccf64389c30003511f00/src/Elastic.Apm/Agent.cs#L37

The configuration snapshot is not however exposed on `IApmAgent`, which would be useful to have, for checking configuration values at runtime where the agent is referenced through the interface.

glucaci:
I will propose to:
1. Mark `ConfigurationReader` as `Obsolate` for 1.x and remove it in 2.x
```csharp
public interface IApmAgent
{
    [Obsolete("Use IApmAgent.ConfigurationSnapshot, this will be removed in 2.x")]
    IConfigurationReader ConfigurationReader { get; }
}
```
2. Expose `IConfigSnapshotProvider`
```csharp
public interface IApmAgent
{
    IConfigSnapshotProvider Configuration { get; }
}
```

pbalwall:
Thanks @russcam  and @glucaci  . Do we any idea when this feature could make it into a dotnet release?

glucaci:
@russcam do you have another proposal on this?

russcam:
@glucaci I haven't put further thought into this yet. At the moment, we're focused on finishing Azure Storage, Service Bus, CosmosDB, and MongoDB integrations for the next release. I'm hoping that we'll have some time to look at this afterwards.

I think configuration could be made easier to work with, though we may be limited in terms of binary compatibility in what we can do in 1.x; the end result might simply end up exposing `IConfigStore` in 1.x, but I think it'd be good to evaluate what the options are.

gregkalapos:
Done in https://github.com/elastic/apm-agent-dotnet/pull/1323

Issue #1144, opened 2021-01-25
Setting from the CentralConfiguration (ie via Kibana) not visible via Agent.Config


pbalwall:
Is there a way to view the latest APM Agent configuration, after incorporating the values retreived via the CentralConfiguration aspect of Agent configuration?

I am using Agent.Config, hoping that it would contain these config values, but it appears to solely have the configuration present in my web.config. I am using Agent 1.7.1.

Based on some debug logging done by the Agent, I can see that it does talk to the APMServer and does retrieve the current 'central configuration'. See below for a example message

Another question: from some of these messages (and from some docs) it implied that CaptureBodyContentTypes is also dynamically configurable (ie via central config). But the Kibana app does not allow its configuration?

--------------
{CentralConfigResponseParser} Processing HTTP response...
+-> Response:
    StatusCode: 200, ReasonPhrase: 'OK', Version: 1.1, Content: System.Net.Http.StreamContent, Headers:
    {
      Access-Control-Expose-Headers: Etag
      X-Content-Type-Options: nosniff
      Cache-Control: must-revalidate, max-age=30
      Date: Mon, 25 Jan 2021 22:45:45 GMT
      ETag: "a087b409681d7542dbded4f4516c67664a3d8aed"
      Content-Length: 212
      Content-Type: text/plain; charset=utf-8
    }
+-> Response body [length: 212]:
    {"capture_body":"errors","capture_headers":"false","log_level":"info","recording":"false","span_frames_min_duration":"10NaNms","stack_trace_limit":"0","transaction_max_spans":"10","transaction_sample_rate":".5"}

pbalwall:
To add more detail. It looks like the dotnet agent is indeed honoring the current central configuration. For example, if I set Recording=false in the central configuration (via Kibana), events dont get sent to Elasticsearch (as expected). My main issue is that the Agent.Config does not reflect the actual active configuration - it appears to be the static configuration present in web.config.

Is there a way to have access to the active configuration that the agent is acting upon?

gregkalapos:
> Is there a way to view the latest APM Agent configuration, after incorporating the values retreived via the CentralConfiguration aspect of Agent configuration?

[...]

PR #1323, opened 2021-06-02
Make IConfigSnapshot public

gregkalapos:
Attempt to solve https://github.com/elastic/apm-agent-dotnet/issues/1244 and https://github.com/elastic/apm-agent-dotnet/issues/1144

This PR makes `IConfigSnapshot ConfigSnapshot` public on `ITransaction`. The thing to keep in mind here is that configs can be changed at any time - but the `ConfigSnapshot` on `ITransaction` is fix (not changing) for the given transaction. The way it works is that when the transaction starts it takes a snapshot from the merged central and local config and attaches that to the transaction. With that callers see the same values throughout the transaction. If central config changes, values will be applied to the next transaction.

TBD:
- Should we also make `IConfigStore` on `Agent` public? (that's what the issue says) Some config may be needed before a transaction is started. The downside here is that this will change, so it only gives the current snapshot and using it within a transaction can be dangerous.

### DIFF
--- a/src/Elastic.Apm/Api/ITransaction.cs
+++ b/src/Elastic.Apm/Api/ITransaction.cs
@@ -4,6 +4,7 @@
 
 using System.Collections.Generic;
 using Elastic.Apm.Api.Constraints;
+using Elastic.Apm.Config;
 
 namespace Elastic.Apm.Api
 {
@@ -34,6 +35,14 @@ namespace Elastic.Apm.Api
 		/// Unlike <see cref="IExecutionSegment.Labels" /> the data in this property is not trimmed.
 		/// </summary>
 		Dictionary<string, string> Custom { get; }
+
+		/// <summary>
+		/// Holds configuration snapshot (which is immutable) that was current when this transaction started.
+		/// We would like transaction data to be consistent and not to be affected by possible changes in agent's configuration
+		/// between the start and the end of the transaction. That is why the way all the data is collected for the transaction
+		/// and its spans is controlled by this configuration snapshot.
+		/// </summary>
+		public IConfigSnapshot ConfigSnapshot { get; }
 
 		/// <summary>
 		/// A string describing the result of the transaction.
--- a/src/Elastic.Apm/Api/ITransaction.cs
+++ b/src/Elastic.Apm/Api/ITransaction.cs
@@ -40,7 +40,7 @@ namespace Elastic.Apm.Api
 		/// A snapshot of configuration from when the transaction started. A snapshot contains values
 		/// from initial configuration combined with dynamic values from central configuration, if enabled.
 		/// </summary>
-		public IConfigurationSnapshot ConfigurationSnapshot { get; }
+		public IConfiguration Configuration { get; }
 
 		/// <summary>
 		/// A string describing the result of the transaction.
--- a/src/Elastic.Apm/Api/ITransaction.cs
+++ b/src/Elastic.Apm/Api/ITransaction.cs
@@ -42,7 +42,7 @@ namespace Elastic.Apm.Api
 		/// between the start and the end of the transaction. That is why the way all the data is collected for the transaction
 		/// and its spans is controlled by this configuration snapshot.
 		/// </summary>
-		public IConfigSnapshot ConfigSnapshot { get; }
+		public IConfigurationSnapshot ConfigurationSnapshot { get; }
 
 		/// <summary>
 		/// A string describing the result of the transaction.
--- a/src/Elastic.Apm/Api/ITransaction.cs
+++ b/src/Elastic.Apm/Api/ITransaction.cs
@@ -37,10 +37,8 @@ namespace Elastic.Apm.Api
 		Dictionary<string, string> Custom { get; }
 
 		/// <summary>
-		/// Holds configuration snapshot (which is immutable) that was current when this transaction started.
-		/// We would like transaction data to be consistent and not to be affected by possible changes in agent's configuration
-		/// between the start and the end of the transaction. That is why the way all the data is collected for the transaction
-		/// and its spans is controlled by this configuration snapshot.
+		/// A snapshot of configuration from when the transaction started. A snapshot contains values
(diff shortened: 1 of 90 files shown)


src/Elastic.Apm/Config/IConfigSnapshotOptions.cs
```diff
@@ -1,8 +0,0 @@
-// Licensed to Elasticsearch B.V under one or more agreements.
-// Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
-// See the LICENSE file in the project root for more information
-
-namespace Elastic.Apm.Config
-{
-	internal interface IConfigSnapshotOptions : IConfigurationReader { }
```
gregkalapos, line 7: I don't really get the point of this interface, so just removed it.

src/Elastic.Apm/Api/ITransaction.cs
```diff
@@ -35,6 +36,14 @@ public interface ITransaction : IExecutionSegment
 		/// </summary>
 		Dictionary<string, string> Custom { get; }
 
+		/// <summary>
+		/// Holds configuration snapshot (which is immutable) that was current when this transaction started.
```
gregkalapos, line 40: This is the original docs, should we change it?

src/Elastic.Apm/Api/ITransaction.cs
```diff
@@ -35,6 +36,14 @@ public interface ITransaction : IExecutionSegment
...
+		/// We would like transaction data to be consistent and not to be affected by possible changes in agent's configuration
+		/// between the start and the end of the transaction. That is why the way all the data is collected for the transaction
+		/// and its spans is controlled by this configuration snapshot.
+		/// </summary>
```
russcam, line 44: I think we could make it more succinct e.g.

```suggestion
		/// <summary>
		/// A snapshot of configuration from when the transaction started. A snapshot contains values
		/// from initial configuration combined with dynamic values from central configuration
		/// </summary>
```

src/Elastic.Apm/Model/NoopTransaction.cs
```diff
@@ -39,6 +44,8 @@ public NoopTransaction(string name, string type, ICurrentExecutionSegmentsContai
 			ReusableContextInstance;
 
 		public Dictionary<string, string> Custom => _custom.Value;
+
+		public IConfigSnapshot ConfigSnapshot => _defaultConfigSnapshot;
```
russcam, line 48: Wondering whether this should also be a real snapshot of configuration, since only `ITransaction` is publicly exposed; I could imagine a scenario where a user might decide to take some action based on a dynamic configuration value from central config, which wouldn't be seen with the stubbed `DefaultConfigSnapshot`

src/Elastic.Apm/Config/IConfigSnapshot.cs
```diff
@@ -1,10 +1,18 @@
...
 {
-	internal interface IConfigSnapshot : IConfigSnapshotOptions
+	/// <summary>
+	/// This represents the snapshot of the merged local and central configuration values.
+	/// An instance of this is attached to each <see cref="Api.ITransaction" /> which holds a snapshot of the
+	/// merged config values from the point in time when the transaction started.
+	/// In case central config changes in the middle of a transaction, this snapshot won't chance. Instead changes will be
+	/// applied when the next transaction is created with its new snapshot.
+	/// </summary>
+	public interface IConfigSnapshot : IConfigurationReader
```
russcam, line 15: nitpick: consistent naming of _configuration_ types. There is `IConfigurationReader` but `IConfigSnapshot`, `IConfigStore`, etc. 

Suggest Config* types are renamed to Configuration*, since `IConfigurationReader` is public.

src/Elastic.Apm/Config/IConfigSnapshot.cs
```diff
@@ -1,10 +1,18 @@
...
 	{
 		string DbgDescription { get; }
```
russcam, line 17: This seems out of place on the public interface. Since it's only used for logging, perhaps can be implemented on concrete types by overriding `ToString()`, or some other non-public approach?

russcam (commented):
I've had an initial review and left some comments

src/Elastic.Apm/Config/IConfigSnapshot.cs
russcam: Moved this to a separate internal interface

src/Elastic.Apm/Config/IConfigSnapshot.cs
russcam: Renamed types to `Configuration*`. Kept the applicable namespaces as `.Config` as this would be a breaking change.

src/Elastic.Apm/Model/NoopTransaction.cs
russcam: Changed to use the real snapshot.

russcam:
Rebased this against master and force pushed

russcam:
I've moved the description for the configuration snapshot to an internal interface, as it is only used for logging.

@gregkalapos Since `IConfigurationSnapshot` is now simply `IConfigurationReader`, it could now also be removed, leaving a single interface to represent agent configuration, `IConfigurationReader`. We need to make a decision on this before merging this PR, as snapshot will become public. I'm in favour of removing `IConfigurationSnapshot` and having a single interface for agent configuration.

gregkalapos:
Summary of a discussion via zoom:

- We decided to rename `IConfigurationSnapshot` to `IConfiguration`, and keep that interface for now. Later in a breaking change we can harmonize `IConfigurationReader` (that's public so we don't touch it now to prevent breaking changes) and `IConfiguration`. Done in: https://github.com/elastic/apm-agent-dotnet/pull/1323/commits/6950435309ec33cd976d78eb2401abeffc112bb2
- Main reason to not just remove `IConfigurationSnapshot` is that by only having `IConfigurationReader` we'd have implementations that really read out configuration (e.g. from env. vars.) and other implementations that really have values in memory (e.g. the snapshot from central config merged with local configs).

PR #1358, opened 2021-07-26
Propagate Trace context in exit spans

### LINKED ISSUE #1344: Header Propagation does not work for Elasticsearch and Azure
Relates to: https://github.com/elastic/apm-agent-dotnet/pull/1306

This prevents unnecessary spans to be created for already instrumented services. The check however precludes these services from further header propagation.

https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerImplBase.cs#L127

### LINKED ISSUE #1350: Always propagate W3C headers to downstream services
Currently in HTTP monitoring if we don't create spans for a specific call (e.g. Elasticsearch or Azure services) we return in the diagnostic listener very early and don't inject any headers.

After discussion in the weekly meeting we concluded that agents should inject the headers for every outgoing HTTP call.

russcam:
This commit propagates the Trace Context for
exit span outgoing requests, without constructing
a child HTTP span, as per current behaviour.

The Trace context propagated is retrieved from
the current span.

Closes #1350
Closes #1344

### DIFF
--- a/src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerImplBase.cs
+++ b/src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerImplBase.cs
@@ -124,25 +124,28 @@ namespace Elastic.Apm.DiagnosticListeners
 
 		private void ProcessStartEvent(TRequest request, Uri requestUrl)
 		{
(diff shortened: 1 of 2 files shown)


gregkalapos (approved):
LGTM

russcam:
jenkins run tests

PR #1362, opened 2021-07-28
Unset parentId if TraceContextIgnoreSampledFalse is active

gregkalapos:
Follow up from https://github.com/elastic/apm-agent-dotnet/pull/1310 

In https://github.com/elastic/apm-agent-dotnet/pull/1310  we introduced the `TraceContextIgnoreSampledFalse` config. By setting that to `true` the agent will make a new sampling decision when the upstream service calls with `traceparent: sampled:false` and it's not from our agent. 

One issue this setting addresses is the problem that for transactions where the caller sends `traceparent: sampled:false` in Kibana we showed this:

image

Now, this problem was fixed in https://github.com/elastic/apm-agent-dotnet/pull/1310 - I doubled checked this. So even if there it's not a root transaction and the root is not in Elasticsearch, Kibana still properly fills that part.

On the other hand, such transaction will not show up under "traces" and it won't be part of any trace:
image

This PR makes sure that such transactions are root transactions by unsetting the parentid.

### DIFF
--- a/src/Elastic.Apm/Model/Transaction.cs
+++ b/src/Elastic.Apm/Model/Transaction.cs
@@ -238,6 +238,10 @@ namespace Elastic.Apm.Model
 				{
 					IsSampled = sampler.DecideIfToSample(idBytes);
 					_traceState?.SetSampleRate(sampler.Rate);
+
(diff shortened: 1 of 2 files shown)


russcam (approved):
LGTM

PR #1365, opened 2021-07-30
Update SystemInfoHelper.cs

gregkalapos:
We read container related infos from '/proc/self/cgroup'. This file exists on Linux, but if it's not a container then reading the container info won't return a result.

When this happens, currently we print a warning, but that is normal behaviour on Linux without containerization. We got reports that the warning log causes additional troubles in some environments.

This PR changes the log to info.


russcam (approved):
LGTM

Issue #1361, opened 2021-07-28
[BUG] Flooded with logs "The limit of 1000 metricsets has been reached, no new metricsets will be created"


tomap:
### APM Agent version
The version of the `Elastic.Apm` 1.11

### Environment

**Operating system and version:** both windows & docker linux

**.NET Framework/Core name and version** : .Net 5.0.8

**Application Target Framework(s)** (e.g. net461, netcoreapp3.1): ASP.Net.Sdk.Web

### Describe the bug
Since the last update, we re flooded with log message "The limit of 1000 metricsets has been reached, no new metricsets will be created"

comming from https://github.com/elastic/apm-agent-dotnet/blob/07a448406c78871a59088f87edaf3bcc191c2e67/src/Elastic.Apm/Metrics/MetricsProvider/BreakdownMetricsProvider.cs#L90

I don't understand the issue... how to fix it, ...

Here is our configuration: 

```
configuration[ConfigConsts.KeyNames.LogLevel] = "Debug";
                configuration[ConfigConsts.KeyNames.Environment] = app.ApplicationServices.GetRequiredService<IHostEnvironment>().EnvironmentName;
                configuration[ConfigConsts.KeyNames.CentralConfig] = "false";
                configuration[ConfigConsts.KeyNames.CloudProvider] = ConfigConsts.SupportedValues.CloudProviderNone;
                configuration[ConfigConsts.KeyNames.GlobalLabels] = $"product={ApiTitle}";
                configuration[ConfigConsts.KeyNames.ServiceVersion] = VersionLazy.Value;
                configuration[ConfigConsts.KeyNames.ServiceName] = ApiTitle;
                configuration[ConfigConsts.KeyNames.SanitizeFieldNames] = "password,passwd,pwd,secret,*key,*token*,*session*,*credit*,*card*,set-cookie,*auth*,*Newrelic*";
                configuration[ConfigConsts.KeyNames.StackTraceLimit] = "0";
                configuration[ConfigConsts.KeyNames.DisableMetrics] = "*";
                configuration[ConfigConsts.KeyNames.UseElasticTraceparentHeader] = "false";
                configuration[ConfigConsts.KeyNames.TransactionSampleRate] = "1"; // No sampling

                app.UseElasticApm(configuration,
                    new HttpDiagnosticsSubscriber(),
                    new SqlClientDiagnosticSubscriber(),
                    new EfCoreDiagnosticsSubscriber(),
[...]

PR #1367, opened 2021-08-02
Make sure BreakdownMetricsProvider prints 1K warning only once per collection

gregkalapos:
Solves https://github.com/elastic/apm-agent-dotnet/issues/1361

Prior to this PR once `BreakdownMetricsProvider` reached  the 1K metricset limit, it pinted a warning after each transaction. With a high number of transactions this can span logs.

This PR changes this to print this log only once in every metics collection round. So in every [MeticInterval](https://www.elastic.co/guide/en/apm/agent/dotnet/current/config-reporter.html#config-metrics-interval) (by default 30sec) we'll print the warning once if the 1K limit is reached. This also reflects reality better, since once the 1K limit is reached, we reset it after the current `List<MetricSet>` is sent to the server.

### DIFF
--- a/src/Elastic.Apm/Metrics/MetricsProvider/BreakdownMetricsProvider.cs
+++ b/src/Elastic.Apm/Metrics/MetricsProvider/BreakdownMetricsProvider.cs
@@ -4,7 +4,6 @@
 // See the LICENSE file in the project root for more information
 
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


russcam (approved):
LGTM

SbiCA:
Any chance that this fix will be released soon? It really creates noise in the logs

Issue #1359, opened 2021-07-27
Config for RequestBodyMaxLength 


fuzolan:
The fixed value of 2048 could be rather limiting. Is there a technical issue, to capture bigger bodies? 
If I'm not misstaken, the java client has a bigger limit here.

It would be really nice, if this would be a config parameter for the client.

sean-jost:
Have a need for this in .net agent.  Would be nice if we can get long_string_max_length implemented, as defined in elastic/apm/#488.

PR #1368, opened 2021-08-03
Use 10K limit for CaptureBody similar to the Java agent

gregkalapos:
From https://github.com/elastic/apm-agent-dotnet/issues/1359.

Change the request body limit to 10K - this is the limit that the Java agent [uses](https://www.elastic.co/guide/en/apm/agent/java/current/config-core.html#config-capture-body).


russcam (commented):
The change LGTM. 

The `http.request.body.content` field in span and transaction index mappings to which request body related is mapped as keyword datatype with ignore_above 1024. We might want to consider looking at changing the mapping, if folks are expecting to search on this field.

russcam (approved):

PR #1377, opened 2021-08-10
Cleanup Azure resource groups

adam-stokes:
Utilize a predefined AZ resource group prefix, we query the available resource
groups utilizing JMESPATH pulling out the group names. Finally passing that
through powershell's ForEach block to remove those remaining groups.

Related https://github.com/elastic/observability-robots/issues/497

Signed-off-by: Adam Stokes <51892+adam-stokes@users.noreply.github.com>

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -564,9 +564,9 @@ def withAzureAuth(Closure body){
     def props = getVaultSecret(secret: 'secret/apm-team/ci/apm-agent-dotnet-azure')
     def authObj = props?.data
     withEnvMask(vars: [
-        [var: 'AZ_CLIENT_ID', password: "${authObj.client_id}"]
-        [var: 'AZ_CLIENT_SECRET', password: "${authObj.client_secret}"]
-        [var: 'AZ_SUBSCRIPTION_ID', password: "${authObj.subscription_id}"]
+        [var: 'AZ_CLIENT_ID', password: "${authObj.client_id}"],
+        [var: 'AZ_CLIENT_SECRET', password: "${authObj.client_secret}"],
+        [var: 'AZ_SUBSCRIPTION_ID', password: "${authObj.subscription_id}"],
         [var: 'AZ_TENANT_ID', password: "${authObj.tenant_id}"]
     ]) {
         cmd label: "Logging into Azure", script: "az login --service-principal --username ${AZ_CLIENT_ID} --password ${AZ_CLIENT_SECRET} --tenant ${AZ_TENANT_ID}"
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -512,11 +512,8 @@ pipeline {
     }
   }
   post {
-        cleanup {
-            withAzureAuth(){
-                sh label: "Checking and removing any Azure related resource groups",
-                    script: "for group in `az group list --query \"[?name | starts_with(@,'${AZURE_RESOURCE_GROUP_PREFIX})']\" --out json|jq .[].name`;do az group delete --name $group --no-wait --yes;done"
-            }
+    cleanup {
+      cleanupAzureResources()
       notifyBuildResult()
     }
   }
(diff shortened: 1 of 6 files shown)


adam-stokes:
@russcam @elastic/observablt-robots Looking for feedback on this approach, if worthwhile I may generalize this into a helper function to be put in our pipeline library.

Jenkinsfile
```diff
@@ -16,6 +16,7 @@ pipeline {
     OPBEANS_REPO = 'opbeans-dotnet'
     BENCHMARK_SECRET  = 'secret/apm-team/ci/benchmark-cloud'
     SLACK_CHANNEL = '#apm-agent-dotnet'
+    AZURE_RESOURCE_GROUP_PREFIX = "ci-dotnet"
```
russcam, line 19: This would need to include some unique identifier for a specific build run, so that the cleanup step only deletes resource groups created in this run. Is the Jenkins `BUILD_ID` unique per run? Could that be included?

Jenkinsfile
```diff
@@ -512,6 +513,10 @@ pipeline {
   }
   post {
     cleanup {
+            withAzureCredentials(path: "${homePath}", credentialsFile: '.credentials.json') {
+                powershell label: "Checking and removing any Azure related resource groups",
+                    script: "az group list --query \"[?name | starts_with(@,'${AZURE_RESOURCE_GROUP_PREFIX})']\" --out json|jq .[].name | ForEach {az group delete --name $_}"
```
russcam, line 518: Before `az group list` can run, 

1. [the Service Principal credentials in `.credentials.json` will need to be used to log in](https://docs.microsoft.com/en-us/cli/azure/authenticate-azure-cli#sign-in-with-a-service-principal). Something like the following (this needs checking/testing to ensure sensitive details are not captured in logs or displayed in CI)

```sh
az login --service-principal --username <clientId> --password <clientSecret> --tenant <tenantId>
```

2. Since a tenant can contain multiple subscriptions, the active subscription to work with should be set as the one from `.credentials.json`

```sh
az account set --subscription <subscriptionId>
```

Now, `az group list` is good to go

russcam:
```
jq .[].name
```
Could avoid needing to install `jq`, if it's just for this, using PowerShell's [`ConvertFrom-Json`](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/convertfrom-json?view=powershell-7.1). Also would want [`--nowait` and `--yes` when deleting groups](https://docs.microsoft.com/en-us/cli/azure/group?view=azure-cli-latest#az_group_delete). Something like

```powershell
.. | ConvertFrom-Json | ForEach { az group delete --name $_.name --no-wait --yes  }
```

russcam (changes requested):

Jenkinsfile
russcam: Would need a bash/sh variant too, for Linux build hosts?

Jenkinsfile
```diff
@@ -511,7 +512,17 @@ pipeline {
...
   }
   post {
-    cleanup {
+        cleanup {
+            withAzureAuth(){
+                whenTrue(isUnix()) {
+                    sh label: "Checking and removing any Azure related resource groups",
+                        script: "for group in `az group list --query \"[?name | starts_with(@,'${AZURE_RESOURCE_GROUP_PREFIX})']\" --out json|jq .[].name`;do az group delete --name $group --no-wait --yes;done"
+                }
+                whenFalse(isUnix()) {
```
adam-stokes, line 521: @russcam I think the cleanup always runs on the parent node which seems to be linux (ubuntu) is there a situation where this may not be the case? If not, we could probably remove this windows portion

russcam: I'm not super familiar with the Jenkins setup so will defer to @elastic/observablt-robots. If it helps, it doesn't matter from what node/host the cleanup task runs from, so long as the input parameters are correct.

Jenkinsfile
```diff
@@ -549,6 +560,25 @@ def dotnet(Closure body){
   }
 }
 
+def withAzureAuth(Closure body){
```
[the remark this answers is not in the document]
russcam: looks like withAzureCredentials will get the right set of field names, though it writes them to file. Ideally want a step that gets the field names from withAzureCredentials, but sets the values as environment variables 🙂 Parsing the file with jq would work though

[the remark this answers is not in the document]
adam-stokes: @v1v I looked at `withAzureEnv` however it looks like it's pulling different key names than what's provided in that closure.

@russcam I decided against `withAzureCredentials` as to not have to worry about masking the output from `jq` or reading the file in general. We already had closures for handling this task via vault so this was more straightforward.

cachedout:
@adam-stokes Just trying to move this one along before we hit the end of our sprint. :)

This looks like it's set to WIP but [your most recent comment suggested that you're waiting for reviews](https://github.com/elastic/observability-robots/issues/497#issuecomment-897633230). If this is ready to go, it's probably best to remove the WIP status and see if we can get the reviews needed to get this moving forward. Thanks.

mdelapenya (approved):

Jenkinsfile
```diff
@@ -549,6 +551,29 @@ def dotnet(Closure body){
...
+    def dockerCmd = "docker run --rm -i -v \$(pwd)/.azure:/root/.azure mcr.microsoft.com/azure-cli:latest"
+    withEnvMask(vars: [
+        [var: 'AZ_CLIENT_ID', password: "${authObj.client_id}"],
+        [var: 'AZ_CLIENT_SECRET', password: "${authObj.client_secret}"],
+        [var: 'AZ_SUBSCRIPTION_ID', password: "${authObj.subscription_id}"],
+        [var: 'AZ_TENANT_ID', password: "${authObj.tenant_id}"]
+    ]) {
+        dir("${BASE_DIR}"){
+            cmd label: "Create storage for Azure authentication",
+                script: "mkdir -p ${BASE_DIR}/.azure || true"
```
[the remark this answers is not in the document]
adam-stokes: Hey @russcam I believe the host is always ephemeral as it's booting a vm with `immutable`. But just to be extra cautious we can remove it after cleanup occurs

adam-stokes:
> The `AZURE_RESOURCE_GROUP_PREFIX` environment variable needs to be used on the .NET side, when spinning up resources with Terraform. Would you like me to push a commit to this PR to update this?

That would be awesome @russcam if you aren't able to get to it during your work hours I can tackle it in the morning as well.

Thanks!

russcam:
I've added a commit to use the `AZURE_RESOURCE_GROUP_PREFIX` environment variable

adam-stokes:
@russcam Ok everything is ready to go, the last hiccup is the resource name itself giving an error:

```
09:31:33  + docker run --rm -i -v /var/lib/jenkins/workspace/net_apm-agent-dotnet-mbp_PR-1377/apm-agent-dotnet/.azure:/root/.azure mcr.microsoft.com/azure-cli:latest az group list --query [?name | starts_with(@,'ci-dotnet-18')] --out json
09:31:34  + pwd
09:31:34  + docker run --rm -i -v /var/lib/jenkins/workspace/net_apm-agent-dotnet-mbp_PR-1377/apm-agent-dotnet/.azure:/root/.azure mcr.microsoft.com/azure-cli:latest az group delete --name "ci-dotnet-18-apm-ci-immutable-ubuntu-1804-1629463347797644918-cosmosdb-test" --no-wait --yes
09:31:35  ERROR: Parameter 'resource_group_name' must conform to the following pattern: '^[-\\w\\._\\(\\)]+$'.
```

The only reference I could find was https://github.com/Azure/azure-cli/issues/1284#issuecomment-476749240

russcam:
The resource name looks OK, [here are the rules](https://docs.microsoft.com/en-us/rest/api/resources/resource-groups/create-or-update#uri-parameters)

> Can include alphanumeric, underscore, parentheses, hyphen, period (except at end), and Unicode characters that match the allowed characters.
> Regex pattern: `^[-\w\._\(\)]+$`

adam-stokes:
> The resource name looks OK, [here are the rules](https://docs.microsoft.com/en-us/rest/api/resources/resource-groups/create-or-update#uri-parameters)
> 
> > Can include alphanumeric, underscore, parentheses, hyphen, period (except at end), and Unicode characters that match the allowed characters.
> > Regex pattern: `^[-\w\._\(\)]+$`

I think it was the `"` as jq without `--raw-output` will quote the strings when printing to stdout. This most recent change should address that

PR #1471, opened 2021-08-16
Sanitize Central config request URI and headers in logs

### LINKED ISSUE #1376: [BUG]  CentralConfigFetcher logs SecretToken and ApiKey from request headers
### APM Agent version
Version: v1.10.0

### Environment
OS: Windows 10
Net Framework version: NET Core 3.1.100

### Describe the bug
When server url is corrupted CentralConfigFetcher can't reach apm server and logs error with serialized request,
which contains sensitive data such apikey or authorization token.

### To Reproduce
1) Set to config "ServerUrls" corrupted address, set "ApiKey" or "SecretToken" configs.
2) Start app.
3) Open logs.
4) We can see ApiKey and SecretToken in logs.

#### Expected behavior
Apm sanitize sensitive data in request's headers before log and we can't see sensitive data.

#### Actual behavior
We can see sensitive data:

> +-> Request:
>     Method: GET, RequestUri: 'http://unavailable_address:8200/config/v1/agents?service.name=service_name&service.environment=Development', Version: 1.1, Content: <null>, Headers:
>     {
>       User-Agent: elasticapm-dotnet/1.10.0+ecfe0d02ee33d22706236b4cb31126d15a02507f
>       User-Agent: System.Net.Http/5.0.721.25508
>       User-Agent: .NET_5/5.0.7
>       Authorization: ApiKey SecretApiKey
>       traceparent: 00-edc8c158612d944da8c3a3b8e3a22322-c490329b63c0ac4a-00
>     }
> +-> Response: N/A
> +-> Response body [length: N/A]: N/A
> System.Net.Http.HttpRequestException: No such host is known. (unavailable_address:8200)

russcam:
This commit sanitizes the request URI and headers of
request to fetch central configuration, to redact username/password info
and sensitive headers.

- Move sanitization methods to separate Santization helper class.
- Add test to assert central config sensitive request details are redacted.
- Use Consts.Redacted wherever "[REDACTED]" is needed.

Fixes #1376

### DIFF
--- a/src/Elastic.Apm/Api/Http.cs
+++ b/src/Elastic.Apm/Api/Http.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Net.Http;
 using Elastic.Apm.Api.Constraints;
+using Elastic.Apm.Helpers;
 using Elastic.Apm.Report.Serialization;
 using Elastic.Apm.Libraries.Newtonsoft.Json;
 
@@ -23,7 +24,7 @@ namespace Elastic.Apm.Api
 		public string Method { get; set; }
 
 		/// <summary>
-		/// The Url in its original form as it was passed to the Agent, without sanitization or any other trimming.
+		/// The Url in its original form as it was passed to the Agent, without sanitization or trimming.
 		/// </summary>
 		internal Uri OriginalUrl => _originalUrl;
 
@@ -39,7 +40,7 @@ namespace Elastic.Apm.Api
 		public string Url
 		{
 			get => _url;
-			set => _url = Sanitize(value, out var newValue, out _originalUrl) ? newValue : value;
+			set => _url = Sanitization.TrySanitizeUrl(value, out var newValue, out _originalUrl) ? newValue : value;
 		}
 
 		/// <summary>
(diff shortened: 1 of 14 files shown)


gregkalapos (approved):
Changes LGTM.

Somewhat related: do we want to maybe change the log level of the log [here](https://github.com/elastic/apm-agent-dotnet/pull/1471/files#diff-9accb79784bd2ccde4608b9ad2348abd98e7678ad2819dc886283a1ed199cb4fR141-R152). We print this as a warning (or even error?) by default and people already asked about it multiple times. When central config is not enabled in Kibana, this is expected and normal behaviour - so at least in that case I think an info log would be enough.

russcam:
Info log level sounds good. It can probably do with some further refactoring to trim the log down to the essential information, and to detect, if possible, if central config is not enabled and log/not log appropriately.

russcam:
I've updated use the log level defined on the `FailedToFetchConfigException` thrown, for logging the exception. This is Debug for 403 and 404 responses and Error for everything else. I think this can be refactored in future, to simplify the flow further.

PR #1481, opened 2021-09-03
Use Kubernetes pod id determined from cgroup file

russcam:
This commit updates the creation of System info reported by the agent, to bring it
more in line with the Java agent implementation.

The implementation did not use the pod id determined from parsing lines in
/proc/self/cgroup file in the resulting Kubernetes metadata, instead using the container id
when Kubernetes environment variables are not set. This is now addressed.

### DIFF
--- a/src/Elastic.Apm/Helpers/SystemInfoHelper.cs
+++ b/src/Elastic.Apm/Helpers/SystemInfoHelper.cs
@@ -80,7 +80,7 @@ namespace Elastic.Apm.Helpers
 				_logger.Info()?.Log("Could not parse container ID from '/proc/self/cgroup' line: {line}", line);
 		}
 
-		internal Api.System ParseSystemInfo(string hostName)
+		internal Api.System GetSystemInfo(string hostName)
 		{
 			var detectedHostName = GetHostName();
 			var system = new Api.System { DetectedHostName = detectedHostName, ConfiguredHostName = hostName };
@@ -109,22 +109,19 @@ namespace Elastic.Apm.Helpers
 			try
 			{
 				using var sr = GetCGroupAsStream();
-				if (sr == null)
+				if (sr is null)
 				{
 					//just debug log, since this is normal on non-docker environments
 					_logger.Debug()?.Log("No /proc/self/cgroup found - the agent will not report container id");
 					return;
 				}
 
-				var line = sr.ReadLine();
-
-				while (line != null)
+				string line;
+				while ((line = sr.ReadLine()) != null)
 				{
 					ParseContainerId(system, reportedHostName, line);
 					if (system.Container != null)
 						return;
-
-					line = sr.ReadLine();
 				}
 			}
 			catch (Exception e)
@@ -154,22 +151,25 @@ namespace Elastic.Apm.Helpers
 				var podUid = Environment.GetEnvironmentVariable(PodUid);
 				var nodeName = Environment.GetEnvironmentVariable(NodeName);
 
-				if (@namespace != null || podName != null || podUid != null || nodeName != null)
+				var podUidNotNullOrEmpty = !string.IsNullOrEmpty(podUid);
+				var podNameNotNullOrEmpty = !string.IsNullOrEmpty(podName);
+
+				if (podUidNotNullOrEmpty || podNameNotNullOrEmpty || !string.IsNullOrEmpty(@namespace) || !string.IsNullOrEmpty(nodeName))
 				{
 					system.Kubernetes ??= new KubernetesMetadata();
 					system.Kubernetes.Namespace = @namespace;
 
-					if (nodeName != null)
+					if (!string.IsNullOrEmpty(nodeName))
 						system.Kubernetes.Node = new Api.Kubernetes.Node { Name = nodeName };
 
-					if (podUid != null || podName != null)
+					if (podUidNotNullOrEmpty || podNameNotNullOrEmpty)
 					{
-						// retain existing pod values when environment variables are null
+						// retain any existing pod values, and overwrite with environment variables values only if not null or empty
 						system.Kubernetes.Pod ??= new Pod();
-						if (podUid != null)
+						if (podUidNotNullOrEmpty)
 							system.Kubernetes.Pod.Uid = podUid;
 
-						if (podName != null)
+						if (podNameNotNullOrEmpty)
 							system.Kubernetes.Pod.Name = podName;
 					}
 				}
--- a/src/Elastic.Apm/Helpers/SystemInfoHelper.cs
+++ b/src/Elastic.Apm/Helpers/SystemInfoHelper.cs
@@ -14,35 +14,30 @@ namespace Elastic.Apm.Helpers
 {
 	internal class SystemInfoHelper
 	{
-		private const string ContainerUidRegexString = "^[0-9a-fA-F]{64}$";
-
-		private const string ShortenedUuidPattern = "^[0-9a-fA-F]{8}\\-[0-9a-fA-F]{4}\\-[0-9a-fA-F]{4}\\-[0-9a-fA-F]{4}\\-[0-9a-fA-F]{4,}";
-
-		private const string PodRegexString = @"(?:^/kubepods[\S]*/pod([^/]+)$)|"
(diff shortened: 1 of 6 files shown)


Parani-Resideo:
Hi - Will this change also address container metrics correlation from the .Net apps APM transaction screen.

russcam:
@Parani-Resideo This change will address issues where the kubernetes pod id is used to correlate metrics. Currently, this pod id is incorrectly set in the metadata that the .NET agent sends- it's incorrectly set to the container id parsed from /proc/self/cgroup, instead of the pod id parsed from  /proc/self/cgroup. Hopefully [the additional tests in the PR help to explain](https://github.com/elastic/apm-agent-dotnet/pull/1481/files#diff-12649fe67fccee669d1979df801c74d87a1c4ff40fc8a33058f2b444e9d9d03eR47-R74).

src/Elastic.Apm/Helpers/SystemInfoHelper.cs
```diff
@@ -161,27 +145,39 @@ protected virtual StreamReader GetCGroupAsStream()
...
-				return containerInfo != null
-					? new KubernetesMetadata { Pod = new Pod { Uid = containerInfo.Id, Name = hostName } }
-					: null;
-			}
+				var @namespace = Environment.GetEnvironmentVariable(Namespace);
+				var podName = Environment.GetEnvironmentVariable(PodName);
+				var podUid = Environment.GetEnvironmentVariable(PodUid);
+				var nodeName = Environment.GetEnvironmentVariable(NodeName);
+
+				if (@namespace != null || podName != null || podUid != null || nodeName != null)
```
gregkalapos, line 157: Maybe check with `string.IsNullOrEmpty` to also handle empty string.

src/Elastic.Apm/Helpers/SystemInfoHelper.cs
```diff
@@ -161,27 +145,39 @@ protected virtual StreamReader GetCGroupAsStream()
...
+					system.Kubernetes ??= new KubernetesMetadata();
+					system.Kubernetes.Namespace = @namespace;
+
+					if (nodeName != null)
+						system.Kubernetes.Node = new Api.Kubernetes.Node { Name = nodeName };
 
-			var kubernetesMetadata = new KubernetesMetadata { Namespace = @namespace };
-			if (podName != null || podUid != null) kubernetesMetadata.Pod = new Pod { Name = podName, Uid = podUid };
-			if (nodeName != null) kubernetesMetadata.Node = new Api.Kubernetes.Node { Name = nodeName };
+					if (podUid != null || podName != null)
```
gregkalapos, line 165: Same as above, maybe use `string.IsNullOrEmpty`.

src/Elastic.Apm/Helpers/SystemInfoHelper.cs
```diff
@@ -161,27 +145,39 @@ protected virtual StreamReader GetCGroupAsStream()
...
+					{
+						// retain existing pod values when environment variables are null
```
gregkalapos, line 167: Is this the intended line for this comment? 🤔 Seems a bit off to me. Here I'd either say 
>  retain existing pod values when environment variables are null, otherwise use them 

(so it indicates that the next lines will use them)
Or just move it to the end of the if block.

src/Elastic.Apm/Helpers/SystemInfoHelper.cs
```diff
@@ -83,26 +74,20 @@ private Container ParseContainerId(string line)
...
-
-			_logger.Info()?.Log("Could not parse container ID from '/proc/self/cgroup' line: {line}", line);
-			return null;
+			if (system.Kubernetes != null || _containerUidRegex.IsMatch(idPart) || _shortenedUuidRegex.IsMatch(idPart))
+				system.Container = new Container { Id = idPart };
+			else
+				_logger.Info()?.Log("Could not parse container ID from '/proc/self/cgroup' line: {line}", line);
 		}
 
 		internal Api.System ParseSystemInfo(string hostName)
```
gregkalapos, line 83: This is really a nit, but maybe call it `GetSystemInfo`? 

Reason I mention is that the signatures changed, all `ParseX` methods are now in `void ParseX(variableToFill)` format in this file, except this one. With that it'd also match the signiture/naming of the `GetHostName` method.

src/Elastic.Apm/Helpers/SystemInfoHelper.cs
```diff
@@ -161,27 +145,39 @@ protected virtual StreamReader GetCGroupAsStream()
...
 
-		internal KubernetesMetadata ParseKubernetesInfo(Container containerInfo, string hostName)
+		internal void ParseKubernetesInfo(Api.System system)
 		{
-			var @namespace = Environment.GetEnvironmentVariable(Namespace);
-			var podName = Environment.GetEnvironmentVariable(PodName);
-			var podUid = Environment.GetEnvironmentVariable(PodUid);
-			var nodeName = Environment.GetEnvironmentVariable(NodeName);
-
-			if (@namespace == null && podName == null && podUid == null && nodeName == null)
```
gregkalapos, line 171: Oh, so here was the bug... with all the `&&`...

gregkalapos (approved):
LGTM 👍 

I added some comments with very minor suggestions.

src/Elastic.Apm/Helpers/SystemInfoHelper.cs
russcam: The comment is in reference to the use of `??=`, to indicate that `new Pod()` is assigned _only_ when `system.Kubernetes.Pod` is null i.e. if it's not null, we want to retain any values that are already set, and will overwrite any existing values with `podUid` or `podName` from environment variables if they're not null.

src/Elastic.Apm/Helpers/SystemInfoHelper.cs
russcam: I considered this, but thought a user might want to set one of the values to an empty string, to overwrite a value parsed from /proc/self/cgroup. Happy to change, as treating an unset variable or variable with empty string the same is likely a better experience.

gregkalapos:
Hi @Parani-Resideo,

could you maybe test the latest agent release? (link: https://github.com/elastic/apm-agent-dotnet/releases/tag/1.11.1 - also available via NuGet). That contains this fix as well - it'd be nice to directly test it.

Parani-Resideo:
HI @gregkalapos , @russcam - We upgraded the agent version from 1.10.x to 1.11.1 and I had attached a document with the observation and issue details.

Metrics correlation is still failing and host details menu is removed from the investigate list (APM transaction screen) after removing the below listed environment variables, however the Pod metrics correlation is working without the below environment variables on the pod yaml configuration explicitly setting. Pls Note: APM .Net agent 1.10.x was having issue with the Pod metrics correlation and the below environment variables fixed the correlation issue, upon version 1.11.1 upgrade even without the below environment variables, pod correlation is working but container metrics correlation is still failing. Attached document got more details on the observation and issue with screen shot for your review. Pls advise with your feedback on these inputs.

- name: KUBERNETES_NODE_NAME
            valueFrom:
              fieldRef:
                fieldPath: spec.nodeName
          - name: KUBERNETES_POD_NAME
            valueFrom:
              fieldRef:
                fieldPath: metadata.name
          - name: KUBERNETES_NAMESPACE
            valueFrom:
              fieldRef:
                fieldPath: metadata.namespace
          - name: KUBERNETES_POD_UID
            valueFrom:
              fieldRef:
                fieldPath: metadata.uid
[Pod, Container Metrics Correlation and Host Details.docx](https://github.com/elastic/apm-agent-dotnet/files/7181392/Pod.Container.Metrics.Correlation.and.Host.Details.docx)

Thanks,
Parani.

russcam:
@Parani-Resideo I believe you are a customer that has an existing open case relating to this issue. If my understanding is correct, please can I request that we continue to communicate through the existing open case; this is the best way to manage and provide support, and share relevant documents. Thanks

Parani-Resideo:
Sure @russcam - I updated the open case as well with these observation and document, will get inputs from support team for next steps.

Thanks,
Parani.

PR #1487, opened 2021-09-13
ADO.NET instrumentation flags

russcam:
This commit

- adds instrumentation flags for ADO.NET
- creates the Tracer instance before creating and starting metrics or central configuration components.
- extracts Db span naming into an internal method to be in other assemblies


gregkalapos (approved):
LGTM

PR #1490, opened 2021-09-13
CgroupMetricsProvider: Only query on Linux, just log on other OSs

gregkalapos:
Inspired by [a discuss entry](https://discuss.elastic.co/t/warning-metriccollector-net5-0-windows/).

`CgroupMetricsProvider` prior to this PR always tried to open `/proc/self/cgroup` and if it failed, it printed a warning. On non-Linux OSs, this always resulted in a warning because the file does not exist in that case. This PR adds an OS check and on non-Linux OSs it just prints a trace log and returns (so no more warning).

### DIFF
--- a/src/Elastic.Apm/Metrics/MetricsProvider/CgroupMetricsProvider.cs
+++ b/src/Elastic.Apm/Metrics/MetricsProvider/CgroupMetricsProvider.cs
@@ -79,6 +79,10 @@ namespace Elastic.Apm.Metrics.MetricsProvider
 			_collectStatsInactiveFileBytes = IsSystemProcessCgroupMemoryStatsInactiveFileBytesEnabled(disabledMetrics);
 			_logger = logger.Scoped(nameof(CgroupMetricsProvider));
 
+			_cGroupFiles = FindCGroupFiles(procSelfCGroup, mountInfo);
+
+			IsMetricAlreadyCaptured = true;
+
 			if (!RuntimeInformation.IsOSPlatform(OSPlatform.Linux) && !_ignoreOs)
 			{
 				_logger.Trace()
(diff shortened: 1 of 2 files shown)


src/Elastic.Apm/Metrics/MetricsProvider/CgroupMetricsProvider.cs
```diff
@@ -82,6 +90,14 @@ internal class CgroupMetricsProvider : IMetricsProvider
 
 		private CgroupFiles FindCGroupFiles(string procSelfCGroup, string mountInfo)
 		{
+			if (!RuntimeInformation.IsOSPlatform(OSPlatform.Linux) && !_ignoreOs)
+			{
+				_logger.Trace()
+					?.Log("{MetricsProviderName} detected a non Linux OS, therefore"
+						+ " Cgroup metrics will not be reported", nameof(CgroupMetricsProvider));
+				return null;
+			}
```
russcam, line 99: This looks like it could moved to the start of the internal ctor, and return early if the condition is matched.

russcam (approved):
The changes look OK, I left a comment for consideration.

For `IMetricsProvider` instances, an instance is instantiated in order to check `IsEnabled(...)` on the instance. If the enabled check were to be a static method instead of an instance method, we could avoid creating an instance altogether when not enabled. Then, the OS check could be part of this check for `CGroupMetricsProvider`. Granted, it's convenient to have `IsEnabled()` be part of the `IMetricsProvider` interface, but the agent could avoid doing work if it were a static method.

src/Elastic.Apm/Metrics/MetricsProvider/CgroupMetricsProvider.cs
gregkalapos: Makes sense - done.

PR #1498, opened 2021-09-22
Activity integration 

gregkalapos:
Solves: https://github.com/elastic/apm-agent-dotnet/issues/1521 

Integration with `System.Diagnostics.Activity` - aka OpenTelemetry bridge.

With this, the agent will automatically listen for `Activity`s and turn those into Spans.


src/Elastic.Apm/DiagnosticListeners/KnownListners.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+	internal static class KnownListeners
+	{
+		public const string MicrosoftAspNetCoreHostingHttpRequestIn = "Microsoft.AspNetCore.Hosting.HttpRequestIn";
+		public const string SystemNetHttpHttpRequestOut = "System.Net.Http.HttpRequestOut";
+		public const string SystemNetHttpDesktopHttpRequestOut = "System.Net.Http.Desktop.HttpRequestOut";
+		public const string ApmTransactionActivityName = "ElasticApm.Transaction";
+
+
+		public static List<string> KnownListenersList => new()
```
russcam, line 18:
```suggestion
		public static HashSet<string> Set => new(StringComparer.Ordinal)
```

This could be a `HashSet<string>` with case sensitive ordinal comparison

src/Elastic.Apm/OpenTelemetry/ElasticActivityListener.cs
```diff
@@ -0,0 +1,173 @@
...
+		private readonly IApmLogger _logger;
+		internal Tracer _tracer;
+
+
+		private Action<Activity> ActivityStarted =>
+			activity =>
+			{
+				_logger.Trace()?.Log($"ActivityStarted: name:{activity.DisplayName} id:{activity.Id} traceId:{activity.TraceId}");
+
+				if (KnownListeners.KnownListenersList.Contains(activity.DisplayName))
```
russcam, line 50: It looks like this is used to not duplicate transactions/spans for Activities for which there is already a DiagnosticListener. For those Activity backed by an ActivitySource, would implementing `listener.ShouldListenTo` perform better?

Should this filtering be extensible to take into account other Elastic APM integrations that are backed by activities, like Azure Service Bus, Storage, MongoDB, and other custom ones a user may have implemented?

src/Elastic.Apm/OpenTelemetry/ElasticActivityListener.cs
```diff
@@ -0,0 +1,173 @@
...
+
+					if (activity.Id != null) ActiveTransactions.TryAdd(activity.Id, transaction);
+				}
+			};
+
+
+		private Action<Activity> ActivityStopped =>
+			activity =>
+			{
+				if (activity == null)
```
russcam, line 108: Is this ever true?

src/Elastic.Apm/OpenTelemetry/ElasticActivityListener.cs
```diff
@@ -0,0 +1,173 @@
...
+					transaction.Duration = activity.Duration.TotalMilliseconds;
+
+					if (activity.Tags.Any())
+						transaction.Otel.Attributes = new Dictionary<string, string>();
+
+					foreach (var tag in activity.Tags) transaction.Otel.Attributes.Add(tag.Key, tag.Value);
+					transaction.End();
+				}
+
+				if (activity.Id != null && ActiveSpans.TryRemove(activity.Id, out var span))
```
russcam, line 130:
```suggestion
				else if (activity.Id != null && ActiveSpans.TryRemove(activity.Id, out var span))
```

Looks like this could be an `else if` to the previous `if` condition (the branching could also be inside an `if (activity.Id !- null) { /** try remove active transaction, else try remove active span  */ }`

src/Elastic.Apm/OpenTelemetry/ElasticActivityListener.cs
```diff
@@ -0,0 +1,173 @@
...
+					InferSpanTypeAndSubType(span, activity);
+					span.End();
+				}
+			};
+
+		public ActivityListener listener { get; }
+
+		private void InferSpanTypeAndSubType(Span span, Activity activity)
+		{
+			if (activity.Kind == ActivityKind.Client && activity.Tags.Any(n => n.Key == "http.url" || n.Key == "http.scheme"))
```
russcam, line 148: I think this could filter on activity.Kind first, then check tags, to avoid checking kind on each condition

src/Elastic.Apm.AspNetCore/Elastic.Apm.AspNetCore.csproj
```diff
@@ -35,6 +35,7 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="5.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="5.0.0" />
     <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="5.0.0" />
+    <ProjectReference Include="..\Elastic.Apm.OpenTelemetry\Elastic.Apm.OpenTelemetry.csproj" />
```
russcam, line 38: project doesn't exist?

src/Elastic.Apm/DiagnosticListeners/KnownListners.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Collections.Generic;
+
+namespace Elastic.Apm.DiagnosticListeners
+{
+	internal static class KnownListeners
+	{
+		public const string MicrosoftAspNetCoreHostingHttpRequestIn = "Microsoft.AspNetCore.Hosting.HttpRequestIn";
+		public const string SystemNetHttpHttpRequestOut = "System.Net.Http.HttpRequestOut";
+		public const string SystemNetHttpDesktopHttpRequestOut = "System.Net.Http.Desktop.HttpRequestOut";
+		public const string ApmTransactionActivityName = "ElasticApm.Transaction";
```
russcam, line 15: Would it make sense to keep these as constants in their respective types, and just collate the set of them here?

src/Elastic.Apm/AgentComponents.cs
```diff
@@ -72,6 +72,13 @@ public class AgentComponents : IApmAgent, IDisposable
 				}
 				else
 					Logger.Info()?.Log("The Elastic APM .NET Agent is disabled - the agent won't capture traces and metrics.");
+
+				TracerInternal = new Tracer(Logger, Service, PayloadSender, ConfigurationStore,
+					currentExecutionSegmentsContainer ?? new CurrentExecutionSegmentsContainer(), ApmServerInfo, breakdownMetricsProvider);
+
+#if NET5_0
+				new OpenTelemetry.ElasticActivityListener(this, TracerInternal);
+#endif
```
russcam, line 81: Since the Elastic.Apm takes a dependency on System.Diagnostics.DiagnosticSource 5.0.0, can the listener be enabled for all supported TFMs?

src/Elastic.Apm/AgentComponents.cs
```diff
@@ -72,6 +72,13 @@ public class AgentComponents : IApmAgent, IDisposable
 				}
 				else
 					Logger.Info()?.Log("The Elastic APM .NET Agent is disabled - the agent won't capture traces and metrics.");
+
+				TracerInternal = new Tracer(Logger, Service, PayloadSender, ConfigurationStore,
+					currentExecutionSegmentsContainer ?? new CurrentExecutionSegmentsContainer(), ApmServerInfo, breakdownMetricsProvider);
+
+#if NET5_0
+				new OpenTelemetry.ElasticActivityListener(this, TracerInternal);
```
russcam, line 80: Since the activity listener is enabled by default, should there be a way to opt-out e.g. through configuration?

src/Elastic.Apm/OpenTelemetry/ElasticActivityListener.cs
```diff
@@ -0,0 +1,173 @@
...
+						span.Otel.Attributes = new Dictionary<string, string>();
+
+					foreach (var tag in activity.Tags) span.Otel.Attributes.Add(tag.Key, tag.Value);
+
+					InferSpanTypeAndSubType(span, activity);
+					span.End();
+				}
+			};
+
+		public ActivityListener listener { get; }
```
russcam, line 144:
```suggestion
		public ActivityListener Listener { get; }
```

src/Elastic.Apm/OpenTelemetry/ElasticActivityListener.cs
```diff
@@ -0,0 +1,173 @@
...
+using Elastic.Apm.Api;
+using Elastic.Apm.DiagnosticListeners;
+using Elastic.Apm.DistributedTracing;
+using Elastic.Apm.Helpers;
+using Elastic.Apm.Logging;
+using Elastic.Apm.Model;
+
+namespace Elastic.Apm.OpenTelemetry
+{
+	public class ElasticActivityListener
```
russcam, line 22: Should this implement `IDisposable`, to dispose the `listener` on disposal?

src/Elastic.Apm/OpenTelemetry/ElasticActivityListener.cs
```diff
@@ -0,0 +1,173 @@
...
+	{
+		internal readonly ConcurrentDictionary<string, Span> ActiveSpans = new();
+		internal readonly ConcurrentDictionary<string, Transaction> ActiveTransactions = new();
```
russcam, line 25: Could one `ConcurrentDictionary<string, IExecutionSegment> ActiveSegments` be used to track active transactions and spans?

src/Elastic.Apm/Elastic.Apm.csproj
```diff
@@ -68,6 +68,9 @@
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="System.Diagnostics.PerformanceCounter" Version="4.5.0" />
   </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'net5.0'">
+    <PackageReference Include="System.Diagnostics.PerformanceCounter" Version="4.5.0" />
+  </ItemGroup>
```
russcam, line 73: Maybe combine into `<ItemGroup Condition="'$(TargetFramework)' != 'net461'">` ?

sample/SampleAspNetCoreApp/SampleAspNetCoreApp.csproj
```diff
@@ -23,6 +23,9 @@
     <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="5.0.0" />
     <PackageReference Include="Microsoft.AspNetCore.Identity.EntityFrameworkCore" Version="5.0.0" />
     <PackageReference Include="Microsoft.AspNetCore.Identity.UI" Version="5.0.0" />
+    <PackageReference Include="OpenTelemetry.Instrumentation.AspNetCore" Version="1.0.0-rc6"/>
+<!--    <PackageReference Include="OpenTelemetry" Version="1.0.0-rc6"/>-->
+    <PackageReference Include="OpenTelemetry.Extensions.Hosting" Version="1.0.0-rc6"/>
```
russcam, line 28: Should we include this in SampleAspNetCoreApp, or create a separate sample to demonstrate AspNetCore + OTel?

sample/OpenTelemetrySample/OpenTelemetrySample.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <OutputType>Exe</OutputType>
+        <TargetFramework>net5.0</TargetFramework>
+        <DockerDefaultTargetOS>Linux</DockerDefaultTargetOS>
```
russcam, line 6: is this needed?

russcam (changes requested):
I've gone through the PR and left some comments

ElasticApmAgent.sln
```diff
@@ -149,6 +149,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Elastic.Apm.Azure.CosmosDb"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Elastic.Apm.Azure.CosmosDb.Tests", "test\Elastic.Apm.Azure.CosmosDb.Tests\Elastic.Apm.Azure.CosmosDb.Tests.csproj", "{0EEA16C4-C7DF-4D90-94C2-009417D765AC}"
 EndProject
+<<<<<<< HEAD
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
gregkalapos: Yeah, fixed.

sample/ApiSamples/Program.cs
```diff
@@ -10,6 +10,10 @@
 using Elastic.Apm;
 using Elastic.Apm.Api;
 
+#if NET5_0
+using OpenTelemetry.Trace;
```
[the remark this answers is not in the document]
gregkalapos: Agreed - removed from here.

sample/OpenTelemetrySample/Program.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace OpenTelemetrySample
+{
+	internal class Program
+	{
+		// ReSharper disable once ArrangeTypeMemberModifiers
+		static void Main(string[] args)
+		{
+			Agent.Setup(new AgentComponents());
+			new OTSamples().Sample1();
```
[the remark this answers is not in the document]
gregkalapos: I don't think this matters - changed it.

src/Elastic.Apm/OpenTelemetry/ElasticActivityListener.cs
gregkalapos: I remember adding this during testing. Unfortunately I don't remember the details anymore, but I saw `null`s here.

src/Elastic.Apm/OpenTelemetry/ElasticActivityListener.cs
gregkalapos: Ok - done.

src/Elastic.Apm/OpenTelemetry/ElasticActivityListener.cs
gregkalapos: I changed it as you suggested, after looking at it, this (at least to me) hurts readability. Ideally this would be part of the spec and I think there would be more advantage to write it the same way as the spec - e.g. for maintenance. I doubt this makes any impact on perf. 

For now let's keep it the way you suggest - I may revisit this with benchmarks if the spec evolves and this part becomes hard to read.

src/Elastic.Apm/AgentComponents.cs
gregkalapos: Good point. I added a config for this.

For the beta phase, I'd make it opt-in. By default it's disabled and people can enable it. Once it's GA I'd flip it to enabled by default.

gregkalapos:
Most points are addressed from the review - in the last weekly we discussed this and agreed that we'll merge, so we can have this out as beta.

PR #1501, opened 2021-09-27
Change log level in CentralConfigurationFetcher.cs

gregkalapos:
From [discuss](https://discuss.elastic.co/t/loglevel-changed-in-centralconfigfetcher/284413). 

The log line changes here is printed in every round - prior to https://github.com/elastic/apm-agent-dotnet/pull/1471 if everything went well, we printed this log as a trace log and https://github.com/elastic/apm-agent-dotnet/pull/1471 changed it to info. Since we log this in every round I think it'd be better to keep this in trace. Regarding https://github.com/elastic/apm-agent-dotnet/pull/1471: that changes the log around the exception, even if an exception happens I'd say this log line is still fine on trace - no info related to the exception is printed in this line.


russcam (approved):
LGTM, thanks for catching @gregkalapos

PR #1511, opened 2021-10-06
Implement DroppedSpanStats

gregkalapos:
Solves https://github.com/elastic/apm/issues/497

According to [the spec](https://github.com/elastic/apm/blob/master/specs/agents/handling-huge-traces/tracing-spans-dropped-stats.md) this PR adds a new field to `Transaction` which contains aggregated data about dropped spans.

We had 1 optimization prior to this PR: we did not keep records of destination and resource for dropped spans, which we won't be able to do once we collect DroppenSpansStats.


src/Elastic.Apm/Model/Transaction.cs
```diff
@@ -436,6 +451,25 @@ private Activity StartActivity()
 			return activity;
 		}
 
+		internal void UpdateDroppedSpanStats(string destinationServiceResource, Outcome outcome, double duration)
+		{
+			_droppedSpanStatsMap ??= new Dictionary<DroppedSpanStatsKey, DroppedSpanStats>();
+
+			if (_droppedSpanStatsMap.Keys.Count >= 128)
```
russcam, line 458:
```suggestion
			if (_droppedSpanStatsMap.Count >= 128)
```

src/Elastic.Apm/Model/Transaction.cs
```diff
@@ -734,5 +768,19 @@ public void SetLabel(string key, long value)
 
 		public void SetLabel(string key, decimal value)
 			=> _context.Value.InternalLabels.Value.InnerDictionary[key] = value;
+
+		private struct DroppedSpanStatsKey
```
russcam, line 772: Implementing `IEquatable<DroppedSpanStatsKey>` would avoid the inherited reflection-based equality approach and could check equality on fields directly.

src/Elastic.Apm/Model/Transaction.cs
```diff
@@ -436,6 +451,25 @@ private Activity StartActivity()
 			return activity;
 		}
 
+		internal void UpdateDroppedSpanStats(string destinationServiceResource, Outcome outcome, double duration)
```
russcam, line 454: nitpick: the amount of work that has to be done when a new dropped span stats dictionary is created vs. one that already exists, is different e.g. a new dropped span stats dictionary doesn't need to be checked if count >= 128 or if it contains a specific value, so could branch based on whether dropped span stats dictionary is null or not. It's a minor optimization for sure, but would still be just as readable in my opinion.

russcam (approved):
LGTM. I left some suggestions for small changes.

One thing that looks missing from this PR is documentation for dropped span stats. Should that be added?

gregkalapos:
> One thing that looks missing from this PR is documentation for dropped span stats. Should that be added?

I don't think we need any extra docs in this repo. This is following [the spec](https://github.com/elastic/apm/blob/master/specs/agents/handling-huge-traces/tracing-spans-dropped-stats.md) which documents this.

gregkalapos:
`Transaction_And_Spans_Captured_When_Large_Request` fails. It's green locally.

gregkalapos:
Introduced a field to store `Context.Destination.Service.Resource` on the top level on `Span`. With that we, we can avoid allocating the whole `Context` object when a span is dropped.

Prior to the commit (benchmark is included - it's 1 transaction with 10 spans, 9 of those are dropped):
```
| Method |     Mean |    Error |   StdDev |   Gen 0 | Gen 1 | Gen 2 | Allocated |
|------- |---------:|---------:|---------:|--------:|------:|------:|----------:|
|      V | 12.07 ms | 0.231 ms | 0.204 ms | 15.6250 |     - |     - | 173.13 KB |
```

After the commit:
```
| Method |     Mean |    Error |   StdDev |   Gen 0 | Gen 1 | Gen 2 | Allocated |
|------- |---------:|---------:|---------:|--------:|------:|------:|----------:|
|      V | 12.76 ms | 2.177 ms | 6.175 ms | 15.6250 |     - |     - | 126.52 KB |
```

gregkalapos:
`Transaction_And_Spans_Captured_When_Large_Request` still fails. 

Another benchmark, 1000spans, 1sampled, the rest isn't.

This PR:
```
|        Method |     Mean |     Error |    StdDev |    Gen 0 |  Gen 1 | Gen 2 | Allocated |
|-------------- |---------:|----------:|----------:|---------:|-------:|------:|----------:|
| Test1000Spans | 3.260 ms | 0.0790 ms | 0.0660 ms | 941.4063 | 7.8125 |     - |   5.64 MB |
```

master:
```
|        Method |     Mean |     Error |    StdDev |    Gen 0 |  Gen 1 | Gen 2 | Allocated |
|-------------- |---------:|----------:|----------:|---------:|-------:|------:|----------:|
| Test1000Spans | 3.201 ms | 0.0390 ms | 0.0365 ms | 941.4063 | 7.8125 |     - |   5.64 MB |
```

So as it seems we don't really add overhead to spans that are not sampled.

gregkalapos:
Some more findings:


This PR  1K spans, 0,5K sampled


|        Method |     Mean |     Error |    StdDev |    Gen 0 |    Gen 1 | Gen 2 | Allocated |
|-------------- |---------:|----------:|----------:|---------:|---------:|------:|----------:|
| Test1000Spans | 5.822 ms | 0.1163 ms | 0.1845 ms | 968.7500 | 250.0000 |     - |    5.8 MB |


  
  master:
  
  |        Method |     Mean |     Error |    StdDev |    Gen 0 |    Gen 1 | Gen 2 | Allocated |
|-------------- |---------:|----------:|----------:|---------:|---------:|------:|----------:|
| Test1000Spans | 6.035 ms | 0.1282 ms | 0.2070 ms | 960.9375 | 242.1875 |     - |   5.79 MB |


This is all .NET Core - I don't really see any sign of a huge overhead added by this PR.

The reason `Transaction_And_Spans_Captured_When_Large_Request` fails is clearly the number of spans it generates - it takes 15+sec already locally on me without this PR.

Now, the `Transaction_And_Spans_Captured_When_Large_Request` should assert that large requests are captured properly and the 1K span generated in the test is just a side effect. I changed the test to generate less spans but kept the large request size.

PR #1520, opened 2021-10-13
Auto-infer destination.service.resource and adapt public API

gregkalapos:
Solves https://github.com/elastic/apm-agent-dotnet/issues/1330

- Obsoletes `Destination.Service.Name` and `Destination.Service.Type` - these fields were never used in APM Server and we should remove those long term according to elastic/apm-agent-dotnet#1330 
- Adds the concept of exit spans and implements auto-infer of `Destination.Service.Resource` for exit spans. 
  - We already had a logic to auto-infer it, but the logic did not fully follow [the spec](https://github.com/elastic/apm/blob/6ced3937252478a3aa4e84a1439ef8905a6c657a/specs/agents/tracing-spans-destination.md#contextdestinationserviceresource). This PR adapts to logic according to the spec.
  - With this PR we only infer `Destination.Service.Resource`  for exit spans. According to the spec this should only happen for exit spans which was not the case prior to this PR
- Extends the `StartSpan` and `CaptureSpan` APIs to also start exit spans and adapts the usage of it in instrumentation logic.

### DIFF
--- a/test/Elastic.Apm.Grpc.Tests/GrpcTests.cs
+++ b/test/Elastic.Apm.Grpc.Tests/GrpcTests.cs
@@ -135,7 +135,7 @@ namespace Elastic.Apm.Grpc.Tests
 			//Make sure all spans are collected
 			Thread.Sleep(500);
 
-			payloadSender.Spans.Should().HaveCountGreaterThan(1);
+			payloadSender.Spans.Should().HaveCountGreaterOrEqualTo(1);
 			payloadSender.Spans.Should()
 				.Contain(span => span.Subtype == ApiConstants.SubTypeGrpc
 					&& span.Outcome == Outcome.Failure


gregkalapos:
> The changes LGTM. A couple of questions:
> 
> 1. Should a property be included on `ISpan` to indicate whether it is an exit span?
> 2. It looks like exit span can be used in places where `InstrumentationFlag` is currently used to suppress child spans, such as in HttpDiagnosticListenerImplBase
>    https://github.com/elastic/apm-agent-dotnet/blob/64c27a874c584ab665cc8ccd07e5a8e44af7f75a/src/Elastic.Apm/DiagnosticListeners/HttpDiagnosticListenerImplBase.cs#L139-L146
>    
>    Should this be part of this PR?

1. So far, there was no public usage of the `isExitSpan` field, we only use used it in `Span.End()`, so I thought let's keep it private as long as we don't have a public usage of it. Point 2 is practically a public usage of it, so I ended up adding it :) 
2. Yeah, that makes sense - I adapted it. But with that, I think we'll still need instrumentation flags, so I kept those.

gregkalapos:
Merging this in - there are some failing tests with `Elastic.Apm.Specification.Validator`, those are unrelated.

Issue #1214, opened 2021-03-01
APM schema tests: Use synchronizes spec files from repo 


gregkalapos:
Follow up from https://github.com/elastic/apm-agent-dotnet/pull/1208

Currently we fetch json specs from the APM Server repo. We agreed across the agent teams that sync PRs will be opened on each spec change and we can use these spec files directly from the repo instead of fetching them in each tests.

The point of this issue is to not forget to update the tests to use these spec files.

PR #1530, opened 2021-10-20
Use checked in APM specs

russcam:
This commit updates the validator to use the checked in
APM specs. The download method is now exposed as a static
method, to allow testing other branches in future.

Closes elastic/apm-agent-dotnet#1214


russcam:
Merging this one in to unblock progress

PR #1538, opened 2021-11-01
Collect .NET Framework GC metrics only when filtering supported

### LINKED ISSUE #1346: Elastic APM - GC related memory leaks
### APM Agent version
Elastic.Apm 1.10.0

### Environment
Windows 10
.NET 4.8

### Describe the bug
I identified that we are having a slow memory leak in production.  Local testing with the server running idle showed that memory was being allocated continually

image

Attached is a screen shot taken from my local dev machine (Windows 10 64)
running a self hosted WebAPI on .NET 4.8 (owin self hosted)

### To Reproduce

1. Add Elastic.Apm to a standard .NET framework Web Api Project
2. Run the API service
3. Observe memory increase over time

I was able to fix this issue by rolling back to version 1.8.1.  Version 1.9 also had the same memory leak occurring

#### Expected behavior
I expect an idle web api services to not increase in memory usage over time

#### Actual behavior
There is a slow memory leak while the server is even idle which all seems to be related to Elastic APM's GC metric collection

russcam:
This commit updates GcMetricsProvider to collect GC metrics
on .NET Framework only when TraceEventSession filtering
is supported (Windows 8.1 / 2012 R2+). Without filtering,
the TraceEventSession keeps track of all TraceProcesses
and DotNetRuntimes in dictionaries that can grow unbounded for
the lifetime of the session, leading to continual memory usage growth.
With filtering, the TraceEventSession can be configured to track
only the current process of the application.

In future, GC heap stats and count could be collected using performance
counters instead of Trace event session. See

https://docs.microsoft.com/en-us/dotnet/framework/debug-trace-profile/performance-counters#memory-performance-counters

Performance counters don't expose a clean way to get GC time, instead
providing a % time in GC. Performance counters however may be a less
intrusive way of collecting GC metrics than TraceEventSession.

Other changes in this commit:

- do not start the trace event session or event listener if GC metrics are disabled

Fixes #1346

### DIFF
--- a/src/Elastic.Apm/Metrics/MetricsProvider/GcMetricsProvider.cs
+++ b/src/Elastic.Apm/Metrics/MetricsProvider/GcMetricsProvider.cs
@@ -198,7 +198,7 @@ namespace Elastic.Apm.Metrics.MetricsProvider
 			_gen3Size = (ulong)a.GenerationSize3;
 		}
 
-		private void RuntimeGCEnd(TraceProcess a, TraceGC gc)
+		private void RuntimeGCEnd(TraceProcess _, TraceGC gc)
 		{
 			if (!_isMetricAlreadyCaptured)
 			{
--- a/src/Elastic.Apm/Metrics/MetricsProvider/GcMetricsProvider.cs
+++ b/src/Elastic.Apm/Metrics/MetricsProvider/GcMetricsProvider.cs
@@ -44,7 +44,6 @@ namespace Elastic.Apm.Metrics.MetricsProvider
 		private readonly bool _collectGcGen2Size;
 		private readonly bool _collectGcGen3Size;
 		private readonly bool _collectGcTime;
-		private readonly int _currentProcessId;
 
 		private readonly GcEventListener _eventListener;
 		private readonly object _lock = new object();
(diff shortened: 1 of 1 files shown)


src/Elastic.Apm/Metrics/MetricsProvider/GcMetricsProvider.cs
```diff
@@ -201,32 +187,26 @@ public void Dispose()
...
+					_isMetricAlreadyCaptured = true;
 			}
+			_gen0Size = (ulong)a.GenerationSize0;
+			_gen1Size = (ulong)a.GenerationSize1;
+			_gen2Size = (ulong)a.GenerationSize2;
+			_gen3Size = (ulong)a.GenerationSize3;
 		}
 
-		private void RuntimeGCEnd(TraceProcess traceProcess, TraceGC gc)
+		private void RuntimeGCEnd(TraceProcess a, TraceGC gc)
```
gregkalapos, line 201:
```suggestion
		private void RuntimeGCEnd(TraceProcess _, TraceGC gc)
```

If we rename, then use `_` to indicate it's unused.

gregkalapos (approved):
LGTM

PR #1542, opened 2021-11-04
Use Traceparent binary format name for Kafka integration

### LINKED ISSUE #1541: Use Traceparent binary format name for Kafka integration
See https://github.com/elastic/apm/blob/master/specs/agents/tracing-distributed-tracing.md#binary-fields.

The integration uses `traceparent` header name, but should use `elasticapmtraceparent`.

russcam:
This commit uses the Traceparent binary format name for Kafka integration, in line with the spec.

Transaction context message queue name and age properties are also populated.

Closes #1541


russcam:
All tests pass; Merging this in

PR #1555, opened 2021-11-10
Documentation improvements

russcam:
This commit improves the documentation by

- Adding NuGet package information to Profiler auto instrumentation.
Assembly versions supported don't always align with NuGet package
versions.
- linking each supported technology to the relevant integration section
- Moving agent added versions to tags
- Adding missing sections for NuGet packages


gregkalapos (approved):
lgtm

PR #1558, opened 2021-11-10
Update ConfigTests.cs

gregkalapos:
The question if we support server url with a path came up in a discussion - e.g. say you have a reverse proxy in front of APM Server which translates "/apm/intake/v2/events" into "/intake/v2/events" on the backend.

This was supported - but there wasn't really a test for it. Adding a test in this PR.


gregkalapos:
This just adds an additional test - merging this in.

PR #1569, opened 2021-11-16
Add CosmosDb package to README

russcam:


russcam:
Merging this small README change in

PR #1576, opened 2021-11-23
Add trace log with callstack in Agent.Setup

gregkalapos:
This could help us diagnose double initialization issues better. This will we'll know the callstack of the first initialization.


gregkalapos:
This is a small change, only adds comment - merging in.

Issue #1590, opened 2021-12-13
[BUG] Startup hooks fail on .NET 6


dstelljes:
### APM Agent version
1.12.1

### Environment

**Operating system and version:** Debian 11 (mcr.microsoft.com/dotnet/aspnet:6.0)

**.NET Framework/Core name and version:** 6.0.0

> Host (useful for support):
>   Version: 6.0.0
>   Commit:  4822e3c3aa
> 
> .NET SDKs installed:
>   No SDKs were found.
> 
> .NET runtimes installed:
>   Microsoft.AspNetCore.App 6.0.0 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
>   Microsoft.NETCore.App 6.0.0 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
> 

**Application Target Framework(s):** net6.0

### Describe the bug

When starting a .NET 6 application instrumented with the [zero code change approach](https://www.elastic.co/guide/en/apm/agent/dotnet/current/setup-dotnet-net-core.html#zero-code-change-setup), the application crashes.

### To Reproduce

With the .NET 6 SDK:

```shell
dotnet new console -n CrashRepro

cat <<EOF > Dockerfile
FROM mcr.microsoft.com/dotnet/sdk:6.0 AS build

RUN apt update && apt install -y unzip && apt clean

WORKDIR /build
ADD https://github.com/elastic/apm-agent-dotnet/releases/download/1.12.1/ElasticApmAgent_1.12.1.zip elastic.zip
RUN unzip elastic.zip -d elastic && rm elastic.zip
COPY ./ ./

WORKDIR /build/CrashRepro
RUN dotnet publish --configuration Release --output out

FROM mcr.microsoft.com/dotnet/aspnet:6.0
WORKDIR /app

COPY --from=build /build/elastic ./elastic
COPY --from=build /build/CrashRepro/out ./
ENTRYPOINT ["dotnet", "CrashRepro.dll"]

ENV DOTNET_STARTUP_HOOKS=/app/elastic/ElasticApmAgentStartupHook.dll
ENV ELASTIC_APM_ENVIRONMENT=test
ENV ELASTIC_APM_SECRET_TOKEN=redacted
ENV ELASTIC_APM_SERVER_URL=redacted
ENV ELASTIC_APM_SERVICE_NAME=crash-repro

USER 1000
EOF

docker build . -t apm-crash-repro
docker run -it --rm apm-crash-repro
```

[...]

PR #1603, opened 2022-01-17
Build startuphook zip containing DiagnosticSource 6.x

gregkalapos:
Solves: https://github.com/elastic/apm-agent-dotnet/issues/1590 

### Root cause of the bug

The zip file with the agent prior to this PR contained 2 folders with 2 different versions of `System.Diagnostics.DiagnosticSource`. One with version `4.0.0`, the other one with version `5.0.0`. During startup the agent decided from which folder to load the agent based on the loaded version of this assembly (logic [here](https://github.com/elastic/apm-agent-dotnet/blob/master/src/ElasticApmAgentStartupHook/StartupHook.cs#L98)). The issue is, .NET 6 ships with a new version (`6.0.0`) of this dll, therefore the `StartupHook` implementation tried to load the agent from the `6.0.0` which did not exits.

### Solution

This PR adds specifically building the agent for .NET6 as well with `System.Diagnostics.DiagnosticSource` v `6.0.0`. It also updates the release scripts to ship a `6.0.0` folder based on the output of the .NET6 build. 

### Alternative 

It'd be also possible to force the agent to load from the `5.0.0` folder which is built agains `System.Diagnostics.DiagnosticSource` v5. Likely this would work, reasons not to do this is: 1) we may use some `Activity` API later that are only present in .NET6, 2) if we know the app is built agains DiagnosticSource 6, I think it's safer to load a version of the agent which was built agains v6 as well.

### DIFF
--- a/build/scripts/Build.fs
+++ b/build/scripts/Build.fs
@@ -16,8 +16,6 @@ open Fake.Core
 open Fake.DotNet
 open Fake.IO
 open Fake.IO.Globbing.Operators
-open Fake.SystemHelper
-open Fake.SystemHelper
 open TestEnvironment
 open Tooling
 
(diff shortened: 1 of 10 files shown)


twogood:
Why isn't the APM agent loading System.Diagnostics.DiagnosticSource from the .NET runtime instead of shipping its own copy?

PR #1604, opened 2022-01-18
Add more context to MongoDB Span

glucaci:
- Before
image

- After
image

### DIFF
--- a/src/Elastic.Apm.MongoDb/DiagnosticSource/MongoDiagnosticListener.cs
+++ b/src/Elastic.Apm.MongoDb/DiagnosticSource/MongoDiagnosticListener.cs
@@ -55,7 +55,7 @@ namespace Elastic.Apm.MongoDb.DiagnosticSource
 				var span = currentExecutionSegment.StartSpan(
 					@event.CommandName,
 					ApiConstants.TypeDb,
-					"mongo", isExitSpan: true);
+					"mongodb", isExitSpan: true);
 
 				if (!_processingQueries.TryAdd(@event.RequestId, span))
 				{
@@ -70,7 +70,7 @@ namespace Elastic.Apm.MongoDb.DiagnosticSource
 				{
 					Statement = @event.Command.ToString(),
 					Instance = @event.DatabaseNamespace.DatabaseName,
-					Type = "mongo"
+					Type = "mongodb"
 				};
 
 				if (@event.ConnectionId?.ServerId?.EndPoint != null)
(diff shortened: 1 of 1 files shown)


src/Elastic.Apm.MongoDb/DiagnosticSource/MongoDiagnosticListener.cs
```diff
@@ -79,10 +79,22 @@ private void HandleCommandStartEvent(CommandStartedEvent @event)
...
+							{
+								Resource = $"{ipEndPoint.Address}:{ipEndPoint.Port}"
+							},
 							Address = ipEndPoint.Address.ToString(),
 							Port = ipEndPoint.Port
 						},
-						DnsEndPoint dnsEndPoint => new Destination { Address = dnsEndPoint.Host, Port = dnsEndPoint.Port },
+						DnsEndPoint dnsEndPoint => new Destination
+						{
+							Service = new Destination.DestinationService
```
gregkalapos, line 91: `Service` does not need to be set here. Two things to this:

1. According to the current spec `Service.Resource` should be just `monodb`. Spec [here](https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-db.md#mongodb)
2. There is a logic in `Span.cs` [here](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm/Model/Span.cs#L489) which fills this. 

So this should be removed.

src/Elastic.Apm.MongoDb/DiagnosticSource/MongoDiagnosticListener.cs
```diff
@@ -55,7 +55,7 @@ private void HandleCommandStartEvent(CommandStartedEvent @event)
 				var span = currentExecutionSegment.StartSpan(
 					@event.CommandName,
 					ApiConstants.TypeDb,
-					"mongo", isExitSpan: true);
+					"mongodb", isExitSpan: true);
```
gregkalapos, line 58: Nice - indeed, this should be `mongodb`.

gregkalapos (changes requested):
Thanks for opening this - I added 1 thing regarding `Service.Resource` which should be removed. 

Additionally, we'd also need to adapt tests to cover this.

The rest looks good - nice catch! 👍

gregkalapos:
This is done in https://github.com/elastic/apm-agent-dotnet/pull/1677.

Issue #1566, opened 2021-11-15
Add .NET 6 to CI and update tests/sample apps


gregkalapos:
.NET6 is now GA, so we can update our CI and tests to actively test against it as part of our CI.

russcam:
I started on this last week- Let's coordinate 🙂

PR #1605, opened 2022-01-18
Add .NET6 to CI scripts

gregkalapos:
Solves https://github.com/elastic/apm-agent-dotnet/issues/1566

- Adding .NET 6 in CI
- Installing the VS2022 version of vs BuildTools
- Running all tests on .NET 6 as well (where applicable)

### DIFF
--- a/.ci/windows/dotnet.bat
+++ b/.ci/windows/dotnet.bat
@@ -16,4 +16,6 @@ dotnet sln remove test/Elastic.Apm.Profiler.Managed.Tests/Elastic.Apm.Profiler.M
 
 dotnet nuget locals all --clear
 
+dotnet nuget add source --name nuget.org https://api.nuget.org/v3/index.json
+
 dotnet build -c Release
--- a/.ci/windows/dotnet.bat
+++ b/.ci/windows/dotnet.bat
@@ -18,4 +18,5 @@ dotnet nuget locals all --clear
 
 dotnet nuget add source --name nuget.org https://api.nuget.org/v3/index.json
 
+:: Build solution. Add `--verbosity detailed` for more detailed logs
 dotnet build -c Release
--- a/.ci/windows/dotnet.bat
+++ b/.ci/windows/dotnet.bat
@@ -14,6 +14,6 @@ dotnet sln remove test/Elastic.Apm.StartupHook.Tests/Elastic.Apm.StartupHook.Tes
 :: Remove profiler tests, which are tested separately- require profiler to be built
 dotnet sln remove test/Elastic.Apm.Profiler.Managed.Tests/Elastic.Apm.Profiler.Managed.Tests.csproj
 
-dotnet --version
+dotnet nuget locals all --clear
 
 dotnet build -c Release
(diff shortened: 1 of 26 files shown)


.ci/windows/dotnet.bat
```diff
@@ -14,4 +14,9 @@ dotnet sln remove test/Elastic.Apm.StartupHook.Tests/Elastic.Apm.StartupHook.Tes
 :: Remove profiler tests, which are tested separately- require profiler to be built
 dotnet sln remove test/Elastic.Apm.Profiler.Managed.Tests/Elastic.Apm.Profiler.Managed.Tests.csproj
 
-dotnet build -c Release --verbosity detailed
+dotnet nuget locals all --clear
+
+dotnet nuget add source --name nuget.org https://api.nuget.org/v3/index.json
+
+:: Build solution. Add `--verbosity detailed` for more detailed logs
```
gregkalapos, line 21: Took out verbose logging - it turned out to generate more noise if I need to look at logs from a failed build. I'd prefer to keep it without verbose logs and only turn it on once we really need it.

.ci/windows/msbuild-tools.ps1
```diff
@@ -6,7 +6,7 @@ $ErrorActionPreference = "Stop"
 
 # Install visualstudio2019 and workloads needed to build ASP.NET and .NET Core apps
 Invoke-WebRequest -UseBasicParsing `
-    -Uri "https://download.visualstudio.microsoft.com/download/pr/cb1d5164-e767-4886-8955-2df3a7c816a8/b9ff67da6d68d6a653a612fd401283cc213b4ec4bae349dd3d9199659a7d9354/vs_BuildTools.exe" `
+    -Uri "https://download.visualstudio.microsoft.com/download/pr/ce8663b0-08ed-410a-9f5d-4f9469d1b2cb/73271b3d53a4e50e65e2e918a8c1100d2681c17bc418e03513c9f0554609ff8a/vs_BuildTools.exe" `
```
gregkalapos, line 9: This updates it to the vs 2022 version.

Jenkinsfile
```diff
@@ -358,9 +358,6 @@ pipeline {
                       unstash 'source'
                       dir("${BASE_DIR}"){
                         powershell label: 'Install test tools', script: '.ci\\windows\\test-tools.ps1'
-                        retry(3) {
-                          bat label: 'Build', script: '.ci/windows/dotnet.bat'
```
gregkalapos, line 362: Instead of building it once and then running `test.bat` in a separate step we can do the whole thing in `test.bat`.

gregkalapos:
https://github.com/elastic/apm-integration-testing/pull/1412 takes care of failing integration tests.

v1v (approved):
Neat!

.ci/windows/test.bat
```diff
@@ -1,8 +1,24 @@
...
+
+:: Remove profiler tests, which are tested separately- require profiler to be built
+dotnet sln remove test/Elastic.Apm.Profiler.Managed.Tests/Elastic.Apm.Profiler.Managed.Tests.csproj
+
 ::
 :: This script runs the tests and stored them in an xml file defined in the
 :: LogFilePath property
 ::
-dotnet test -c Release --no-build ^
+dotnet test -c Release ^
```
v1v, line 21: Out of curiosity, if the test.bat does also run the build, are the `dotnet sln remove` needed for the building?

gregkalapos: Yes, that's intentional.

Those projects that we remove with `dotnet sln remove` need more than what we have in this context (e.g. ASP.NET Classic), so the `dotnet` tool isn't enough for those. We use those projects in the `Windows .NET Framework` stage.

I only know this from experience :) If we don't remove those, it'll fail. Prior to this PR removing happened in the `dotnet.bat` file.

v1v: Thanks Greg for the clarification :)

Issue #1602, opened 2022-01-14
Support Npgsql 6


ajuch:
We started using Elastic Cloud recently and were quite happy with APM. But since upgrading Npgsql to 6.0.1, the Postgres spans disappeared. https://www.elastic.co/guide/en/apm/agent/dotnet/current/setup-auto-instrumentation.html states that NpgsqlCommand is only instrumented for Npgsql up to 5.*, so I suspect this is the reason.

**Describe the solution you'd like**
We'd like to see our database queries in Elastic Cloud even when we use Npgsql 6.*, so we hope that a newer version of the APM agent will bring this functionality back.

**Describe alternatives you've considered**
Maybe it's possible to use a DB independent way of monitoring SQL queries?

gregkalapos:
Thanks for reporting!

There wasn't any specific reason to not support Npgsql v6 - it's just that it was new and we haven't had tests for it. https://github.com/elastic/apm-agent-dotnet/pull/1609 introduces support for `6.*`, and also adds test covering it. This will be in the next release.

ajuch:
Great, thank you for your work!

PR #1609, opened 2022-01-19
Bump supported Npgsql version

gregkalapos:
Solves https://github.com/elastic/apm-agent-dotnet/issues/1602 

Add Npgsql v6 tests and bump supported version to 6.x.


russcam:
Support for npgsql 6 can be achieved with:

- Update `ProfiledApplication` to support [passing MSBuild properties to `dotnet publish`](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-publish#msbuild) so that `-p:NpgsqlVersion=6.0.2` can be passed from a test for Npgsql 6.x:

   https://github.com/elastic/apm-agent-dotnet/blob/bab776194d91a8e65be077d3ad1f5db3f38f81b4/test/Elastic.Apm.Profiler.Managed.Tests/ProfiledApplication.cs#L68

   Might want to append a unique hash of passed MSBuild properties to the output directory, if it'd be helpful to keep compiled published artifacts on disk for different passed test parameters.
- Add another test to `NpgSqlCommandTests` that passes NpgsqlVersion=6.0.2 as an MsBuild property to ProfiledApplication, and is parameterized only for the TFMs for which Npgsql 6.0.2 is supported.
- Update `InstrumentNpgsqlAttribute` to specify `"6.*.*"` as the maximum version:

   https://github.com/elastic/apm-agent-dotnet/blob/bbfa788a2074c1006e1ca11e90f8d7d847c8b136/src/Elastic.Apm.Profiler.Managed/Integrations/AdoNet/AdoNetTypeNames.cs#L29

- Regenerate intergrations.yml and integrations.asciidoc by running the following from the project root

   ```sh
   cargo make generate-integrations
   ```

Parameterizing the MsBuild properties for instrumented assembly version dependencies and passing them when publishing sample applications removes the need to create shared projects and shared project items, which will add maintenance overhead, and will only get more awkward to maintain as newer versions of instrumented assemblies are released.

russcam:
I've opened https://github.com/elastic/apm-agent-dotnet/pull/1611, which follows the steps above  (https://github.com/elastic/apm-agent-dotnet/pull/1609#issuecomment-1017237532)

PR #1611, opened 2022-01-20
Support NpgSql 6.x

### LINKED ISSUE #1602: Support Npgsql 6

russcam:
This commit adds support for Npgsql 6.x, along with tests.
Update ProfiledApplication to allow passing MsBuild properties
to be able to test multiple dependent assembly versions.

Closes #1602


gregkalapos:
Thank! Planning to merge once https://github.com/elastic/apm-agent-dotnet/pull/1605 is all green.

gregkalapos (approved):

PR #1618, opened 2022-01-24
[Docs] Update wording in the "how to collect logs" part

gregkalapos:
This came up in discussion.

The wording with "When the Agent is activated..." was misleading. What we try to say in this part is that if the agent is added to the monitored application on a specific way (and this specific way is now better specified), then the agent just uses the ASP.NET Core logging system.

cc @predogma


bmorelli25 (approved):
👍

Issue #782, opened 2020-03-20
Proxy support


killnine:
**Is your feature request related to a problem? Please describe.**
My employer has a proxy that outbound traffic is routed through. Unfortunately, the .NET agent doesn't seem to support proxy credentials like, say, the Ruby agent does. 

**Describe the solution you'd like**
I'd like to help implement proxy support for the client where users can supply a client URI, a plaintext username, and password for the agent to use in its outbound requests to APM servers (in this case, cloud instances). Alternatively, supplying a ClientCredential or NetworkCredential might be better from a security standpoint. 

**Describe alternatives you've considered**
Long-term, we whitelist URLs or IPs for resources we commonly use. However, we're prototyping our solution and authenticating through the proxy would normally be faster for a short-term solution. However, based on what our secops folks are saying, it doesn't seem like traffic is consistently routing to and from the same locations (maybe because of load balancing or something?) so their attempts to whitelist haven't worked.

**Additional context**
We're evaluating APM Cloud solutions right now but stopped in our tracks because we can't reach out to the URL provided by the startup documentation. We're a .NET shop so this would really help things out and at least make it consistent with some other agents.

gregkalapos:
Hi @killnine,

definitely sounds like a valid use-case.

Configs similar to what [the Ruby agent has](https://www.elastic.co/guide/en/apm/agent/ruby/current/configuration.html#config-proxy-address) would be preferred. 

To be very open here - this is not one of our top priorities right now, but as always we are super happy to take community PRs and I can also try helping with it.

Some ideas and hints in this regard:
- So, I'd suggest trying to implement configs similar to what the Ruby agent has, maybe `proxy_headers` is a bit more complicated, but if some configs are missing, it's totally fine. If the PR achieves useful functionality it'll be fine.
- Adding other configs to cover your use case (ClientCredential or NetworkCredential) is also fine.
- [`IConfigurationReader`](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm/Config/IConfigurationReader.cs) is a good place to start if you want to add a new config. You'll need to add the new properties to multiple classes, the best approach is to just follow how other configs are implemented - once you add the new properties to `IConfigurationReader` build errors will show what other classes you need to extend. Some configs are runtime changeable from Kibana, that's why configs are not trivial. I'd suggest ignoring that for now, these don't have to be live changeable in the first iteration. 
[...]

PR #1629, opened 2022-02-17
.net Agent support for outbound traffic using Proxy

m-hafi:
In this PR proxy support for outbound traffic is added to the HTTP Client Handler. This issue originated as our firm uses proxy for outbound traffic and the current .net Agent lacks this feature.

### DIFF
--- a/src/Elastic.Apm/BackendComm/BackendCommUtils.cs
+++ b/src/Elastic.Apm/BackendComm/BackendCommUtils.cs
@@ -176,7 +176,18 @@ namespace Elastic.Apm.BackendComm
 				};
 			}
 
-			return new HttpClientHandler { ServerCertificateCustomValidationCallback = serverCertificateCustomValidationCallback };
+			IWebProxy proxy = null;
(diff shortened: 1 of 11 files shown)


gregkalapos:
Hi @m-hafi,

it seems the [CLA is not yet signed](https://github.com/elastic/apm-agent-dotnet/pull/1629#issuecomment-1043016214) - we'd need that before we proceed.

More importantly, we'll need tests which cover this feature and the new config should be also documented. We have a doc about the PR process [here](https://github.com/elastic/apm-agent-dotnet/blob/main/CONTRIBUTING.md).

The idea with this feature already came up, but it was low prio: https://github.com/elastic/apm-agent-dotnet/issues/782 

I'll bring this up in our planning meeting and see how we proceed - if you have time and you could invest into covering this with tests and add the docs I'm happy to review - otherwise we'll need to invest more time into this, which we did not plan short term.

bjola:
You should consider adding support for DefaultProxyCredentials in addition to the URL

PR #1630, opened 2022-02-18
Fix handling of HttpClient client side timeouts

Lionhunter3k:
Hello :)

We are using the apm-agent-dotnet in several of our applications, and have discovered that the agent will stop calling the endpoints  "intake/v2/events" and "config/v1/agents" if the response latency exceeds the HttpClient timeout. Therefore, if the APM server has higher latency than expected (and it can happen), our applications will no longer send their data and fetch their settings at all, because the logic  client side HTTP timeouts isn't handled.

This PR addresses this.

Perhaps the BackendCommComponentBase should actually handle the distinction between cancelled operations that short circuit the work loop iteration and those that don't, but the most straightforward fix was that to check if the CancellationToken is actually cancelled inside each implementation of BackendCommComponentBase.

Another possible solution is to drop the "servicePoint.ConnectionLeaseTimeout", and use a composite cancellation token comprised out of the current ".BackendCommComponentBase.CancellationTokenSource" property and a newly created one from a Task.Delay inside the "FetchConfigHttpResponseImplAsync"/"ProcessQueueItems" methods.

### DIFF
--- a/src/Elastic.Apm/BackendComm/CentralConfig/CentralConfigurationFetcher.cs
+++ b/src/Elastic.Apm/BackendComm/CentralConfig/CentralConfigurationFetcher.cs
@@ -94,7 +94,7 @@ namespace Elastic.Apm.BackendComm.CentralConfig
 		protected override async Task WorkLoopIteration()
 		{
 			++_dbgIterationsCount;
-			WaitInfoS waitInfo;
+			WaitInfoS waitInfo = default;
 			HttpRequestMessage httpRequest = null;
 			HttpResponseMessage httpResponse = null;
 			string httpResponseBody = null;
@@ -115,7 +115,10 @@ namespace Elastic.Apm.BackendComm.CentralConfig
 			}
 			catch (OperationCanceledException)
 			{
-				throw;
+				if (CancellationTokenSource.IsCancellationRequested)
+				{
+					throw;
+				}
 			}
 			catch (Exception ex)
 			{
--- a/src/Elastic.Apm/Report/PayloadSenderV2.cs
+++ b/src/Elastic.Apm/Report/PayloadSenderV2.cs
@@ -347,7 +347,10 @@ namespace Elastic.Apm.Report
 						, string.Join($",{Environment.NewLine}{TextUtils.Indentation}", queueItems));
 
 				// throw to allow Workloop to handle
-				throw;
+				if(CancellationTokenSource.IsCancellationRequested)
+				{
+					throw;
+				}
 			}
 			catch (Exception e)
 			{


gregkalapos:
Thank you for the PR @Lionhunter3k.


This is indeed the bug - thanks for the super useful description here.

> Perhaps the BackendCommComponentBase should actually handle the distinction between cancelled operations that short circuit the work loop iteration and those that don't

Yeah, the `BackendCommComponentBase` should handle this - In fact, that was the intention, but it simply doesn't work as it should be. That seems to be the root cause of this bug.

[Here it should handle this case](https://github.com/elastic/apm-agent-dotnet/blob/20fea7888fd3a044e13fb34c3e657d26502a6911/src/Elastic.Apm/Helpers/ExceptionUtils.cs#L57) - the only issue is that this is within a task and even though there is a catch block to handle the exception, that task itself is already faulted - and that's why the [this task in the workloop](https://github.com/elastic/apm-agent-dotnet/blob/20fea7888fd3a044e13fb34c3e657d26502a6911/src/Elastic.Apm/BackendComm/BackendCommComponentBase.cs#L87-L94) runs to completion as well - and that's why the work loop basically finishes.

I think the `BackendCommComponentBase.cs` should be fixed, with that this problem will go away as well.

Let me do another iteration on this - at this point I'd feel better with directly fixing `BackendCommComponentBase.cs` which would take care of this issue as well.

Lionhunter3k:
Thank you for your answer @gregkalapos .
Since we cannot migrate our AppInsights based alerts to the Elasticsearch Alerts stack because the apm agent can stop sending actionable metrics, is it possible to disclose a rough timeframe for a fix? This information would help my team decide how to proceed further regarding our alerts migration.

gregkalapos:
> Thank you for your answer @gregkalapos .
> Since we cannot migrate our AppInsights based alerts to the Elasticsearch Alerts stack because the apm agent can stop sending actionable metrics, is it possible to disclose a rough timeframe for a fix? This information would help my team decide how to proceed further regarding our alerts migration.

Usually we try to avoid committing to specific dates in public (because that usually doesn't end well), but I understand your situation. My goal is to have this fixed in the next agent release; so far we don't have a specific date for that release. We also have the option to push a patch release with the fix - that'd be much faster. A patch could happen within a few weeks. Would that help you?

Lionhunter3k:
Yes, a patch release would be perfect! Thank you!

gregkalapos:
I merged https://github.com/elastic/apm-agent-dotnet/pull/1634 which addresses this and it also covers it with a test.

Sorry that I did not merge your PR @Lionhunter3k, but I thought it'd be better to fix this directly in `BackendCommComponentBase` and according to [our processes](https://github.com/elastic/apm-agent-dotnet/blob/main/CONTRIBUTING.md#code-contributions-please-read-this-before-your-first-pr) I also added a test.

Closing this one. A patch will be released with this.

PR #1634, opened 2022-03-01
Prevent RanToCompletion in WorkingLoop when APM Server timeouts

gregkalapos:
Solves the issue reported in https://github.com/elastic/apm-agent-dotnet/pull/1630


Lionhunter3k:
Great! Will use the new version! Thanks!

PR #1651, opened 2022-03-10
ci: speed up tag releases

v1v:
### What

Run the build/tests for branches and Pull Requests as long as changes are not related to docs only

### Why

This will help to run the releases on tag basis as tags are created from commits that have been verified in the CI previously

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -61,7 +61,13 @@ pipeline {
         stage('Parallel'){
           when {
             beforeAgent true
-            expression { return env.ONLY_DOCS == "false" }
+            allOf {
+              expression { return env.ONLY_DOCS == "false" }
+              anyOf {
+                changeRequest()
+                branch '**/*'
+              }
+            }
           }
           parallel{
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -61,7 +61,13 @@ pipeline {
         stage('Parallel'){
           when {
             beforeAgent true
-            expression { return env.ONLY_DOCS == "false" }
+            allOf {
+              expression { return env.ONLY_DOCS == "false" }
```
v1v, line 65: Keep the same logic as used to be and add the condition to exclude any tags

Jenkinsfile
```diff
@@ -61,7 +61,13 @@ pipeline {
...
+              anyOf {
+                changeRequest()
+                branch '**/*'
```
v1v, line 68: I could not find any other way to filter but using the `glob` 

image

gregkalapos (approved):
Great - thanks!

PR #1674, opened 2022-04-06
Fix InvalidCastException in AspNetCoreDiagnosticListener

aheubusch:
Fixes ``System.InvalidCastException: 'Unable to cast object of type'<>f__AnonymousType1`3[ Microsoft.AspNetCore.Http.HttpContext, System.Int64, System.Exception ]' to type  '<>f__AnonymousType0`2[ Microsoft.AspNetCore.Http.HttpContext, System.Exception ]'.'``.
This exception happens when both `UnhandledException` cases are executed. One event has an anonymous type with two values, the other with three values. The anonymous type is cached in the `PropertyFetcher`.


gregkalapos (approved):
Thank you!

I appreciate the nice test you added.

Issue #1688, opened 2022-04-21
Use Open Telemetry Kind Property Default Transaction Type


alfkonee:
Instead of defaulting to "unknown" for the transaction type of OpenTelemetry Activities  I'd like to Propose we use the activity kind Property instead where the transaction type or span type cannot be introspected

**Describe the solution you'd like**
[Proof Of Concept](https://github.com/elastic/apm-agent-dotnet/pull/1687)

gregkalapos:
Hey @alfkonee 

we have a spec on how to bridge the OTel Activity instances into our Elastic format. The spec can be found [here](https://github.com/elastic/apm/blob/main/specs/agents/tracing-api-otel.md#transaction-type) - the linked part defines how to infer `Transaction.Type`- and then we have the same for `Span.Type`.

That logic runs at the end of each span/transaction. The reason those run at the end is that the result also depend on the presence of some attributes which may not be set at the start of the `Activity`. On the other hand, since you raised this, it makes me think: in some cases this algorithm would still return a better result than `unknown` (which we have now).

But just setting it to `Activity.Kind` would not follow the spec and it'd create very different `Transaction.Type`s compares to other transactions you typically have with Elastic APM.

So, I'll do the following: I'll run the type inference methods both at transaction/span start (that may or may not give us what we want) and then run it again at the end (at that point we have all the attributes we need).

gregkalapos:
Closing for inactivity.

PR #1690, opened 2022-04-27
OTel bridge: set Transaction.Type and Span.Type/Subtype at start

gregkalapos:
Solves: https://github.com/elastic/apm-agent-dotnet/issues/1688

Prior to this PR we only inferred `Transaction.Type` and `Span.Type`/`Span.Subtype` at the end of the `Activity`. This PR runs the inference logic at the beginning of each span/transaction and runs it again when the `Activity` finishes. 

This way users will have proper `Transaction.Type` and `Span.Type`/`Span.Subtype` when they look at the bridged OTel spans through our API before the OTel span (aka `Activity`) finishes.


adnang:
What's the status of this?

gregkalapos:
@adnang it did not get merged - we thought it's not relevant.

Do you need this functionality? Could you describe your use case?

PR #1692, opened 2022-05-05
ci: use new azure secret

v1v:
### What

Use a new service 

### Important

With some expiration date!


gregkalapos:
I'm closing this - this was just a try to migrate to another Azure AD instance, but turned out to be not needed.

PR #1696, opened 2022-05-12
Change tag pattern to start tags with `v` to be in sync with our SLSA specification.

gregkalapos:
Without this change, CI does not pick up release tags to build.

All other Elastic APM Agents start their tag names with `v` - so this is to align with that.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -461,7 +461,7 @@ pipeline {
                 allOf {
                   anyOf {
                     branch 'main'
-                    tag pattern: '\\d+\\.\\d+\\.\\d+.*', comparator: 'REGEXP'
+                    tag pattern: 'v\\d+\\.\\d+\\.\\d+.*', comparator: 'REGEXP'
                     expression { return params.Run_As_Main_Branch }
                     expression { return env.BENCHMARK_UPDATED != "false" }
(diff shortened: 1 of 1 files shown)


kuisathaverat (approved):

v1v:
no! :(

v1v (changes requested):
There are some other changes to be done:

- In the JJBB -> https://github.com/elastic/apm-agent-dotnet/blob/main/.ci/jobs/defaults.yml#L32
- In the https://github.com/elastic/apm-agent-dotnet/blob/03d5f13cfbcb431ef0ee7cec6c88734dbf6d7540/Jenkinsfile#L591-L592

and maybe https://github.com/elastic/opbeans-dotnet/blob/main/.ci/bump-version.sh if the version of the agent is now v based

gregkalapos:
> and maybe https://github.com/elastic/opbeans-dotnet/blob/main/.ci/bump-version.sh if the version of the agent is now v based

No, this is only about the tag name - the agent version itself is not effected by this.

gregkalapos:
@v1v could you please open the PR for the 2 missing things, because I don't fully understand what it'd need. 

The JJBB pattern is too complex to me and the other one in the jenkinsfile already had the `v`, so I don't fully get this.

PR #1702, opened 2022-05-17
Enable AzureCredentialsFactAttribute

v1v:
Reverts elastic/apm-agent-dotnet#1694 partially to enable `AzureCredentialsFactAttribute`


gregkalapos (commented):
We'll need to uncomment this line as well: https://github.com/elastic/apm-agent-dotnet/blob/main/Jenkinsfile#L601 

Tests are still skipped, all logs for these tests look like this:
```
[xUnit.net 00:00:03.02]   Starting:    Elastic.Apm.Azure.ServiceBus.Tests
[xUnit.net 00:00:03.13]       Azure credentials not available. If running locally, run `az login` to login
[xUnit.net 00:00:03.16]       Azure credentials not available. If running locally, run `az login` to login
```

@v1v did you update the secret in vault?

gregkalapos:
Hahh, they work on Linux, but don't work on Windows.

I see this pattern, e.g.: https://apm-ci.elastic.co/job/apm-agent-dotnet/job/apm-agent-dotnet-mbp/job/PR-1702/2/testReport/Elastic.Apm.Azure.ServiceBus.Tests/AzureMessagingServiceBusDiagnosticListenerTests/ 

Logs above are from Windows.

I'm not sure if this is intentional... I believe these tests were running on both.

v1v:
> We'll need to uncomment this line as well: https://github.com/elastic/apm-agent-dotnet/blob/main/Jenkinsfile#L601

Done

> I'm not sure if this is intentional... I believe these tests were running on both.

As far as I see, terraform is not configured for Windows, while all the linux stages in the CI (build, test, profiler) use the [`dotnet`](https://github.com/elastic/apm-agent-dotnet/blob/2cfc941e2b8e8e21106ef3af7f860eafee0d9ccd/Jenkinsfile#L611-L631) function which configures terraform. Is it required? or only azure?

gregkalapos:
> > We'll need to uncomment this line as well: https://github.com/elastic/apm-agent-dotnet/blob/main/Jenkinsfile#L601
> 
> Done
> 
> > I'm not sure if this is intentional... I believe these tests were running on both.
> 
> As far as I see, terraform is not configured for Windows, while all the linux stages in the CI (build, test, profiler) use the [`dotnet`](https://github.com/elastic/apm-agent-dotnet/blob/2cfc941e2b8e8e21106ef3af7f860eafee0d9ccd/Jenkinsfile#L611-L631) function which configures terraform. Is it required? or only azure?

I'd assume both are required. But I think terraform is configured on Windows - at least I see sign of the intention of doing so. [Here we apply `.credentials.json`](https://github.com/elastic/apm-agent-dotnet/blob/2cfc941e2b8e8e21106ef3af7f860eafee0d9ccd/Jenkinsfile#L367) and in `.ci\\windows\\test-tools.ps1` we install terraform (that's the line above).

gregkalapos (approved):
tldr: I'd say let's merge this as it is.

I looked into the C# code that reads the Azure authentication related things (e.g. `.credentials.json`)  and ends up with skipping the tests with: `Azure credentials not available. If running locally, run az login to login`.

I think either the `.credentials.json` is missing (or is not at the right place) or something goes wrong when we parse it. Problem is that this whole thing uses console logging and that logging is very limited anyway and I don't see those in any of the log files. I'd change the logger in this area and pass the XUnit logger into it, so we have more info on where it goes wrong on Windows.

Right now I'm working on other stuff, so I'd just open an issue to make the logger changes and investigate what's going on on Windows later, and merge this now as it is, so we have these tests at least on Linux, because that'd already be a big progress on the overall issue (which is that these azure tests weren't running at all)

PR #1729, opened 2022-06-09
Fix if routeData is null to prevent throwing Exception

jadanah:
When updating from 1.15.0 to 1.16.0 some exceptions were being thrown when running in some legacy services running aspnet core 2.2 using net48.


jadanah:
Null check was previously there but was removed in commit 47540a6578fe2f0bce393452816a603d7d72591f

gregkalapos (approved):
Thank you!

PR #1772, opened 2022-07-29
Include PathBase in the transaction name and details (#1771)

### LINKED ISSUE #1771: Include PathBase when logging a transaction
In our k8s-based microservice we are logging requests across the request pipeline, with the main focus on two places - our ingress controller and ASP.NET Core application. Ingress controller rewrites the request path, excluding the prefix, which is then forwarded to the service in `X-Forwarded-Prefix` header and included in the `HttpContext.Request.PathBase` property.

Unfortunately the APM agent doesn't include the `PathBase` when logging, which prevents us from correlating spans across a transaction. Including it by default should ease the integration of Elastic APM into services sitting behind any kind of reverse proxy, but might affect users that specifically moved the APM agent further down the ASP.NET middleware pipeline.

Given the fact that the combination of `PathBase` and `Path` canonically represents the full request path (excluding the protocol and domain parts) I'd like the APM to include `PathBase` when logging the transaction. The change should preferably be configurable to avoid introducing breaking changes.

As a workaround we've temporarily stopped path rewriting by the ingress controller and moved APM further up in the middleware pipeline, but this way we are forcing our microservices into a situation where they need to know their target environment.

**Additional context**
I'm preparing a PR with the changes that would enable the agent to include PathBase during transaction logging - it should be posted tomorrow.

TrueGoric:
This PR introduces automatic inclusion of PathBase in the transaction building process. The change is managed by a new configuration entry.

Closes #1771

### DIFF
--- a/src/Elastic.Apm/Config/AbstractConfigurationWithEnvFallbackReader.cs
+++ b/src/Elastic.Apm/Config/AbstractConfigurationWithEnvFallbackReader.cs
@@ -161,6 +161,9 @@ namespace Elastic.Apm.Config
 		public bool UseElasticTraceparentHeader => ParseUseElasticTraceparentHeader(Read(KeyNames.UseElasticTraceparentHeader,
 			EnvVarNames.UseElasticTraceparentHeader));
 
+		public bool UseFullPathRequestMatching => ParseUseElasticTraceparentHeader(Read(KeyNames.UseFullPathRequestMatching,
+			EnvVarNames.UseFullPathRequestMatching));
+
 		public virtual bool VerifyServerCert =>
 			ParseVerifyServerCert(Read(KeyNames.VerifyServerCert, EnvVarNames.VerifyServerCert));
 
--- a/test/Elastic.Apm.AspNetCore.Tests/AspNetCoreDiagnosticListenerTest.cs
+++ b/test/Elastic.Apm.AspNetCore.Tests/AspNetCoreDiagnosticListenerTest.cs
@@ -194,5 +194,45 @@ namespace Elastic.Apm.AspNetCore.Tests
 				context?.Request.Method.Should().Be(HttpMethod.Get.Method);
 			}
 		}
+
+
+		[Fact]
+		public async Task RecordFullRequestPathWhenFullPathRequestMatchingIsConfigured()
+		{
+			var capturedPayload = new MockPayloadSender();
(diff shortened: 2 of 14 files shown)


src/Elastic.Apm/Config/AbstractConfigurationWithEnvFallbackReader.cs
```diff
@@ -161,6 +161,9 @@ public virtual IReadOnlyList<Uri> ServerUrls
 		public bool UseElasticTraceparentHeader => ParseUseElasticTraceparentHeader(Read(KeyNames.UseElasticTraceparentHeader,
 			EnvVarNames.UseElasticTraceparentHeader));
 
+		public bool UseFullPathRequestMatching => ParseUseElasticTraceparentHeader(Read(KeyNames.UseFullPathRequestMatching,
```
gregkalapos, line 164:
```suggestion
		public bool UseFullPathRequestMatching => ParseUseFullPathRequestMatching(Read(KeyNames.UseFullPathRequestMatching,
```

test/Elastic.Apm.AspNetCore.Tests/AspNetCoreDiagnosticListenerTest.cs
```diff
@@ -194,5 +194,45 @@ public async Task CaptureDiagnosticAndHostingUnhandledExceptionEvent()
...
+				var builder = _factory
+					.WithWebHostBuilder(n =>
+					{
+						n.Configure(app =>
+						{
+							app.Use((context, next) =>
+							{
+								context.Request.PathBase = "/ReverseProxyPrefix";
+								return next();
+							});
```
gregkalapos, line 215: This will not test `AspNetCoreDiagnosticListener` - it'll only test the `ApmMiddleware`.

`AspNetCoreDiagnosticListener` is a diagnostic source based implementation of ASP.NET Core tracing without an ASP.NET Core middleware. The main reason we have it is because that's the only way to enable the agent with zero code change (we can't register a middleware - or it'd be fairly hard).

The test should cover both.

gregkalapos (commented):
Thank you for the PR! 

I added a comment in the issue https://github.com/elastic/apm-agent-dotnet/issues/1771#issuecomment-1201456099 - I'd suggest continuing the discussion on the issue first.

I looked at the PR - just adding 2 things I found here.

gregkalapos:
This is blocked by open questions on https://github.com/elastic/apm-agent-dotnet/issues/1771 - closing it for now.

PR #1784, opened 2022-08-16
ci: skip stdout by default

v1v:
### What

As per the Jenkins docs

> keepLongStdio : boolean (optional)
If checked, any standard output or error from a test suite will be retained in the test results after the build completes. (This refers only to additional messages printed to console, not to a failure stack trace.) Such output is always kept if the test failed, but by default lengthy output from passing tests is truncated to save space. Check this option if you need to see every log message from even passing tests, but beware that Jenkins's memory consumption can substantially increase as a result, even if you never look at the test results!

### Why

There a dozens of lines in the `system-out`

image

And produces memory heaps when digesting the test results

Discussed offline so this is not needed by default

### DIFF
--- a/.ci/windows/test.bat
+++ b/.ci/windows/test.bat
@@ -23,9 +23,8 @@ dotnet sln remove test/Elastic.Apm.StaticExplicitInitialization.Tests/Elastic.Ap
 :: LogFilePath property
 ::
 dotnet test -c Release ^
- --verbosity normal ^
+ --verbosity quiet ^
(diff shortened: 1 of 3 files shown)


gregkalapos (approved):
As discussed via slack: based on the docs of `keepLongStdio` we don't really need those logs - fine by me to turn this off.

amannocci (approved):

v1v:
This is not required anymore, after changing the debug traces printed in the console.

PR #1787, opened 2022-08-17
Fix FillApmServerInfo : Invalid ElasticApm_ApiKey throws Exception

### LINKED ISSUE #1735: [BUG] FillApmServerInfo : Invalid ElasticApm_ApiKey throws object not set exception
### APM Agent version
Latest Main branch code on 16/6/2022
### Environment

**Operating system and version:**

**.NET Framework/Core name and version** (e.g. .NET 4.6.1, NET Core 3.1.100) :
.Net 6

**Application Target Framework(s)** (e.g. net461, netcoreapp3.1):
.Net 6

### Describe the bug
When connecting using an ElasticApm_ApiKey that is invalid, an exception is thrown when calling app.UseAllElasticApm(Configuration);

reason is that the Elastic.Apm.ServerInfo.ApmServerInfoProvider FillApmServerInfo does a request that succeeds, but does not return any data, and therefore the deSerialisation returns metadata null, then doing metadata["version"]  throws an "object not set to an instance of an object" exception.

					var metadata = serializer.Deserialize<JObject>(jsonReader);
					**var version = metadata["version"];**

### To Reproduce
Configure an invalid Api Key in ElasticApm_ApiKey
use the app.UseAllElasticApm(Configuration);  as explained for .Net core.
an exception is thrown

#### Expected behavior
improve the code : 
- include the Deserialize in the foreseen try catch
- what if strVersion is null, no callbackInFinish is called 

#### Actual behavior
throws Object not set to an instance of an object exception

z1c0:
Fixes #1735

When connecting with an invalid `ElasticApm_ApiKey` an exception was thrown when calling `UseAllElasticApm(Configuration)`.

The unit test makes sure that we output a log message for the case when no APM Server version can be retrieved.

### DIFF
--- a/test/Elastic.Apm.Tests/BackendCommTests/PayloadSenderTests.cs
+++ b/test/Elastic.Apm.Tests/BackendCommTests/PayloadSenderTests.cs
@@ -349,6 +349,20 @@ namespace Elastic.Apm.Tests.BackendCommTests
 		}
 
 		[Fact]
+		public void Find_a_better_name()
+		{
+			var logger = new InMemoryBlockingLogger(LogLevel.Warning);
+			var configReader = new MockConfiguration(logger, apiKey: "*** INVALID_API_KEY ***");
+			var mockHttpMessageHandler = new MockHttpMessageHandler((r, c) => Task.FromResult(new HttpResponseMessage(HttpStatusCode.OK)));
+			var service = Service.GetDefaultService(configReader, logger);
+
+			using (new PayloadSenderV2(logger, configReader, service, new Api.System(), null, mockHttpMessageHandler, TestDisplayName))
+			{
+				logger.Lines.Should().Contain(l => l.Contains("Failed parsing APM Server version - version string not available"));
+			};
+		}
+
+		[Fact]
 		public void Dispose_stops_the_thread()
 		{
 			PayloadSenderV2 lastPayloadSender = null;
--- a/test/Elastic.Apm.Tests/BackendCommTests/PayloadSenderTests.cs
+++ b/test/Elastic.Apm.Tests/BackendCommTests/PayloadSenderTests.cs
@@ -349,20 +349,6 @@ namespace Elastic.Apm.Tests.BackendCommTests
 		}
 
 		[Fact]
-		public void Find_a_better_name()
-		{
-			var logger = new InMemoryBlockingLogger(LogLevel.Warning);
-			var configReader = new MockConfiguration(logger, apiKey: "*** INVALID_API_KEY ***");
-			var mockHttpMessageHandler = new MockHttpMessageHandler((r, c) => Task.FromResult(new HttpResponseMessage(HttpStatusCode.OK)));
-			var service = Service.GetDefaultService(configReader, logger);
-
-			using (new PayloadSenderV2(logger, configReader, service, new Api.System(), null, mockHttpMessageHandler, TestDisplayName))
-			{
-				logger.Lines.Should().Contain(l => l.Contains("Failed parsing APM Server version - version string not available"));
-			};
-		}
-
-		[Fact]
 		public void Dispose_stops_the_thread()
 		{
 			PayloadSenderV2 lastPayloadSender = null;
--- a/src/Elastic.Apm/Model/Span.cs
+++ b/src/Elastic.Apm/Model/Span.cs
@@ -547,7 +547,7 @@ namespace Elastic.Apm.Model
 		private bool IsSameKind(Span other) => Type == other.Type
 			&& Subtype == other.Subtype
 			&& _context.IsValueCreated && other._context.IsValueCreated
-			&& Context.Service.Target == other.Context.Service.Target;
+			&& Context?.Service?.Target == other.Context?.Service?.Target;
 
 		private bool TryToCompressRegular(Span sibling)
 		{
--- a/src/Elastic.Apm/Report/PayloadSenderV2.cs
+++ b/src/Elastic.Apm/Report/PayloadSenderV2.cs
@@ -240,7 +240,18 @@ namespace Elastic.Apm.Report
 			_eventQueue.TryReceive(null, out var receivedItems);
 
 			if (_flushInterval == TimeSpan.Zero)
+			{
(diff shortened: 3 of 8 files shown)


test/Elastic.Apm.Tests/BackendCommTests/PayloadSenderTests.cs
```diff
@@ -348,6 +348,20 @@ internal async void FlushInterval_test(TestArgs args, int numberOfEventsToSend)
 			}
 		}
 
+		[Fact]
+		public void Find_a_better_name()
+		{
+			var logger = new InMemoryBlockingLogger(LogLevel.Warning);
+			var configReader = new MockConfiguration(logger, apiKey: "*** INVALID_API_KEY ***");
+			var mockHttpMessageHandler = new MockHttpMessageHandler((r, c) => Task.FromResult(new HttpResponseMessage(HttpStatusCode.OK)));
```
gregkalapos, line 356: We'd need to return a content here, otherwise `ApmServerInfoProvider.cs` will fail already when reading the content [here](https://github.com/elastic/apm-agent-dotnet/pull/1787/files#diff-3f17e138c3b0c6eacad18b7496cc3ba19fe91dcd8e6d43269697ba9b28e41995R35).

gregkalapos (commented):
I think the test won't be stable (added some thoughts on the specific lines).

I'd test it this way:

```
[Fact]
public async Task ApmServerInfo_Without_ServerVersion()
{
	var waitHandle = new ManualResetEvent(false);
	var logger = new InMemoryBlockingLogger(LogLevel.Warning);
	var configReader = new MockConfiguration(logger, apiKey: "*** INVALID_API_KEY ***");
	var mockHttpMessageHandler = new MockHttpMessageHandler((r, c) =>
	{
		var response = new HttpResponseMessage(HttpStatusCode.OK);
		// A response without `version` which makes parsing server version fail.
		response.Content = new StringContent("{'foo': 'bar'}", Encoding.UTF8, "application/json");
		return Task.FromResult(response);
	});
	var service = Service.GetDefaultService(configReader, logger);
	var apmServerInfo = new ApmServerInfo();
	var httpClient =
		BackendCommUtils.BuildHttpClient(logger, configReader, service, nameof(ApmServerInfo_Without_ServerVersion), mockHttpMessageHandler);
	await ApmServerInfoProvider.FillApmServerInfo(apmServerInfo, logger, configReader, httpClient, (result, serverInfo) =>
	{
		result.Should().BeFalse();
		waitHandle.Set();
	});
	
        waitHandle.WaitOne();
	logger.Lines.Should().Contain(l => l.Contains("Failed parsing APM Server version - version string not available"));
}
```


- The line `var configReader = new MockConfiguration(logger, apiKey: "*** INVALID_API_KEY ***");` doesn't really make a difference - the point is that we have a response without `version`.  Nevertheless, since https://github.com/elastic/apm-agent-dotnet/issues/1735 described that the problem was caused by API key, we can leave the config there... but that'd only really make a difference with a real APM Server. 
- With this, we don't really utilize `PayloadSenderV2`, so if we do it this way, let's not put this into `PayloadSenderTests.cs`. I see we don't have much tests around fetching ApmServerInfo, maybe we could create a new file for this test - or in case we have such tests and I overlooked, then I'd just put it into the file of those tests.

test/Elastic.Apm.Tests/BackendCommTests/PayloadSenderTests.cs
z1c0: Agreed, a dedicated test file for `ApmServerInfo` makes more sense.

test/Elastic.Apm.Tests/BackendCommTests/PayloadSenderTests.cs
```diff
@@ -348,6 +348,20 @@ internal async void FlushInterval_test(TestArgs args, int numberOfEventsToSend)
 			}
 		}
 
+		[Fact]
+		public void Find_a_better_name()
```
z1c0, line 352: Obviously, I forgot to find a better name ;-) - will be addressed in the upcoming commit.

gregkalapos: I thought it's a challenge for me :D

z1c0: The (updated) test method is now in `ApmServerInfoProviderTests.FillApmServerInfo_warns_on_missing_server_version`

test/Elastic.Apm.Tests/BackendCommTests/PayloadSenderTests.cs
```diff
@@ -348,6 +348,20 @@ internal async void FlushInterval_test(TestArgs args, int numberOfEventsToSend)
...
+		{
+			var logger = new InMemoryBlockingLogger(LogLevel.Warning);
+			var configReader = new MockConfiguration(logger, apiKey: "*** INVALID_API_KEY ***");
+			var mockHttpMessageHandler = new MockHttpMessageHandler((r, c) => Task.FromResult(new HttpResponseMessage(HttpStatusCode.OK)));
+			var service = Service.GetDefaultService(configReader, logger);
+
+			using (new PayloadSenderV2(logger, configReader, service, new Api.System(), null, mockHttpMessageHandler, TestDisplayName))
+			{
+				logger.Lines.Should().Contain(l => l.Contains("Failed parsing APM Server version - version string not available"));
```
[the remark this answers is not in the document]
z1c0: It should succeed as long as *one* log line matches the expected pattern, so I guess the expected log line was not present in that case at all.
That being said, the test method was updated to specifically test `ApmServerInfoProvider` and moved into a dedicated test class.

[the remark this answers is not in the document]
gregkalapos: > It should succeed as long as one log line matches the expected pattern, so I guess the expected log line was not present in that case at all.

Yeah, that was it - because we had no content in the HTTP call, it never reached the log line we assert on here.

gregkalapos (approved):
LGTM! 👍 

_(Let's wait for CI to finish before hitting merge, code looks good to me, should be green)._

PR #1798, opened 2022-08-24
CentralConfigResponseParser - test with empty json

gregkalapos:
Add `{}` to test input.

This came up in another discussion. Empty json response from central config should not cause any issue to the agent.

This is already the case, but it wasn't tested - therefore adding `{}` to test inputs to cover this case.


z1c0 (approved):
LGTM

PR #1800, opened 2022-08-25
Fix incorrect transaction name in ASP.NET Web Api

### LINKED ISSUE #1645: Transaction name wrongly assigned NET461
The transaction name in NET461 is wrongly assigned when there are two methods sharing a similar route. When the transaction name is set due to the hierarchy of the routes, it always takes the first one when calling the second endpoint.

Package: `Elastic.Apm.AspNetFullFramework`
Class: `ElasticApmModule`
Method: `ProcessEndRequest`

Steps to reproduce:
1. Clone repo
2. Replace the sample class `AttributeRoutingWebApiController` in `sample/AspNetFullFrameworkSampleApp`
3. Set the `LogLevel` to `Debug`
4. Run project and call endpoint `route1/{route1:guid}/routes/process`
5. Check the transaction name is `route1/{route1:guid}/routes/{route2}` instead of the one requested

```
namespace AspNetFullFrameworkSampleApp.Controllers
{
	[RoutePrefix(RoutePrefix)]
	public class AttributeRoutingWebApiController : ApiController
	{
		public const string RoutePrefix = "route1/{route1:guid}/routes";
		public const string Route = "process";

		[HttpGet]
		[Route("{route2}")]
		public IHttpActionResult Get1([FromUri] Guid route1, Guid route2) =>
			Ok($"attributed routed web api controller {route1} and {route2}");

		[HttpGet]
		[Route(Route)]
		public IHttpActionResult Get2([FromUri] Guid route1) =>
			Ok($"attributed routed web api controller {route1}");
	}
}

```

z1c0:
Fixes #1645

When selecting the matching route template and multiple candidates exist, the *precedence* value of the route must be considered.

Previously, we always selected the first route (see [here](https://github.com/elastic/apm-agent-dotnet/blob/20fea7888fd3a044e13fb34c3e657d26502a6911/src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs#L330)) which could lead to the undesired behavior when multiple potential route matches are present.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -107,7 +107,7 @@ pipeline {
                           whenTrue(isPR()) {
                             // build nuget packages and profiler
                             sh(label: 'Package', script: '.ci/linux/release.sh true')
-                            sh label: 'Rustup', script: 'rustup default 1.57.0'
+                            sh label: 'Rustup', script: 'rustup default 1.59.0'
                             sh label: 'Cargo make', script: 'cargo install --force cargo-make'
                             sh(label: 'Build profiler', script: './build.sh profiler-zip')
                           }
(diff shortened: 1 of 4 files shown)


gregkalapos (approved):
Nice, lgtm!

Issue #1802, opened 2022-08-26
TraceContinuationStrategy=restart_external without TraceParent header causes NullReferenceException


Vaahik:
### APM Agent version
elastic_apm_profiler_1.17.0-beta.1.zip

### Environment

**Operating system and version:**
CentOS Stream 8

**.NET Framework/Core name and version** 
Host (useful for support):
  Version: 5.0.17
  Commit:  6a98414363

.NET runtimes installed:
  Microsoft.AspNetCore.App 5.0.17 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 5.0.17 [/usr/share/dotnet/shared/Microsoft.NETCore.App]

**Application Target Framework(s)** (netcoreapp5.0):

### Describe the bug
+-> Exception: System.NullReferenceException: Object reference not set to an instance of an object.
at Elastic.Apm.Model.Transaction..ctor(IApmLogger logger, String name, String type, Sampler sampler, DistributedTracingData distributedTracingData, IPayloadSender sender, IConfiguration configuration, ICurrentExecutionSegmentsContainer currentExecutionSegmentsContainer, IApmServerInfo apmServerInfo, BreakdownMetricsProvider breakdownMetricsProvider, Boolean ignoreActivity, Nullable`1 timestamp, String id, String traceId, IEnumerable`1 links) in C:\Users\jenkins\workspace\net_apm-agent-dotnet-mbp_PR-1797\apm-agent-dotnet\src\Elastic.Apm\Model\Transaction.cs:line 319
at Elastic.Apm.Api.Tracer.StartTransactionInternal(String name, String type, DistributedTracingData distributedTracingData, Boolean ignoreActivity, Nullable`1 timestamp, String id, String traceId, IEnumerable`1 links) in C:\Users\jenkins\workspace\net_apm-agent-dotnet-mbp_PR-1797\apm-agent-dotnet\src\Elastic.Apm\Api\Tracer.cs:line 87
at Elastic.Apm.Api.Tracer.StartTransaction(String name, String type, DistributedTracingData distributedTracingData, Boolean ignoreActivity, IEnumerable`1 links) in C:\Users\jenkins\workspace\net_apm-agent-dotnet-mbp_PR-1797\apm-agent-dotnet\src\Elastic.Apm\Api\Tracer.cs:line 66
[...]

PR #1803, opened 2022-08-26
Fix potential NullReferenecException in TraceContinuationStrategy implementation

gregkalapos:
Potential fix for https://github.com/elastic/apm-agent-dotnet/issues/1802  Not verified yet if it's the same bug, but this PR definitely fixes a real bug.

This feature was introduced in https://github.com/elastic/apm-agent-dotnet/pull/1739 (released in `1.17.0`). 

The added test shows the case which caused the `NullReferenceException`: It's when`TraceContinuationStrategy` was set to `restart_external` and an incoming request did not have a traceparent header.

Fix: adding a null-check.

### DIFF
--- a/src/Elastic.Apm/Model/Transaction.cs
+++ b/src/Elastic.Apm/Model/Transaction.cs
@@ -134,7 +134,7 @@ namespace Elastic.Apm.Model
 
 			var shouldRestartTrace = configuration.TraceContinuationStrategy == ConfigConsts.SupportedValues.Restart ||
 				configuration.TraceContinuationStrategy == ConfigConsts.SupportedValues.RestartExternal
-				&& !distributedTracingData.TraceState.SampleRate.HasValue;
+				&& distributedTracingData is { TraceState: { SampleRate: null } };
 
 			// For each new transaction, start an Activity if we're not ignoring them.
 			// If Activity.Current is not null, the started activity will be a child activity,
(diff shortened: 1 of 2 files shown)


gregkalapos:
Regarding red CI:

The only thing making this fail is:
```
[2022-08-26T17:00:05.470Z]   package `time v0.3.14` cannot be built because it requires rustc 1.59.0 or newer, while the currently active rustc version is 1.57.0
[2022-08-26T17:00:05.470Z] [cargo-make] ERROR - Error while executing command, exit code: 101
```

Which is fixed over in https://github.com/elastic/apm-agent-dotnet/pull/1800 - once that merged and this is updated to have that commit, I think this will be green.

src/Elastic.Apm/Model/Transaction.cs
```diff
@@ -134,7 +134,7 @@ internal Transaction(ApmAgent agent, string name, string type, long? timestamp =
 
 			var shouldRestartTrace = configuration.TraceContinuationStrategy == ConfigConsts.SupportedValues.Restart ||
 				configuration.TraceContinuationStrategy == ConfigConsts.SupportedValues.RestartExternal
-				&& !distributedTracingData.TraceState.SampleRate.HasValue;
+				&& distributedTracingData is { TraceState: { SampleRate: null } };
```
z1c0, line 137: Nice - pattern matching ftw.

z1c0 (approved):
lgtm 👍

z1c0:
#1800 was merged and the CI should be green now.
I'll merge your PR - hope that's fine with you.

PR #1804, opened 2022-08-28
Fix in handling TraceContinuationStrategy=restart_external

gregkalapos:
Follow up to https://github.com/elastic/apm-agent-dotnet/pull/1803 after https://github.com/elastic/apm-agent-dotnet/issues/1802#issuecomment-1229194614. After thinking about it and looking at it again, I realized we needed to tweak the `shouldRestartTrace` check in `Transaction.cs`. 

Prior to this PR with https://github.com/elastic/apm-agent-dotnet/pull/1803 we already handled the missing `traceparent` and/or `tracestate` header, but the following case wasn't handled correctly:

- TraceContinuationStrategy=restart_external
- TraceParent is present and it's valid
- TraceState is not present

Prior to this PR those traces were never restarted (despite `restart_external`), which is fixed here.

Also adding a test case to cover this.

### DIFF
--- a/src/Elastic.Apm/Model/Transaction.cs
+++ b/src/Elastic.Apm/Model/Transaction.cs
@@ -132,10 +132,16 @@ namespace Elastic.Apm.Model
 			var spanLinks = links as SpanLink[] ?? links?.ToArray();
 			Links = spanLinks;
 
+			// Restart the trace when:
+			// - `TraceContinuationStrategy == Restart` OR
+			// - `TraceContinuationStrategy == RestartExternal` AND
+			//		- `TraceState` is not present (Elastic Agent would have added it) OR
+			//		- `TraceState` is present but the SampleRate is not present (Elastic agent adds SampleRate to TraceState)
 			var shouldRestartTrace = configuration.TraceContinuationStrategy == ConfigConsts.SupportedValues.Restart ||
-				configuration.TraceContinuationStrategy == ConfigConsts.SupportedValues.RestartExternal
-				&& distributedTracingData is { TraceState: { SampleRate: null } };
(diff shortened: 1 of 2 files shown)


src/Elastic.Apm/Model/Transaction.cs
```diff
@@ -132,10 +132,16 @@ internal Transaction(ApmAgent agent, string name, string type, long? timestamp =
...
+			// - `TraceContinuationStrategy == RestartExternal` AND
+			//		- `TraceState` is not present (Elastic Agent would have added it) OR
+			//		- `TraceState` is present but the SampleRate is not present (Elastic agent adds SampleRate to TraceState)
 			var shouldRestartTrace = configuration.TraceContinuationStrategy == ConfigConsts.SupportedValues.Restart ||
-				configuration.TraceContinuationStrategy == ConfigConsts.SupportedValues.RestartExternal
-				&& distributedTracingData is { TraceState: { SampleRate: null } };
+				(configuration.TraceContinuationStrategy == ConfigConsts.SupportedValues.RestartExternal
+					&& (distributedTracingData?.TraceState == null || distributedTracingData is { TraceState: { SampleRate: null } }));
 
+			// For each new transaction, start an Activity if we're not ignoring them.
```
z1c0, line 144: This comment was duplicated from the line below.

z1c0 (approved):
LGTM

(I'm not sure if you wanted to add an additional comment (see above) so I'll leave the PR open for you.)

src/Elastic.Apm/Model/Transaction.cs
gregkalapos: Hahh, thanks! Fixed

PR #1814, opened 2022-09-09
Use correct default value for ExitSpanMinDuration

### LINKED ISSUE #1789: [BUG] DefaultValues.ExitSpanMinDurationInMilliseconds has incorrect value
### APM Agent version
1.16.1

### Environment
N/A

### Describe the bug
According to the [documentation](https://www.elastic.co/guide/en/apm/agent/dotnet/master/config-core.html#config-exit-span-min-duration) `ExitSpanMinDuration` has a default value of `1ms`.

However the constant `DefaultValues.ExitSpanMinDurationInMilliseconds` has a value of `1000` instead of `1`: https://github.com/elastic/apm-agent-dotnet/blob/main/src/Elastic.Apm/Config/ConfigConsts.cs#L30-L31

z1c0:
Fixes #1789

Both the [documentation](https://www.elastic.co/guide/en/apm/agent/dotnet/master/config-core.html#config-exit-span-min-duration) for `ExitSpanMinDuration` and the implementation of `DefaultValues.ExitSpanMinDurationInMilliseconds` (see [here](https://github.com/elastic/apm-agent-dotnet/blob/main/src/Elastic.Apm/Config/ConfigConsts.cs?rgh-link-date=2022-08-17T15%3A07%3A39Z#L30-L31)) are in violation of the [specification](https://github.com/elastic/apm-agent-java/blob/198cfac574fcbaae2d9a374702637bbcf44c46f8/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/SpanConfiguration.java#L68).

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -589,7 +589,7 @@ Additionally, spans that lead to an error can't be discarded.
 [options="header"]
 |============
 | Default | Type
-| `1ms`  | TimeDuration
+| `0ms`  | TimeDuration
 |============
 
 
--- a/src/Elastic.Apm/Config/ConfigConsts.cs
+++ b/src/Elastic.Apm/Config/ConfigConsts.cs
@@ -27,8 +27,8 @@ namespace Elastic.Apm.Config
 			public const bool CentralConfig = true;
 			public const string CloudProvider = SupportedValues.CloudProviderAuto;
 			public const bool EnableOpenTelemetryBridge = false;
-			public const string ExitSpanMinDuration = "1ms";
-			public const int ExitSpanMinDurationInMilliseconds = 1000;
+			public const string ExitSpanMinDuration = "0ms";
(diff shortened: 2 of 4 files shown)


gregkalapos (approved):
LGTM - I think this can be merged as is.

There is some additional info, which I add here for documentation.

The original default was `1ms`, so the "dropping fast exit spans" feature was released with using `1ms` as default for this config.

Then this default got discussed, and cross-agent we decided to use `0ms` by default - meaning we decided to make dropping fast exit spans opt-in (`0ms` means it's turned off be default). That happened here: https://github.com/elastic/apm/pull/610 - but so far that wasn't implemented for .NET (and I don't find an issue to do so either).

Having said all that, https://github.com/elastic/apm-agent-dotnet/issues/1789 was indeed a real bug - but as this PR already does, I also think we should go with the current default (`0ms`).

Issue #1678, opened 2022-04-11
[BUG] BreakdownMetricsProvider hits internal limit of 1000 items with more than 1000 transaction within [MetricsInterval] seconds


fschoening:
### APM Agent version
1.13.0

### Environment

**Operating system and version:**
Windows 10 Enterprise, v21H2

**.NET Framework/Core name and version**
.NET 4.6.1, NET Core 3.1.100

**Application Target Framework(s)**
net461, netcoreapp3.1

### Describe the bug
Exception _The limit of 1000 metricsets has been reached, no new metricsets will be created until the current set is sent to APM Server._ is thrown frequently when transactions end, even with few distinct `SpanTimerKey` combinations in internal Spans.

Looking at class `Elastic.Apm.Metrics.MetricsCollector` metrics are collected on a fixed interval configured by setting [`MetricsInterval`](https://www.elastic.co/guide/en/apm/agent/dotnet/current/config-reporter.html#config-metrics-interval) (defaults to 30s). In this case it does so by calling `Elastic.Apm.Metrics.MetricsProvider.BreakdownMetricsProvider.GetSamples()` that returns the value of field `List<MetricSet> _itemsToSend`.

Span timings on the transaction are aggregated on the field of class `Elastic.Apm.Model.Transaction`:

```internal readonly ConcurrentDictionary<SpanTimerKey, SpanTimer> SpanTimings = new();```

In the course of a transaction, spans are thus aggregated from the `Elastic.Apm.Model.Span` with the dictionary key composed by span + span type:

```
if (_enclosingTransaction.SpanTimings.ContainsKey(new SpanTimerKey(Type, Subtype)))
  _enclosingTransaction.SpanTimings[new SpanTimerKey(Type, Subtype)].IncrementTimer(SelfDuration);
else
  _enclosingTransaction.SpanTimings.TryAdd(new SpanTimerKey(Type, Subtype), new SpanTimer(SelfDuration));
```

When the transaction ends, `_breakdownMetricsProvider?.CaptureTransaction(this);` is called, that iterates the span timings in the following manner (`Elastic.Apm.Metrics.MetricsProvider.BreakdownMetricsProvider.CaptureTransaction(...)`):

```
foreach (var item in transaction.SpanTimings)
{
  var metricSet =
    new MetricSet(timestampNow,
      new List<MetricSample>
      {
        new(SpanSelfTimeCount, item.Value.Count),
[...]

PR #1816, opened 2022-09-14
Group MetricSets in BreakdownMetricsProvider

gregkalapos:
In `BreakdownMetricsProvider` we collected metrics in a `List<MetricSet>` - each span and transaction added a new item into this list without aggregating those. The result was technically correct, but the list was growing quickly and we also have a 1k limit which was very easy to reach.

This PR fixes it and groups MetricSets according to the spec. The new tests in `test/Elastic.Apm.Tests/BreakdownTests.cs` show test cases which were failing with the original code - in all those scenarios we quickly reached the 1k limit.

Additionally this PR also changes the timestamp of the reported `MetricSet`s - as discussed in the weekly meeting, the timestamp should be the timestamp when we report the metrics (and not when the transaction finishes; as it was before). 

Addresses https://github.com/elastic/apm-agent-dotnet/issues/1678

### DIFF
--- a/src/Elastic.Apm/Metrics/MetricSet.cs
+++ b/src/Elastic.Apm/Metrics/MetricSet.cs
@@ -2,6 +2,7 @@
 // Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information
 
+using System;
 using System.Collections.Generic;
 using Elastic.Apm.Api;
 using Elastic.Apm.Api.Constraints;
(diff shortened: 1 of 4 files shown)


src/Elastic.Apm/Metrics/MetricSet.cs
```diff
@@ -27,19 +30,77 @@ public MetricSet(long timestamp, IEnumerable<MetricSample> samples)
 		public SpanInfo Span { get; set; }
 	}
 
-	public class TransactionInfo
+	public class TransactionInfo : IEquatable<TransactionInfo>
```
gregkalapos, line 33: Ideally this could be a `struct`. But this is a `public` type and it'd be a breaking change to change it.

z1c0: Right, but one could argue that implementing `IEquatable` and changing the `Equals` implementation is equally breaking. So we might still consider that and document it.
(We could also make those public types `sealed` in the course of that?)

src/Elastic.Apm/Metrics/MetricsProvider/BreakdownMetricsProvider.cs
```diff
@@ -28,7 +43,6 @@ internal class BreakdownMetricsProvider : IMetricsProvider
 		private bool _loggedWarning;
 
 		private readonly object _lock = new();
```
z1c0, line 45: I know this part was unchanged, but a comment (or variable naming) on which field(s) is/are protected by that `lock` could be helpful.
Maybe considering a thread-safe collection would also be worthwhile here?

test/Elastic.Apm.Tests/BreakdownTests.cs
```diff
@@ -613,16 +684,34 @@ public void BreakdownLogTest()
 				.Be(2);
 		}
 
+		/// <summary>
+		/// According to the spec, the timestamp of the reported MetricSet should be the timestamp when the MetricSet gets reported.
+		/// </summary>
+		[Fact]
+		private async Task TimeStampTest()
```
z1c0, line 691: Good test!

src/Elastic.Apm/Metrics/MetricsProvider/BreakdownMetricsProvider.cs
```diff
@@ -91,12 +120,14 @@ public IEnumerable<MetricSet> GetSamples()
...
-				retVal.AddRange(_itemsToSend);
+				retVal.AddRange(_itemsToSend.Values.ToList());
 				_itemsToSend.Clear();
-				_transactionCount = 0;
 				_loggedWarning = false;
 			}
 
+			var timestampNow = TimeUtils.TimestampNow();
+			// According to the spec, timestampNow should be the time when we report the metrics.
+			foreach (var item in retVal) item.Timestamp = timestampNow;
```
z1c0, line 130: Since we need to touch each `item` anyway, we could replace the `AddRange` above with this loop and add each item individually to `retVal`.

src/Elastic.Apm/Metrics/MetricSet.cs
gregkalapos: As discussed over zoom, we keep it a `class`, just to be on the safe side. The current change is less risky than making it to a `struct`. At the same time I opened https://github.com/elastic/apm-agent-dotnet/issues/1817 - the change itself is trivial, we can do that once we bump the major version.

src/Elastic.Apm/Metrics/MetricsProvider/BreakdownMetricsProvider.cs
gregkalapos: Very good point on naming - I changed it and also added a comment.

As mentioned over zoom, a thread-safe collection would not be enough, since the unit of operation is a bit more than reading/writing from/to the collection - e.g. in `GetSamples` we need to clear the collection as well in the same round. So we keep the lock with better naming.

src/Elastic.Apm/Metrics/MetricsProvider/BreakdownMetricsProvider.cs
gregkalapos: Thanks, good point! I change it.

Issue #1819, opened 2022-09-19
Docker tests are failing


gregkalapos:
Already saw 2 fails right after each other on `main` and now it also failed in one PR.

Example [here](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-dotnet%2Fapm-agent-dotnet-mbp/detail/PR-1818/1/tests).

Log:
```
Error
System.TypeInitializationException : The type initializer for 'DotNet.Testcontainers.Client.DockerApiClient' threw an exception.
---- Docker.DotNet.DockerApiException : Docker API responded with status code=BadRequest, response=400 Bad Request
Stacktrace
System.TypeInitializationException : The type initializer for 'DotNet.Testcontainers.Client.DockerApiClient' threw an exception.
---- Docker.DotNet.DockerApiException : Docker API responded with status code=BadRequest, response=400 Bad Request
Stack Trace:
   at DotNet.Testcontainers.Client.DockerApiClientImage.ByPropertyAsync(String property, String value)
   at DotNet.Testcontainers.Client.DockerApiClientImage.ByNameAsync(String name)
   at DotNet.Testcontainers.Client.AbstractContainerImageClient`1.ExistsWithNameAsync(String name)
   at DotNet.Testcontainers.Client.TestcontainersClient.RunAsync(TestcontainersConfiguration config, CancellationToken ct)
   at DotNet.Testcontainers.Containers.Modules.TestcontainersContainer.Create()
   at DotNet.Testcontainers.Containers.Modules.TestcontainersContainer.StartAsync()
   at Elastic.Apm.SqlClient.Tests.SqlServerFixture.InitializeAsync() in /var/lib/jenkins/workspace/net_apm-agent-dotnet-mbp_PR-1818/apm-agent-dotnet/test/Elastic.Apm.SqlClient.Tests/SqlServerFixture.cs:line 35
----- Inner Stack Trace -----
   at Docker.DotNet.DockerClient.HandleIfErrorResponseAsync(HttpStatusCode statusCode, HttpResponseMessage response, IEnumerable`1 handlers)
   at Docker.DotNet.DockerClient.MakeRequestAsync(IEnumerable`1 errorHandlers, HttpMethod method, String path, IQueryString queryString, IRequestContent body, IDictionary`2 headers, TimeSpan timeout, CancellationToken token)
   at Docker.DotNet.SystemOperations.GetSystemInfoAsync(CancellationToken cancellationToken)
[...]

PR #1820, opened 2022-09-20
Fix failing docker tests

gregkalapos:
Fixes https://github.com/elastic/apm-agent-dotnet/issues/1819 

Prior to this PR we used `DotNet.TestContainer` `Version="1.0.0"`. That [package is already deprecated](https://www.nuget.org/packages/DotNet.Testcontainers/1.7.0-beta.2269) in favor of [Testcontainers](https://www.nuget.org/packages/Testcontainers/2.1.0).

This PR updates to the latest stable release of `Testcontainers` and adapts user code to breaking changes.

Reason for updating the package:
- Most importantly, it fixes the failing tests - unlike the current `main`, this PR with `Testcontainers` can set up the docker environments and tests are green.
- I also noticed it gives better error messages (details on  https://github.com/elastic/apm-agent-dotnet/issues/1819)

### DIFF
--- a/test/Elastic.Apm.Elasticsearch.Tests/ElasticsearchTestContainer.cs
+++ b/test/Elastic.Apm.Elasticsearch.Tests/ElasticsearchTestContainer.cs
@@ -3,14 +3,18 @@
 // Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information
 
-using DotNet.Testcontainers.Containers.Configurations;
-using DotNet.Testcontainers.Containers.Modules.Abstractions;
+
+using System.Linq;
+using DotNet.Testcontainers.Configurations;
+using DotNet.Testcontainers.Containers;
+using Microsoft.Extensions.Logging;
 
 namespace Elastic.Apm.Elasticsearch.Tests
 {
-	public class ElasticsearchTestContainer : HostedServiceContainer
+	public sealed class ElasticsearchTestContainer : TestcontainerDatabase
 	{
-		internal ElasticsearchTestContainer(TestcontainersConfiguration configuration) : base(configuration) => Hostname = "localhost";
+		internal ElasticsearchTestContainer(ITestcontainersConfiguration configuration, ILogger logger)
+			: base(configuration, logger) => ContainerPort = int.Parse(configuration.ExposedPorts.First().Value);
 
 		public override string ConnectionString => $"http://{Hostname}:{Port}";
 	}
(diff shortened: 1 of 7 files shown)


test/Elastic.Apm.SqlClient.Tests/SqlServerFixture.cs
```diff
@@ -38,9 +39,21 @@ public async Task InitializeAsync()
 		public async Task DisposeAsync()
 		{
 			await _container.StopAsync();
-			_container.Dispose();
+			await _container.DisposeAsync();
 		}
 
-		public void Dispose() => _container?.Dispose();
+#if !NET5_0_OR_GREATER
+		private readonly ValueTask _completedValueTask = default;
```
z1c0, line 46: Is there a reason for making this a field?

test/Elastic.Apm.Elasticsearch.Tests/ElasticsearchTestContainer.cs
```diff
@@ -3,14 +3,18 @@
...
+
+using System.Linq;
+using DotNet.Testcontainers.Configurations;
+using DotNet.Testcontainers.Containers;
+using Microsoft.Extensions.Logging;
 
 namespace Elastic.Apm.Elasticsearch.Tests
 {
-	public class ElasticsearchTestContainer : HostedServiceContainer
+	public sealed class ElasticsearchTestContainer : TestcontainerDatabase
```
z1c0, line 14: Just out of curiosity: is there a reason to make this class `sealed` now?

z1c0 (approved):
LGTM
Thanks for looking into this and resolving this so quickly!

test/Elastic.Apm.Elasticsearch.Tests/ElasticsearchTestContainer.cs
gregkalapos: It's because we set `ContainerPort` in the `.ctor` and `ContainerPort` is `virtual`. So this is to make sure the virtual member call in a constructor is fine.

test/Elastic.Apm.SqlClient.Tests/SqlServerFixture.cs
gregkalapos: Probably just me overthinking :)

I got obsessed with `ValueTask.CompletedTask`, but that's only in .NET5+. So I wanted to have something similar and not new up a `ValueTask` directly. On the other hand, this is in the `DisposeAsync` method, which runs once, and the `_completedValueTask` field does not make any difference.

So I ended up removing it, and just do `return new ValueTask();`... it's much simpler in this scenario.

test/Elastic.Apm.Elasticsearch.Tests/ElasticsearchTestContainer.cs
z1c0: Right, makes sense!

test/Elastic.Apm.SqlClient.Tests/SqlServerFixture.cs
z1c0: Yes, easier to read. Thanks!

PR #1824, opened 2022-09-20
Remove warnings in Elastic.Apm(.Tests)

z1c0:
This is the first PR that addresses the load of warnings produced by our build.

Projects.`Elastic.Apm` and `Elastic.Apm.Tests` are close to warning-free now (3 warnings remain that need follow-up).

The biggest chunk of warnings could be removed by introducing a dedicated `.editorconfig` file for the `Libraries` folder which holds `Ben.Demystifier` and `Newtonsoft.Json`).


gregkalapos (approved):
Nice! 🎉

PR #1826, opened 2022-09-23
.ci: remove Integration Tests stage

axw:
See https://github.com/elastic/apm-integration-testing/issues/1523

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -11,8 +11,6 @@ pipeline {
     NOTIFY_TO = credentials('notify-to')
     JOB_GCS_BUCKET = credentials('gcs-bucket')
     CODECOV_SECRET = 'secret/apm-team/ci/apm-agent-dotnet-codecov'
-    GITHUB_CHECK_ITS_NAME = 'Integration Tests'
-    ITS_PIPELINE = 'apm-integration-tests-selector-mbp/main'
     OPBEANS_REPO = 'opbeans-dotnet'
     BENCHMARK_SECRET  = 'secret/apm-team/ci/benchmark-cloud'
     SLACK_CHANNEL = '#apm-agent-dotnet'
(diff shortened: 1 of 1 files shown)


cachedout (approved):

z1c0:
@gregkalapos what's your take on these tests?
Are we relying on them in some way or can we do without them?

v1v (approved):

gregkalapos:
Hmm - I have a bit of a mixed feeling about removing these.

These were the only reel tests that tested against a real APM server and asserted against real data in Elasticsearch for our PRs in this repo. To me that gave confidence that the agent works in a real environment, so I'm a bit sad to see these being removed.

This was the only end-to-end test we had when we opened PRs in this repo.

Basically the biggest value I see in these tests is this: there are these super embarrassing things when an agent is built, it starts up without any error, but then when it connects to a real setup (real APM Server, real Elasticsearch), then it doesn't send data (I saw that happening). That scenario is well tested with these integration tests.

> The agent integration tests are of limited value. The tests are fairly basic,

That is definitely true - these tests did not cover most of the agent features. On the other hand they catch trivial issues quickly in a real setup. I see huge value in having a test which validates the basic features of the agent against a real setup.

>  and we have since added fake APM Server implementations to the agent repos, which validate payloads.

Yes, I also agree with that, and it's definitely a better way to validate payload. But that's still not a real end-to-end integration test. In general once something is mocked in a test I think it's good for testing specific component, but in my opinion a real end-to-end test is necessary to make sure faking/mocking any component did not cause any untested (or not correctly tested) code.

> The cost of maintaining the integration tests exceeds the value they are providing, therefore we should remove them. Later on we might replace them with something a little more maintainable. For example: disentangling stack/server provisioning, and distributing the tests to each agent repository. We can come back to that as needed.

That's the point where I can't disagree, because I'm in a different position. For .NET we did not had to maintain this much, so the cost of maintaining these was totally acceptable for us (.NET Team) - but I don't know how much of an effort this is for others. If it's too much, then of course I accept it.

So... overall, I see this a bit risky (we loose our only end-to-end testing here in the .NET Agent pipeline), but if other teams can't maintain these tests, then I guess there is no way around it.

z1c0:
Thanks, @gregkalapos for so eloquently summarizing my gut feeling in your  [comment](https://github.com/elastic/apm-agent-dotnet/pull/1826#issuecomment-1255988572).

I'm also really hesitant about removing those tests if they are the only *real* end-to-end smoke tests we have.

axw:
@gregkalapos @z1c0 I understand where you're coming from.

My initial thought was not to remove them, but to replace them with an alternative that would live in the agent repo, and be maintained by the agent developers: https://github.com/elastic/apm-dev/issues/831. (Apologies to external folks for referencing a private issue; the gist is to move all the smoke testing logic out of the all-in-one apm-integration-testing repo and into the agent repo, and reuse existing Terraform automation for creating an Elastic Cloud deployment.)

I haven't heard from the RUM agent team yet, but no other agent team has expressed an interest in maintaining the status quo.

If you would like to maintain some smoke tests that include a real APM Server and real Elasticsearch, I would suggest implementing something along the lines of what we did for the AWS Lambda extension: https://github.com/elastic/apm-aws-lambda/tree/main/tf (which is basically what I described in the issue linked above). If you agree with that, then the question is: can we remove this first, or do we need to wait for a replacement? What's the level of risk? When was the last time a bug escaped to apm-integration-testing?

gregkalapos:
Thanks for explaining @axw.

I should have commented on the linked internal issue (it seems I have a different opinion compared to others), but too late.. :) 

> If you would like to maintain some smoke tests that include a real APM Server and real Elasticsearch, I would suggest implementing something along the lines of what we did for the AWS Lambda extension: https://github.com/elastic/apm-aws-lambda/tree/main/tf (which is basically what I described in the issue linked above).

Yeah, I strongly think those tests are valuable. Thanks for the suggestion - sounds good to me.

>  If you agree with that, then the question is: can we remove this first, or do we need to wait for a replacement? 

If that helps the overall team (less maintenance effort for others), then of course I agree. I'd say, in order to unblock this, let's merge this and I'll open an issue in this repo for creating the replacement, which we can then discuss in the .NET planning to see what others think. 

> What's the level of risk? When was the last time a bug escaped to apm-integration-testing?

I'd say it's acceptable. I couldn't recall when was the last time we had that happening - definitely not recently.

gregkalapos:
Side note: ignore the one failing test in the PR, I'm handling it here: https://github.com/elastic/apm-agent-dotnet/issues/1819#issuecomment-1256026300  - it's ok to merge.

PR #1827, opened 2022-09-23
Remove warnings in Elastic.Apm(.Tests)

z1c0:
This removes the remaining warnings in `Elastic.Apm` and `Elastic.Apm.Tests`.

* `CloudMetadataProviderCollection` produced a warning due to hiding the method `TryGetValue` from its base class. However, this method was only used in tests and was not really required. Hence, I just removed it and updated the tests accordingly.
* I added a `pragma` to suppress the obsolete warning of `ServicePointManager`.


gregkalapos (approved):
LGTM

PR #1828, opened 2022-09-23
Fix line endings

z1c0:
Apparently, there was some inconsistency in this file when it comes to whitespace (line endings).
I inadvertently committed this change with #1806 when working in a Windows environment.
Since then, this file permanently shows up in my changeset. This PR should fix that.

### DIFF
--- a/src/Elastic.Apm.MongoDb/LICENSE
+++ b/src/Elastic.Apm.MongoDb/LICENSE
@@ -199,7 +199,7 @@ Apache License
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
-
+
 ==========
 Elastic.Apm.MongoDb
 ----------


gregkalapos (approved):
Ahh, this is known and happens when we jump between macOS/Linux and Windows - so far I just ignored and did not commit this file, which is probably not the best thing :)

I also overlooked it in https://github.com/elastic/apm-agent-dotnet/pull/1806 - since the whole file changed GitHub collapsed it, so it's easy to not spot.

Anyways... I think we can merge this and then long term try to fix it.

Some thoughts quickly:
- These `LICENSE` files are generated here: https://github.com/elastic/apm-agent-dotnet/blob/main/src/Directory.Build.targets#L7-L15
- Our `.editorconfig` fixes the line endings of it: https://github.com/elastic/apm-agent-dotnet/blob/main/.editorconfig#L215-L216 
- I guess the combination of these 2 causes the trouble, because on Windows (depending on git configs) you may end up with CRLF which then regenerates the LICENSE file on build (because it changed...). 
- Maybe we should just remove https://github.com/elastic/apm-agent-dotnet/blob/main/.editorconfig#L215-L216? 🤔

z1c0:
Interestingly, this has always and only happened with this one file (`src/Elastic.Apm.MongoDb/LICENSE`) so far.
Let's see if the problem keeps popping up again and then find a proper fix (like you suggested [here](https://github.com/elastic/apm-agent-dotnet/pull/1828#pullrequestreview-1118247841)).

Issue #1823, opened 2022-09-20
Build produces warnings


z1c0:
Currently, building `apm-dotnet-agent` produces many warnings, which makes it difficult to spot newly introduced warnings.

Ideally, after getting rid of all current warnings, we enable "treat warnings as errors" to fail a build as soon as  PR introduces new warnings.

PR #1846, opened 2022-09-29
Use json-specs/cgroup_parsing.json as test input

z1c0:
This PR introduces the `JsonFileDataAttribute` xUnit test attribute.
It allows specifying:
* the path to a JSON file which represents the test input data.
* a `Type` into which a single test input dataset will be converted.

`SystemInfoHelperTests.ParseKubernetesInfo_FromCGroupLine` (the "group parsing test") is the first test that makes use of this attribute.

### DIFF
--- a/test/Elastic.Apm.Tests.Utilities/JsonFileDataAttribute.cs
+++ b/test/Elastic.Apm.Tests.Utilities/JsonFileDataAttribute.cs
@@ -1,0 +1,46 @@
+// Licensed to Elasticsearch B.V under one or more agreements.
+// Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using Elastic.Apm.Libraries.Newtonsoft.Json.Linq;
+using Xunit.Sdk;
+
+namespace Elastic.Apm.Tests.Utilities;
+
+public class JsonFileDataAttribute : DataAttribute
+{
+	private readonly string _fileName;
+	private readonly Type _inputDataType;
+
+	public JsonFileDataAttribute(string fileName, Type inputDataType)
+	{
+		_fileName = fileName;
+		_inputDataType = inputDataType;
+	}
+
+	public override IEnumerable<object[]> GetData(MethodInfo testMethod)
+	{
+		if (!File.Exists(_fileName))
+			throw new ArgumentException($"JSON input file {_fileName} does not exist");
+
+		var jToken = JToken.Parse(File.ReadAllText(_fileName));
+		switch (jToken.Type)
+		{
+			case JTokenType.Array:
+				foreach (var t in jToken)
+					yield return new []{ t.ToObject(_inputDataType) };
+				break;
+			case JTokenType.Object:
+				foreach (var kvp in (JObject)jToken)
+					yield return new [] { kvp.Key, kvp.Value.ToObject(_inputDataType) };
+				break;
+			default:
+				throw new Exception($"Unexpected JSON input: '{jToken}");
+		}
+	}
+}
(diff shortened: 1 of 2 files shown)


test/Elastic.Apm.Tests/SystemInfoHelperTests.cs
```diff
@@ -9,6 +9,7 @@
 using Elastic.Apm.Helpers;
 using Elastic.Apm.Tests.Utilities;
 using FluentAssertions;
+using Newtonsoft.Json.Linq;
```
gregkalapos, line 12:
```suggestion
```

and `using Elastic.Apm.Api;` can be removed as well (I can't change it here, because it's untouched)

test/Elastic.Apm.Tests/SystemInfoHelperTests.cs
```diff
@@ -44,25 +45,21 @@ public void ParseKubernetesInfo_ShouldReturnNull_WhenNoEnvironmentVariablesAreSe
...
-			"22949dce-fd8b-11ea-8ede-98f2b32c645c")]
-		// ubuntu cgroup
-		[InlineData("1:name=systemd:/user.slice/user-1000.slice/user@1000.service/apps.slice/apps-org.gnome.Terminal" +
-			".slice/vte-spawn-75bc72bd-6642-4cf5-b62c-0674e11bfc84.scope", null, null)]
-		// AWS ECS cgroup
-		[InlineData("1:name=systemd:/ecs/03752a671e744971a862edcee6195646/03752a671e744971a862edcee6195646-4015103728",
-			"03752a671e744971a862edcee6195646-4015103728", null)]
-		public void ParseKubernetesInfo_FromCGroupLine(string line, string containerId, string podId)
+		[JsonFileData("./TestResources/json-specs/cgroup_parsing.json", typeof(CGroupTestData))]
+		public void ParseKubernetesInfo_FromCGroupLine(string name, CGroupTestData data)
```
gregkalapos, line 57: `string name` is unused in the method and Rider shows a warning. On the other hand we can't remove it, because it's in the data that gets passed into the method, plus all the debug output shows these parameters and that's super useful, because if it fails, we know which test-case fails.

Adding `#pragma warning disable xUnit1026` (disabling the warning) would maybe just make the code more noisy, I'm not sure if it's better than just accepting as is.

This is probably fine as is, I'm just adding here what I was thinking.

gregkalapos (approved):
Nice, thank you! 

I added some comments, all are just my thoughts, none of those is a blocker.

test/Elastic.Apm.Tests.Utilities/JsonFileDataAttribute.cs
```diff
@@ -0,0 +1,46 @@
...
+		var jToken = JToken.Parse(File.ReadAllText(_fileName));
+		switch (jToken.Type)
+		{
+			case JTokenType.Array:
+				foreach (var t in jToken)
+					yield return new []{ t.ToObject(_inputDataType) };
+				break;
+			case JTokenType.Object:
+				foreach (var kvp in (JObject)jToken)
+					yield return new [] { kvp.Key, kvp.Value.ToObject(_inputDataType) };
```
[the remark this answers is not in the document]
z1c0: My plan was to evolve this helper class with every adaption of another `json-spec` test.
If you're ok with that, I'll leave it as it is for now and iron out any quirks that might come along.

test/Elastic.Apm.Tests/SystemInfoHelperTests.cs
z1c0: Thanks, I actually missed that warning.
I guess it's really time we get our warnings down to zero in https://github.com/elastic/apm-agent-dotnet/issues/1823 and flag them as errors from then on.
I will add the suggested `#pragma`.

test/Elastic.Apm.Tests.Utilities/JsonFileDataAttribute.cs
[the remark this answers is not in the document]
gregkalapos: 👍 sounds good.

>  to evolve this helper class with every adaption of another `json-spec` test.

I was thinking exactly the same :) We control the data that gets parsed here (the new json files), if this is an issue, we can handle it on-demand

PR #1847, opened 2022-09-30
Use json-specs/w3c_distributed_tracing.json test input

z1c0:
* Extend `DistributedTracingTests` to leverage the test cases from `json-specs`.
* Small fixes to `TraceContext` and `TraceState.cs` that were uncovered by those test cases.
* Make the `inputDataType` parameter optional in `JsonFileDataAttribute` and fall back to generic `JToken`.
  * Note that this is not necessarily required by the test in this PR (`DistributedTracingTests.cs`) but it will be handy in future tests.

### DIFF
--- a/test/Elastic.Apm.Tests/DistributedTracing/DistributedTracingTests.cs
+++ b/test/Elastic.Apm.Tests/DistributedTracing/DistributedTracingTests.cs
@@ -31,7 +31,7 @@ namespace Elastic.Apm.Tests.DistributedTracing
 				.Select(h => h[1]);
 			var traceStates = data.Headers
 				.Where(h => h[0].Equals("tracestate", StringComparison.InvariantCultureIgnoreCase))
-				.Select(h => h?[1]);
+				.Select(h => h[1]);
 			// Multiple occurrences of the trace headers are automatically invalid.
 			if (traceParents.Count() < 2 && traceStates.Count() < 2)
 			{
--- a/test/Elastic.Apm.Tests/DistributedTracing/DistributedTracingTests.cs
+++ b/test/Elastic.Apm.Tests/DistributedTracing/DistributedTracingTests.cs
@@ -2,7 +2,12 @@
 // Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information
 
+using System;
+using System.Collections.Generic;
+using System.Linq;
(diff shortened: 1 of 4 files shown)


test/Elastic.Apm.Tests/DistributedTracing/DistributedTracingTests.cs
```diff
@@ -2,14 +2,56 @@
...
+			[JsonProperty("is_tracestate_valid")] public bool? IsTraceStateValid { get; set; }
+		}
+
+		[Theory]
+		[JsonFileData("./TestResources/json-specs/w3c_distributed_tracing.json", typeof(TraceStateTestData))]
+		public void TestCasesFromJsonSpec(TraceStateTestData data)
+		{
+			var traceParents = data.Headers
+				.Where(h => h[0].Equals("traceparent", StringComparison.InvariantCultureIgnoreCase))
+				.Select(h => h?[1]);
```
gregkalapos, line 31:
```suggestion
				.Select(h => h[1]);
```

nit: I think the `?` null-checks the array itself (which is never `null`), but not the 1. item. Maybe something like this would make it: ~`.Select(h => h.Length < 2 ? string.Empty : h[1]);`~ Update: `h.ElementAtOrDefault(1)`. 

But I think no null-check is needed here due to the nature of the data from the json file (there is always a 1. item in that array).

test/Elastic.Apm.Tests/DistributedTracing/DistributedTracingTests.cs
```diff
@@ -2,14 +2,56 @@
...
+			var traceStates = data.Headers
+				.Where(h => h[0].Equals("tracestate", StringComparison.InvariantCultureIgnoreCase))
+				.Select(h => h?[1]);
```
gregkalapos, line 34:
```suggestion
				.Select(h => h[1]);
```

Same as above.

gregkalapos (approved):
Lgtm

test/Elastic.Apm.Tests/DistributedTracing/DistributedTracingTests.cs
z1c0: Nice catch!
I was using `SelectFirstOrDefault` initially (which could produce actual `null` values for the array) before I realized that there were test cases with multiple hits. The `?` is a relict from that.

Issue #1794, opened 2022-08-19
Log reason for transaction to end unsampled


ornic:
It would be much help to separate transactions dropped from sampling for different reasons.

**Is your feature request related to a problem? Please describe.**
Yes. We have an instance of Framework 4,8.2 app that connected to company kibana server.

And all requests from our outsource partners marked as unsampled, we got tons of `{Transaction} Dropping unsampled transaction - unsampled transactions won't be sent on APM Server v8+. TransactionId: ...` messages in logs. And in APM there is not a single transaction, but stable statistics for time spent in spans.

But when I try to make those same requests from my browser / insomnia - they all appear perfectly as transactions in kibana with pretty spans of internal api / db requests (the single peak in the picture after service redeploy).

![image](https://user-images.githubusercontent.com/3350070/185641230-f5e75309-cbb7-467a-a31b-8acd2f4db580.png)

**Describe the solution you'd like**
Option in settings to produce Trace level logs: transaction with id=... was dropped from sampling due to head-based sampling rate settings. 
Option in settings to produce Debug level logs: transaction with id=... was dropped from sampling due to `some other reason`.

**Describe alternatives you've considered**
Gooooooogling. Got me nothing, unfortunately

ornic:
In our case the reason for unsampling was incoming net5 W3C header, that we [started to ignore](https://www.elastic.co/guide/en/apm/agent/dotnet/current/config-supportability.html#config-trace-context-ignore-sampled-false).

gregkalapos:
Thanks for reporting this @ornic.

I think I understand the overall issue and the suggestion here, but I'd have one question to think this through completely: where do you expect to see the reason for not sampling a transaction?

In Kibana you don't see it, but you'll see it in the agent logs. [Here we log this info](https://github.com/elastic/apm-agent-dotnet/blob/main/src/Elastic.Apm/Model/Transaction.cs#L307-L323) when the transaction is created. But for that you'll need to look into the agent logs.

Changing logs in the agent is fairly easy and if that could be improved, we can definitely do it quickly. On the other hand, having a reason for not sampling a transaction in Kibana would require more work (that involves APM Server and Kibana as well, not only the agent).

ornic:
> where do you expect to see the reason for not sampling a transaction?

Ideally: inside the log entry about dropping unsampled transaction that is logged on agent side.

[...]

PR #1850, opened 2022-10-04
Use debug log level when dropping spans

z1c0:
As suggested by @felixbarny in https://github.com/elastic/sdh-apm/issues/738#issuecomment-1266545799, let's use a more conservative log level (`Debug`) when dropping spans to prevent log flooding.

### DIFF
--- a/src/Elastic.Apm/Model/Transaction.cs
+++ b/src/Elastic.Apm/Model/Transaction.cs
@@ -650,7 +650,7 @@ namespace Elastic.Apm.Model
 			{
 				if (!CompressionBuffer.IsSampled && _apmServerInfo?.Version >= new ElasticVersion(8, 0, 0, string.Empty))
 				{
-					_logger?.Info()
+					_logger?.Debug()
 						?.Log("Dropping unsampled compressed span - unsampled span won't be sent on APM Server v8+. SpanId: {id}",
 							CompressionBuffer.Id);
 				}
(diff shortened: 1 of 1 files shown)


gregkalapos (approved):
LGTM 👍 

I agree this can potentially log a lot with a lower sample rate and debug seems to be better.

Just mentioning, since it's related: we had this discussion that users have a hard time having an overview on why and/or how many transactions were unsampled. One discussion here: https://github.com/elastic/apm-agent-dotnet/issues/1794 (it wasn't about this log line, but a similar issue). But that problem wasn't not solved by this `info` log (as discussed in a weekly, it's potentially better covered with a metric in the future) and I agree logging on `debug` is better to avoid log spamming.

felixbarny (approved):

felixbarny:
> as discussed in a weekly, it's potentially better covered with a metric in the future

The new storage explorer can also show the effective sample rate of a service. Is that what you had in mind?

gregkalapos:
> > as discussed in a weekly, it's potentially better covered with a metric in the future
> 
> The new storage explorer can also show the effective sample rate of a service. Is that what you had in mind?

I didn't know that - that definitely helps on having an overview about the number of unsampled transactions. But I guess it doesn't cover the reason why those were unsampled. There can be multiple reasons - sample rate, traceparent sampled flag, something in the APM server, etc. In https://github.com/elastic/apm-agent-dotnet/issues/1794 the reason for dropping transactions was challenging to figure out.

PR #1851, opened 2022-10-04
Use json-specs/wildcard_matcher_tests.json test input

z1c0:
Use `dynamic` type for the test data as the object's property names represent the test cases here.

### DIFF
--- a/test/Elastic.Apm.Tests/WildcardMatcherTests.cs
+++ b/test/Elastic.Apm.Tests/WildcardMatcherTests.cs
@@ -5,7 +5,10 @@
 
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq;
 using Elastic.Apm.Helpers;
+using Elastic.Apm.Libraries.Newtonsoft.Json.Linq;
+using Elastic.Apm.Tests.Utilities;
 using FluentAssertions;
 using Xunit;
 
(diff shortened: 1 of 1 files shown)


test/Elastic.Apm.Tests/WildcardMatcherTests.cs
```diff
@@ -5,14 +5,36 @@
...
+#pragma warning disable xUnit1026
+		[Theory]
+		[JsonFileData("./TestResources/json-specs/wildcard_matcher_tests.json")]
+		public void TestCasesFromJsonSpec(string testCaseName, dynamic data)
+		{
+			var testData = (JObject)data;
+			var key = testData.Properties().First();
+			var value = (JObject)key.Last;
+			var matcher = WildcardMatcher.ValueOf(key.Name);
+			foreach (var kvp in value)
```
gregkalapos, line 29: Since Rider shows this on my machine, I'm adding it here:
`value` could technically be `null` - same for `kvp.Value.Value<>` below.

image

On the other hand, I don't think we should have totally bulletproof null-checks in test code. I personally think it's much better to have good readability, and if the test code throws nullref exceptions, then we'll look at the test data.

I leave this up to you.

gregkalapos (approved):
LGTM

test/Elastic.Apm.Tests/WildcardMatcherTests.cs
z1c0: If it's fine for you, I'll just leave this as it is.
As you [said in a related PR](https://github.com/elastic/apm-agent-dotnet/pull/1846#discussion_r984407624), the input data is in our control anyway and it would be an easy fix should a future test case yield this problem.

PR #1858, opened 2022-10-06
Use json-specs/service_resource_inference.json test input

z1c0:
Add support for the service inference test cases provided via json spec.
Even though the discussion regarding default port handling (see https://github.com/elastic/apm/issues/703) is not resolved yet, I figured this PR already provides value.
* It uncovered minor inconsistencies and spec violations when deducting service targets (e.g. priorities of `db`, `message`, `http`).
* As soon as the default port discussion is resolved we need to either update our tests or update the json spec.


src/Elastic.Apm/Model/Span.cs
```diff
@@ -734,7 +734,7 @@ private void DeduceServiceTarget()
 			if (!IsExitSpan)
 				return;
 
-			if (!_context.IsValueCreated)
+			if (!_context.IsValueCreated && !IsExitSpan)
```
gregkalapos, line 737: - What here actually happens is that we store `Context.Service.Target` (and resource) on the top level in `DroppedSpanStatCache` for dropped spans and then return - with this we avoid initialization of the huge `Context` object, which would happen later in the method. At least that's how I interpret this :)  So... shouldn't be an `&& _isDropped` in this check? Because if the span is not dropped, we don't want to set `DroppedSpanStatCache`.
- I feel a one liner comment would help here a lot. It took me a while to understand this :) Maybe let's add something like this before the line (just an idea, feel free to change the text): "In order to avoid the creation of `Context`, set `target` and `resource` on the top level `DroppedSpanStatCache` and return".

gregkalapos (approved):
LGTM!

I added 2 thoughts to the modified check below - I don't see those as blocker, just something that I thought when I looked at the code.

src/Elastic.Apm/Model/Span.cs
z1c0: * You mean, this check for `_isDropped` should have been here in the first place? But yes, I'll add it locally and run the tests again.
* Yes, an explanatory comment is definitely a good idea here.

gregkalapos: > You mean, this check for _isDropped should have been here in the first place? 

Yes. So the original check wasn't very correct - I guess that code was from me.

z1c0: > > You mean, this check for _isDropped should have been here in the first place?
> 
> Yes. So the original check wasn't very correct - I guess that code was from me.

No blaming ;-) - this was just for my understanding.

PR #1859, opened 2022-10-06
Verify spans in MockPayloadSender using json-spec information

z1c0:
Add the ability to `MockPayloadSender` of **strict span checking** based on the information in `TestResources/json-specs/span_types.json`.

For now, the strict span check is turned off as it would require too many changes in existing test code (where span type names like "test", "foo", "bar", ... are used). As an example, `IsStrictSpanCheckEnabled` is activated for `GrpcTests`.

The long-term goal is to make `IsStrictSpanCheckEnabled` an opt-out setting (see #1893).


gregkalapos (approved):
LGTM

PR #1863, opened 2022-10-10
Move json-spec TestResources into into Elastic.Apm.Tests.Utilities

z1c0:


z1c0:
> Maybe one small thing left: in `test/Elastic.Apm.Tests/Elastic.Apm.Tests.csproj` we have the same section with ` <Content Include="TestResources\**\**\*.*">` - that can be removed as well, right?

There are still other resources (`db`, `proc`) in there, so I think we need to leave it as is.

gregkalapos:
> > Maybe one small thing left: in `test/Elastic.Apm.Tests/Elastic.Apm.Tests.csproj` we have the same section with ` <Content Include="TestResources\**\**\*.*">` - that can be removed as well, right?
> 
> There are still other resources (`db`, `proc`) in there, so I think we need to leave it as is.

Oh, you are totally right. My bad, sorry for the confusion.

z1c0:
> > > Maybe one small thing left: in `test/Elastic.Apm.Tests/Elastic.Apm.Tests.csproj` we have the same section with ` <Content Include="TestResources\**\**\*.*">` - that can be removed as well, right?
> > 
> > 
> > There are still other resources (`db`, `proc`) in there, so I think we need to leave it as is.
> 
> Oh, you are totally right. My bad, sorry for the confusion.

No worries 😄

PR #1864, opened 2022-10-11
Move custom .NET SQL parsing tests to Benchmark project

z1c0:
Move the custom test input data file `db/sql_signature_examples.json` from `Elastic.Apm.Tests` to `Elastic.Apm.Benchmarks`.
The test cases are mostly duplicates (except for one) and we cover those in the the `json-spec` tests now anyway.
The one additional test case is a very long SQL query which is why it is relevant for benchmarking.

So the remaining question is: should we leave the rest of the test cases in the file as well? Are those relevant for benchmarking? Or are we only interested in this one additional test case? I'll leave this decision to you @gregkalapos 😉 .


gregkalapos (approved):
> So the remaining question is: should we leave the rest of the test cases in the file as well? Are those relevant for benchmarking? Or are we only interested in this one additional test case? I'll leave this decision to you @gregkalapos 😉 .

Those tests aren't relevant for benchmarking. From what I remember, those never really revealed any issue, we really needed the long sql statement to clearly hit the performance issue. I'd suggest we remove the other samples, and only keep the long one - that way, we'll only have relevant data for the benchmark.

Otherwise this is good to merge in my opinion.

z1c0:
> ... I'd suggest we remove the other samples, and only keep the long one - that way, we'll only have relevant data for the benchmark.

Perfect 👍 , I'll update the PR.

PR #1867, opened 2022-10-11
Improve handling of multiple agent initialization

gregkalapos:
Result of our debugging session from today.

Goal of the PR:
- Add the missing `return` in `Agent.Setup`.
- Improve logging, which can help us debug issues around agent initialization in the future.

What this PR changes
- `Agent.Setup(AgentComponents)` will return on the 2. (and subsequent) calls early and really reject the new `AgentComponents` parameter (basically added the missing `return`)
- In above case, we log both on the existing logger and on the rejected one as well (the idea is to make sure the error shows up). We also log the call-stack in case we reject the parameter - this way we'll see the code-path that leads to the second/rejected call to `Agent.Setup`. 
- Also log in the initialization of the `LazyApmAgent` including call stack. This will show the code path which leads to the implicit initialization (typically without calling `Agent.Setup()`, but calling e.g. `Agent.Tracer` directly as the 1. interaction with the agent).
- Added `lock` to the initialization logic to avoid running it in parallel in case there are multiple threads calling it.
- Extended the test to cover above changes.
- Add more logging to [AutoInstrumentation.cs](https://github.com/elastic/apm-agent-dotnet/blob/main/src/Elastic.Apm.Profiler.Managed/AutoInstrumentation.cs)


z1c0 (commented):
Could you add some logging to https://github.com/elastic/apm-agent-dotnet/blob/main/src/Elastic.Apm.Profiler.Managed/AutoInstrumentation.cs as well?
The value of `skipInstantiation` will be useful to have in our logs.

src/Elastic.Apm/Agent.cs
```diff
@@ -162,16 +171,31 @@ private static bool CheckAndAddFilter(Action<PayloadSenderV2> action)
...
+				if (_isConfigured)
+				{
+					Components?.Logger?.Error()
+						?.Log("The singleton APM agent has" +
+							" already been instantiated and can no longer be configured. Reusing existing instance. "
+							+ "Callstack: {callstack}", new StackTrace().ToString());
+
+					// Above line logs on the already configured `Components`
+					// In order to let the caller know, we also log on the logger of the rejected `agentComponents`
+					agentComponents.Logger?.Error()
```
z1c0, line 185:
```suggestion
					agentComponents?.Logger?.Error()
```

src/Elastic.Apm/Agent.cs
```diff
@@ -58,12 +58,21 @@ public static class Agent
 	{
 		private static readonly Lazy<ApmAgent> LazyApmAgent = new Lazy<ApmAgent>(() =>
 		{
-			_isConfigured = true;
-			return new ApmAgent(Components);
+			lock (InitializationLock)
+			{
+				_isConfigured = true;
+				var agentComponents = new ApmAgent(Components);
```
z1c0, line 64:
```suggestion
				var agent = new ApmAgent(Components);
```

src/Elastic.Apm/Agent.cs
```diff
@@ -58,12 +58,21 @@ public static class Agent
...
+
+				agentComponents?.Logger?.Trace()
+					?.Log("Initialization - Agent instance initialized. Callstack: {callstack}", new StackTrace().ToString());
+
+				return agentComponents;
```
z1c0, line 69:
```suggestion
				return agent;
```

src/Elastic.Apm/Agent.cs
```diff
@@ -162,16 +171,31 @@ private static bool CheckAndAddFilter(Action<PayloadSenderV2> action)
...
-
-			Components = agentComponents;
-			_isConfigured = true;
+			lock (InitializationLock)
+			{
+				if (_isConfigured)
+				{
+					Components?.Logger?.Error()
+						?.Log("The singleton APM agent has" +
+							" already been instantiated and can no longer be configured. Reusing existing instance. "
```
z1c0, line 180: I just realized that technically this error message will be wrong/misleading when you call `Agent.Setup` **twice**.
In that case, no agent is actually instantiated at all, but the log message says so.
That was probably the reason for the existing code to check on the `LazyApmAgent.IsValueCreated` value.

Maybe bringing these two code parts (`LazyApmAgent` and `Setup`) together would make our code more readable.

src/Elastic.Apm/Agent.cs
```diff
@@ -162,16 +171,31 @@ private static bool CheckAndAddFilter(Action<PayloadSenderV2> action)
...
-				?.Log("Initialization - Agent.Setup called. Callstack: {callstack}", new StackTrace().ToString());
-
-			Components = agentComponents;
-			_isConfigured = true;
+			lock (InitializationLock)
+			{
+				if (_isConfigured)
+				{
+					Components?.Logger?.Error()
+						?.Log("The singleton APM agent has" +
```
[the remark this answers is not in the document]
gregkalapos: I tried, but then I realized this is a parametrized log. So we pass the callstack as a parameter to the `Log` method and that applies a formatter. This can then keep the parameters as key-value pairs (in this case `callstack` as key and then the callstack itself as value) and ship the log that way. E.g. in this case if you ship to elasticsearch this will be searchable.

If I prepare a string we loose the formatter which is applied by the logger.

No strong opinion, but after looking into this, I'd keep it this way.

src/Elastic.Apm/Agent.cs
gregkalapos: 👍  this is a good point.

I'm not sure the `_isConfigured` variable is needed. I removed it and now it only relies on `LazyApmAgent`.

src/Elastic.Apm/Agent.cs
```diff
@@ -162,16 +168,32 @@ private static bool CheckAndAddFilter(Action<PayloadSenderV2> action)
...
+					agentComponents?.Logger?.Error()
+						?.Log("The singleton APM agent has" +
+							" already been instantiated and can no longer be configured. Reusing existing instance. "
+							+ "Callstack: {callstack}", new StackTrace().ToString());
+
+					return;
+				}
+
+				agentComponents?.Logger?.Trace()
+					?.Log("Initialization - Agent.Setup called");
```
gregkalapos, line 191: We don't print the callstack here, because below we force initialization with `LazyApmAgent.Value` and that getter prints the callstack.

z1c0 (approved):

gregkalapos:
@z1c0 sorry, for pushing one more change, but the logs you mentioned in [AutoInstrumentation.cs](https://github.com/elastic/apm-agent-dotnet/blob/main/src/Elastic.Apm.Profiler.Managed/AutoInstrumentation.cs) were not yet done.

I pushed that in https://github.com/elastic/apm-agent-dotnet/pull/1867/commits/29b5acb8def29e053873642a73fc2b6d58540e55 - could you please look at that as well?

z1c0:
> @z1c0 sorry, for pushing one more change, but the logs you mentioned in [AutoInstrumentation.cs](https://github.com/elastic/apm-agent-dotnet/blob/main/src/Elastic.Apm.Profiler.Managed/AutoInstrumentation.cs?rgh-link-date=2022-10-17T11%3A20%3A04Z) were not yet done.
> 
> I pushed that in [29b5acb](https://github.com/elastic/apm-agent-dotnet/commit/29b5acb8def29e053873642a73fc2b6d58540e55) - could you please look at that as well?

Right, those logs will definitely be useful.
Thanks for adding them.

PR #1875, opened 2022-10-14
Remove `Opbeans` stage from pipeline

gregkalapos:
This is a stage which consistently fails when we release. An example can be seen [here](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-dotnet%2Fapm-agent-dotnet-mbp/detail/v1.18.0/1/pipeline/):

```
[2022-10-13T20:28:11.078Z] error: Unable to find package Elastic.Apm.NetCoreAll with version (>= 1.18.0)
[2022-10-13T20:28:11.078Z] error:   - Found 30 version(s) in nuget.org [ Nearest version: 1.17.0 ]
```
We pushed out `1.18.0`, but it's not yet available.

Same story on all releases.

### What's this stage
We run this when we release (only on releases based on tag name). The intention with it is to update [opbeans-dotnet](https://github.com/elastic/opbeans-dotnet) to the agent version we release in the same pipeline. So e.g. we release `v1.18.0` which we publish to nuget.org, and then we want opbeans-dotnet to use this version.

### Why I suggest to remove
Unfortunately this consistently fails, because we try to update the agent packages in opbeans-dotnet to a version which we just pushed a few seconds earlier in an earlier step. This never works, because NuGet needs some time to index the new versions. Therefore we need another way to automatically bump up the agent version in opbeans-dotnet.

Since this makes release pipeline red, I think the best is to just remove it (failing releases due to this is in my opinion really not ok), and then at the same time I opened an issue so we can think about a better way to automate this later: https://github.com/elastic/apm-agent-dotnet/issues/1876

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -548,36 +548,6 @@ pipeline {
         }
       }
     }
-    stage('AfterRelease') {
-      options {
-        skipDefaultCheckout()
-      }
-      when {
-        anyOf {
-          tag pattern: 'v\\d+\\.\\d+\\.\\d+', comparator: 'REGEXP'
-        }
-      }
-      stages {
-        stage('Opbeans') {
-          environment {
-            REPO_NAME = "${OPBEANS_REPO}"
-          }
-          steps {
-            deleteDir()
-            dir("${OPBEANS_REPO}"){
-              git(credentialsId: 'f6c7695a-671e-4f4f-a331-acdce44ff9ba',
-                  url: "git@github.com:elastic/${OPBEANS_REPO}.git",
-                  branch: 'main')
-              sh script: ".ci/bump-version.sh ${env.BRANCH_NAME.replaceAll('^v', '')}", label: 'Bump version'
-              // The opbeans pipeline will trigger a release for the main branch
-              gitPush()
-              // The opbeans pipeline will trigger a release for the release tag with the format v<major>.<minor>.<patch>
-              gitCreateTag(tag: "${env.BRANCH_NAME}")
-            }
-          }
-        }
-      }
-    }
   }
   post {
     cleanup {


v1v (commented):
Is it possible to query nuget for any given release? I saw `nuget list` maybe that could work well with a wait for ?

gregkalapos:
> Is it possible to query nuget for any given release? I saw `nuget list` maybe that could work well with a wait for ?

I'm not sure that's the best way to go. 

We don't have the `nuget` tool in this context (`nuget.exe` is windows only - maybe we could run it with some hack, but I'm not sure if it's a good idea) - what we have is the `dotnet` CLI, which has built in `nuget` support, but from what I found does not support listing versions (e.g. [StackOverflow](https://stackoverflow.com/questions/57742638/how-to-list-available-package-versions-using-the-dotnet-cli)) and there is [`dotnet nuget list`](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-nuget-list-source), but that's different from `nuget list`.

Having said that, we can actually test if a specific version exists with the exact same command we already use. We can just run:

```
dotnet add package Elastic.Apm.NetCoreAll -v [Version]'
```

If you see the output above (specifically `Unable to find package Elastic.Apm.NetCoreAll with version (>= [Version])`), then the package exists, but not the new version. So based on this output we can wait and retry.

But do we really want to hold up a release pipeline just because of this? 🤔

z1c0:
> But do we really want to hold up a release pipeline just because of this? 🤔

Although I could not find any authoritative information on how long it takes for a package to become available for use, from practical experience I would say it's in the ballpark of 5-10 minutes.
So, the question is if we are comfortable with adding this additional delay to our release pipeline.

My tendency would be to do so and wait for the package to become availalbe.

gregkalapos:
>  from practical experience I would say it's in the ballpark of 5-10 minutes.

Yeah, same experience.

> So, the question is if we are comfortable with adding this additional delay to our release pipeline.

Here is why I'd not do it: in theory this all seems good, but let's be very honest: our CI is not 100% stable at this point - here and there we have issues. And that in my opinion is not good. So things that can be potentially unstable and aren't 100% necessary I'd just remove. We publish the packages before and of that doesn't return `ok` our pipeline reports that - so if pushing the package fails, we'll know that. I personally would not try to go further (in this case also trying to install the packages) in the same pipeline because I'm just worried that what we'll end up is just more failed builds.

> My tendency would be to do so and wait for the package to become available.

It seems I'm outvoted at this point, which is totally fine :) Happy to continue with this path if that's the team decision.

So my interpretation is that @v1v and @z1c0 is for implementing a wait step here, correct?

In that case so far my idea is something like this (this is a draft, feel free to add/change things):

- We start with waiting 5 minutes before this step
- We run `dotnet add package Elastic.Apm.NetCoreAll -v [Version]` as we do already
- We check for the output, if it fails, AND it prints `Unable to find package Elastic.Apm.NetCoreAll with version (>= [Version])` then we wait (let's say) 1 minute and then retry. 
- We retry (let's say) 25 times (roughly 30 minutes max. wait time until it fails)

What are your thoughts on that?

z1c0:
> ...
> What are your thoughts on that?

I'm definitely fine with your suggestion.

Regarding the concerns you phrased: If we discover that we made a bad situation worse, we can always go back and remove this step like you initially suggested here. If however, this approach works as we hope, even better 😄 .

gregkalapos:
> > ...
> > What are your thoughts on that?
> 
> I'm definitely fine with your suggestion.
> 
> Regarding the concerns you phrased: If we discover that we made a bad situation worse, we can always go back and remove this step like you initially suggested here. If however, this approach works as we hope, even better 😄 .

Agreed - I'll update this PR accordingly.

AlexanderWert:
Closing --> will be solved for GitHub actions

Issue #1642, opened 2022-03-04
Azure Functions Support


CoverID:
Right now i'm using Azure Application Insight and need another alternative for monitoring.
Is there any information regarding future developments on azure functions support?

TODO:

- [ ] Update/Add specification:
  * https://github.com/elastic/apm/pull/716
  * https://github.com/elastic/apm/pull/712).
- [x] Support tracing `dotnet-isolated` (out-of-process) functions with a custom middleware.
- [ ] Support `dotnet` (in-process) runtime (or decide not to).
- [x] Prepare CI to support tests using the **"Azure Functions Core Tools"** (aka `func`tool).
- [ ] Create sample applications
- [ ] Create unit tests based on the sample applications and `func` tool.
- [ ] Make a decision whether we need to support "flushing" or we rely on "continuous export".
  * Update spec accordingly
  * Find out what "continuous export" could mean in terms of billing or added cost.

MichaelKoch:
would love to have the feature and use apm agent in dotnet azure functions

marzoukali:
Any updates here??

z1c0:
I opened [this issue ](https://github.com/Azure/Azure-Functions/issues/2317) in the Azure Function repository, asking about process runtime lifetimes, suspension, billing ...

lincolncoe:
We are excited to use this! Is there an estimate for when it will be released?

z1c0:
Hi @lincolncoe,
while there is no final date for the upcoming agent release, you can expect it to happen soon.

PR #1877, opened 2022-10-14
Enable CloudMetadataProvider on Azure Functions

z1c0:
This is the first PR for the Azure Functions support feature (#1642).

I'm adding this as a separate PR because it adds value already when using the `Tracer` API in an Azure Functions environment as the agent picks up cloud metadata correctly now.

* Make sure that setting cloud metadata also works in Azure Functions. There are subtle differences to the existing `AzureAppServiceMetadataProvider`, which is why a dedicated provider class was implemented for that.
* Existing code could be reused and was refactored into the newly introduced base class `EnvironmentBasedAzureMetadataProvider`.

### DIFF
--- a/src/Elastic.Apm/Cloud/AzureAppServiceMetadataProvider.cs
+++ b/src/Elastic.Apm/Cloud/AzureAppServiceMetadataProvider.cs
@@ -13,11 +13,12 @@ namespace Elastic.Apm.Cloud
 	/// <summary>
 	/// Provides cloud metadata for Microsoft Azure App Services
 	/// </summary>
-	internal class AzureAppServiceMetadataProvider : EnvironmentBasedAzureMetadataProvider
+	public class AzureAppServiceMetadataProvider : EnvironmentBasedAzureMetadataProvider
 	{
 		internal const string Name = "azure-app-service";
 
-		public AzureAppServiceMetadataProvider(IApmLogger logger, IDictionary environmentVariables) : base(Name, logger,
+		internal AzureAppServiceMetadataProvider(IApmLogger logger, IDictionary environmentVariables) : base(Name,
+			logger,
 			environmentVariables)
 		{
 		}
@@ -41,11 +42,11 @@ namespace Elastic.Apm.Cloud
 
 			return Task.FromResult(new Api.Cloud
 			{
-				Account = new CloudAccount { Id = tokens.Value.subscriptionId },
+				Account = new CloudAccount { Id = tokens.Value.SubscriptionId },
 				Instance = new CloudInstance { Id = websiteInstanceId, Name = websiteSiteName },
 				Project = new CloudProject { Name = websiteResourceGroup },
 				Provider = "azure",
-				Region = tokens.Value.region
+				Region = tokens.Value.Region
 			});
 		}
 	}
--- a/src/Elastic.Apm/Cloud/AzureAppServiceMetadataProvider.cs
+++ b/src/Elastic.Apm/Cloud/AzureAppServiceMetadataProvider.cs
@@ -3,7 +3,6 @@
 // Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information
 
-using System;
 using System.Collections;
 using System.Threading.Tasks;
 using Elastic.Apm.Api;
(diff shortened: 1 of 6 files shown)


z1c0:
Note the related spec PR: https://github.com/elastic/apm/pull/712

src/Elastic.Apm/Cloud/AzureAppServiceMetadataProvider.cs
```diff
@@ -14,110 +13,40 @@ namespace Elastic.Apm.Cloud
 	/// <summary>
 	/// Provides cloud metadata for Microsoft Azure App Services
 	/// </summary>
-	public class AzureAppServiceMetadataProvider : ICloudMetadataProvider
```
gregkalapos, line 17: Hmm.. I also think this type should be `internal` and I don't know why this was `public` in the first place.

But making it `internal` from `public` is technically a breaking change. I don't think this class was useful for users (as a `public` class) and I don't think they used it. Nevertheless, we already had surprises with such changes.

I looked at the other types in the `Cloud` folder and I see some others which are `public` as well (`CloudMetadataProviderCollection` and `ICloudMetadataProvider`) - from what I see there is no way users can do anything useful with those. I think when we introduced cloud detection, we did not really think about the visibility of these types.

I don't have strong opinion on this specific change, only thing I'd add is:
- If we make this technically breaking change, then probably it'd make sense to think about making those other types `internal` as well.
- Or we just leave it `public`, since actually that's not really the point of this PR - and we may or may not clean up this later. With this we are definitely on the safe side.

gregkalapos (commented):
Overall looks good to me 👍 

I added 2 points to discuss - I suggest we make a decision on those before we merge this.

src/Elastic.Apm/Cloud/AzureAppServiceMetadataProvider.cs
z1c0: While I also think that the types in `Elastic.Apm.Cloud` should be `internal` rather than `public`, I agree that this is not the right PR to make this change. I will revert `AzureAppServiceMetadataProvider` to `public` visibility.

I will leave the newly introduced type `AzureFunctionsMetadataProvider` `internal` though, if that's ok with you.

I'll create an issue for that with the `breaking change` label. Then we can make these changes at a later point if we decide to do so.

src/Elastic.Apm/Cloud/EnvironmentBasedAzureMetadataProvider.cs
```diff
@@ -0,0 +1,110 @@
...
+			_logger.Trace()?.Log(
+				"Unable to get {Provider} cloud metadata as no {EnvironmentVariable} environment variable exists. The application is likely not running in {Provider}",
+				Provider,
+				key,
+				Provider);
+
+			return true;
+		}
+
+		protected (string subscriptionId, string resourceGroup, string region)? TokenizeWebSiteOwnerName(
```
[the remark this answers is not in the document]
z1c0: Honestly, I did not think about the problems with tuples on Full Framework and used them for their sheer convenience.
That being said, I'll simply introduce a custom type here and I need to remind myself in the future to abstain from using tuples.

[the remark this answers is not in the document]
gregkalapos: > Honestly, I did not think about the problems with tuples on Full Framework and used them for their sheer convenience.

It was the same for me ...until people started reporting the linked problem :)

src/Elastic.Apm/Cloud/AzureAppServiceMetadataProvider.cs
gregkalapos: > Update: While changing back the visibility, I remembered the reason for this change. Leaving `AzureAppServiceMetadataProvider` `public` requires me to make the newly introduced base class `EnvironmentBasedAzureMetadataProvider` `public` as well, which I wanted to avoid. I suppose, I have to create a bit more legacy here.

Oh, I see. Yeah I think the current status of the PR is fine - with keeping the `AzureAppServiceMetadataProvider` public and accordingly making the new base public public as well is the safest way to proceed (definitely not breaking users). https://github.com/elastic/apm-agent-dotnet/issues/1890 is useful, let's handle it there...

gregkalapos (approved):
Thanks for the updates @z1c0.

Issue #1881, opened 2022-10-19
[BUG] Cannot correlate 2 distributed OpenTelemetry activity series into a same APM trace


kuanpak:
### APM Agent version
The version of the `Elastic.Apm` nuget package used: 1.18

### Environment

**Operating system and version: Windows 11

**.NET Framework/Core name and version** (e.g. .NET 4.6.1, NET Core 3.1.100) : .NET 6

**Application Target Framework(s)** (e.g. net461, netcoreapp3.1): net6.0

### Describe the bug
Cannot correlate 2 distributed OpenTelemetry activity series into a same APM trace when the first transaction was created via ElasticActivityListenser.

### To Reproduce

1. Use ElasticApm:EnableOpenTelemetryBridge = true
2. Call ActivitySource.StartActivity() to start Activity "s1" and also child activity "s2"
3. Extract "traceparent" value from "s2" and then stop "s2" and "s1" activities
4. Call ActivitySource.StartActivity() with TraceContext from s2 traceparent to start activity "s3"
5. "s3" will be a separated trace shown in Kibana. 

#### Expected behavior
Expect we can see the full trace like "s1" -> "s2" -> "s3" in kibana

#### Actual behavior
2 distributed traces could not be correlated

#### Root cause assumption
Tracer.StartTransactionInternal()'s `traceId` parameter is not passed into Transaction internal constructor, so that the first transaction's TraceId will be different from Activity.TraceId (`s1`), but `s3` parent id is referring to `s1` trace id.
https://github.com/elastic/apm-agent-dotnet/blob/9067636b28bffab9c5aa4d2d1592b1ee26e0a755/src/Elastic.Apm/Api/Tracer.cs#L81-L83

z1c0:
Hi @kuanpak,
thanks for reporting this and even more so for coming up with a fix PR as well!

gregkalapos:
Fixed in https://github.com/elastic/apm-agent-dotnet/pull/1882 by @kuanpak. Thanks for the contribution!

PR #1882, opened 2022-10-22
Fix transaction trace id not aligned when transaction is created from OTel bridge without parent

kuanpak:
The is the PR for fixing OTel bridged transactions correlation bug #1881 

Just pass Tracer.StartTransactionInternal()'s traceId parameter into Transaction internal constructor in order to fix the issue, so that the elastic transaction created from OTel bridge will use the same traceId as the OTel span (Activity).

Related Unit tests are also added to verify the results.

### DIFF
--- a/test/Elastic.Apm.Feature.Tests/Features/otel_bridge.feature
+++ b/test/Elastic.Apm.Feature.Tests/Features/otel_bridge.feature
@@ -17,6 +17,7 @@ Feature: OpenTelemetry bridge
     Then Elastic bridged transaction is a root transaction
     # outcome should not be inferred from the lack/presence of errors
     Then Elastic bridged transaction outcome is "unknown"
+	Then Elastic bridged transaction trace id is equal to OTel trace id
 
   Scenario: Create span from OTel span
     Given an agent
--- a/test/Elastic.Apm.Feature.Tests/Features/otel_bridge.feature
+++ b/test/Elastic.Apm.Feature.Tests/Features/otel_bridge.feature
@@ -17,7 +17,6 @@ Feature: OpenTelemetry bridge
     Then Elastic bridged transaction is a root transaction
     # outcome should not be inferred from the lack/presence of errors
     Then Elastic bridged transaction outcome is "unknown"
-	Then Elastic bridged transaction trace id is equal to OTel trace id
 
   Scenario: Create span from OTel span
     Given an agent
--- a/test/Elastic.Apm.Feature.Tests/OpenTelemetryBridgeStepDefinitions.cs
+++ b/test/Elastic.Apm.Feature.Tests/OpenTelemetryBridgeStepDefinitions.cs
@@ -68,7 +68,8 @@ namespace Elastic.Apm.Feature.Tests
 		public void GivenOTelSpanIsCreatedWithoutParent()
 		{
 			var src = new ActivitySource("Test");
-			src.StartActivity("foo");
+			var activity = src.StartActivity("foo");
+			_scenarioContext.Add("traceId", activity.TraceId);
 		}
 
 		[Given(@"OTel span ends")]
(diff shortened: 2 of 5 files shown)


test/Elastic.Apm.Feature.Tests/Features/otel_bridge.feature
```diff
@@ -17,6 +17,7 @@ Feature: OpenTelemetry bridge
     Then Elastic bridged transaction is a root transaction
     # outcome should not be inferred from the lack/presence of errors
     Then Elastic bridged transaction outcome is "unknown"
+	Then Elastic bridged transaction trace id is equal to OTel trace id
```
gregkalapos, line 20: These files are synced from https://github.com/elastic/apm/tree/main/tests/agents/gherkin-specs. We sync these files into every elastic APM Agent repo. So these files are managed in the linked repo and not here.

This means, if we change this, first we'd need to open a PR over in that repo and then it'll be synced into this one.

I just realize if you only look at this repo, you'd have no chance to know this - I'll open a PR and add a readme file to `test/Elastic.Apm.Feature.Tests`. This is totally on us and I'm sorry about this.

On the other hand, I don't think this level of detail needs to be specified. 

So, in order to proceed, I'd suggest to remove the change from this file ([otel_bridge.feature](https://github.com/elastic/apm-agent-dotnet/pull/1882/files#diff-ea36b655c118bd668317b2016c43a4e1b75b4367ba5b84ba9e99d25048b07b1f)) and from [test/Elastic.Apm.Feature.Tests/OpenTelemetryBridgeStepDefinitions.cs](https://github.com/elastic/apm-agent-dotnet/pull/1882/files#diff-4368859739d912974d2225c73bb3bd9c1f06b574ed959fde10244b4d2d944fd3) and proceed with this PR without the gherkin changes.

test/Elastic.Apm.Feature.Tests/OpenTelemetryBridgeStepDefinitions.cs
```diff
@@ -89,6 +90,15 @@ public void ThenElasticBridgedTransactionOutcomeIs(string outcome)
 			payloadSender.FirstTransaction.Outcome.ToString().ToLower().Should().Be(outcome);
 		}
 
+		[Then(@"Elastic bridged transaction trace id is equal to OTel trace id")]
```
gregkalapos, line 93: As mentioned above, I suggest to remove these.

gregkalapos (changes requested):
Thank you for opening this @kuanpak - super useful.

The changes look good to me, very nice PR.

The only blocker I see is the change in `oteg_bridge.feature` - we haven't communicated clearly how those work, and those changes need to be reverted (more on that in a comment below). Could you please do that?

Regarding the added test: After looking into this myself, I guess the core issue is that we did not use the `traceid` in case of a root activity was started. The test you added covers this very well, and I'd definitely keep that test. But additionally I'd also add a smaller test which tests the core issue, something like this (feel free to just add this to the PR):

```
[Fact]
public void ActivityAndTransactionTraceIdSynced()
{
	var payloadSender = new MockPayloadSender();
	using var agent = new ApmAgent(new TestAgentComponents(payloadSender: payloadSender, apmServerInfo: MockApmServerInfo.Version716,
		configuration: new MockConfiguration(enableOpenTelemetryBridge: "true")));
	var src = new ActivitySource("Test");
	string traceId;
	using (var activity = src.StartActivity("foo", ActivityKind.Server)) traceId = activity.TraceId.ToString();
	payloadSender.FirstTransaction.TraceId.Should().Be(traceId);
}
```

Again, sorry about the inconvenience regarding the gherkin tests - and thanks for the PR!

kuanpak:
Thanks @gregkalapos for reviewing this PR and providing the nice comments. I've reverted `oteg_bridge.feature` changes and aligned with your team's coding style.

z1c0 (commented):
Thanks for the PR. LGTM!
I have a small ask about moving the test logic.
Also, @gregkalapos suggested adding an additional test in [this comment](https://github.com/elastic/apm-agent-dotnet/pull/1882#pullrequestreview-1153476487). Do you think you could add that as well? Very much appreciated!

z1c0 (approved):

gregkalapos:
CI green, I hit merge on this one.

The [small test](https://github.com/elastic/apm-agent-dotnet/pull/1882#pullrequestreview-1153476487) isn't in this PR, but that's fine - I'll quickly open a small follow up PR.

Thanks again @kuanpak! 🎉

PR #1887, opened 2022-10-25
Add test which asserts the traceId is the same on a root activity and the corresponding transaction

gregkalapos:
Follow up from https://github.com/elastic/apm-agent-dotnet/pull/1882#pullrequestreview-1152603638. 

Add small test to make sure the traceId is synced across a root activity and the transaction created based on that root activity.

 cc @kuanpak


z1c0 (approved):
Perfect, thanks for taking care of this additional test case @gregkalapos!

PR #1883, opened 2022-10-24
Add readme file to  test/Elastic.Apm.Feature.Tests

gregkalapos:
After looking at https://github.com/elastic/apm-agent-dotnet/pull/1882 I realize we don't communicate very well how the `.feature` files are synced from the `apm` repo.

To fix this, this PR adds a short readme explaining how those files are synced.

PR #1888, opened 2022-10-27
Add README regarding json-spec files

z1c0:
Similar to https://github.com/elastic/apm-agent-dotnet/pull/1883, I added a `README.md` explaining soure and usage of the files in `json-spec`.

### DIFF
--- a/test/Elastic.Apm.Tests.Utilities/README.md
+++ b/test/Elastic.Apm.Tests.Utilities/README.md
@@ -1,0 +1,15 @@
+# Elastic.Apm.Tests.Utilities
+
+This project contains test specifications in JSON format that are used as test inputs in
+different projects (e.g. `Elastic.Apm.Tests`).
+
+Those specifications can be found in the `/TestResources/json-specs` folder and are synchronized
+from the [`github.com/elastic/apm`](https://github.com/elastic/apm) repository from the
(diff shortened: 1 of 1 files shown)


test/Elastic.Apm.Tests.Utilities/README.md
```diff
@@ -0,0 +1,15 @@
+# Elastic.Apm.Tests.Utilities
+
+This project contains test specifications in JSON format that are used as test inputs in
```
gregkalapos, line 3: Since this is in the root of the `Elastic.Apm.Tests.Utilities` could we maybe add 1 general sentence before we go into the JSON specs here?

Thinking behind this is that if people open the folder and look at this readme they may get the impression this project is primarily about the JSON files.

So I'd start with something like this:

```
This project contains reusable utilities (e.g. in-memory or mock implementation of agent components,  test inputs, etc.) which are used in other test projects.
```

gregkalapos (approved):
Nice! 👍 

I added one small suggestion below.

test/Elastic.Apm.Tests.Utilities/README.md
z1c0: Yes, that makes sense. I'll update the PR.

PR #1891, opened 2022-11-02
Add Middleware for tracing Azure Functions (dotnet-isolated).

z1c0:
This is not the final/complete PR for Azure Functions support but - I think - a legit and self-contained unit of code for review.

What's included in this PR:

* Support for tracing `dotnet-isolated` (out-of-process) functions via a custom middleware.
* A sample Azure Functions application (also used for testing).
* CI test support on Windows using "Azure Functions Core Tools" (`func`).
  * *Note*: Tests running on Linux are deliberately skipped at the moment.
* An exemplary unit test (further tests to be added later, as soon as the spec is approved and stable).

What's still to do:

* Implement the missing parts of the [spec](https://github.com/elastic/apm/pull/716) as soon as it's approved and ready.
* ~~Prepare Linux CI to support tests using "Azure Functions Core Tools" (`func`).~~
* Improve/add unit tests
* Support `dotnet` (in-process) runtime (or decide not to).

(I added those points to the issue #1642 as well)


z1c0:
Update: The `func` tool is now also available on the Linux CI and the unit test is running successfully.

src/Elastic.Apm.Azure.Functions/ElasticApmMiddleware.cs
```diff
@@ -0,0 +1,126 @@
...
+	private static HttpRequestData? GetHttpRequestData(FunctionContext functionContext)
+	{
+		var inputData = GetProperty<IReadOnlyDictionary<string, object>>(functionContext, "InputData");
+		return inputData?.Values.SingleOrDefault(o => o is HttpRequestData) as HttpRequestData;
+	}
+
+	private static T? GetProperty<T>(FunctionContext functionContext, string name) where T : class
+	{
+		var feature = functionContext.Features.SingleOrDefault(f => f.Key.Name == "IFunctionBindingsFeature").Value;
+		return feature?.GetType().GetProperties().SingleOrDefault(p => p.Name == name)?.GetValue(feature) as T;
```
gregkalapos, line 108: Rider suggests me that you can remove conditional access on `feature?.`:

image

Which I don't really get.. `feature` is `object?`, so it can be null.
Although `.Value` in the previous line returns `TValue` and not `TValue?`, so I don't know why Rider thinks it's `object?` and not `object`.

Anyways, I'm confused by Rider, and I think the code is ok. I just wanted to let you know.

src/Elastic.Apm.Azure.Functions/ElasticApmMiddleware.cs
```diff
@@ -0,0 +1,126 @@
...
+using Elastic.Apm.Logging;
+using Elastic.Apm.Model;
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Azure.Functions.Worker.Http;
+using Microsoft.Azure.Functions.Worker.Middleware;
+using TraceContext = Elastic.Apm.DistributedTracing.TraceContext;
+
+namespace Elastic.Apm.Azure.Functions;
+
+public class ElasticApmMiddleware : IFunctionsWorkerMiddleware
```
gregkalapos, line 23: Maybe call `ApmMiddleware`, because we already have `Elastic` in the namespace? We do the same for the `Elastic.Apm.AspNetCore.Middleware`. 

```suggestion
public class ApmMiddleware : IFunctionsWorkerMiddleware
```

src/Elastic.Apm.Azure.Functions/Elastic.Apm.Azure.Functions.csproj
```diff
@@ -0,0 +1,23 @@
...
+    <Nullable>enable</Nullable>
+
+    <AssemblyName>Elastic.Apm.Azure.Functions</AssemblyName>
+    <RootNamespace>Elastic.Apm.Azure.Functions</RootNamespace>
+    <PackageId>Elastic.Apm.Azure.Functions</PackageId>
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Elastic.Apm\Elastic.Apm.csproj"/>
```
gregkalapos, line 14: I think we should take dependency on more packages and turn on other listeners - the other Azure packages (CosmosDb, ServiceBus, Storage) are especially interesting here, since those are more likely to be used in an Azure function.

src/Elastic.Apm.Azure.Functions/ElasticApmMiddleware.cs
```diff
@@ -0,0 +1,126 @@
...
+	public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next)
+	{
+		try
+		{
+			Logger.Trace()?.Log($"{nameof(Invoke)} - {context.FunctionDefinition.Name}");
+
+			var data = GetTriggerSpecificData(context);
+			await Agent.Tracer.CaptureTransaction(data.Name, ApiConstants.TypeRequest, async t =>
+			{
+				if (t is Transaction transaction) // TODO: Expose FaaS property on ITransaction?
```
gregkalapos, line 50: > // TODO: Expose FaaS property on ITransaction?

Yeah, I'd do it in this PR. The `Faas` type is already public, so let's properly expose it - it's better for us (no need to cast here) and better for users (we don't expose public types that they can't access on our interfaces and use e.g. in filters).

src/Elastic.Apm.Azure.Functions/ElasticApmMiddleware.cs
```diff
@@ -0,0 +1,126 @@
...
+						Id = $"{FaasIdPrefix}{context.FunctionDefinition.Name}",
+						Trigger = new Trigger { Type = data.TriggerType },
+						Execution = context.InvocationId,
+						ColdStart = IsColdStart()
+					};
+				}
+
+				await next(context);
+
+				t.Outcome = Outcome.Success;
```
gregkalapos, line 64: This seems to be always `Success`, which doesn't look correct.

We already have logic which sets `Transaction.Outcome` based on HTTP response [here](https://github.com/elastic/apm-agent-dotnet/blob/main/src/Elastic.Apm.AspNetCore/WebRequestTransactionCreator.cs#L239). That's currently in `Elastic.Apm.AspNetCore`, but we could move that method into a more central place. For the next line I added a comment suggesting to use [this](https://github.com/elastic/apm-agent-dotnet/blob/main/src/Elastic.Apm/Model/Transaction.cs#LL793), we could just put the `SetOutcomeForHttpResult` method from `WebRequestTransactionCreator.cs` next to that method.

src/Elastic.Apm.Azure.Functions/ElasticApmMiddleware.cs
```diff
@@ -0,0 +1,126 @@
...
+				t.Result = GetTriggerSpecificResult(context);
```
gregkalapos, line 65: We have something similar [here](https://github.com/elastic/apm-agent-dotnet/blob/main/src/Elastic.Apm/Model/Transaction.cs#LL793) - we could maybe just call that.

test/Elastic.Apm.Azure.Functions.Tests/AzureFunctionsTests.cs
```diff
@@ -0,0 +1,117 @@
...
+using System.Threading.Tasks;
+using Elastic.Apm.Api;
+using Elastic.Apm.Logging;
+using Elastic.Apm.Tests.MockApmServer;
+using Elastic.Apm.Tests.Utilities;
+using FluentAssertions;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Elastic.Apm.Azure.Storage.Tests;
```
gregkalapos, line 20:
```suggestion
namespace Elastic.Apm.Azure.Functions.Tests;

```

src/Elastic.Apm.Azure.Functions/ElasticApmMiddleware.cs
```diff
@@ -0,0 +1,126 @@
...
+	private static bool IsColdStart() => Interlocked.Exchange(ref ColdStart, 0) == 1;
+
+	public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next)
+	{
+		try
+		{
+			Logger.Trace()?.Log($"{nameof(Invoke)} - {context.FunctionDefinition.Name}");
+
+			var data = GetTriggerSpecificData(context);
+			await Agent.Tracer.CaptureTransaction(data.Name, ApiConstants.TypeRequest, async t =>
```
gregkalapos, line 48: Did you test what happens when `await next(context);`  throws?

If `await next(context);` within this block throws an exceptions, then the `CaptureTransaction` will deal with it (end the transaction properly, capture the exception, etc..), but we have 2 lines after `await next(context);` which set `t.Outcome` and `t.Result` - those will not run if `await next(context);` throws.

I haven't tested this and I don't know if those 2 lines would otherwise do something relevant in that case (I'd expect to see HTTP500 on the `context`, but maybe that's not the case).

If this is really an issue, we could just do try-catch-finally here and start the transaction by calling `Agent.Tracer.StartTransaction` and manually end it, plus run the necessary lines to set `Outcome` and `Result` in `finally`.

gregkalapos (commented):
Looks very nice! I had some comments and thoughts for a few things.

Jenkinsfile
```diff
@@ -618,6 +620,37 @@ def dotnet(Closure body){
...
+    unzip -d azure-functions-cli "\${latest_v4_release_file}"
+
+    # Make required executables ... executable.
+    chmod +x ./azure-functions-cli/func
+    chmod +x ./azure-functions-cli/gozip
+    """)
+    body()
+  }
+}
+
```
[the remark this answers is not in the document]
z1c0: I considered both approaches and decided on the `Jenkinsfile` "inline version" because:
* It follows the example of the `dotnet` block we already use here:
* The `PATH` modification needs to happen in `withEnv` either way and by having everything in this block it's more discoverable what actually happens.
* The code can be pulled out almost verbatim into a script should we ever have the need.
Let me know what you think.

sample/Elastic.AzureFunctionApp.Isolated/host.json
```diff
@@ -0,0 +1,11 @@
+{
+    "version": "2.0",
+    "logging": {
+        "applicationInsights": {
+            "samplingSettings": {
+                "isEnabled": true,
```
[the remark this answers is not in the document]
z1c0: oops

sample/Elastic.AzureFunctionApp.Isolated/Elastic.AzureFunctionApp.Isolated.csproj
```diff
@@ -0,0 +1,31 @@
+<Project Sdk="Microsoft.NET.Sdk">
+    <PropertyGroup>
+        <TargetFramework>net6.0</TargetFramework>
+        <AzureFunctionsVersion>V4</AzureFunctionsVersion>
+        <OutputType>Exe</OutputType>
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
```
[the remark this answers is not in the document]
z1c0: Yes, that was unintentional.

src/Elastic.Apm.Azure.Functions/Elastic.Apm.Azure.Functions.csproj
```diff
@@ -0,0 +1,23 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
+    <Nullable>enable</Nullable>
+
+    <AssemblyName>Elastic.Apm.Azure.Functions</AssemblyName>
+    <RootNamespace>Elastic.Apm.Azure.Functions</RootNamespace>
+    <PackageId>Elastic.Apm.Azure.Functions</PackageId>
+    <IsPackable>false</IsPackable>
```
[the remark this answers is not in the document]
z1c0: Thanks for the reminder! I saw that, but don't think I would have remembered it tbh.
Is that the (your) usual approach, however:
a) Set `IsPackable` to `true` and add the project to the script in one step?
b) Or do you set `IsPackable` from the beginning?
c) Something else? ;-)

src/Elastic.Apm.Azure.Functions/ElasticApmMiddleware.cs
z1c0: Hmm, this seems like a Rider issue to me.
The call to `functionContext.Features.SingleOrDefault` returns a `KeyValuePair<Type, object>` which is a `struct`, hence never `null`. The `Value` of this `KeyValuePair` is of type `object` though (the `feature`) variable, so the `check` is required.

test/Elastic.Apm.Azure.Functions.Tests/AzureFunctionsTests.cs
z1c0: Thanks for this catch!

src/Elastic.Apm.Azure.Functions/ElasticApmMiddleware.cs
z1c0: The idea was to also support other trigger types (non-HTTP) here over time, but thanks for the hint. I will reuse the existing code for HTTP.

src/Elastic.Apm.Azure.Functions/Elastic.Apm.Azure.Functions.csproj
z1c0: Good point. However, let's do this in another iteration as you suggested 👍 .

src/Elastic.Apm.Azure.Functions/ElasticApmMiddleware.cs
z1c0: Sure 👍

src/Elastic.Apm.Azure.Functions/ElasticApmMiddleware.cs
z1c0: Right, this part was definitely in an unfinished state. I have should have mentioned in the PR description that I planned to improve on that once I had more test cases.
However, it was not a good idea to leave it like this in the first place.
I tried to clean this code up so that it makes more sense, but it will probably still need refinement once we have more test scenarios and the [spec](https://github.com/elastic/apm/pull/716) is stable.

src/Elastic.Apm.Azure.Functions/ElasticApmMiddleware.cs
z1c0: I improved and refactored that code. PTAL
(see also my [comment](https://github.com/elastic/apm-agent-dotnet/pull/1891#discussion_r1026156310) above)

src/Elastic.Apm.Azure.Functions/Elastic.Apm.Azure.Functions.csproj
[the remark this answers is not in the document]
gregkalapos: I usually did option `a` before.

Typically when the package is in development similarly to what you did here, I set `IsPackable` to `false` (just to be 100% sure we don't accidentally publish it, even if it's not in the script - this is also good to communicate that it's not ready yet for being published).

Once the package is ready,  I usually flip `IsPackable` to `true` and add it to the script as one step.

src/Elastic.Apm.Azure.Functions/ElasticApmMiddleware.cs
gregkalapos: > Hmm, this seems like a Rider issue to me.

Yeah, I also think that.

src/Elastic.Apm.Azure.Functions/ElasticApmMiddleware.cs
gregkalapos: > The idea was to have this dedicated method so we can support other trigger types (non-HTTP) in there over time. 

That's a good point, and honestly during my review I was very much focused on HTTP. The `SetTriggerSpecificResult` you added handled this nicely.

gregkalapos (approved):
Thanks for the updates.

I think this is ready to merge 🚀

PR #1897, opened 2022-11-04
Add basic agent logging preamble

z1c0:
This follows the spec for a logging preambel in this [draft PR](https://github.com/elastic/apm/pull/713/).

### DIFF
--- a/src/Elastic.Apm/AgentComponents.cs
+++ b/src/Elastic.Apm/AgentComponents.cs
@@ -188,7 +188,7 @@ namespace Elastic.Apm
 			CentralConfigurationFetcher?.Dispose();
 		}
 
-		private static void PrintAgentLogPreambel(IApmLogger logger)
+		private static void PrintAgentLogPreamble(IApmLogger logger)
 		{
 			if (logger?.Info() != null)
 			{
--- a/src/Elastic.Apm/AgentComponents.cs
+++ b/src/Elastic.Apm/AgentComponents.cs
@@ -48,7 +48,7 @@ namespace Elastic.Apm
 				var tempLogger = logger ?? ConsoleLogger.LoggerOrDefault(configurationReader?.LogLevel);
 				ConfigurationReader = configurationReader ?? new EnvironmentConfigurationReader(tempLogger);
 				Logger = logger ?? ConsoleLogger.LoggerOrDefault(ConfigurationReader.LogLevel);
-				PrintAgentLogPreambel(Logger);
+				PrintAgentLogPreamble(Logger);
 				Service = Service.GetDefaultService(ConfigurationReader, Logger);
 
 				var systemInfoHelper = new SystemInfoHelper(Logger);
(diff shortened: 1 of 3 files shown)


src/Elastic.Apm/AgentComponents.cs
```diff
@@ -182,5 +187,24 @@ public void Dispose()
...
 		}
+
+		private static void PrintAgentLogPreambel(IApmLogger logger)
+		{
+			if (logger?.Info() != null)
+			{
+				var info = logger.Info().Value;
+				info.Log("********************************************************************************");
+				info.Log(
+					$"Elastic APM .NET Agent, version: {Assembly.GetExecutingAssembly().GetName().Version}, build date: {File.GetCreationTime(Assembly.GetExecutingAssembly().Location).ToUniversalTime()} UTC");
```
z1c0, line 198: The solution for the "build date" is not yet ideal, but a good first approximation.
A better mid- to long-term solution could be to introduce a custom assembly attribute that gets actually set a build time.

src/Elastic.Apm/AgentComponents.cs
```diff
@@ -182,5 +187,24 @@ public void Dispose()
 
 			CentralConfigurationFetcher?.Dispose();
 		}
+
+		private static void PrintAgentLogPreambel(IApmLogger logger)
```
gregkalapos, line 191: nit: small typo I think.

```suggestion
		private static void PrintAgentLogPreamble(IApmLogger logger)
```

src/Elastic.Apm/AgentComponents.cs
gregkalapos: 👍  I also think this is a good first start. Just one thing maybe: I think it'd be better to print "agent file creation time" (or something along those lines) to make sure there is no misunderstanding by this log line.

gregkalapos (commented):
I see some of the properties can throw exceptions which we access.  According to the doc, all are `InvalidOperationException`s. 

Here are the ones:
- `Process.GetCurrentProcess().Id` (the `Id` property)
- `Process.GetCurrentProcess().ProcessName` (again the `ProcessName` property)
- `Environment.MachineName`

Super unlikely cases, but I'd feel better by having a try-catch on those.

src/Elastic.Apm/AgentComponents.cs
z1c0: Agreed, then this will be unambiguous 👍

z1c0:
> I see some of the properties can throw exceptions which we access. According to the doc, all are `InvalidOperationException`s.
> 
> Here are the ones:
> 
> * `Process.GetCurrentProcess().Id` (the `Id` property)
> * `Process.GetCurrentProcess().ProcessName` (again the `ProcessName` property)
> * `Environment.MachineName`
> 
> Super unlikely cases, but I'd feel better by having a try-catch on those.

Good catch (pun intended)!
I'll wrap the whole preamble block in a `try/catch` statement - that way, it's less likely that we (I 😉) miss exceptions that might be thrown by log statements added in the future.

gregkalapos (approved):
Thanks!

PR #1898, opened 2022-11-06
Debug Agent: Loosen the constraints on calling Agent.Setup()

gregkalapos:
The main purpose of this PR is to create an agent for debugging.

We see issues with `Agent.Setup()` being called multiple times in specific environments. In order to debug this more, this PR flips the behaviour in `Agent.Setup()`: Instead of rejecting the new AgentComponents instance, the agent will accept and use the new instance and abandon the old one.

Essentially with this, the agent can be initialized multiple times and always the last setup wins.

With this we may have new insights on the specific issues we see - depending on the outcome we may investigate if this behaviour is better than the old one.

Although, that'd open up a bunch of other issues, e.g.: Multiple payloadsenders can be initialized with multiple server urls, if multiple tracers are present, they may capture things multiple times, multiple MetricsCollector will also likely run in parallel. None of these are addressed here and if we keep the behaviour in this PR we'll need to investigate and address all those potential issues.

### DIFF
--- a/src/Elastic.Apm/Agent.cs
+++ b/src/Elastic.Apm/Agent.cs
@@ -170,25 +170,15 @@ namespace Elastic.Apm
 		{
 			lock (InitializationLock)
 			{
-				if (LazyApmAgent.IsValueCreated)
-				{
-					Components?.Logger?.Error()
-						?.Log("The singleton APM agent has" +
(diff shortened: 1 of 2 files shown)


gregkalapos:
Not relevant anymore and we are not following up on this. Closing now.

Issue #1665, opened 2022-03-24
Kubernetes attachment


felixbarny:
- [x] Publish a docker image that contains the agent distribution, similar to what the Java agent is already doing (see [build](https://github.com/elastic/apm-agent-java/blob/3187b4fbbf349041f95080e49b194848b1c33098/.ci/release/Jenkinsfile#L245-L257) and the [observability/apm-agent-java](https://container-library.elastic.co/r/observability/apm-agent-java) image in our Docker registry)
- [x] Integrate with the webhook in APM Server (https://github.com/elastic/apm-server/issues/7386) to copy the agent from the image to a volume via an init container and add environment variables to attach the agent.

gregkalapos:
Done - this was implemented in multiple PRs - all linked above.

PR #1907, opened 2022-11-16
Publish docker image with agent

gregkalapos:
Next step for: https://github.com/elastic/apm-agent-dotnet/issues/1665 

Open question:
- [x] How do we publish the image first? I think the step to publish the image should be in the `Release` stage (that's how it is in the PR right now), but that runs only when we release next (it won't run on a normal PR or a merge to `main`). 
Currently my best idea: once we are fine with the PR, I'll temporarily move the new `Publish Docker image` stage outside `Release` (likely right after `Create Docker image`) and let it run once, so we publish the image. If someone has a better idea, I'm happy to explore. 
Update: I opened https://github.com/elastic/apm-agent-dotnet/pull/1915 - that PR is branched away from the last release (`1.18.0`) and I ran the `push_docker` step which successfully pushed the docker image.


Jenkinsfile
```diff
@@ -546,6 +546,15 @@ pipeline {
             }
           }
         }
+        stage('Publish Docker image') {
```
gregkalapos, line 549: Question for Jenkins experts:

This is in the `Release` stage, because we only want to publish the image, when we do a release.

But the building of the image happens earlier in the `Create Docker image` stage which run after our tests on each build - but that's outside of the `Release` stage. The idea here is that we always build the image (to make sure we don't break something, but only publish when we release).

Question: will the created docker image, which we create in the `Create Docker image` stage visible here?

v1v: That's correct, all the stages are using the top-level agent one defined after the `pipeline {`, unless they are override with `agent { label `, and I see it's not the case unless running for Windows or the Benchmarks.

So, it's visible

Jenkinsfile
```diff
@@ -546,6 +546,15 @@ pipeline {
...
+          steps {
+            withGithubNotify(context: 'Create Docker image - Linux') {
+              dir("${BASE_DIR}"){
```
v1v, line 552: There is a need to configure the required secrets to interact with `docker.elastic.co`

```suggestion
              dir("${BASE_DIR}"){
                dockerLogin(secret: env.DOCKER_SECRET, registry: env.DOCKER_REGISTRY)
```

Jenkinsfile
v1v, line 549: Set the env variables with the docker specific details:

```suggestion
        stage('Publish Docker image') {
          environment {
            DOCKER_SECRET = 'secret/apm-team/ci/docker-registry/prod'
            DOCKER_REGISTRY = 'docker.elastic.co'
```

v1v (changes requested):
In order to push to docker.elastic.co is required to configure the docker context, for such, see the suggestions

v1v (approved):

z1c0 (approved):

gregkalapos:
I pushed 2 changes to the PR:
- https://github.com/elastic/apm-agent-dotnet/pull/1907/commits/272f235ad9eaad049a229cf9d651e316bfe15252: Without those 2 new lines, this step failed in https://github.com/elastic/apm-agent-dotnet/pull/1915, so those seem to be needed (keep in mind this step will only run in `Release`, so we had no chance to test this in CI before). 
- https://github.com/elastic/apm-agent-dotnet/pull/1907/commits/8c6b7fb9bd00469f1e9a6a8477ba133e39043644: I noticed the `push_docker.sh` file changed in the Java repo - the core of the change is that it sets both the current tag and `latest` (I remember, we actually discussed we should make sure we set `latest` as well). Accordingly, I updated our file to be in-sync with Java.

With that, I think this is ready. CI runs currently, once it's green I'll merge, unless someone raises something.

PR #1922, opened 2022-11-23
Use Span timing instead of cumulative SqlCommand statistics

### LINKED ISSUE #1869: SqlClientDiagnosticListener - change how we measure duration
[Reported via discuss](https://discuss.elastic.co/t/mssql-span-durations-inaccurate-when-sourced-from-sqlclientdiagnosticlistener/316225)

- Currently we use `Statistics.ExecutionTime` in `SqlClientDiagnosticListener` to get the duration of the given span for a given sql call.
- This value seems to be an aggregated value of multiple statements - I don't fully understand how it's aggregated yet, but as it seems it's not a 1-1 mapping to spans we generate. Maybe this part needs to be researched and understood a little bit better.

Suggestion for the fix:
- Let's just rely on the internal timing of our `Span` type. So instead of getting the duration via the diagnostic source listener, we can just measure it as we do for spans by default.

_If someone picks this up: please also try to add a test for this._

bertvan:
Closes #1869 

* Removed usage of SqlCommand statistics (`ExecutionTime`) for span duration, as that statistic is cumulative (see: https://learn.microsoft.com/en-us/dotnet/framework/data/adonet/sql/provider-statistics-for-sql-server).
* Removal of `duration` parameter on `EndSpan()`, results in the Span timing being used instead.
* Added test to check if span duration with multiple operations is not cumulative.

### DIFF
--- a/src/Elastic.Apm.SqlClient/SqlClientDiagnosticListener.cs
+++ b/src/Elastic.Apm.SqlClient/SqlClientDiagnosticListener.cs
@@ -85,13 +85,7 @@ namespace Elastic.Apm.SqlClient
 				{
 					if (!_spans.TryRemove(operationId, out var span)) return;
 
-					TimeSpan? duration = null;
-
-					if (propertyFetcherSet.Statistics.Fetch(payloadData) is IDictionary<object, object> statistics &&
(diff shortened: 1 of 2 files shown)


z1c0 (approved):
LGTM!
Thanks for the contribution @bertvan

PR #1930, opened 2022-11-29
Improve SOAP action parsing

z1c0:
This PR aims to make the SOAP action parsing logic less error-prone. 

The reason is that we recently noticed side effects with other application/extension-logic trying to parse SOAP body data as well.

Notable changes:
* `GetSoap12ActionFromInputStream` now works directly on `request.InputStream`, copies it to a `MemoryStream` and resets `Stream.Position` to `0`.
  * This seems to be the recommended approach (see e.g. [here](https://stackoverflow.com/questions/1678846/how-to-log-request-inputstream-with-httpmodule-then-reset-inputstream-position)).
  * Before that, we used `request.GetBufferedInputStream()` which could result in failing subsequent reads as the underlying `Stream` was not seekable any more.
* The code has an additional check that the `Stream` is actually seekable.
* More logging was added in error cases.
* `TryExtractSoapAction` was moved to `ProcessEndRequest` to further reduce any side-effect potential with application logic reading the body data.

### DIFF
--- a/src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
+++ b/src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
@@ -146,8 +146,6 @@ namespace Elastic.Apm.AspNetFullFramework
 			}
 
 			var transactionName = $"{request.HttpMethod} {request.Unvalidated.Path}";
-			if (SoapRequest.TryExtractSoapAction(_logger, request, out var soapAction))
-				transactionName += $" {soapAction}";
 
 			var distributedTracingData = ExtractIncomingDistributedTracingData(request);
 			ITransaction transaction;
@@ -292,6 +290,9 @@ namespace Elastic.Apm.AspNetFullFramework
 			var application = (HttpApplication)sender;
 			var context = application.Context;
 			var response = context.Response;
+
+			if (SoapRequest.TryExtractSoapAction(_logger, context.Request, out var soapAction))
+				transaction.Name += $" {soapAction}";
 
 			// update the transaction name based on route values, if applicable
 			if (transaction is Transaction t && !t.HasCustomName)
--- a/src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
+++ b/src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
@@ -291,9 +291,6 @@ namespace Elastic.Apm.AspNetFullFramework
 			var context = application.Context;
 			var response = context.Response;
 
-			if (SoapRequest.TryExtractSoapAction(_logger, context.Request, out var soapAction))
-				transaction.Name += $" {soapAction}";
-
 			// update the transaction name based on route values, if applicable
 			if (transaction is Transaction t && !t.HasCustomName)
 			{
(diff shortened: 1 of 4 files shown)


src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
```diff
@@ -293,6 +291,9 @@ private void ProcessEndRequest(object sender)
 			var context = application.Context;
 			var response = context.Response;
 
+			if (SoapRequest.TryExtractSoapAction(_logger, context.Request, out var soapAction))
```
gregkalapos, line 294: 👍 good idea to move this to `ProcessEndRequest`.

test/Elastic.Apm.AspNetFullFramework.Tests/Soap/SoapParsingTests.cs
```diff
@@ -184,7 +184,7 @@ public class SoapParsingTests
 		public void Soap12Parser_ParsesHeaderAndBody(string soap, string expectedAction)
 		{
 			var requestStream = new MemoryStream(Encoding.UTF8.GetBytes(soap));
-			var action = SoapRequest.GetSoap12ActionFromInputStream(requestStream);
+			var action = SoapRequest.GetSoap12ActionFromInputStream(null, requestStream);
```
gregkalapos, line 187: nit: we have `NoopLogger` (this will need a `using` at the top of the file)
```suggestion
			var action = SoapRequest.GetSoap12ActionFromInputStream(new NoopLogger(), requestStream);
```

src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
```diff
@@ -293,6 +291,9 @@ private void ProcessEndRequest(object sender)
...
+				transaction.Name += $" {soapAction}";
```
gregkalapos, line 295: There is an edge case we should consider here: transactions can be renamed by users via the public API - super rare, but that's possible (just do `Agent.Tracer.CurrentTransaction.Name = "foo"`). If a custom transaction name is used, I think we should not add the SOAP action name. We can do something like that:

```
if (transaction is Transaction realTransaction)
{
	if(!realTransaction.HasCustomName)
		transaction.Name += $" {soapAction}";
}
else // If it's some other implementation of `ITransaction`, let's just append the soap action name
	transaction.Name += $" {soapAction}";
```

Parsing the `soapAction` name can be also saved if we know the transaction has a custom name. Also, I see there is already a casting to `Transaction` going on later in this method - maybe just do the check+parsing there, if possible.

src/Elastic.Apm.AspNetFullFramework/Extensions/SoapRequest.cs
```diff
@@ -82,19 +74,50 @@ private static bool IsSoap12Action(NameValueCollection headers)
...
+		{ 
+			var shared = ArrayPool<byte>.Shared;
+			var bytes = shared.Rent((int)stream.Length);
 			try
 			{
-				streamReader = new StreamReader(stream);
+				stream.Position = 0;
+				stream.Read(bytes, 0, bytes.Length);
+				stream.Position = 0;
+				var content = Encoding.ASCII.GetString(bytes);
```
gregkalapos, line 112: Seems to be unused. 

```suggestion
```

gregkalapos (commented):
Nice! I think this will address the issues we saw.

I found some smaller things - see below.

src/Elastic.Apm.AspNetFullFramework/Extensions/SoapRequest.cs
z1c0: Thanks, a leftover from my dozens of different experiments 😄 .

src/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
z1c0: Thanks, very good point. I was not aware of that.

gregkalapos (approved):
Thanks for the updates!

PR #1933, opened 2022-12-01
Enable DOTNET_STARTUP_HOOKS for .NET 7

z1c0:
Enable the `DOTNET_STARTUP_HOOKS` functionality for .NET 7.

The important change is this [line](https://github.com/elastic/apm-agent-dotnet/pull/1933/files#diff-9e2a2dda2e5c74dfba09c0a6ea8be7e1f84bc637a10d44795b5887a95c27e271R98) - with that startup hooks functionality will be upwards compatible by default. That way we don't need to create an additional (redundant) folder for .NET 7.

The [test input mechanism](https://github.com/elastic/apm-agent-dotnet/pull/1933/files#diff-c07906b942fb1af3cb70e6198e8591a21fc0b6bfc391d99321e02272d2bfc362) was also refactored to allow for simple test extension (e.g. NET 8.0) in the future.

### DIFF
--- a/build/scripts/scripts.fsproj
+++ b/build/scripts/scripts.fsproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>net7.0</TargetFramework>
     <OutputType>Exe</OutputType>
     <NoWarn>$(NoWarn);NU1701</NoWarn>
     <IsPackable>false</IsPackable>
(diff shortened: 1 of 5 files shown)


build/scripts/scripts.fsproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>net7.0</TargetFramework>
```
z1c0, line 3: FYI @gregkalapos - this change is not required. However, I had repeated ([Apple Silicon-related](https://stackoverflow.com/questions/68204296/mac-m1-chip-issues-with-rosetta-attachment-of-code-signature-supplement-failed)) problems when running the `./build agent-zip` command, so I just kept it in.

gregkalapos (approved):

PR #1935, opened 2022-12-05
Support for `Elastic.Clients.Elasticsearch`

gregkalapos:
This PR is covering the rest of the things needed to support the new Elasticsearch .NET Client.

The core of the change was done directly in the client code: https://github.com/elastic/elasticsearch-net/pull/6990

This PR adds:
- Integration-tests with docker by using the new es client and by using the OTel bridge in the agent.
- Implements suppression of HTTP span creation when `elasticsearch` spans are already created by the OTel bridge.

Regarding release: I think this should not block a release - if this ends up being merged after the next release, it's still fine.

### DIFF
--- a/src/Elastic.Apm/OpenTelemetry/ElasticActivityListener.cs
+++ b/src/Elastic.Apm/OpenTelemetry/ElasticActivityListener.cs
@@ -29,7 +29,7 @@ namespace Elastic.Apm.OpenTelemetry
 
 		private readonly IApmLogger _logger;
 		private Tracer _tracer;
-		private HttpTraceConfiguration _httpTraceConfiguration;
+		private readonly HttpTraceConfiguration _httpTraceConfiguration;
 
 		internal void Start(Tracer tracerInternal)
 		{
@@ -158,8 +158,6 @@ namespace Elastic.Apm.OpenTelemetry
 							case ActivityStatusCode.Error:
 								transaction.Outcome = Outcome.Failure;
 								break;
-							default:
-								break;
 						}
 #endif
 
@@ -191,8 +189,6 @@ namespace Elastic.Apm.OpenTelemetry
 							case ActivityStatusCode.Error:
 								span.Outcome = Outcome.Failure;
 								break;
-							default:
-								break;
 						}
 #endif
 						span.End();
@@ -207,17 +203,11 @@ namespace Elastic.Apm.OpenTelemetry
 			var isMessaging = activity.Tags.Any(n => n.Key == "messaging.system");
 
 			if (activity.Kind == ActivityKind.Server && (isRpc || isHttp))
-			{
 				transaction.Type = ApiConstants.TypeRequest;
-			}
 			else if (activity.Kind == ActivityKind.Consumer && isMessaging)
-			{
 				transaction.Type = ApiConstants.TypeMessaging;
-			}
 			else
-			{
 				transaction.Type = "unknown";
-			}
 		}
 
 		private ActivityListener Listener { get; set; }
--- a/src/Elastic.Apm/OpenTelemetry/ElasticActivityListener.cs
+++ b/src/Elastic.Apm/OpenTelemetry/ElasticActivityListener.cs
@@ -24,13 +24,17 @@ namespace Elastic.Apm.OpenTelemetry
 		private readonly ConcurrentDictionary<string, Span> ActiveSpans = new();
 		private readonly ConcurrentDictionary<string, Transaction> ActiveTransactions = new();
 
-		internal ElasticActivityListener(IApmAgent agent) => _logger = agent.Logger?.Scoped(nameof(ElasticActivityListener));
+		internal ElasticActivityListener(IApmAgent agent, HttpTraceConfiguration httpTraceConfiguration) => (_logger, _httpTraceConfiguration) =
+			(agent.Logger?.Scoped(nameof(ElasticActivityListener)), httpTraceConfiguration);
 
 		private readonly IApmLogger _logger;
 		private Tracer _tracer;
+		private HttpTraceConfiguration _httpTraceConfiguration;
 
 		internal void Start(Tracer tracerInternal)
 		{
+			_httpTraceConfiguration?.AddTracer(new ElasticSearchHttpNonTracer());
+
 			_tracer = tracerInternal;
 			Listener = new ActivityListener
 			{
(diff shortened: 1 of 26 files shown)


gregkalapos:
Looking at

```
[2022-12-05T08:48:41.400Z] /var/lib/jenkins/workspace/net_apm-agent-dotnet-mbp_PR-1935/apm-agent-dotnet/src/Elastic.Apm/OpenTelemetry/ElasticSearchHttpNonTracer.cs(29,35): error CS1061: 'Activity' does not contain a definition for 'DisplayName' and no accessible extension method 'DisplayName' accepting a first argument of type 'Activity' could be found (are you missing a using directive or an assembly reference?) [/var/lib/jenkins/workspace/net_apm-agent-dotnet-mbp_PR-1935/apm-agent-dotnet/src/Elastic.Apm/Elastic.Apm.csproj::TargetFramework=netstandard2.0]
```

test/Elastic.Clients.Elasticsearch.Tests/ElasticSearchTestFixture.cs
```diff
@@ -0,0 +1,51 @@
...
+	public async Task InitializeAsync()
+	{
+		await Container.StartAsync();
+
+
+		var settings = new ElasticsearchClientSettings(new Uri(Container.ConnectionString))
+			.ServerCertificateValidationCallback(CertificateValidations.AllowAll)
+			.Authentication(new BasicAuthentication(Container.Username, Container.Password));
+
+		Cleint = new ElasticsearchClient(settings);
```
z1c0, line 35:
```suggestion
		Client = new ElasticsearchClient(settings);
```

test/Elastic.Clients.Elasticsearch.Tests/ElasticSearchTestFixture.cs
```diff
@@ -0,0 +1,51 @@
...
+		if (Cleint == null)
```
z1c0, line 36:
```suggestion
		if (Client == null)
```

test/Elastic.Clients.Elasticsearch.Tests/ElasticSearchTests.cs
```diff
@@ -0,0 +1,241 @@
...
+{
+	private readonly ITestOutputHelper _testOutputHelper;
+	private readonly ElasticSearchTestFixture _esClientListenerFixture;
+	private ElasticsearchClient _client;
+
+	public ElasticSearchTests(ITestOutputHelper testOutputHelper, ElasticSearchTestFixture esClientListenerFixture)
+	{
+		_testOutputHelper = testOutputHelper;
+		_esClientListenerFixture = esClientListenerFixture;
+		_client = _esClientListenerFixture.Cleint ?? throw new Exception("ElasticsearchClient is `null`");
```
z1c0, line 22:
```suggestion
		_client = _esClientListenerFixture.Client ?? throw new Exception("ElasticsearchClient is `null`");
```

src/Elastic.Apm/OpenTelemetry/ElasticSearchHttpNonTracer.cs
```diff
@@ -0,0 +1,34 @@
...
+	public class ElasticSearchHttpNonTracer : IHttpSpanTracer
+	{
+		public bool IsMatch(string method, Uri requestUrl, Func<string, string> headerGetter) => false;
+		public ISpan StartSpan(IApmAgent agent, string method, Uri requestUrl, Func<string, string> headerGetter) => null;
+
+		public bool SuppressSpanCreation()
+		{
+			if (Activity.Current == null || Activity.Current.Parent == null)
+				return false;
+			return Activity.Current.Parent.DisplayName.StartsWith("Elasticsearch:") && Activity.Current.Parent.Tags.Any(n => n is { Key: "db.system", Value: "elasticsearch" });
```
z1c0, line 29: There are cases where `SuppressSpanCreation` returns `false`,  however `IsMatch` always returns `false` and `StartSpan` is a no-op. Is this intentional?

src/Elastic.Apm/OpenTelemetry/ElasticActivityListener.cs
```diff
@@ -24,13 +24,17 @@ public class ElasticActivityListener : IDisposable
...
+		internal ElasticActivityListener(IApmAgent agent, HttpTraceConfiguration httpTraceConfiguration) => (_logger, _httpTraceConfiguration) =
+			(agent.Logger?.Scoped(nameof(ElasticActivityListener)), httpTraceConfiguration);
 
 		private readonly IApmLogger _logger;
 		private Tracer _tracer;
+		private HttpTraceConfiguration _httpTraceConfiguration;
 
 		internal void Start(Tracer tracerInternal)
 		{
+			_httpTraceConfiguration?.AddTracer(new ElasticSearchHttpNonTracer());
```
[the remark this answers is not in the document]
gregkalapos: No, it should not be. But technically it's impossible and I just wanted to be on the safe side.

src/Elastic.Apm/OpenTelemetry/ElasticSearchHttpNonTracer.cs
gregkalapos: Yes, that is intentional and (from my understanding) is the expected behaviour we want.

`SuppressSpanCreation` returns `false` means we deal with an http call and it's not coming from the Elasticsearch (es) client (e.g. a normal call via `HttpClient` to some web service).

`IsMatch` always returns `false`: so if it's not from the es client, the `ElasticSearchHttpNonTracer.cs` never matches any any outgoing HTTP call. In `HttpDiagnosticListenerImplBase` you can see, if `IsMatch` returns `false`, then the normal outgoing HTTP capturing will run (so we capture the call as a normal http span).

`StartSpan` is a no-op: this will never run because `IsMatch` returns false. And even if it runs, we return `null` which makes the normal HTTP capturing to kick in.

So in case of an HTTP call, the result is:
- if it's from the es client: in `HttpDiagnosticListenerImplBase` we suppress span creation completely - no span for those http calls.
- if it's not not from the es client: it'll never match any url, so `HttpDiagnosticListenerImplBase` will run the normal HTTP span creation code path.

z1c0: Makes sense, thanks for the explanation 👍.

z1c0 (approved):
LGTM 👍

PR #1942, opened 2022-12-06
Fix flaky tests

z1c0:
This PR aims at making the GRPC tests less flaky by:
* Updating GRPC-related NuGet packages (e.g. `Grpc.Tools`) to their latest version.
* Add proper cleanup/shutdown to test cases (also in case of failures).

Also, run this test on more .NET versions.

### DIFF
--- a/test/Elastic.Apm.Grpc.Tests/GrpcTests.cs
+++ b/test/Elastic.Apm.Grpc.Tests/GrpcTests.cs
@@ -14,14 +14,23 @@ using GrpcServiceSample;
 using Microsoft.Extensions.Hosting;
 using Xunit;
 
-#pragma warning disable CS0436 // Type conflicts with imported type
-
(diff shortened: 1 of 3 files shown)


test/Elastic.Apm.Grpc.Tests/GrpcTests.cs
```diff
@@ -14,14 +14,23 @@
 using Microsoft.Extensions.Hosting;
 using Xunit;
 
-#pragma warning disable CS0436 // Type conflicts with imported type
-
 namespace Elastic.Apm.Grpc.Tests
 {
-	public class GrpcTests
+	public class GrpcTests : IAsyncLifetime
```
gregkalapos, line 19: Definitely makes sense.

gregkalapos (approved):

PR #1950, opened 2022-12-13
Continue implementation of Azure Functions spec

z1c0:
This PR implements the remaining items of the (yet still in progress) [Azure Functions tracing specification](https://github.com/elastic/apm/pull/716):

## Disabled Functionality

The agent disables certain functionality (e.g. metrics collection, remote configuration, ...) when running in Azure Functions.
For that, the concept of **Agent Features** (see [here](https://github.com/elastic/apm-agent-dotnet/pull/1950/files#diff-f0d3baa9742b47d8b2af711c92fb241b689d51c36c9f9d343c422c1fea524463)) was introduced.
This concept is meant for enabling/disabling certain agent features based on an agent's habitat (e.g. Azure Functions, AWS Lambda, ...) but could be extended to a more generic feature (e.g. review/beta) mechanism over time.

## MetaData

The agent now provides these additional metadata fields in [`UpdateServiceInformation`](https://github.com/elastic/apm-agent-dotnet/pull/1950/files#diff-0cda3819b069d7a4b554472497a8afd42a76ffac36faea4609113d76fde9604dR69):
* `service.framework.*`
* `service.runtime.*`

## HTTP Trigger

The agent now populates the `context.request` and `context.response` fields for HTTP invocations.

## Other noteworthy changes:

*  The **CloudMetadataProvider** mechanism for Azure Functions (and AppService as a consequence) was refactored to work together with the new above-mentioned `AgentFeatures` mechanism.
* Unit tests cover more scenarios and details now.


src/Elastic.Apm.Azure.Functions/ApmMiddleware.cs
```diff
@@ -114,6 +136,9 @@ private static void SetTriggerSpecificResult(ITransaction transaction, bool succ
 		}
 	}
 
+	private static Dictionary<string, string> CreateHeadersDictionary(HttpHeadersCollection httpHeadersCollection) =>
+		httpHeadersCollection.ToDictionary(h => h.Key, h => string.Join(',', h.Value));
```
gregkalapos, line 140: This needs to be sanitized. We have the [`SanitizeFieldNames`](https://www.elastic.co/guide/en/apm/agent/dotnet/current/config-core.html#config-sanitize-field-names) config, I think that should apply here as well.

[Here is](https://github.com/elastic/apm-agent-dotnet/blob/main/src/Elastic.Apm.AspNetCore/WebRequestTransactionCreator.cs#L126-L132) how we do this in case of ASP.NET Core.

gregkalapos (changes requested):
Looks very nice. 

The AgentFeature idea definitely makes sense - I think we'll run into other situations as well when we can take advantage of this.

I added a few comments, the only (in my opinion) serious one is about the HTTP header sanitization (details below), otherwise it's ready-to-go.

src/Elastic.Apm/AgentComponents.cs
```diff
@@ -137,12 +143,30 @@ public class AgentComponents : IApmAgent, IDisposable
...
-					MetricsCollector.StartCollecting();
+					var agentFeatures = AgentFeaturesProvider.Get(Logger);
+					//
+					// Central configuration
+					//
+					if (centralConfigurationFetcher != null)
+						CentralConfigurationFetcher = centralConfigurationFetcher;
+					else if (agentFeatures.Check(AgentFeature.RemoteConfiguration))
+						CentralConfigurationFetcher =
+							new CentralConfigurationFetcher(Logger, ConfigurationStore, Service);
```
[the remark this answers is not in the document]
z1c0: Sorry for that. After after upgrading Rider, the automatic code cleanup/formatting I had configured was broken.
I only realized that later :-(

src/Elastic.Apm.Azure.Functions/ApmMiddleware.cs
```diff
@@ -73,20 +66,42 @@ public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next
...
+	{
+		if (service == null)
+		{
+			Logger.Warning()?.Log($"{nameof(UpdateServiceInformation)}: service is null");
+			return;
+		}
+
+		service.Framework = new() { Name = "Azure Functions", Version = MetaData.FunctionsExtensionVersion };
+		var runtimeVersion = service.Runtime?.Version ?? "n/a";
+		service.Runtime = new() { Name = MetaData.FunctionsWorkerRuntime, Version = runtimeVersion };
```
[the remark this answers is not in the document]
z1c0: Yes, that was my intention.
The runtime version is the generic .NET version (which is correct IMHO) but with the runtime name, we can be more specific to differentiate between the actual .NET runtime (`dotnet` vs `dotnet-isolated`) used.

src/Elastic.Apm.Azure.Functions/ApmMiddleware.cs
z1c0: I actually thought about sanitization, but then I saw [this code](https://github.com/elastic/apm-agent-dotnet/blob/5672eb4b9e1cd3508e0c3827359374f9cf5459fd/src/Elastic.Apm/Report/PayloadSenderV2.cs#L131) and thought sanitization would happen generically?
(I forgot to test this however tbh.)

gregkalapos: Hahh, you are right! 

...which means then we should remove it in `WebRequestTransactionCreator.cs` (unrelated to this PR, so no need to follow up here).

Sorry for the confusion.

gregkalapos (approved):
LGTM

src/Elastic.Apm.Azure.Functions/ApmMiddleware.cs
z1c0: I also saw that and suspected it was a remnant. I'll put it on my TODO list as a follow-up task 👍 .

z1c0:
> LGTM

Now I only need to get the CI (tests) stable again 😆.

z1c0:
@gregkalapos the tests are finally passing now.

I had to refactor https://github.com/elastic/apm-agent-dotnet/blob/main/test/Elastic.Apm.Azure.Functions.Tests/AzureFunctionsTests.cs by introducing https://github.com/elastic/apm-agent-dotnet/blob/main/test/Elastic.Apm.Azure.Functions.Tests/AzureFunctionsTestFixture.cs which leaves the `func` process (and `MockApmServer`) alive throughout the runtime of **all** Azure Functions tests now (as opposed to starting/stopping for each single invocation before).

This makes the tests less brittle already. Also, I made sure that the `func` tool is fully operational before running the actual tests by waiting for its log output [here](https://github.com/elastic/apm-agent-dotnet/blob/main/test/Elastic.Apm.Azure.Functions.Tests/AzureFunctionsTestFixture.cs#L66).

As no more changes to production code were made, I took the liberty of merging the PR right away. I hope that's fine with you.

PR #1959, opened 2022-12-22
Implement latest Azure Functions spec changes

z1c0:
This PR implements the latest changes made to the [Azure Functions spec](https://github.com/elastic/apm/pull/716):
* Set `service.name` to `WEBSITE_SITE_NAME` (if no explicit service name was provided in the configuration).
* Set `cloud.service.name` to `functions`
* Set `service.node.configured_name` to `WEBSITE_INSTANCE_ID`.

### DIFF
--- a/src/Elastic.Apm.Azure.Functions/ApmMiddleware.cs
+++ b/src/Elastic.Apm.Azure.Functions/ApmMiddleware.cs
@@ -5,6 +5,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Elastic.Apm.Api;
 using Elastic.Apm.Cloud;
+using Elastic.Apm.Config;
 using Elastic.Apm.Extensions;
 using Elastic.Apm.Logging;
 using Elastic.Apm.Model;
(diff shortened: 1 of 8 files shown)


src/Elastic.Apm.Azure.Functions/ApmMiddleware.cs
```diff
@@ -74,9 +75,16 @@ private static void UpdateServiceInformation(Service? service)
...
+		if (service.Name == AbstractConfigurationReader.AdaptServiceName(AbstractConfigurationReader.DiscoverDefaultServiceName()))
+		{
+			// Only override the service name if it was set to default.
+			service.Name = MetaData.WebsiteSiteName;
+		}
 		service.Framework = new() { Name = "Azure Functions", Version = MetaData.FunctionsExtensionVersion };
 		var runtimeVersion = service.Runtime?.Version ?? "n/a";
 		service.Runtime = new() { Name = MetaData.FunctionsWorkerRuntime, Version = runtimeVersion };
+		service.Node ??= new Node();
+		service.Node.ConfiguredName = MetaData.WebsiteInstanceId;
```
gregkalapos, line 87: There is a config for this field: [docs here](https://www.elastic.co/guide/en/apm/agent/dotnet/current/config-core.html#config-service-node-name).

Now question is, what does that mean in case of Azure Functions? :) 

With the current code, we'd just overwrite that value. I think in most cases, users will not set it and all good.

If they set it, maybe we can do one of 2 things:
- (a bit similar to `service.Name` above): Check here if it's already set (that happens [here](https://github.com/elastic/apm-agent-dotnet/blob/9795170d906add11efba094eb5d7e97f56737752/src/Elastic.Apm/Api/Service.cs#L64) if the config returns a value), and only use `MetaData.WebsiteInstanceId` if it's not set.
- Overwrite it, but maybe print a warning to let users know.

gregkalapos (approved):
LGTM

src/Elastic.Apm.Azure.Functions/ApmMiddleware.cs
z1c0: Thanks for the hint 👍 .
We should definitely address that in the spec.

z1c0: As discussed today via Zoom, the latest commit prints a log message if we override `Node.ConfiguredName`.
I'll put that in the spec as well.

PR #1960, opened 2022-12-22
Remove redundant header sanitization code

z1c0:
As HTTP headers are sanitized generically in `PayloadSender` we can remove the explicit sanitation in `WebRequestTransactionCreator` (as suggested [here](https://github.com/elastic/apm-agent-dotnet/pull/1950#discussion_r1053696287)).


gregkalapos (approved):
lgtm

Issue #1895, opened 2022-11-04
AppSettings ElasticApm:Enabled is not fully honored in ASP.NET Classic


xecrets:
Without going in to the full details, there is a code path executed early during initialization that causes the  `Agent.Instance.ConfigurationReader` to be set to an instance of `EnvironmentConfigurationReader`, thus bypassing the setting in Web.Config AppSettings. I believe the problem has it's root in:

    Elastic.Apm.AspNetFullFramework.ElasticApmModule.SafeAgentSetup():

    AgentComponents agentComponents = CreateAgentComponents(dbgInstanceName);
    if (agentComponents.ConfigurationReader.Enabled)
    {
        Agent.Setup(agentComponents);
    }

This causes the `Agent.Setup()` call to be skipped (because the `Enabled` setting tested here does report 'false' from Web.Config), which in turn causes the later evalutation of `Elastic.Apm.Agent.LazyApmAgent.Value` to call:

    return new ApmAgent(Components);

with `Components == null` (because of the above skip of `Agent.Setup()`). This in turn calls:

    internal ApmAgent(AgentComponents agentComponents)
    {
        Components = agentComponents ?? new AgentComponents();
    }

and since `agentComponents` is `null` here, a new `AgentComponents` is instantatiated like this:

    public AgentComponents(IApmLogger logger = null, IConfigurationReader configurationReader = null, IPayloadSender payloadSender = null)
        : this(logger, configurationReader, payloadSender, null, null, null, null)
    {
    }

where the important part is `IConfigurationReader configurationReader = null` and this in turn results in:

    internal AgentComponents(IApmLogger logger, IConfigurationReader configurationReader, IPayloadSender payloadSender, IMetricsCollector metricsCollector, ICurrentExecutionSegmentsContainer currentExecutionSegmentsContainer, ICentralConfigurationFetcher centralConfigurationFetcher, IApmServerInfo apmServerInfo, BreakdownMetricsProvider breakdownMetricsProvider = null)
    {
        try
        {
            IApmLogger tempLogger = logger ?? ConsoleLogger.LoggerOrDefault(configurationReader?.LogLevel);
            ConfigurationReader = configurationReader ?? new EnvironmentConfigurationReader(tempLogger);
            ...

et voilà! We have a `ConfigurationReader` of type `EnvironmentConfigurationReader` not honoring the AppSettings value .

jzelos:
[...]

PR #1961, opened 2022-12-22
Prevent discarding of agent components if agent is disabled

z1c0:
Could you take a look at this, please @gregkalapos?

It's definitely a bug that's been reported in the linked issue #1895.

If the agent is disabled via config, all the agent components get discarded and we end up with default agent components (e.g. environment-variable based configuration instead of `web.config`.based).

The change that I made now makes sure that `Agent.Setup(agentComponents)` is still called in that case.

I could find any negative side effects so far, but it is definitely a hairy change.

Also, it's not really feasible to test this other than creating a whole new sample application with a disabled agent, since the code relies on statically initialized data. So I only ran a few manual tests.


gregkalapos (approved):
lgtm

PR #1965, opened 2023-01-03
Fixed the broken Dockerfile

philippkahr:
apt-get is not available on alpine.
curl was also missing.
The url to the downloads was wrong as the version is prefixed with a `v` as in `/v1.19.0` instead of `/1.19.0`

### DIFF
--- a/docs/setup-auto-instrumentation.asciidoc
+++ b/docs/setup-auto-instrumentation.asciidoc
@@ -120,19 +120,18 @@ can be used as part of a https://docs.docker.com/develop/develop-images/multista
 
 [source,sh]
 ----
-ARG AGENT_VERSION=1.11.1
+ARG AGENT_VERSION=1.19.0
 
 FROM alpine:latest AS build
 ARG AGENT_VERSION
 WORKDIR /source
 
 # install unzip
-RUN apt-get -y update && apt-get -yq install unzip
+RUN apk update && apk add zip curl
 
 # pull down the zip file based on ${AGENT_VERSION} ARG and unzip
-RUN curl -L -o elastic_apm_profiler_${AGENT_VERSION}.zip https://github.com/elastic/apm-agent-dotnet/releases/download/${AGENT_VERSION}/elastic_apm_profiler_${AGENT_VERSION}.zip && \ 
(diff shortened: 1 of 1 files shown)


philippkahr:
Only thing, would it be possible to automatically detect the latest version of the agent for the docs? Currently this is hardcoded in line 1:

```
ARG AGENT_VERSION=1.19.0
```

I think the docs should auto update the 1.19.0, every time there is a new release.

gregkalapos (approved):
LGTM.

@bmorelli25 can you answer [this part](https://github.com/elastic/apm-agent-dotnet/pull/1965#issuecomment-1369666048)? - unfortunately I don't know if we can do that.

docs/setup-auto-instrumentation.asciidoc
```diff
@@ -120,19 +120,18 @@ can be used as part of a https://docs.docker.com/develop/develop-images/multista
...
 ARG AGENT_VERSION
 WORKDIR /source
 
 # install unzip
-RUN apt-get -y update && apt-get -yq install unzip
+RUN apk update && apk add zip curl
 
 # pull down the zip file based on ${AGENT_VERSION} ARG and unzip
-RUN curl -L -o elastic_apm_profiler_${AGENT_VERSION}.zip https://github.com/elastic/apm-agent-dotnet/releases/download/${AGENT_VERSION}/elastic_apm_profiler_${AGENT_VERSION}.zip && \ 
+RUN curl -L -o elastic_apm_profiler_${AGENT_VERSION}.zip https://github.com/elastic/apm-agent-dotnet/releases/download/v${AGENT_VERSION}/elastic_apm_profiler_${AGENT_VERSION}.zip && \
```
gregkalapos, line 133: Ahh, nice!

At some point we added `v` before the version numbers in the .NET agent, in order to be aligned with other APM agents. So the url without `v` worked for older versions, but it's now needed here for newer versions.

bmorelli25:
Apologies, I've been on PTO.

> Only thing, would it be possible to automatically detect the latest version of the agent for the docs?

Hmmm. Unfortunately, there's no way to do this that I'm aware of.

PR #1966, opened 2023-01-03
Document OTel bridge

gregkalapos:
- The OTel bridge got released in v1.13.0 as experimental, but we have not documented it back then.
- This PR adds documentation including samples for the OTel Bridge.

### DIFF
--- a/docs/api-opentelemetry.asciidoc
+++ b/docs/api-opentelemetry.asciidoc
@@ -19,7 +19,6 @@ In .NET, there is an https://opentelemetry.io/docs/instrumentation/net/shim/[Ope
 
 The OpenTelemetry bridge in the Elastic .NET APM Agent targets the https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity?view=net-5.0[Activity API]. Since the https://opentelemetry.io/docs/instrumentation/net/shim/[OpenTelemetry .NET shim] builds on top of the https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity?view=net-5.0[Activity API], the shim is implicitly supported as well, although we don't directly test it, because the Activity API is the recommended API.
 
-
 [float]
 [[otel-getting-started]]
 === Getting started
@@ -47,7 +46,6 @@ If you configured the agent via the `appsettings.json` file, then set `ElasticAp
     }
 }
 ----
-
 
 [float]
 [[create-activity-source-and-spans]]
--- a/docs/api-opentelemetry.asciidoc
+++ b/docs/api-opentelemetry.asciidoc
@@ -113,7 +113,7 @@ The Metrics API is currently not supported.
 [[otel-span-events]]
 ===== Span Events
 Span events (https://open-telemetry.github.io/opentelemetry-js-api/interfaces/span.html#addevent[`Span#addEvent()`])
-is not currently supported. Events will be silently dropped.
+are not currently supported. Events will be silently dropped.
 
 [float]
 [[otel-baggage]]
--- a/docs/api-opentelemetry.asciidoc
+++ b/docs/api-opentelemetry.asciidoc
@@ -1,0 +1,123 @@
+ifdef::env-github[]
+NOTE: For the best reading experience,
+please view this documentation at https://www.elastic.co/guide/en/apm/agent/dotnet[elastic.co]
+endif::[]
+
+[[opentelemetry-bridge]]
+== OpenTelemetry Bridge
+
+NOTE: Added as experimental in v1.13.0.
+To enable it, set <<opentelemetry-bridge-enabled, `opentelemetryBridgeEnabled`>> to `true`.
+
+NOTE: The OpenTelemetry bridge is supported on .NET 5 and newer versions.
+
+The OpenTelemetry Bridge in the Elastic .NET APM Agent bridges OpenTelemetry spans into Elastic APM transactions and spans. The Elastic APM OpenTelemetry bridge allows one to use the vendor-neutral OpenTelemetry Tracing API to manually instrument one's code, and have the Elastic .NET APM agent handle those API calls. This allows one to use the Elastic APM agent for tracing, without any vendor lock-in from adding manual tracing using the APM agent’s own <<public-api, Public API>>.
+
+OpenTelemetry in .NET is implemented via the https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity?view=net-5.0[Activity API].
+
+In .NET, there is an https://opentelemetry.io/docs/instrumentation/net/shim/[OpenTelemetry shim] which is built on top of the Activity API. 
+
+The OpenTelemetry bridge in the Elastic .NET APM Agent targets the https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity?view=net-5.0[Activity API]. Since the https://opentelemetry.io/docs/instrumentation/net/shim/[OpenTelemetry .NET shim] builds on top of the https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity?view=net-5.0[Activity API], the shim is implicitly supported as well, although we don't directly test it, because the Activity API is the recommended API.
+
+
+[float]
+[[otel-getting-started]]
+=== Getting started
+
+The OpenTelemetry bridge is part of the core agent pacakge (https://www.nuget.org/packages/Elastic.Apm[`Elastic.Apm`]), so you don't need to add any additional dependency.
+
+[float]
+[[otel-enable-bridge]]
+==== Enable the OpenTelemetry Bridge
+
+Currently the bridge is opt-in, so you need to enable it.
+
+If you configured the agent via environment variables, set the `ELASTIC_APM_ENABLEOPENTELEMETRYBRIDGE` environment variable to `true``.
+
+If you configured the agent via the `appsettings.json` file, then set `ElasticApm:EnableOpenTelemetryBridge` to `true`.
+
+[source,js]
+----
+{
+  "ElasticApm":
+    {
+      "ServerUrl":  "http://myapmserver:8200",
+      "SecretToken":  "apm-server-secret-token",
+      "EnableOpenTelemetryBridge": true
+    }
+}
+----
+
+
+[float]
+[[create-activity-source-and-spans]]
+==== Create an ActivitySource and start spans
+
+Once you enabled the bridge, you can create OpenTelemetry spans, or in .NET terminology you can start created new activities via the activity source and the agent will bridge those spans automatically.
+
+[source,csharp]
+----
+public static void Sample()
+{
+	var src = new ActivitySource("Test");
+	using var activity1 = src.StartActivity(nameof(Sample), ActivityKind.Server);
+	Thread.Sleep(100);
+	using var activity2 = src.StartActivity("foo");
+	Thread.Sleep(150);
+}
+----
+
+The code-snippet above creates a span called `Sample` and a child span on `Sample` called `foo`. The bridge will create a transaction from `Sample` and a child span called `foo`.
+
+[float]
+[[mixing-apis]]
+==== Mixing OpenTelemetry and the Elastic APM Public API
+
+You can also mix the Activity API with the <<public-api, Public API>>, the OpenTelemetry bridge will take care of putting the spans into the right place. The advantage of this is that if you already have some libraries that you instrumented via the <<public-api, Public API>>, but going forward, you'd like to use the vendor-independent OpenTelemetry API, you don't need to replace all Public API calls in one go.
+
+[source,csharp]
+----
+/// ElasticTransaction
+/// -
+/// ---> OTelSpan
+///           -
+///           ---> ElasticSpan
+
+var src = new ActivitySource("Test");
+tracer.CaptureTransaction( nameof(Sample4), "test", t =>
+{
+	Thread.Sleep(100);
+	using (var activity = src.StartActivity("foo"))
+	{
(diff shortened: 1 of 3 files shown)


docs/configuration.asciidoc
```diff
@@ -593,6 +593,26 @@ Additionally, spans that lead to an error can't be discarded.
...
+[float]
+[[config-enable-open-telemetry-bridge]]
+==== `EnableOpenTelemetryBridge` (added[1.13])
+
+Setting this option to true will enable the <<opentelemetry-bridge, OpenTelemetry Bridge>>. Briefly, the OpenTelemetry Bridge allows one to use the vendor-neutral OpenTelemetry Tracing API (which is the https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity?view=net-5.0[Activity API] in .NET) to manually instrument your code, and have the Elastic .NET APM agent handle those API calls.
+
+[options="header"]
+|============
+| Environment variable name            | IConfiguration key
+| `ELASTIC_APM_ENABLEOPENTELEMETRYBRIDGE` | `ElasticApm:EnableOpenTelemetryBridge`
```
gregkalapos, line 605: Let's discuss the naming, because I don't think this is ok.

What I documented is the current status - so this is in the agent code and right now users use these values.

Issues I see:
- For the environment variable we should use `_`, so it should be `ELASTIC_APM_ENABLE_OPENTELEMETRY_BRIDGE` (or `OPEN_TELEMETRY`?)
- The Node.js agent uses [`ELASTIC_APM_OPENTELEMETRY_BRIDGE_ENABLED`](https://github.com/elastic/apm-agent-nodejs/blob/main/docs/configuration.asciidoc#opentelemetrybridgeenabled) - if we change this in .NET, I think we should be in-sync with other agents. This config is not part of the spec, because most agents integrate with OTel differently ([see here](https://www.elastic.co/guide/en/apm/agent/java/current/opentelemetry-bridge.html#otel-getting-started)).
- If we change it, that'll be a breaking change for existing users. This feature is not yet GA if I remember correctly, so technically we can do it, but it's still not nice.

What are other's thoughts on this?

I tend to prefer renaming to `ELASTIC_APM_OPENTELEMETRY_BRIDGE_ENABLED` (and adapt the `IConfiguration key` as well), but no strong feelings...

bmorelli25: Not sure if my vote counts, but I vote for `ELASTIC_APM_OPENTELEMETRY_BRIDGE_ENABLED`. If it's not GA yet, I think we should take advantage of that fact and align on naming. That's what preview/beta is for, right?

bmorelli25: Also, hi. I'm back from vacation now and taking a look at this PR 👀

docs/api-opentelemetry.asciidoc
```diff
@@ -0,0 +1,121 @@
...
+
+In .NET, there is an https://opentelemetry.io/docs/instrumentation/net/shim/[OpenTelemetry shim] which is built on top of the Activity API. 
+
+The OpenTelemetry bridge in the Elastic .NET APM Agent targets the https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity?view=net-5.0[Activity API]. Since the https://opentelemetry.io/docs/instrumentation/net/shim/[OpenTelemetry .NET shim] builds on top of the https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity?view=net-5.0[Activity API], the shim is implicitly supported as well, although we don't directly test it, because the Activity API is the recommended API.
+
+[float]
+[[otel-getting-started]]
+=== Getting started
+
+The OpenTelemetry bridge is part of the core agent pacakge (https://www.nuget.org/packages/Elastic.Apm[`Elastic.Apm`]), so you don't need to add any additional dependency.
```
bmorelli25, line 26:
```suggestion
The OpenTelemetry bridge is part of the core agent package (https://www.nuget.org/packages/Elastic.Apm[`Elastic.Apm`]), so you don't need to add an additional dependency.
```

docs/api-opentelemetry.asciidoc
```diff
@@ -0,0 +1,121 @@
...
+      "EnableOpenTelemetryBridge": true
+    }
+}
+----
+
+[float]
+[[create-activity-source-and-spans]]
+==== Create an ActivitySource and start spans
+
+Once you enabled the bridge, you can create OpenTelemetry spans, or in .NET terminology you can start created new activities via the activity source and the agent will bridge those spans automatically.
```
bmorelli25, line 54:
```suggestion
After enabling the bridge, you can create OpenTelemetry spans, or in .NET terminology, you can start creating new activities via the activity source, and the agent will bridge those spans automatically.
```

docs/api-opentelemetry.asciidoc
```diff
@@ -0,0 +1,121 @@
...
+{
+	var src = new ActivitySource("Test");
+	using var activity1 = src.StartActivity(nameof(Sample), ActivityKind.Server);
+	Thread.Sleep(100);
+	using var activity2 = src.StartActivity("foo");
+	Thread.Sleep(150);
+}
+----
+
+The code-snippet above creates a span called `Sample` and a child span on `Sample` called `foo`. The bridge will create a transaction from `Sample` and a child span called `foo`.
```
bmorelli25, line 68:
```suggestion
The code snippet above creates a span named `Sample` and a child span on `Sample` named `foo`. The bridge will create a transaction from `Sample` and a child span named `foo`.
```

docs/api-opentelemetry.asciidoc
```diff
@@ -0,0 +1,121 @@
...
+	Thread.Sleep(100);
+	using (var activity = src.StartActivity("foo"))
+	{
+		tracer.CurrentSpan.CaptureSpan("ElasticApmSpan", "test", () => Thread.Sleep(50));
+		Thread.Sleep(150);
+	}
+});
+----
+
+The code-snippet above creates a transaction with the Elastic .NET APM Agent's <<public-api, Public API>>. Then it creates an activity called `foo`. This activity is going to be a child of the previously created transaction. Then finally, a span is created again by using the Elastic .NET APM Agent's <<public-api, Public API>>. This span is going to be a child span of the OpenTelemetry span.
```
bmorelli25, line 96:
```suggestion
The code snippet above creates a transaction with the Elastic .NET APM Agent's <<public-api, Public API>>. Then it creates an activity called `foo`; this activity will be a child of the previously created transaction. Finally, a span is created again using the Elastic .NET APM Agent's <<public-api, Public API>>; this span will be a child span of the OpenTelemetry span.
```

docs/configuration.asciidoc
```diff
@@ -593,6 +593,26 @@ Additionally, spans that lead to an error can't be discarded.
 |============
 
 
+[float]
+[[config-enable-open-telemetry-bridge]]
+==== `EnableOpenTelemetryBridge` (added[1.13])
+
+Setting this option to true will enable the <<opentelemetry-bridge, OpenTelemetry Bridge>>. Briefly, the OpenTelemetry Bridge allows one to use the vendor-neutral OpenTelemetry Tracing API (which is the https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity?view=net-5.0[Activity API] in .NET) to manually instrument your code, and have the Elastic .NET APM agent handle those API calls.
```
bmorelli25, line 600:
```suggestion
Setting this option to true will enable the <<opentelemetry-bridge, OpenTelemetry Bridge>>. This enables the use of the vendor-neutral OpenTelemetry Tracing API (the https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity?view=net-5.0[Activity API] in .NET) to manually instrument your code, and have the Elastic .NET APM agent handle those API calls.
```

docs/api-opentelemetry.asciidoc
```diff
@@ -0,0 +1,121 @@
...
+endif::[]
+
+[[opentelemetry-bridge]]
+== OpenTelemetry Bridge
+
+NOTE: Added as experimental in v1.13.0.
+To enable it, set <<config-enable-open-telemetry-bridge, `OpentelemetryBridgeEnabled`>> to `true`.
+
+NOTE: The OpenTelemetry bridge is supported on .NET 5 and newer versions.
+
```
bmorelli25, line 13: I'd prefer to use our built-in `preview::[]` as it includes legal jargon about how we support technical preview features. What about:

```suggestion
preview::[]

NOTE: This feature was added in v1.13.0. It is supported on .NET 5 and newer versions.

```

docs/api-opentelemetry.asciidoc
```diff
@@ -0,0 +1,121 @@
...
+The OpenTelemetry Bridge in the Elastic .NET APM Agent bridges OpenTelemetry spans into Elastic APM transactions and spans. The Elastic APM OpenTelemetry bridge allows one to use the vendor-neutral OpenTelemetry Tracing API to manually instrument one's code, and have the Elastic .NET APM agent handle those API calls. This allows one to use the Elastic APM agent for tracing, without any vendor lock-in from adding manual tracing using the APM agent’s own <<public-api, Public API>>.
```
bmorelli25, line 14:
```suggestion
The OpenTelemetry Bridge in the Elastic .NET APM Agent bridges OpenTelemetry spans into Elastic APM transactions and spans. The Elastic APM OpenTelemetry bridge allows you to use the vendor-neutral OpenTelemetry Tracing API to manually instrument your code and have the Elastic .NET APM agent handle those API calls. This means you can use the Elastic APM agent for tracing, without any vendor lock-in from adding manual tracing using the APM agent’s own <<public-api, Public API>>.
```

docs/api-opentelemetry.asciidoc
```diff
@@ -0,0 +1,121 @@
...
+
+OpenTelemetry in .NET is implemented via the https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity?view=net-5.0[Activity API].
+
+In .NET, there is an https://opentelemetry.io/docs/instrumentation/net/shim/[OpenTelemetry shim] which is built on top of the Activity API. 
+
+The OpenTelemetry bridge in the Elastic .NET APM Agent targets the https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity?view=net-5.0[Activity API]. Since the https://opentelemetry.io/docs/instrumentation/net/shim/[OpenTelemetry .NET shim] builds on top of the https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity?view=net-5.0[Activity API], the shim is implicitly supported as well, although we don't directly test it, because the Activity API is the recommended API.
```
bmorelli25, line 20: A couple of thoughts here.

1. This seems repetitive
2. Are these implementation details?

Can we move this to the end? Maybe a "how this feature works" section?

bmorelli25 (commented):
Looking good! Mostly grammar suggestions. Structure looks good—just one or two suggestions there.

docs/api-opentelemetry.asciidoc
gregkalapos: I moved it into the end.

What I wanted to say here is this:
- OTel has its predefined API (like `Tracer`, `Span`, etc.) - of course .NET does not follow this :D In .NET there is this thing called `Activity` which is the equivalent of a span, so the API (especially the name of specific types, and functions) for .NET OTel is different - things are naming differently and the .NET implementation doesn't follow the naming of the spec. (I know... not nice, but that's how it is). So I wanted to say that when we talk about the OTel bridge in .NET, what we mean is that we bridge the `Activity` API (which is de-facto OTel in .NET).
- To make things more complicated there is actually an alternative API that strictly follows the spec - that's the "shim" I talk about. That's built on top of `Activity`, and this shim is not recommended to use (the doc of it literally says "use `Activity` instead"). But I wanted to point out, that we indirectly support this shim as well.

> Are these implementation details?

So I would not say this is implementation detail - I'm just trying to communicate what API we actually support by the bridge.

I named the new section at the end `Supported OpenTelemetry implementations` - I think that better describes what this is.

> This seems repetitive

I changed the text a little bit.

docs/configuration.asciidoc
gregkalapos: Of course your vote counts @bmorelli25! Welcome back 👏 

Cool - so far, the plan is to change this into `ELASTIC_APM_OPENTELEMETRY_BRIDGE_ENABLED`. @z1c0 ok with that? If so, I'll update this PR to use the new name and open another one where I change it in the code as well.

gregkalapos: Discussed in the .NET weekly -> updated the config name to `ELASTIC_APM_OPENTELEMETRY_BRIDGE_ENABLED`.

docs/api-opentelemetry.asciidoc
```diff
@@ -0,0 +1,122 @@
...
+
+The OpenTelemetry bridge is part of the core agent package (https://www.nuget.org/packages/Elastic.Apm[`Elastic.Apm`]), so you don't need to add an additional dependency.
+
+[float]
+[[otel-enable-bridge]]
+==== Enable the OpenTelemetry Bridge
+
+Currently the bridge is opt-in, so you need to enable it.
+
+If you configured the agent via environment variables, set the `ELASTIC_APM_ENABLEOPENTELEMETRYBRIDGE` environment variable to `true``.
```
bmorelli25, line 27:
```suggestion
If you configured the agent via environment variables, set the `ELASTIC_APM_OPENTELEMETRY_BRIDGE_ENABLED ` environment variable to `true``.
```

docs/api-opentelemetry.asciidoc
bmorelli25: Aha! Thanks. This makes sense to me now. I like it at the end if you're okay with it there. And the rewording is great.

bmorelli25 (approved):
One env update otherwise looks good. Thanks, Greg!

z1c0 (approved):

PR #1972, opened 2023-01-12
Rename config to enable/disable OTel bridge

gregkalapos:
Fix the name of the config which enabled/disabled Otel Bridge.

Discussion: https://github.com/elastic/apm-agent-dotnet/pull/1966#discussion_r1060551833

Spec PR: https://github.com/elastic/apm/pull/743

Let's wait with merging until the spec PR is not merged.

### DIFF
--- a/src/Elastic.Apm/Config/AbstractConfigurationWithEnvFallbackReader.cs
+++ b/src/Elastic.Apm/Config/AbstractConfigurationWithEnvFallbackReader.cs
@@ -175,7 +175,7 @@ namespace Elastic.Apm.Config
 			ParseVerifyServerCert(Read(KeyNames.VerifyServerCert, EnvVarNames.VerifyServerCert));
 
 		public bool EnableOpenTelemetryBridge =>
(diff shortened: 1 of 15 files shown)


src/Elastic.Apm/Config/AbstractConfigurationWithEnvFallbackReader.cs
```diff
@@ -175,7 +175,7 @@ public virtual IReadOnlyList<Uri> ServerUrls
 			ParseVerifyServerCert(Read(KeyNames.VerifyServerCert, EnvVarNames.VerifyServerCert));
 
 		public bool EnableOpenTelemetryBridge =>
```
z1c0, line 177: Renaming the property would be a breaking change to the public API, right?
While it would be nice for consistency's sake, it's probably not worth the hassle.

z1c0 (approved):
Thanks for taking this! :shipit:

src/Elastic.Apm/Config/AbstractConfigurationWithEnvFallbackReader.cs
gregkalapos: Hmm - that's a good point, I have not thoughts about that part.

Yes, I agree with you, I'd keep it as is to minimize impact on users.

PR #1976, opened 2023-01-13
Set agent activation method

z1c0:
Set agent activation method for following scenarios:
* `nuget` (this is assumed to be the default unless we know better).
* `profiler`
* `startup-hook`

(see: [Activation method](https://github.com/elastic/apm/blob/main/specs/agents/metadata.md#activation-method))


src/Elastic.Apm/Api/ApiConstants.cs
```diff
@@ -29,5 +29,9 @@ public struct ApiConstants
 		public const string TypeStorage = "storage";
 		public const string TypeApp = "app";
 		public const string TypeUnknown = "unknown";
+
+		public const string ActivationMethodNuGet = "nuget";
```
gregkalapos, line 33: I'd move this to another place. The `ApiConstants` was meant to hold things that you pass to the public API - specifically transaction/span type/subtype/action. 

I think activation method is a bit different. I'd not even make these values public.

gregkalapos (commented):
Look ok to me, just one suggestion around `ApiConstants`.

As mentioned in the weekly meeting, let's not forget about the `webhook` in a follow up PR.

src/Elastic.Apm/Api/ApiConstants.cs
z1c0: Yes, makes sense. I'll move these constants somewhere else and tighten their visibility.

z1c0:
Since I had to update this PR anyway ([moving the constants](https://github.com/elastic/apm-agent-dotnet/pull/1976#pullrequestreview-1253937534)), I also added the K8S hook detection here.
Also, see the related PR: https://github.com/elastic/apm-mutating-webhook/pull/68

gregkalapos (approved):

PR #1986, opened 2023-01-24
Allow solution to be build on non windows platforms again

Mpdreamz:
Opening the solution in linux resulted in loading source files that should not be loaded under linux. 

The fix we did in https://github.com/elastic/apm-agent-dotnet/pull/426#issuecomment-542348863
did not extend to code under subfolders.


z1c0:
Interesting that this would be a problem on Linux but not on macOS 🤔.
I usually run the `./build.sh` on the command line which generates `Elastic.ApmAgent.NetCore.sln` that I then use in Rider.

Mpdreamz:
Ahh that makes sense!

I am hoping to kill `Elastic.ApmAgent.NetCore.sln` and the manual removal of projects in the Jenkins build from the solution.

z1c0 (approved):

PR #2013, opened 2023-02-22
Trace in-process Azure Functions invocations

z1c0:


src/Elastic.Apm.Azure.Functions/AzureFunctionsExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Elastic.Apm.DiagnosticSource;
+using Microsoft.Azure.Functions.Extensions.DependencyInjection;
+
+namespace Elastic.Apm.Azure.Functions
+{
+	public static class AzureFunctionsExtensions
+	{
+		public static IFunctionsHostBuilder AddElasticApm(this IFunctionsHostBuilder builder)
```
z1c0, line 14: - [ ] This definitely needs another iteration:
  * documentation
  * additional arguments (configuration, subscribers, logger, ...)

src/Elastic.Apm.Azure.Functions/AzureFunctionsDiagnosticSubscriber.cs
```diff
@@ -0,0 +1,39 @@
...
+// Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information
+
+using System;
+using Elastic.Apm.DiagnosticSource;
+using Elastic.Apm.Logging;
+
+namespace Elastic.Apm.Azure.Functions
+{
+	public class AzureFunctionsDiagnosticSubscriber : IDiagnosticsSubscriber
```
Mpdreamz, line 11: Why can't we reuse our normal AspNetCore subscriber here? https://github.com/elastic/apm-agent-dotnet/blob/31736e4a8ec3ff81cdac514f44f6b745488e5b9d/src/Elastic.Apm.AspNetCore/DiagnosticListener/AspNetCoreDiagnosticListener.cs#L19

@gregkalapos

Mpdreamz (changes requested):

src/Elastic.Apm.Azure.Functions/AzureFunctionsDiagnosticSubscriber.cs
gregkalapos: Very good point - I think we could.

We set `Transaction.FaaS` here, that needs to be done, but other than that I think this is the same as the ASP.NET Core one.

Issue #1999, opened 2023-01-30
NullReferenceException in span compression


gregkalapos:
Reported via https://discuss.elastic.co/t/elastic-apm-profiler-system-data-sqlclient-sqlcommand-error

### APM Agent version
`1.19.0`

### Environment

IIS + Profiler + `System.Data.SqlClient` 

### Callstack

```
[2023-01-30T14:30:17.8818747+01:00] [ERROR] exception whilst instrumenting integration <Elastic.Apm.Profiler.Managed.Integrations.AdoNet.CommandExecuteReaderWithBehaviorIntegration, System.Data.SqlClient.SqlCommand>
System.NullReferenceException: Object reference not set to an instance of an object.
at Elastic.Apm.Model.Span.TryToCompressRegular(Span sibling) in C:\Users\jenkins\workspace\net_apm-agent-dotnet-mbp_v1.19.0\apm-agent-dotnet\src\Elastic.Apm\Model\Span.cs:line 613
at Elastic.Apm.Model.Span.TryToCompress(Span sibling) in C:\Users\jenkins\workspace\net_apm-agent-dotnet-mbp_v1.19.0\apm-agent-dotnet\src\Elastic.Apm\Model\Span.cs:line 573
at Elastic.Apm.Model.Span.End() in C:\Users\jenkins\workspace\net_apm-agent-dotnet-mbp_v1.19.0\apm-agent-dotnet\src\Elastic.Apm\Model\Span.cs:line 504
at Elastic.Apm.Model.DbSpanCommon.EndSpan(ISpan span, IDbCommand dbCommand, Outcome outcome, Nullable1 duration) in C:\Users\jenkins\workspace\net_apm-agent-dotnet-mbp_v1.19.0\apm-agent-dotnet\src\Elastic.Apm\Model\DbSpanCommon.cs:line 77 at Elastic.Apm.Profiler.Managed.Integrations.AdoNet.DbSpanFactory1.EndSpan(ApmAgent agent, IDbCommand command, ISpan span, Exception exception) in C:\Users\jenkins\workspace\net_apm-agent-dotnet-mbp_v1.19.0\apm-agent-dotnet\src\Elastic.Apm.Profiler.Managed\Integrations\AdoNet\DbSpanFactory.cs:line 56
[...]

PR #2014, opened 2023-02-22
Attempt to fix nullref found in #1999

Mpdreamz:
- Fix different possible nullreference
- Fix possible nullref inside TryCompressRegular

I could not quite replicate the conditions for the nullref to trigger the same way as it did for #1999


z1c0 (approved):
Right, that looks like the culprit 👍.

PR #2017, opened 2023-02-23
Capture Azure Function Invoke Exception on Transaction

Mpdreamz:
- Fix snapshotty github action
- Capture the azure function delegate invoke exception on Transaction


src/Elastic.Apm.Azure.Functions/ApmMiddleware.cs
```diff
@@ -57,6 +57,7 @@ public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next
 			catch (Exception ex)
 			{
 				success = false;
+				t.CaptureException(ex);
```
z1c0, line 60: Thanks for catching that!

z1c0 (approved):

PR #2036, opened 2023-03-22
Remove ETW powered GC metrics on FullFramework

Mpdreamz:
This requires more than the documented `Performance Monitor Users`
privileges.

To read from ETW the user also needs to be part of the `Performance Log
Users` group.

However to start a new ETW session the user needs to be LocalSystem,
LocalService, NetworkService or part of the Local Admin group.

This goes against IIS best practices and not something we feel
comfortable telling our users to run their Application Pools under.

Since IIS is a big deployment target for full framework and errors
related to this metric collection come up alot in this context we opt to
pulling this feature out and making it a .NET core only feature.

Currently this metric is not used in the APM UI.

### DIFF
--- a/src/Elastic.Apm/Elastic.Apm.csproj
+++ b/src/Elastic.Apm/Elastic.Apm.csproj
@@ -78,7 +78,6 @@
     <PackageReference Include="System.Diagnostics.PerformanceCounter" Version="6.0.1" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="2.0.2" />
     <PackageReference Include="System.Threading.Tasks.Dataflow" Version="4.9.0" />
     <!-- Used by Ben.Demystifier -->
     <PackageReference Include="System.Reflection.Metadata" Version="5.0.0" />
--- a/src/Elastic.Apm/Helpers/PlatformDetection.cs
+++ b/src/Elastic.Apm/Helpers/PlatformDetection.cs
@@ -33,6 +33,9 @@ namespace Elastic.Apm.Helpers
 			RuntimeInformation.FrameworkDescription.Length >= 6
 			&& char.IsDigit(RuntimeInformation.FrameworkDescription[5]);
 
+		/// <summary> Checks if runtime is .NET Core or .NET 6 and up </summary>
+		internal static readonly bool IsModernDotnet = IsDotNetCore || IsDotNet;
+
 		internal static readonly string DotNetRuntimeDescription = RuntimeInformation.FrameworkDescription;
 
 		internal static string GetDotNetRuntimeVersionFromDescription(string frameworkDescription, IApmLogger loggerArg, string expectedPrefix)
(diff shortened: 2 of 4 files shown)


src/Elastic.Apm/Elastic.Apm.csproj
```diff
@@ -78,7 +78,6 @@
     <PackageReference Include="System.Diagnostics.PerformanceCounter" Version="6.0.1" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="2.0.2" />
```
gregkalapos, line 81: It's good to get rid of this. 👍

src/Elastic.Apm/Helpers/PlatformDetection.cs
```diff
@@ -33,6 +33,9 @@ internal static class PlatformDetection
 			RuntimeInformation.FrameworkDescription.Length >= 6
 			&& char.IsDigit(RuntimeInformation.FrameworkDescription[5]);
 
+		/// <summary> Checks if runtime is .NET Core or .NET 6 and up </summary>
```
gregkalapos, line 36: nit: `IsDotNetCore` is true for .NET 5 as well.

```suggestion
		/// <summary> Checks if runtime is .NET Core or .NET 5 and up </summary>
```

src/Elastic.Apm/Metrics/MetricsProvider/GcMetricsProvider.cs
```diff
@@ -68,54 +58,18 @@ public GcMetricsProvider(IApmLogger logger, IReadOnlyList<WildcardMatcher> disab
...
-					});
-
-					_traceEventSessionTask = Task.Run(() => _traceEventSession.Source.Process());
-				}
-				catch (Exception e)
-				{
-					_logger.Warning()?.LogException(e, "TraceEventSession initialization failed - GC metrics won't be collected");
-					return;
-				}
+				_logger.Info()?.Log("Gc metrics are only available on .NET Core, disabling metric collection");
```
gregkalapos, line 67:
```suggestion
				_logger.Info()?.Log("GC metrics are only available on .NET Core, disabling metric collection");
```

Naming makes this so hard... We'd print `...only available on .NET Core...`, but we actually mean .NET Core and .NET5+. And over time due to the naming change .NET Core is less relevant, and .NET X will be the only supported one. I don't know how to address this. Should we maybe print `.NET Core and .NET 5+` ?

src/Elastic.Apm/Metrics/MetricsProvider/GcMetricsProvider.cs
```diff
@@ -68,54 +58,18 @@ public GcMetricsProvider(IApmLogger logger, IReadOnlyList<WildcardMatcher> disab
...
-				// unbounded, leading to the application consuming ever more memory.
-				if (!TraceEventProviderOptions.FilteringSupported)
-				{
-					_logger.Info()?.Log("TraceEventSession does not support filtering on this operating system. GC metrics won't be collected");
-					return;
-				}
-
-				try
-				{
-					TraceEventSessionName = SessionNamePrefix + Guid.NewGuid();
```
gregkalapos, line 89: `SessionNamePrefix` still defined above (can't add suggestion, because that line is unchanged). We should remove it.

gregkalapos (commented):
Thanks, looks nice. A few minor things below.

gregkalapos (approved):

Mpdreamz:
Two test failures are unrelated merging.

PR #2037, opened 2023-03-22
Clarify 64 bit only support in the profiler docs.

gregkalapos:
As discussed in our weekly meeting - to avoid misunderstandings we clearly state that 32 bit is not supported by the profiler based agent.

### DIFF
--- a/docs/setup-auto-instrumentation.asciidoc
+++ b/docs/setup-auto-instrumentation.asciidoc
@@ -31,7 +31,9 @@ This approach works with the following
 .NET 5+
 |===
 
-and instruments the following assemblies
+NOTE: The Profiler based agent only supports 64 bit applications. 32 bit applications aren't supported by our builds.
+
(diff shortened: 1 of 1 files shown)


docs/setup-auto-instrumentation.asciidoc
```diff
@@ -31,7 +31,9 @@ This approach works with the following
 .NET 5+
 |===
 
-and instruments the following assemblies
+NOTE: The Profiler based agent only supports 64 bit applications. 32 bit applications aren't supported by our builds.
```
Mpdreamz, line 34:
```suggestion
NOTE: The Profiler based agent only supports 64 bit applications. 32 bit applications aren't supported.

```

Mpdreamz (approved):
LGTM left one suggestion.  👍

bmorelli25 (approved):

PR #2040, opened 2023-04-01
Move IIS tests to github actions

Mpdreamz:
- move IIS tests over to github actions


Mpdreamz:
Note for prosperity this consistently fails with `7.0.302` this needs a lower version e.g `7.0.203`

Issue #2038, opened 2023-03-29
[BUG] Elastic.Apm.AspNetFullFramework does not capture SQL operations


ddgpablo:
### APM Agent version

APM 8.6

### Environment

.NET Framework 4.7.2

### Describe the bug

Elastic Apm is not able to automatically capture SqlClient operations.

### To Reproduce

To set up elastic APM:

```
            var agentComponents = ElasticApmModule.CreateAgentComponents();
            Agent.Setup(agentComponents);
            Agent.Subscribe(new SqlClientDiagnosticSubscriber());
```

Then when one of the endpoints is hit, we introduced the following code to test if it's able to capture SQL operations automatically:

```
        private void TestSql()
        {
            string queryString = "SELECT Email FROM [dotmailer].[dbo].[UserAccounts]";
            string connectionString = "YOUR_CONECTION_STRING";
            using (SqlConnection connection = new SqlConnection(connectionString))
            {
                SqlCommand command = new SqlCommand(queryString, connection);
                connection.Open();
                SqlDataReader reader = command.ExecuteReader();
                try
                {
                    while (reader.Read())
                    {
                        Console.WriteLine(reader["Email"]);// etc
                    }
                }
                finally
                {
                    // Always call Close when done reading.
                    reader.Close();
                }
            }
        }
```

#### Expected behavior
We expect to see a transaction which contains a span for the SQL operation.

#### Actual behavior
We see a transaction without any span.

gregkalapos:
Hi @ddgpablo, thanks for reporting this.

Could you please attach agent logs on `trace` level? Docs on how to collect logs can be found [here](https://www.elastic.co/guide/en/apm/agent/dotnet/current/troubleshooting.html#collect-logs-classic).

sksk571:
The important detail is that we use Microsoft.Data.SqlClient library in a .NET Framework project.

The code in [SqlEventListener](https://github.com/elastic/apm-agent-dotnet/blob/main/src/Elastic.Apm.SqlClient/SqlEventListener.cs#L31) mentions issue https://github.com/dotnet/SqlClient/issues/436 that has been fixed more than 3 years ago. Should SqlEventListener be updated to use the new event source name for Microsoft.Data.SqlClient?

ddgpablo:
This is the current error I'm getting from `OnEventWritter`:

```
[...]

PR #2050, opened 2023-04-17
Enable listening to Microsoft.Data.SqlClient.EventSource on .NET full framework

Mpdreamz:
This should fix https://github.com/elastic/apm-agent-dotnet/issues/2038

Still working on reproducing this.

### DIFF
--- a/sample/Sample.Microsoft.Data.SqlClient/Program.cs
+++ b/sample/Sample.Microsoft.Data.SqlClient/Program.cs
@@ -1,0 +1,62 @@
+using Elastic.Apm;
+using Elastic.Apm.Logging;
+using Elastic.Apm.SqlClient;
+using Elastic.Apm.Tests.Utilities;
+using Microsoft.Data.SqlClient;
+using Testcontainers.MsSql;
+
+var agentConfig = new MockConfiguration(centralConfig: "false");
+Agent.Setup(new AgentComponents(ConsoleLogger.LoggerOrDefault(LogLevel.Trace), agentConfig));
+Agent.Subscribe(new SqlClientDiagnosticSubscriber());
+using var instance = Agent.Instance;
+
+Console.WriteLine("Starting mssql container");
+await using var msSql = new MsSqlBuilder().Build();
+await msSql.StartAsync();
+var connectionString = msSql.GetConnectionString();
+
+Console.WriteLine("Performing database calls");
+
+var transaction = Agent.Tracer.StartTransaction("test-sql-client-calls", "app");
+DoSqlClientCall(connectionString);
+transaction.End();
+
+Console.WriteLine("Done!");
+
+static void DoSqlClientCall(string connectionString)
+{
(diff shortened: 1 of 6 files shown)


gregkalapos (approved):

ddgpablo:
Hello,

To reproduce this issue you just need to create a .NETFramework 4.7.2 web app and perform a SQL operation when hitting one of the endpoints. 

I tested this fix and it's still not creating the spans due to these checks:

https://github.com/elastic/apm-agent-dotnet/blob/20fea7888fd3a044e13fb34c3e657d26502a6911/src/Elastic.Apm.SqlClient/SqlEventListener.cs#L75

https://github.com/elastic/apm-agent-dotnet/blob/20fea7888fd3a044e13fb34c3e657d26502a6911/src/Elastic.Apm.SqlClient/SqlEventListener.cs#L121

The payload looks to be always 5 in `ProcessBeginExecute` and 4 in `ProcessEndExecute`. Is this specific for .NET Framework? Should we alter the count check?

Kind regards,
Pablo

Mpdreamz:
> The payload looks to be always 5 in ProcessBeginExecute and 4 in ProcessEndExecute. Is this specific for .NET Framework? Should we alter the count check?

Thanks for reporting this @ddgpablo found the same while working on a reproduction. 
This took a bit longer because I was fighting: https://github.com/dotnet/SqlClient/issues/801 which was caused because I was refencing `net461` instead of `net462`. 

Our payload checks are too stringent, `SqlClientEventSource` reserves the first sequence of the payload but MAY add more tokens to the payload in the future see:

https://github.com/dotnet/SqlClient/blob/3a41288f2b67307a1f816761deb73785247c85c9/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlClientEventSource.cs#L1009

ddgpablo:
Many thanks for your answer @Mpdreamz. Is the PR ready to review? Let me know and I can test it again.

Mpdreamz:
@ddgpablo it is!

ddgpablo:
Great, that works! Many thanks for your help mate.

sample/Sample.Microsoft.Data.SqlClient/Program.cs
```diff
@@ -0,0 +1,62 @@
...
+
+Console.WriteLine("Performing database calls");
+
+var transaction = Agent.Tracer.StartTransaction("test-sql-client-calls", "app");
+DoSqlClientCall(connectionString);
+transaction.End();
+
+Console.WriteLine("Done!");
+
+static void DoSqlClientCall(string connectionString)
```
gregkalapos, line 26: Could we move this into a test? From what I understand, we do not run this currently anywhere. I think `Elastic.Apm.SqlClient.Tests` would be a better place for this.

gregkalapos (approved):
LGTM, just one thing about testing.

PR #2051, opened 2023-04-17
Profiler docs: add note about server url and secret token

gregkalapos:
Based on feedback from multiple real-world confusions - adding a note on how to configure server url and secret token with the profiler.


Mpdreamz (approved):
LGTM :+1:

PR #2054, opened 2023-04-19
fix/eager config load

Mpdreamz:
Removes all code from the getters and eagerly loads configuration in the constructors.

 This change was sparked by the stackoverflow discovered in ECS.NET where we log ServiceVersion but the getter itself also logs: https://github.com/elastic/ecs-dotnet/issues/292

 This further simplifies lazy caching (no need for it anymore) and removes all lusages of our own `LazyContextualInit` construct in the configuration initialization. 

- Rename FullFrameworkConfigReader to ElasticApmModuleConfiguration
- Rename MicrosoftExtensionsConfig to ApmConfiguration

- Merge ConfigurationSnapshotFromReader and WrappingConfiguration into one `RuntimeConfigurationSnapshot`


gregkalapos (commented):
I think the overall change is ok - especially the eager config loading part as discussed in the last weekly.

The only thing I'm a bit skeptical about is the renaming of the configs:

> Rename FullFrameworkConfigReader to ElasticApmModuleConfiguration

This one I see fairly neutral 

> Rename MicrosoftExtensionsConfig to ApmConfiguration

I think the `ApmConfiguration` is a very generic name - everything is APM Configuration in the config related things in the APM Agent. 

However, the namespace clearly tells what that config is for.

So I'd suggest we take one of 2 approaches for all concrete implementations:
- either the class name says what the specific config implementation is for (e.g. keep `MicrosoftExtensionsConfig`) 
- or just use a generic name for all config implementation (like `ApmConfiguration`, or just `Configuration`), but have a namespace for it, which reflects the purpose of the class (e.g. `Elastic.Apm.AspNetFullFramework.System.Configuration` and `Elastic.Apm.Extensions.Hosting.Config`).

Mpdreamz:
> Rename FullFrameworkConfigReader to ElasticApmModuleConfiguration

I did this primarily because this config is **only** used by the `ElasticApmModule` and this more clearly indicates that. 

> Rename MicrosoftExtensionsConfig to ApmConfiguration

We don't quite follow the [Options pattern](https://learn.microsoft.com/en-us/dotnet/core/extensions/options) but `ApmOptions` to me would have been the best name if we did. Since it models the configuration section for `Apm`. 

`ApmConfiguration` is to me therefor the closest alternative. `MicrosoftExtensionsConfig` to me was too generic :smile_cat: in that in that it does not communicate it encapsulates APM configuration data.

PR #2069, opened 2023-05-16
Ensure Agent.Setup on .NET FullFramework behaves the same

Mpdreamz:
Ensures that on .NET FullFramework we always use ConfigurationManager. 

Right now this is only the when you use our `ElasticApmModule` on IIS. 

By moving the basis of this configuration to `Elastic.Apm` we can switch the default `IConfigurationReader` on .NET full framework from `EnvironmentConfiguration` to `AppsettingsConfiguration`


stevejgordon (approved):
LGTM

PR #2074, opened 2023-05-23
remove all stages from JenkinsFile

Mpdreamz:


Mpdreamz:
Less rigorous as https://github.com/elastic/apm-agent-dotnet/pull/2073 there are still a few updatecli.d configurations that need to be ported to github actions

PR #2109, opened 2023-06-06
Agent.Configuration now always points to ConfigurationStore's configuration

Mpdreamz:
This ensure no matter how you get to the configuration you will always get the latest version e.g one that is possibly updated through remote config.

This also deprecates `.ConfigurationReader` property on `IApmAgent` in favor of `.Configuration`.


gregkalapos (approved):
Looks ok to me.

One thing I was thinking about when I looked at this: one concept we followed in the .NET Agent (and most other agents do this as well) is that a config snapshot is immutable for a given transaction. Meaning the config system takes all the values at the beginning of the transaction and those don't change until the end of the transaction. If central config changes during the transaction, then the new value only applies to the next transaction. The idea behind this is simply consistency and to decrease complexity (it's much easier to not deal with question like: what if I disable the agent in the middle of a transaction? What if I disable stack trace capturing in the middle of a transaction? etc. - the answer to all those is that the already started transaction always finishes with the original configsnapshot).

From [what I see](https://github.com/elastic/apm-agent-dotnet/blob/main/src/Elastic.Apm/Api/Tracer.cs#L79) that's still true. We should just make sure to only access `Agent.Configuration` directly if it's not something which should be immutable during a transaction.

Sorry for the noise here, just wanted to share this.

PR #2128, opened 2023-07-05
isolate azure tests

Mpdreamz:


.github/workflows/test-linux.yml
```diff
@@ -67,8 +64,56 @@ jobs:
...
+        with:
+          url: ${{ secrets.VAULT_ADDR }}
+          roleId: ${{ secrets.VAULT_ROLE_ID }}
+          secretId: ${{ secrets.VAULT_SECRET_ID }}
+          method: approle
+          secrets: |
+            secret/apm-team/ci/apm-agent-dotnet-azure client_id | ARM_CLIENT_ID ;
+            secret/apm-team/ci/apm-agent-dotnet-azure client_secret | ARM_CLIENT_SECRET ;
+            secret/apm-team/ci/apm-agent-dotnet-azure subscription_id | ARM_SUBSCRIPTION_ID ;
+            secret/apm-team/ci/apm-agent-dotnet-azure tenant_id | ARM_TENANT_ID
```
kuisathaverat, line 96: This will work only on feature branches, secrets are not accessible from forks 

cc @elastic/ci-robots

stevejgordon (approved):
LGTM

Mpdreamz:
The test reporter needs to be tweaked slightly,

Currently any test suite that is no match will still emit a xunit log file with no tests.

https://github.com/elastic/apm-agent-dotnet/runs/14863078120

We should add a step that removes these prior to the upload so that we don't end up overwriting  a previous valid run. 

Will tackle this in a separate PR though.

PR #2147, opened 2023-08-03
Add Baggage support

gregkalapos:
Solves https://github.com/elastic/apm-agent-dotnet/issues/2001

- We integrate with `Activity.Baggage` - Activity takes care of parsing the incoming `baggage` header, so we take advantage of that.
- Unfortunately for outgoing HTTP request, activity does not use the `baggage` header, so this PR also attaches a `baggage` header to outgoing HTTP request based on `Activity.Baggage`.
- Adds config~s~ `BaggageToAttach` ~`BaggageToAttachOnTransactions`, `BaggageToAttachOnSpans`, `BaggageToAttachOnErrors`~ including docs and tests.
- Tiny refactoring for ASP.NET Core tests: Moved the general part of `DistributedTracingAspNetCoreTests` into `MultiApplicationTestBase`, so we can reuse it, since the setup of the new baggage tests are very similar to the existing distributed tracing tests. `MultiApplicationTestBase` is now the base for `DistributedTracingAspNetCoreTests` and the new `BaggageAspNetCoreTests`.


Mpdreamz:
> Unfortunately for outgoing HTTP request, activity does not use the baggage header, so this PR also attaches a baggage header to outgoing HTTP request based on Activity.Baggage.

Are we 100% sure about this? I would have expected the propagators to do this ootb:

https://github.com/dotnet/runtime/issues/50658

See e.g: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/DistributedContextPropagator.cs#L109C30-L109C43

Mpdreamz:
Extending from previous comment should we enforce a global `DistributedContextPropagator`?

https://learn.microsoft.com/en-us/dotnet/api/System.Diagnostics.DistributedContextPropagator.Current?view=net-7.0

Or even a local one on `SocketsHttpHandler` through our AcitivityListener.

gregkalapos:
> > Unfortunately for outgoing HTTP request, activity does not use the baggage header, so this PR also attaches a baggage header to outgoing HTTP request based on Activity.Baggage.
> 
> Are we 100% sure about this? I would have expected the propagators to do this ootb:
> 
> [dotnet/runtime#50658](https://github.com/dotnet/runtime/issues/50658)
> 
> See e.g: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/DistributedContextPropagator.cs#L109C30-L109C43

Yeah, looking at those actually verifies what I observed during testing.

The [`setter`](https://github.com/dotnet/runtime/blob/b35572540e0904c6a23dc5466f73c361ed877dc1/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/DistributedContextPropagator.cs#L123C21-L123C106) still uses the old header name `CorrelationContext`:

```
setter(carrier, CorrelationContext, baggageList.ToString(0, baggageList.Length - 2));
``` 

which is [later defined](https://github.com/dotnet/runtime/blob/b35572540e0904c6a23dc5466f73c361ed877dc1/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/DistributedContextPropagator.cs#L132) as:

```
internal const string CorrelationContext = "Correlation-Context";
```

In the [linked issue](https://github.com/dotnet/runtime/issues/50658) they also reason on why:
> As an example, the W3C baggage specification changed the name of the header from Correlation-Context to Baggage. We'd need a way to change to version and change propagation behavior without requiring every library to react to spec changes

and then:
> * ~Add implementation for W3C propagator W3C baggage specs still not finalized.~ we'll wait to expose this propagator till the specs are final.

This is still all true - the code is the same with the old header name on `main` and the W3C baggage is still `W3C Working Draft`.


> Extending from previous comment should we enforce a global `DistributedContextPropagator`?
> https://learn.microsoft.com/en-us/dotnet/api/System.Diagnostics.DistributedContextPropagator.Current?view=net-7.0
> Or even a local one on `SocketsHttpHandler` through our AcitivityListener.

Hmm - that's worth discussing, but I think currently it would not change much. Right now we manually add the `baggage` header and we can't rely on this because it uses the old header name. Once they use the final header name (likely `baggage`) then this will be more relevant.

Mpdreamz:
> Hmm - that's worth discussing, but I think currently it would not change much. Right now we manually add the baggage header and we can't rely on this because it uses the old header name. Once they use the final header name (likely baggage) then this will be more relevant.

Discussed offline, we'll go with relying on our own approach for now while 

We will never touch DistributedContextPropagator but will up our logging of whats configured.
https://github.com/elastic/apm-agent-dotnet/issues/2155

gregkalapos:
Adapted the PR to the latest spec changes. This is now ready to review.

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
```diff
@@ -98,6 +98,28 @@ protected IReadOnlyList<WildcardMatcher> ParseSanitizeFieldNames(ConfigurationKe
...
+			=> ParseWildcardMatcher(kv?.Value ?? DefaultValues.BaggageToAttach, "BaggageToAttach");
+
+		private IReadOnlyList<WildcardMatcher> ParseWildcardMatcher(string stringValue, string configName)
+		{
+			try
+			{
+				_logger?.Trace()?.Log("Try parsing {ConfigName}, values: {Values}", configName, stringValue);
+				var values = stringValue.Split(',').Where(n => !string.IsNullOrEmpty(n)).ToList();
+
+				var retVal = new List<WildcardMatcher>(values.Count);
```
Mpdreamz, line 111:
```suggestion
				var matchers = new List<WildcardMatcher>(values.Count);
```

This is nitpicky but `retVal` does not name what we are actually returning.

Mpdreamz (approved):
LGTM I have one small nitpick though.

👍  to `MultiApplicationTestBase`

PR #2104, opened 2023-06-06
[WIP] Backend dependencies granularity for NoSQL and Messaging

stevejgordon:
Fixes #1754

PR #2148, opened 2023-08-04
Move Redis Profiler registration to ConditionalWeakTable

Mpdreamz:
This should ensure we don't leak our registrations if span/transaction's Ended
event never gets called but the spans/transaction themselves get GC'ed.

### DIFF
--- a/src/instrumentations/Elastic.Apm.StackExchange.Redis/ElasticApmProfiler.cs
+++ b/src/instrumentations/Elastic.Apm.StackExchange.Redis/ElasticApmProfiler.cs
@@ -8,6 +8,7 @@ using System.Collections.Concurrent;
 using System.Globalization;
 using System.Net;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 using Elastic.Apm.Api;
 using Elastic.Apm.Helpers;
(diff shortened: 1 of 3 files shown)


src/instrumentations/Elastic.Apm.StackExchange.Redis/ElasticApmProfiler.cs
```diff
@@ -22,29 +23,10 @@ namespace Elastic.Apm.StackExchange.Redis
 	/// </summary>
 	public class ElasticApmProfiler
 	{
-		private readonly ConcurrentDictionary<string, ProfilingSession> _executionSegmentSessions = new();
+		private readonly ConditionalWeakTable<IExecutionSegment, ProfilingSession> _executionSegmentSessions = new();
 
 		private readonly Lazy<IApmLogger> _logger;
 		private readonly Lazy<IApmAgent> _agent;
-		private static readonly Func<object, object> MessageFetcher;
```
Mpdreamz, line 29: This is a continuation of https://github.com/elastic/apm-agent-dotnet/pull/2104 since we no longer set `db.instance` cc @stevejgordon

gregkalapos (approved):

PR #2150, opened 2023-08-07
Extend release action to automatically upload artifacts to release.

Mpdreamz:
Uploads the profiler and startup hooks from the linux build.

Adds a windows build after the full release runs to also attach the
windows profiler build to the release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -57,7 +57,7 @@ jobs:
     - name: Attach Profiler And Startup Hooks
       continue-on-error: true #continue for now until we see it working in action
       run: |
-        gh release upload ${{ github.ref_name }} "ElasticApmAgent_*.zip" "elastic_apm_profiler_*.zip"
+        gh release upload ${{ github.ref_name }} "build/output/ElasticApmAgent_*.zip" "build/output/elastic_apm_profiler_*.zip"
 
     - if: ${{ success() }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -76,3 +81,23 @@ jobs:
         message: |
           :large_yellow_circle: [${{ github.repository }}] Release *${{ github.ref_name }}* could not be published.
           Build: (<${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|here>)
+          
+  release-windows:
+    runs-on: ubuntu-latest
```
stevejgordon, line 86: Should this run on a Windows image?

Mpdreamz: Doh!

.github/workflows/release.yml
```diff
@@ -76,3 +81,23 @@ jobs:
...
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Bootstrap Action Workspace
+        uses: ./.github/workflows/bootstrap
+        with:
+          rust: 'true'
+
+      - name: Package
+        run: ./build.sh pack
```
stevejgordon: Do we need to switch to build.bat here now

stevejgordon (approved):
LGTM

PR #2153, opened 2023-08-08
Add more explicit logging to ApmAgent.Subscribe.

Mpdreamz:
This so that it shows up earlier during the agent setup and not when diagnosticlisteners start firing


stevejgordon (approved):
LGTM

PR #2159, opened 2023-08-16
Fix docker build during release

Mpdreamz:
The previous release failed early because docker push was called without building the image prior. 

This embeds the docker commands and removes the scripts from `.ci` which is a folder we want to get rid of near short term. 

This also extends our bootstrap action to output the current agent version.


.github/workflows/release.yml
```diff
@@ -47,12 +48,12 @@ jobs:
...
+    - name: Build/Push Profiler Docker Image
+      continue-on-error: true #continue for now until we see it working in action
       run: |
-        chmod +x .ci/linux/push_docker.sh
-        .ci/linux/push_docker.sh
-      env:
-        TAG_NAME: ${{ github.ref_name }}
+        docker build . -t docker.elastic.co/observability/apm-agent-dotnet:${{ steps.bootstrap.outputs.agent-version }} \
+            --build-arg AGENT_ZIP_FILE=build/output/elastic_apm_profiler_${{ steps.bootstrap.outputs.agent-version }}-linux-x64.zip
+        docker push docker.elastic.co/observability/apm-agent-dotnet:${{ steps.bootstrap.outputs.agent-version }}
```
v1v, line 56: I'd say to use two different steps, one for building and another one for pushing. So it can retry the pushing step if it failed - sometimes docker.elastic.co can be down or failed.

Mpdreamz: done!

Mpdreamz: Although steps can not be retried in isolation from the UI, only jobs.

.github/workflows/release.yml
```diff
@@ -47,12 +48,15 @@ jobs:
...
+        docker build . -t docker.elastic.co/observability/apm-agent-dotnet:${{ steps.bootstrap.outputs.agent-version }} \
+            --build-arg AGENT_ZIP_FILE=build/output/elastic_apm_profiler_${{ steps.bootstrap.outputs.agent-version }}-linux-x64.zip
+    - name: Push Profiler Docker Image
+      continue-on-error: true #continue for now until we see it working in action
       run: |
-        chmod +x .ci/linux/push_docker.sh
-        .ci/linux/push_docker.sh
-      env:
-        TAG_NAME: ${{ github.ref_name }}
+        docker push docker.elastic.co/observability/apm-agent-dotnet:${{ steps.bootstrap.outputs.agent-version }}
```
v1v, line 59: what about using something like this
```suggestion
        for i in $(seq 1 3); do [ $i -gt 1 ] && sleep 15; docker push docker.elastic.co/observability/apm-agent-dotnet:${{ steps.bootstrap.outputs.agent-version }} && s=0 && break || s=$?; done; (exit $s)
```

I took the loop/sleep/retry from [here](https://unix.stackexchange.com/a/82610)

Mpdreamz: Love it!

v1v (approved):

PR #2171, opened 2023-08-30
Elastic.Apm.Ingest, new PayloadSender implementation

Mpdreamz:
This introduces `ApmChannel` an `Elastic.Ingest.Tranport` implementation geared to APM Servers v2 intake API. 

Opening this early to iterate on as time allows. 

`MockPayloadSender` is now a subclass of `ApmChannel` so it will attempt to do the same work as it does.

Still need to work on ensuring the data that will get written is correct.

### DIFF
--- a/build/scripts/Build.fs
+++ b/build/scripts/Build.fs
@@ -21,6 +21,7 @@ open Tooling
 
 module Build =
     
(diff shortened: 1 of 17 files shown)


stevejgordon:
Closing as unplanned

PR #2178, opened 2023-09-07
DroppedSpanStats: do not flatten duration

gregkalapos:
APM Server expects objects and flattening those as the agent did prior to this PR is not accepted by the server.

Furthermore, in APM Server  8.9.x, there was an issue that in such cases (specifically when the [`duration`](https://github.com/elastic/apm-server/blob/79533702ef0e73b9dd9c6684cd83f78926aba6f6/docs/spec/v2/transaction.json#L735) field was empty (or sending it flattened)) caused a panic. This was fixed in 8.10, but the APM Server would still ignored the flattened `duration` object on `dropped_spans_stats`.

Therefore, this PR changes `DroppedSpanStats.cs` to use sub-objects for the `Duration` object. We do this in all other cases as well.

### DIFF
--- a/src/Elastic.Apm/Model/Transaction.cs
+++ b/src/Elastic.Apm/Model/Transaction.cs
@@ -571,8 +571,10 @@ namespace Elastic.Apm.Model
 
 				if (_droppedSpanStatsMap.TryGetValue(new DroppedSpanStatsKey(serviceTargetType, serviceTargetName, outcome), out var item))
 				{
-					item.DurationCount++;
-					item.DurationSumUs += duration;
+					item.Duration ??= new DroppedSpanStats.DroppedSpanDuration {Sum = new DroppedSpanStats.DroppedSpanDuration.DroppedSpanDurationSum()};
+
+					item.Duration.Count++;
+					item.Duration.Sum.Us += duration;
 				}
 				else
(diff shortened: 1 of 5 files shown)


src/Elastic.Apm/Model/Transaction.cs
```diff
@@ -21,943 +21,945 @@
 using Elastic.Apm.Report;
 using Elastic.Apm.ServerInfo;
 
-namespace Elastic.Apm.Model
+namespace Elastic.Apm.Model;
+
+internal class Transaction : ITransaction
```
gregkalapos, line 26: Changes in this file are minimal, but due to my usual fight with `dotnet format` I hit full reformat on it (after going through multiple fails). Now it seems to be ok, but there are a bunch of formatting changes here. Sorry about that.

Mpdreamz: Reviewing with `?w=1` always helps in these cases :) Let me know if you want to pair over why `dotnet format` is acting up. It might be because your checkout predates some of the cleanup.

Mpdreamz (approved):
LGTM, this looks complete but is still in draft?

Mpdreamz (approved):

PR #2185, opened 2023-09-19
action: override git checks when only docs

v1v:
This should avoid the situation that we have when a docs changes are blocked by

image


.github/workflows/test-docs.yml
```diff
@@ -0,0 +1,33 @@
...
+  pack:
+    runs-on: ubuntu-latest
+    steps:
+      - run: 'echo "Not required for docs"'
+
+  # dummy steps that allow to bypass those mandatory checks for tests
+  tests:
+    runs-on: ubuntu-latest
+    steps:
+      - run: 'echo "Not required for docs"'
```
reakaleek, line 33:
```suggestion
      - run: 'echo "Not required for docs"'

```

reakaleek (dismissed):

reakaleek (dismissed):

.github/workflows/test-docs.yml
```diff
@@ -0,0 +1,33 @@
+# This workflow sets the test / pack and test / tests status check to success in case it's a docs only PR and test-linux.yml is not triggered
```
reakaleek, line 1:
```suggestion
# This workflow sets the 'test-linux / pack' and 'test-linux / tests' status check to success in case it's a docs only PR and test-linux.yml is not triggered
```

.github/workflows/test-docs.yml
```diff
@@ -0,0 +1,33 @@
...
+# https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks
+name: test-linux # The name must be the same as in test.yml
```
reakaleek, line 3:
```suggestion
name: test-linux # The name must be the same as in test-linux.yml
```

reakaleek:
mhmm. Looking at the created status checks, there is also `test-windows / tests`

v1v:
> Which now might run erroneously? e.g snapshotty.

That's a good point. In this particular case, it's not necessary since this is trying to solve the GitHub checks in the branch protection. 

We might need to revisit the current implementation so snapshoty or similar ones run only when needed, aka when everything run successfully. At the moment, the CI workflows, in this particular project, are a bit decentralised. I'll keep a note so we can review the implementation.

Mpdreamz:
Opened https://github.com/elastic/apm-agent-dotnet/pull/2186 to attempt to make `cargo make` happy. 

See also: https://github.com/sagiegurari/cargo-make/issues/67

Mpdreamz (approved):

PR #2191, opened 2023-09-21
action: use GH_TOKEN

v1v:
This should fix the error when using gh

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -61,7 +61,7 @@ jobs:
     - name: Attach Profiler And Startup Hooks
       continue-on-error: true #continue for now until we see it working in action
       env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


Mpdreamz (approved):
GITHUB_TOKEN and GH_TOKEN should be equivalent according to: https://cli.github.com/manual/gh_help_environment

Added that env just after our latest release. Happy to settle/standardize on GH_TOKEN though!

Let me know if I missed anything.

v1v:
> GITHUB_TOKEN and GH_TOKEN

That was my understanding too... 


I saw it failed recently in https://github.com/elastic/apm-agent-dotnet/actions/runs/6249887768/job/16967549984#step:10:17

```
gh: To use GitHub CLI in a GitHub Actions workflow, set the GH_TOKEN environment variable. Example:
  env:
    GH_TOKEN: ${{ github.token }}
```

That's the reason for this PR

PR #2193, opened 2023-10-05
json-specs: cgroup_parsing has been removed

v1v:
## What

I created an empty json file so the `updatecli` will be the one updating it in the next run.

## Why

Fix the failures when updating the json-specs using `updatecli`

## Further details

Likely `test/Elastic.Apm.Tests/SystemInfoHelperTests.cs` will need to be adjusted since `cgroup_parsing.json` does not exist anymore in the `elastic/apm.git`

## Issues

Caused by https://github.com/elastic/apm/pull/807


Mpdreamz (approved):
LGTM

PR #2194, opened 2023-10-10
Document IIS minimum requirements.

Mpdreamz:
And add the most common errors we see due to these restrains not being met under our troubleshooting section


stevejgordon (approved):
LGTM

PR #2196, opened 2023-10-10
Support OnExecuteRequestStep available in new .NET versions for IIS modules.

Mpdreamz:
This more aggressively logs and attempts to restore HttpContext items.

This also more gracefully exits if we detect the module running under classic pipeline mode


stevejgordon (dismissed):
LGTM

stevejgordon (approved):
LGTM

PR #2201, opened 2023-10-17
action: post-release steps

v1v:
### What

Automate the post-release steps so the `<major>.x` gets updated through a Pull Request. Create `<major>.x` if it's not available.

### Why

More automation that follows the github branch protection.

### Further details

I need to figure out how to test this out.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -131,6 +131,7 @@ jobs:
 
       - name: Create GitHub Pull Request if minor release.
         run: |
+          # as long as there is a major.x branch
           existed_in_local=$(git ls-remote --heads origin ${TARGET_BRANCH})
           if [ -z ${existed_in_local} ]; then
             echo "Target branch '${TARGET_BRANCH}' does not exist."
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -10,6 +10,8 @@ permissions:
 
 env:
   NUGET_PACKAGES: ${{ github.workspace }}/.nuget/packages
+  JOB_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+  SLACK_CHANNEL: "#apm-agent-dotnet"
```
v1v, line 14: Unrelated, just a refactoring

.github/workflows/release.yml
```diff
@@ -108,3 +114,45 @@ jobs:
...
+
+      - name: Create GitHub Pull Request if minor release.
+        run: |
+          echo "as long as there is a major.x branch"
+          existed_in_local=$(git ls-remote --heads origin ${TARGET_BRANCH})
+          if [ -z ${existed_in_local} ]; then
+            echo -e "::warning::Target branch '${TARGET_BRANCH}' does not exist."
+            exit 0
+          fi
+          git checkout $GIT_TAG -b ${NEW_BRANCH}
```
Mpdreamz, line 140: I was thinking something like this might be nicer?

```suggestion
          git checkout ${TARGET_BRANCH} -b ${NEW_BRANCH}
          git format-patch -k --stdout ${TARGET_BRANCH}...main -- docs  | git am -3 -k
```

This will only apply changes from the `docs` directory to the backport to `M.x`. 

Several benefits would be:

- speeding up required checks on the PR. The code in the commit was already tested on main and as part of the release, and we never release from `M.x` 
- The PR will only go up if `docs` actually had any changes.


From: https://stackoverflow.com/questions/19821749/git-cherry-pick-or-merge-specific-directory-from-another-branch

Mpdreamz (changes requested):

Mpdreamz:
@v1v https://github.com/elastic/apm-agent-dotnet/pull/2203 is an example manual run of the changes suggested.

Mpdreamz (approved):
LGTM

PR #2213, opened 2023-11-03
Prevent server certificate callback runtime exception

### LINKED ISSUE #2212: [BUG] MissingMethodException: `set_ServerCertificateCustomValidationCallback`
When using the agent with .NET Framework 4.6.2, the user may experience the following exception.

```
System.MissingMethodException: Method not found: 'Void System.Net.Http.HttpClientHandler.set_ServerCertificateCustomValidationCallback(System.Func`5<System.Net.Http.HttpRequestMessage,System.Security.Cryptography.X509Certificates.X509Certificate2,System.Security.Cryptography.X509Certificates.X509Chain,System.Net.Security.SslPolicyErrors,Boolean>)'.
```

This API is technically [missing](https://github.com/dotnet/standard/issues/524) (see [also](https://stackoverflow.com/a/46577035)) from .NET Framework 4.6.2, despite being in `netstandard2.0`.

This doesn't appear at compile time due to the build picking up a specific `system.net.http` package (see #2112). It can sometimes be worked around with binding redirects but this is not always possible in no-code instrumentation scenarios.

Options:
- We could review switching to `System.Net.Http.WinHttpHandler` (see [docs](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http.WinHttpHandler/src/PACKAGE.md)) on `net462 which may avoid the issue while still supporting the server callback API usage.
- We could drop support for certificate calls back on net462 (or fallback to `ServicePointManager`) to avoid this exception at runtime.
- We could review manually depending on a specific NuGet package version (see #2112).

The easiest option, for now, would be removing the certificate callback code on `net462` as this shouldn't impact many/any consumers and should avoid the exception.

References:
- https://github.com/dotnet/runtime/issues/29314
- https://github.com/Microsoft/dotnet/issues/737
- https://stackoverflow.com/questions/49598467/how-to-choose-target-framework-from-nuget-package

stevejgordon:
On .NET Full framework <4.7.2, HttpClientHandler.ServerCertificateCustomValidationCallback is not available. Due to dependency binding issues introduced with .NET standard, the incorrect System.Net.Http package may be resolved and lead to exceptions at runtime.

This fix prevents calling this API on net462 targets which includes any .NET Framework version < .NET 4.7.2.

This introduces the net472 target so that we can resume using this API from that version, where it was included in the box.

Fixes #2212

### DIFF
--- a/docs/supported-technologies.asciidoc
+++ b/docs/supported-technologies.asciidoc
@@ -28,7 +28,15 @@ the agent will not capture transactions.
 === .NET versions
 
 The agent works on every .NET flavor and version that supports .NET Standard 2.0.
-This means .NET Core 2.0 or newer, and .NET Framework 4.6.2 or newer.
+This means .NET Core 2.0 or newer, and .NET Framework 4.6.2* or newer.
+
+[IMPORTANT]
+====
+While this library should work on .NET Core 2.0+, we limit our support to only those 
+versions currently supported by Microsoft - .NET 6.0+ currently.
+====
+
+_* Due to binding issues introduced by Microsoft, we recommend at least .NET Framework 4.7.2 for best compatibility._
 
 [float]
 [[supported-web-frameworks]]
--- a/docs/supported-technologies.asciidoc
+++ b/docs/supported-technologies.asciidoc
@@ -30,13 +30,13 @@ the agent will not capture transactions.
 The agent works on every .NET flavor and version that supports .NET Standard 2.0.
 This means .NET Core 2.0 or newer, and .NET Framework 4.6.2* or newer.
 
+_* Due to binding issues introduced by Microsoft, we recommend at least .NET Framework 4.7.2 for best compatibility._
+
 [IMPORTANT]
 ====
-While this library should work on .NET Core 2.0+, we limit our support to only those 
-versions currently supported by Microsoft - .NET 6.0+ currently.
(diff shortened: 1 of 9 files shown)


gregkalapos (approved):
Changes LGTM.

We have related docs here: https://github.com/elastic/apm-agent-dotnet/blob/main/docs/configuration.asciidoc

With this PR, `VerifyServerCert` and `ServerCert` won't work on `net462` anymore - which was problematic anyways, so all intentional.

Do we maybe want to add a note about `net462` to the doc?

docs/supported-technologies.asciidoc
```diff
@@ -28,7 +28,15 @@ the agent will not capture transactions.
...
 
 The agent works on every .NET flavor and version that supports .NET Standard 2.0.
-This means .NET Core 2.0 or newer, and .NET Framework 4.6.2 or newer.
+This means .NET Core 2.0 or newer, and .NET Framework 4.6.2* or newer.
+
+_* Due to binding issues introduced by Microsoft, we recommend at least .NET Framework 4.7.2 for best compatibility._
+
+[IMPORTANT]
+====
+While this library *should* work on .NET Core 2.0+, we limit our support to only those
```
gregkalapos, line 37: 👍 Good idea to communicate this so clearly.

gregkalapos (approved):

Issue #2006, opened 2023-02-09
Do not always set request.body to [REDACTED] if reading body is not possible


gregkalapos:
Our [doc](https://www.elastic.co/guide/en/apm/agent/dotnet/current/config-http.html#config-capture-body) for CaptureBody says:

> If the request has a body and this setting is disabled, the body will be shown as [REDACTED].

However, [here in the code](https://github.com/elastic/apm-agent-dotnet/blob/6b4d10f52786ee774c4aee9706c9cf4be10ee63a/src/Elastic.Apm/Extensions/TransactionExtensions.cs#L45) we set the body to `[REDACTED]` always if we can't read the body. There are cases where based on the `CaptureBody` setting we'd read the body, but the agent is unable to do so. E.g. [here](https://github.com/elastic/apm-agent-dotnet/blob/6b4d10f52786ee774c4aee9706c9cf4be10ee63a/src/Elastic.Apm.AspNetFullFramework/AspNetHttpRequest.cs#L35-L36).

So the situation is:
- Request body is set to something other than 'off' and the body would be captured
- Agent tries to capture it, but it fails to do so (e.g. due to buffering setting).

Current behaviour: we set body to `[REDACTED]` as if the setting would be `off`. This is confusing.

Expected behaviour: the body should remain empty. (Plus the agent print waring log already).

redcinelli:
Hi folks, I have been facing this exact issue today.

It is indeed very confusing.
Should we update the documentation ?

Should we fix this ?

PR #2225, opened 2023-11-21
Update: As per #2006 Empty body being redacted is confusing

redcinelli:
As per #2006 

When the body is empty, It feels surprising to set it to redacted.
Even more so when we have the `CaptureBody: true`.


Mpdreamz:
Thanks for tackling this @redcinelli and apologies for taking forever to get around to it. 


I've amended the PR slightly to forcibly set the body to redacted if the config is set to `off` and made sure the default is not an empty string because we attempt to read the body multiple times at various stages in the pipeline. 


LGTM but my colleague @stevejgordon will need to review now that I also added code 😸

stevejgordon (approved):
LGTM, thanks.

PR #2237, opened 2023-12-12
Reenable IIS tests

Mpdreamz:
These were disabled because they started failing on CI possibly due to space constraints.

Re-enabling this again in a PR so that we don't forget this is not running


sample/Sample.Microsoft.Data.SqlClient/Sample.Microsoft.Data.SqlClient.csproj
```diff
@@ -1,9 +1,8 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net462;net6.0;net7.0</TargetFrameworks>
```
gregkalapos, line 4: Is this change needed here? Mainly asking, because I remember SqlClient tracing works very differently between `net462` vs new .NET.

gregkalapos (commented):
Nice! Just 1 small question.

sample/Sample.Microsoft.Data.SqlClient/Sample.Microsoft.Data.SqlClient.csproj
Mpdreamz: Its for a sample thats not used by testing, literally a playground I created a while back. I was even debating deleting it alltogether :)

`/sample` no longer holds applications in use by our tests, those have all moved to the `/test` folder.

gregkalapos (approved):

PR #2239, opened 2023-12-12
Move StartupHooks over to netstandard2.0

Mpdreamz:
Since netcoreapp2.2 is deprecated and has severe security bugs against it


stevejgordon (approved):
LGTM

Issue #2234, opened 2023-12-07
High high memory usage traced to ScopedLogger


solomon-okolie:
### APM Agent version
1.25.1

### Environment

**Operating system and version:** WindowsServer, Windows10

**.NET Framework/Core name and version** : .NET 6

**Application Target Framework(s)**: .NET 6

### Describe the bug
We have console a app that processes messages from a queue and we capture transactions manually like below
```Agent.Tracer.CaptureTransaction("TestProcessor", "MQP", () => { // code here });``` and it's eating quite a lot of memory after a couple of hours. This looks very similar to [#2145.](https://github.com/elastic/apm-agent-dotnet/issues/2145) but this time happening in ScopedLogger.
![image](https://github.com/elastic/apm-agent-dotnet/assets/17110967/87ddad34-93fc-4c49-92ea-1e2f8eb92a9b)

We did a little test based on the commit fix for [#2145.](https://github.com/elastic/apm-agent-dotnet/issues/2145). 

Before change
![WithConcurrentDictionary](https://github.com/elastic/apm-agent-dotnet/assets/17110967/486d791e-36bd-4b94-989b-a243c38f5fd4)
Heap snapshot after 11,000 transactions, and a forced a GC

After change
![WithConditionalWeakTable](https://github.com/elastic/apm-agent-dotnet/assets/17110967/e4d8468b-7f83-4489-a23a-10975e03565b)
Heap snapshot after 11,000 transactions, and a forced a GC

Mpdreamz:
Thanks for raising this, I added an initial fix for this here: 

https://github.com/elastic/apm-agent-dotnet/pull/2242

Including a few other logging related fixes. 

I'll do a pass over our logging to find any high cardinality suspects too.

solomon-okolie:
Any idea when a new nuget version will be available?

AlfonsRoerdink:
I will be great if a new hotfix release can be made because my pods are crashing after a few days because out-of-memory. 
Is there an estimation when the new nuget version will be available?

Mpdreamz:
This is now released under 1.25.3. 

So sorry and apologies you got bit too by this @AlfonsRoerdink let us know if the update did not resolve your issues.

AlfonsRoerdink:
I installed the new version and the application is already running 16 hours without a remarkable memory leakage, 
so it seems to be fixed! Thank you very much!

Mpdreamz:
Thanks a ton for looping back with us! 

[...]

PR #2242, opened 2023-12-13
Move logging caching over to ConditionalWeaktable as fix for #2234

Mpdreamz:
This also ensures we no longer use captured state in the lambda provided to `Log()` by ensuring the callback is static.

### DIFF
--- a/src/Elastic.Apm/Logging/IApmLoggingExtensions.cs
+++ b/src/Elastic.Apm/Logging/IApmLoggingExtensions.cs
@@ -78,7 +78,7 @@ internal static class LoggingExtensions
 #else
 		lock (_lock)
 		{
-			if (!Formatters.TryGetValue(message, out var f))
+			if (Formatters.TryGetValue(message, out var f))
 				return f;
 			Formatters.Add(message, formatter);
 			return formatter;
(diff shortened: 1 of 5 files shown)


src/Elastic.Apm/Logging/IApmLoggingExtensions.cs
```diff
@@ -3,157 +3,179 @@
...
+internal static class LoggingExtensions
 {
-	internal static class LoggingExtensions
+	private static ConditionalWeakTable<string, LogValuesFormatter> Formatters { get; } = new();
+
+	internal static ScopedLogger Scoped(this IApmLogger logger, string scope) => new(logger is ScopedLogger s ? s.Logger : logger, scope);
+
+	private static void DoLog(this IApmLogger logger, LogLevel level, string message, Exception e, params object[] args)
 	{
-		private static ConcurrentDictionary<string, LogValuesFormatter> Formatters { get; } = new ConcurrentDictionary<string, LogValuesFormatter>();
```
Mpdreamz, line 17: As a reviewing note, the .NET runtime implements this with a `ConcurrentDictionary` and protects it by dropping anything if more than N items are cached. For now this seems like a good middle ground though.

gregkalapos (approved):

PR #2247, opened 2024-01-02
Add additional logging to outgoing http call propagation

Mpdreamz:
In particular adds logging when
- outgoing http call is already otherwise instrumented
- Outgoing call is part of an exit span and we don't create a child span for it


stevejgordon (approved):
LGTM

PR #2249, opened 2024-01-08
address multiple structured logging violations

Mpdreamz:


src/azure/Elastic.Apm.Azure.Functions/AzureFunctionsDiagnosticListener.cs
```diff
@@ -27,7 +27,7 @@ internal class AzureFunctionsDiagnosticListener : DiagnosticListenerBase
 
 	protected override void HandleOnNext(KeyValuePair<string, object> kv)
 	{
-		Context.Logger.Trace()?.Log($"'{nameof(AzureFunctionsDiagnosticListener)}.{nameof(HandleOnNext)}': {kv.Key}");
+		Context.Logger.Trace()?.Log($"'{nameof(AzureFunctionsDiagnosticListener)}.{nameof(HandleOnNext)}': {{DiagnosticsEventName}}", kv.Key);
```
stevejgordon, line 30: Should we move the concatenation of the class and method name and also pass that as an argument here, too?

Mpdreamz: Since its a constant I think it makes sense to leave it in like it is.

stevejgordon: Yeah, fair point.

stevejgordon (approved):

PR #2264, opened 2024-01-27
Switch to Licence expression rather than file

### LINKED ISSUE #2258: Nuget Packages should have licence info
When my packages are analyzed in Dependency Track using SBOM files the licence is not able to be determined for any of the Elastic.APM packages.

The PackageLicenseExpression property is set in the csproj so that the NUGET package has the information. This can then be included in the SBOM so that analysis can occur in the appropriate tools. This would be just like Elastic.CommonSchema does it.

Ignore the errors.

**Additional context**
N/a

thompson-tomo:
This moves the packages to use a licence expression rather than the file and is consistent with the other Elastic packages ie Elastic.CommonSchema.

Closes #2258


Mpdreamz:
run elasticsearch-ci/docs

thompson-tomo:
No problem at all @Mpdreamz thanks for merging it.

Issue #2266, opened 2024-01-28
Cleanup package reference in Elastic.Apm.AspNetCore for net 6 TFM


thompson-tomo:
**Is your feature request related to a problem? Please describe.**
I would like Elastic.Apm.AspNetCore to make full usage of the Framework reference by allowing removal of the packages provided by the framework.

**Describe the solution you'd like**
I would like the below dependencies removed from the project when targeting net 6 as the provided by the framework:

- Microsoft.Extensions.Configuration.Abstractions
- Microsoft.Extensions.Logging.Abstractions
- Microsoft.Extensions.Hosting.Abstractions

**Describe alternatives you've considered**
Leave the dependencies as they are

**Additional context**
N/A

Mpdreamz:
Merged #2267

PR #2267, opened 2024-01-28
#2266 Cleanup of packages thanks to Framework reference

thompson-tomo:
This request removes the explicit dependencies for a number of packages and that is made possible due to the framework natively providing those dependencies hence explicit is not needed.


Mpdreamz:
run elasticsearch-ci/docs

Mpdreamz (approved):
LGTM ty @thompson-tomo !

thompson-tomo:
No probs @Mpdreamz note in #2264 I have also added that the licence type is set for the packages.

PR #2270, opened 2024-01-30
github-action: use wildcards for discovering all the workflows

v1v:
No need to maintain the static list of Workflows to be monitored with the CI/CD Observability using Opentelemetry


reakaleek (approved):

Mpdreamz (approved):
LGTM 👍

PR #2276, opened 2024-02-02
Address a few synchronization issues in the codebase

Mpdreamz:
- ensure Add() is locked on composite disposable
- Ensure _dynamicMethods is readonly in ILHelpersExtensions
- fix synchronization issue in ModuleLookup
- _reflectionObject should not be static in BinaryConverter


stevejgordon (approved):
LGTM

PR #2278, opened 2024-02-02
Ensure exposes default constants are readonly.

Mpdreamz:
Technically a breaking change but since noone should be modifying these directly we treat this as a BUG not a breaking change


stevejgordon (approved):
LGTM

PR #2299, opened 2024-02-23
Further logging refinements in ElasticApmModule

stevejgordon:
This continues to work to improve the diagnostics in `ElasticApmModule` by attempting to create the logger earlier in the `Init` method. This allows any exceptions to be logged to the `IApmLogger`, which may be helpful if something critical occurs during the rest of the `Init` method.

It adds some additional log messages to trace the progress of the `Init` method.

### DIFF
--- a/src/integrations/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
+++ b/src/integrations/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
@@ -59,6 +59,13 @@ namespace Elastic.Apm.AspNetFullFramework
 		/// <inheritdoc />
 		public void Init(HttpApplication application)
 		{
+			// We create the logger as early as possible so we can log failures to it.
+			AssignScopedLogger();
+
+			// If the preceeding code to initialize and assign a logger failed, this line will be omitted from the written logs.
+			// Note the null change on the _logger field.
+			_logger?.Trace()?.Log($"{nameof(ElasticApmModule)}.{nameof(Init)} was invoked and called {nameof(AttemptAgentInitialization)}.");
+
 			try
 			{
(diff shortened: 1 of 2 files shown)


Mpdreamz (approved):
LGTM!

src/integrations/Elastic.Apm.AspNetFullFramework/ElasticApmModule.cs
```diff
@@ -109,15 +123,47 @@ public void Init(HttpApplication application)
...
+		private void AssignScopedLogger()
+		{
+			try
+			{
+				var logger = AgentDependencies.Logger ?? FullFrameworkDefaultImplementations.CreateDefaultLogger(null);
+				_logger = logger.Scoped(_dbgInstanceName);
+			}
+			catch (Exception ex)
+			{
+				const string message = $"Exception thrown by {nameof(ElasticApmModule)}.{nameof(AssignScopedLogger)}.";
```
Mpdreamz, line 158: Q? Why would this ever throw an exception? Might be good to note this too. 


Having to document this is less than ideal but if necessary c'est la vie.
https://github.com/elastic/apm-agent-dotnet/pull/2302/files

stevejgordon: I was over-cautious here as everything was guarded by the original try/catch. However, having re-checked the code path, I can't see a scenario where an exception could occur when creating the logger. I guess we could assume that at least the logger creation will succeed and avoid the fallback. The only other place where we fall back to `Trace.WriteLine` is in `LoggingExtensions`.

stevejgordon: Cool. I'll amend this PR, and then the docs one can be closed out.

Mpdreamz (approved):

Issue #2305, opened 2024-03-03
Removal of polyfil packages


thompson-tomo:
**Is your feature request related to a problem? Please describe.**
I want that the packages which i am adding to my projects only has dependencies on those packages which it needs & can not obtain via Framework.

**Describe the solution you'd like**
I would like that for the net 6 TFM polyfil packages such as those listed below are not included so that dependencies are kept to a minimum in the case of Elastic.APM the dependencies will be reduced by 80%.
- System.Diagnostics.DiagnosticSource
- System.Reflection.Metadata
- System.Threading.Tasks.Dataflow
- System.Threading.Tasks.Extensions

**Describe alternatives you've considered**
Accept that the polyfil packages are included

**Additional context**
Continuation of #2266

thompson-tomo:
Note happy to submit a PR for the changes

stevejgordon:
@thompson-tomo This sounds reasonable, and we'd very much welcome a PR for this.

thompson-tomo:
Great I will aim to do PR in the next week or so.

thompson-tomo:
@stevejgordon pr is all done with the tweaked conditions to reduce dependencies.

PR #2308, opened 2024-03-10
#2305 Remove polyfill packages

thompson-tomo:
The polyfil packages listed below will only be included in frameworks less than net 6 to reduce dependency graph

- System.Diagnostics.DiagnosticSource
- System.Reflection.Metadata
- System.Threading.Tasks.Dataflow
- System.Threading.Tasks.Extensions

Closes: #2305


thompson-tomo:
@v1v is this something you could review?

thompson-tomo:
Thanks @v1v

Mpdreamz:
Hey @thompson-tomo we will get to this PR as soon as our current priorities allow!

The PR is much appreciated (like the others on our other repositories). 

The build of this repos is a bit more complicated so we have to carve out a little more time to review this then we did on the other PR's. 

I'm off on PTO now so can't dive in yet :)

thompson-tomo:
Ok thanks for the update @Mpdreamz e joy the time off.

Mpdreamz (approved):
This LGTM!

Would love to clean this up a tad since we built this in a few different ways now. 

A regular nuget build which results in no dependencies (as verified on https://nuget.info/).

And a self contained build for the startup hooks zip files which will still include the dependency which I think is fine for now. 

Many thanks for taking this on @thompson-tomo ❤️

thompson-tomo:
No problem at all @Mpdreamz glad I could help. 🙂

Mpdreamz:
run docs-build

Issue #2320, opened 2024-04-01
[BUG] Azure Functions - Service Name can't be the Artifact Name


rfavreau:
### APM Agent version
The version of the `Elastic.Apm` nuget package used is **1.26**

### Environment

**Operating system and version:**
Azure Functions
**.NET Framework/Core name and version** (e.g. .NET 4.6.2, NET Core 3.1.100) :
.Net Core 8
**Application Target Framework(s)** (e.g. net462, netcoreapp3.1):
dotnet-isolated

### Describe the bug
It's not possible to set the service name with the discover default service name (Artifact name). It's always overwrite by the Azure Functions name.

### To Reproduce

1. Deploy an Azure Function with the APM Agent like described in the [documentation](https://www.elastic.co/guide/en/apm/agent/dotnet/current/setup-azure-functions.html)
2. Then call your function trigger (for exemple an HttpTrigger)
3. Then check the Service Name into the APM page
4. This Service Name will be the Azure Functions name and not the Artifact name

#### Expected behavior
With the `ELASTIC_APM_SERVICE_NAME` environment variable, we can update this Service Name but never with the Artifact Name. To have the same namming convention that the .Net Applications (into the APM page), i expecte to have the Artifact Name.

stevejgordon:
Hi, @rfavreau. I'm sorry for not getting back to you sooner. I've been focusing on the Azure Functions instrumentation this week and noticed this issue. 

The current behaviour is by design. When `ELASTIC_APM_SERVICE_NAME` is not configured, it prefers using the Function App name read from `WEBSITE_SITE_NAME` in Azure. When running locally, you can set that via the `local.settings.json`. There is a bug right now that I'll address to ensure that when it is null in local development, we handle it, probably by falling back to the discovered service name.

The expectation is that generally, you should explicitly set `ELASTIC_APM_SERVICE_NAME` in Azure if you want a name different from the name of the Function App the function lives under. Is that possible in your case?

I see you opened a PR to add a new config. I don't think that achieves what you intended, and I'm also reluctant to add another configuration option specifically for this scenario. `ELASTIC_APM_SERVICE_NAME` should be sufficient to control the service name explicitly.

stevejgordon:
Closing as stale

PR #2321, opened 2024-04-01
Azure functions overwite discover default service name

rfavreau:
Hello team,

To fix the issue #2320 with the _Service Name_ generated by the Azure Functions agent, i added a configuration `OverwriteDiscoverDefaultServiceName` item to be able to overwrite or not the service name.

So it will be possible to set the _Service Name_ with the _Artifact Name_ but also with the _Azure Functions_ name.
To not break the compatibility with the previous release, this setting is set to `true` and the Azure Functions name is set by default.

UnitTests and Documentation have been also updated.

Thank you for your approval.

Richard.


stevejgordon:
@rfavreau - Did you see my last comment on #2320?

stevejgordon:
Closing based on issue comment.

PR #2331, opened 2024-04-12
Add support for `transaction_name_groups` and `use_path_as_transaction_name`

### LINKED ISSUE #2318: Support `transaction_name_groups` and `use_path_as_transaction_name`
I think it makes sense for us to implement [transaction_name_groups](https://github.com/elastic/apm/blob/main/specs/agents/tracing-transaction-grouping.md).

As part of implementing this, we need to consider what our default should be for scenarios such as WCF where the `ElasticApmModule` should default to `$method unknown route` and also support: `use_path_as_transaction_name` to opt into this behavior. This would be a sensitive change to our existing behavior, though.

stevejgordon:
Adds support for these new config options and updates the `ElasticApmModule` implementation to apply them accordingly.

Closes #2318


stevejgordon:
It looks like having the WCF sample included is a bit of a problem for the CI builds, as the target might not be there by default. I'll strip it for now, as it's nice but not worth much effort.

src/Elastic.Apm/BackendComm/CentralConfig/RuntimeConfigurationSnapshot.cs
```diff
@@ -117,12 +117,17 @@ internal RuntimeConfigurationSnapshot(IConfigurationReader mainConfiguration, Ce
 		public IReadOnlyList<WildcardMatcher> TransactionIgnoreUrls =>
 			_dynamicConfiguration?.TransactionIgnoreUrls ?? _mainConfiguration.TransactionIgnoreUrls;
 
+		public IReadOnlyCollection<WildcardMatcher> TransactionNameGroups =>
+			_mainConfiguration.TransactionNameGroups;
```
Mpdreamz, line 121: Does it make sense to discuss making this dynamic with the rest of the Agents?

src/Elastic.Apm/Model/Transaction.cs
```diff
@@ -697,7 +697,28 @@ public void End()
 		}
 
 		if (IsSampled || _apmServerInfo?.Version < new ElasticVersion(8, 0, 0, string.Empty))
+		{
```
Mpdreamz, line 700: Implementation is fine and follows established coding style but we should probably split this long method into smaller methods, e.g this whole block.

Not important to do now just a general note.

Mpdreamz (approved):
Awesome! LGTM 🚢 🇮🇹

src/Elastic.Apm/BackendComm/CentralConfig/RuntimeConfigurationSnapshot.cs
stevejgordon: Potentially, yes. Will add to the weekly agenda.

stevejgordon: Discussed and we'll make this dynamic. I'll update this PR.

Mpdreamz (approved):

PR #2356, opened 2024-05-16
Added check that Open Telemetry Bridge is enable and only then displa…

zviedris:
Add additional check so warning is only displayed if Open Telemetry Bridge is enabled - now it is displayed also in cases if Open Telemetry Bridge is disabled and it is misleading.


ZZCharlieC (approved):

Mpdreamz (approved):
LGTM, will eliminate some noise in our logs. Thank you @zviedris for the contribution!

PR #2371, opened 2024-06-03
Enable global file logging

Mpdreamz:
This ensures we can specify file logging globally and uniformly. Greatly simplifying the debug-ability of the agent.

The preferred way is to set any or all of the following:

* `ELASTIC_OTEL_LOG_TARGETS` (to anything but `none`) 
  * if only set to `stdout` no file will be created but global logging will kicking )
* `OTEL_DOTNET_AUTO_LOG_DIRECTORY`
* `OTEL_LOG_LEVEL`  
  * `trace` `debug` will enable global file logging if the other two variables are not set  explicitly.

This ensure we have one way to debug both our proprietary agent as well as the [Elastic OpenTelemetry Distribution](https://github.com/elastic/elastic-otel-dotnet).

See: https://github.com/elastic/elastic-otel-dotnet/pull/129

For backwards compatible reasons the profiler variables are also supported:

* `ELASTIC_APM_PROFILER_LOG`
* `ELASTIC_APM_PROFILER_LOG_DIR`
* `ELASTIC_APM_PROFILER_LOG_TARGETS` 

Globally setting 

* `ELASTIC_APM_LOG_LEVEL` and `ELASTIC_APM_LOG_DIRECTORY` is also supported but not preferred.

Setting these in any of our supported deploy scenarios:

* Manual instrumentation (nuget)
  * ASP.NET (classic)
  * ASP.NET core
    * NOTE: we now log to both the configured ILogger and our global logging.
* Auto Instrumentation
  * Profiler
  * Startup hooks
    To keep support existing deploys this now always globally logs to file if only `ELASTIC_APM_STARTUP_HOOKS_LOGGING` is set as well.
 

This further updates our docs for the profiler and troubleshooting to prefer `ELASTIC_OTEL_*` variables. 

The profiler is updated to read the same environment variables as managed code.


stevejgordon (commented):
A couple of minor fix suggestions. The only thing I'm a little unsure about is using the `ELASTIC_OTEL_` env vars for this agent. I like that migration is easier, but not sure if it's subtly confusing why OTEL is mentioned in their names.

Mpdreamz:
> A couple of minor fix suggestions. The only thing I'm a little unsure about is using the `ELASTIC_OTEL_` env vars for this agent. I like that migration is easier, but not sure if it's subtly confusing why OTEL is mentioned in their names.

Aye, it's not ideal but I'd love to be in a situation where we can point customers to 3 variables that work across the board and in mixed environments.

We could use:

* `ELASTIC_APM_LOG_LEVEL`
* `ELASTIC_APM_LOG_DIRECTORY`
* `ELASTIC_APM_LOG_TARGETS`

For both? Or:

* `ELASTIC_APM_FILE_LOG_LEVEL`
* `ELASTIC_APM_FILE_LOG_DIRECTORY`
* `ELASTIC_APM_LOG_TARGETS`

Mpdreamz:
Yeah it might make sense `ELASTIC_OTEL_FILE_LOG_LEVEL` on its own is enough to start logging to file today. 

`ELASTIC_OTEL_LOG_LEVEL` on its own might be slightly confusing because the default for `ELASTIC_OTEL_LOG_TARGETS` is to include `file`. 

However I think its benefits negate the possible confusion.

If we take this further we can also rename `ELASTIC_OTEL_FILE_LOG_DIRECTORY` to `ELASTIC_OTEL_LOG_DIRECTORY` since this only makes sense to file logging anyhow.


I'll update the PR to the following set:

* `ELASTIC_OTEL_LOG_LEVEL`
* `ELASTIC_OTEL_LOG_DIRECTORY`
* `ELASTIC_OTEL_LOG_TARGETS`

I like symmetry in the naming and its more terser 😸

stevejgordon (approved):
LGTM

src/Elastic.Apm/Logging/GlobalLogConfiguration.cs
```diff
@@ -0,0 +1,198 @@
...
+		bool IsSet(string k, string v)
+		{
+			var b = k.Trim().Equals(v, StringComparison.InvariantCultureIgnoreCase);
+			if (b)
+				found = true;
+			return b;
+		}
+	}
+
+	internal static string GetDefaultLogDirectory() =>
```
stevejgordon, line 99: This is subtly different to the default directory for the otel distro. Is that expected?

src/profiler/Elastic.Apm.Profiler.Managed.Loader/Logger.cs
```diff
@@ -5,134 +5,92 @@
...
-							? Path.Combine(programData, "elastic", "apm-agent-dotnet", "logs")
-							: ".";
-					}
-					else
-						logDirectory = "/var/log/elastic/apm-agent-dotnet";
-				}
-
-				Directory.CreateDirectory(logDirectory);
-				return logDirectory;
+				using var stream = File.Open(LogFile, FileMode.Append, FileAccess.Write, FileShare.Read);
```
stevejgordon, line 75: I'm not sure how opening the file for each log entry performs compared to holding the file open while the process is running. However, as this already exists, I guess we can keep it for now as we've not had any specific reports of issues.

stevejgordon (commented):
One comment and observation.

src/Elastic.Apm/Logging/GlobalLogConfiguration.cs
Mpdreamz: It's known, since the Otel distro was reworked in its PR for global file logging.

I was going to create a follow up PR to align this.

src/profiler/Elastic.Apm.Profiler.Managed.Loader/Logger.cs
Mpdreamz: Aye not ideal however the loader is only short lived.

src/Elastic.Apm/Logging/GlobalLogConfiguration.cs
Mpdreamz: The reason for the follow up PR is that `Elastic.Apm.Profiler.Managed.Loader.csproj` currently is dependency free and `System.Runtime.InteropServices.RuntimeInformation` is not available without a nuget reference. 

Want to tackle that in isolation.

stevejgordon (dismissed):
LGTM

stevejgordon (approved):
LGTM

PR #2384, opened 2024-06-27
Keyless Azure Auth in test-linux

reakaleek:
Use OIDC to login to azure.

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -77,10 +77,9 @@ jobs:
   azure-tests:
     runs-on: ubuntu-latest
     needs: [ 'format', 'tests' ]
-    if: ${{ true }}
-    #if: |
-    #  github.event_name != 'pull_request'
-    #  || github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork == false
+    if: |
+      github.event_name != 'pull_request'
+      || github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork == false
     env:
       ARM_CLIENT_ID: ${{ secrets.ARM_CLIENT_ID }}
       ARM_SUBSCRIPTION_ID: ${{ secrets.ARM_SUBSCRIPTION_ID }}
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -86,7 +86,7 @@ jobs:
       ARM_SUBSCRIPTION_ID: ${{ secrets.ARM_SUBSCRIPTION_ID }}
       ARM_TENANT_ID: ${{ secrets.ARM_TENANT_ID }}
       ARM_USE_OIDC: true
-      AZURE_RESOURCE_GROUP_PREFIX: ci-dotnet-${GITHUB_RUN_ID}
+      AZURE_RESOURCE_GROUP_PREFIX: ci-dotnet-${{ github.run_id }}
     permissions:
       contents: read
       id-token: write
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -77,15 +77,19 @@ jobs:
   azure-tests:
     runs-on: ubuntu-latest
     needs: [ 'format', 'tests' ]
-    if: ${{ false }}
+    if: ${{ true }}
     #if: |
     #  github.event_name != 'pull_request'
     #  || github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork == false
     env:
       ARM_CLIENT_ID: ${{ secrets.ARM_CLIENT_ID }}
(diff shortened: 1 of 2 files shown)


.github/workflows/test-linux.yml
```diff
@@ -77,31 +77,36 @@ jobs:
   azure-tests:
     runs-on: ubuntu-latest
     needs: [ 'format', 'tests' ]
-    if: ${{ false }}
+    if: ${{ true }}
```
reakaleek, line 80:
```suggestion
    if: ${{ false }}
```

revert to previous state

reakaleek:
I will disable the `azure-tests` job again. 

But I was able to validate that the OIDC login works and terraform resources are created as expected in https://github.com/elastic/apm-agent-dotnet/actions/runs/9700716496/job/26773022726

.github/workflows/test-linux.yml
```diff
@@ -83,25 +83,30 @@ jobs:
     #  || github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork == false
     env:
       ARM_CLIENT_ID: ${{ secrets.ARM_CLIENT_ID }}
-      ARM_CLIENT_SECRET: ${{ secrets.ARM_CLIENT_SECRET }}
       ARM_SUBSCRIPTION_ID: ${{ secrets.ARM_SUBSCRIPTION_ID }}
       ARM_TENANT_ID: ${{ secrets.ARM_TENANT_ID }}
+      ARM_USE_OIDC: true
+      AZURE_RESOURCE_GROUP_PREFIX: ci-dotnet-${{ github.run_id }}
```
v1v, line 89: Can you explain these two env variables? Or a link to where we can read more about it?

.github/workflows/test-linux.yml
```diff
@@ -83,25 +83,30 @@ jobs:
...
-        run: |
-          az login --service-principal --username ${{ env.ARM_CLIENT_ID }} --password ${{ env.ARM_CLIENT_SECRET }} --tenant ${{ env.ARM_TENANT_ID }}
-          az account set --subscription ${{ env.ARM_SUBSCRIPTION_ID }}
-          echo "AZURE_RESOURCE_GROUP_PREFIX=ci-dotnet-${GITHUB_RUN_ID}" >> ${GITHUB_ENV}
+      - name: 'Az CLI login'
+        uses: azure/login@6c251865b4e6290e7b78be643ea2d005bc51f69a # v2.1.1
+        with:
+          client-id: ${{ secrets.ARM_CLIENT_ID }}
+          tenant-id: ${{ secrets.ARM_TENANT_ID }}
+          subscription-id: ${{ secrets.ARM_SUBSCRIPTION_ID }}
```
v1v, line 105: Do we even need these secrets when using OIDC/Keyless?

.github/workflows/test-linux.yml
reakaleek: `AZURE_RESOURCE_GROUP_PREFIX` is actually a custom env used by the test in https://github.com/elastic/apm-agent-dotnet/blob/70fedbb01716d832adc244809ee503739715c17b/test/Elastic.Apm.Tests.Utilities/Azure/AzureResources.cs#L16

It was there before. I only moved to `env`.

and `ARM_USE_OIDC` docs is here https://developer.hashicorp.com/terraform/language/settings/backends/azurerm#use_oidc

reakaleek: and here: https://developer.hashicorp.com/terraform/language/settings/backends/azurerm#backend-azure-ad-service-principal-or-user-assigned-managed-identity-via-oidc-workload-identity-federation

.github/workflows/test-linux.yml
reakaleek: Yes, they are needed.

See: https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-azure#requesting-the-access-token

.github/workflows/test-linux.yml
```diff
@@ -77,31 +77,35 @@ jobs:
   azure-tests:
     runs-on: ubuntu-latest
     needs: [ 'format', 'tests' ]
-    if: ${{ false }}
-    #if: |
-    #  github.event_name != 'pull_request'
-    #  || github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork == false
+    if: |
```
reakaleek, line 80: Re-enabling the test again because I was able to verify it works again when increasing the timeout in https://github.com/elastic/apm-agent-dotnet/pull/2385

reakaleek: cc @Mpdreamz

Mpdreamz (approved):

v1v (approved):

PR #2389, opened 2024-07-02
Fix linux build dependency of glibc

### LINKED ISSUE #2383: [BUG] Auto-instrumentation using the profiler-based agent installation fails inside Docker containers since 1.23.0
When building a container image containing the profiler agent for auto-instrumentation of ASP.NET Core applications, the agent is no longer working. No logs appear from the profiler, suggesting that this fails extremely early when the CLR attempts to load the profiler. We see no obvious logs from the .NET runtime to identify the reason for the failure. 

We need to identify what changed between 1.22.0 and 1.23.0 that may have caused this. No obvious changes were made to the managed profiler code.

stevejgordon:
Fixes #2383

Update Linux builds of the Rust profiler to use zigbuild so that we can pin to an older version of GLIBC to avoid crashing on older distributions. This PR also removes some redundant files from the Linux profiler zip and updates the docs.

### DIFF
--- a/.github/workflows/install-dependencies/action.yml
+++ b/.github/workflows/install-dependencies/action.yml
@@ -1,78 +1,89 @@
 ---
-name: Install Dependencies
-description: Ensures an action has the appropiate non .NET dependencies installed
-
-inputs:
-  rust:
-    description: 'Install rust toolchain ("true" or "false")'
-    required: false
-    default: "false"
-  azure:
-    description: 'Install azure functions tool chain ("true" or "false")'
-    required: false
-    default: "false"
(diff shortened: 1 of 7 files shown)


Mpdreamz (approved):
Love the docs and contributing updates!

Also neatly done creating an isolated cargo build task 👍

stevejgordon:
Thanks, @Mpdreamz. The MongoDb license slipped through into my commit, plus I fixed a typo. I am ready for a final review and will merge and try a release.

Mpdreamz (approved):

Mpdreamz (approved):

PR #2399, opened 2024-07-16
release: fix adding latest tag

reakaleek:
## Context

According to https://github.com/docker/metadata-action?tab=readme-ov-file#latest-tag the latest tag is only created on specific configurations.

image

## What Changed

In this change, I'm replacing the type `raw` with `semver` so that the `latest` tag is created automatically.

## Testing

I was able to test and verify the correct behavior in a test repo at https://github.com/unelastisch/test-publish-docker/actions/runs/9960071835/job/27518417577, where the action is producing both the semver and the latest tag.

image

Here is the workflow file I used to reproduce it: https://github.com/unelastisch/test-publish-docker/actions/runs/9960071835/workflow

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -59,10 +59,8 @@ jobs:
         flavor: |
           latest=auto
         tags: |
-          # set latest tag for default branch
-          type=raw,value=latest,enable={{is_default_branch}}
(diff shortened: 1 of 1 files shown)


Mpdreamz (approved):

Mpdreamz:
Thank you @reakaleek !

PR #2418, opened 2024-08-15
Adding filters should not force initialization of Agent.

Mpdreamz:
Instead if the agent is not yet setup we simply buffer the filters until the agent gets constructed.

This allows you to add filters before initializing the static agent.

This manifested itself in the hosted service integrations (.NET Core) where the agent get's constructed slightly delayed.

And calling

```
Agent.AddFilter((ISpan span) =>
{
    return span;
});
```

Would win the race for `Agent.Instance` over the ASP.NET core integrations.


stevejgordon (approved):
Lgtm

stevejgordon (approved):
Lgtm

PR #2427, opened 2024-08-19
Ensure we capture baggage when capturing Errors during unsampled transactions

Mpdreamz:


src/Elastic.Apm/Model/Error.cs
```diff
@@ -55,19 +55,23 @@ private Error(Transaction transaction, string parentId, IApmLogger loggerArg, Di
...
 
-		private void CheckAndCaptureBaggage()
+		private void CheckAndCaptureBaggage(Transaction transaction)
 		{
 			if (Activity.Current == null || !Activity.Current.Baggage.Any())
 				return;
 
+			//if context was not set prior we set it now to ensure we capture baggage for errors
+			//occuring during unsampled transactions
+			Context ??= transaction.Context.DeepCopy();
```
gregkalapos, line 73: Did the line below with `var labels = Context.InternalLabels.Value;` fail with a `NullReferenceException`? Or what was the behaviour before?

Mpdreamz: Context is null because we never copy it from transaction when its an unsampled transaction.

gregkalapos (approved):

src/Elastic.Apm/Model/Error.cs
Mpdreamz: Thought about it but this way we only do it if there is actual baggage to set :)

gregkalapos: > Thought about it but this way we only do it if there is actual baggage to set :)

Yeah, that's a good point.

aheubusch, line 73: Requires a null check. `transaction` can be null.

PR #2431, opened 2024-08-28
Exclude default processes and services

stevejgordon:
This also updates the comparison for excluded process and service names to be case insensitive.


Mpdreamz (approved):
LGTM!

PR #2436, opened 2024-09-09
Ensure isEnabled = false won't start worker loop

Mpdreamz:
Inside PayloadSenderV2, we also now early exit when events get queued.

Removed early exit from constructor, allow readonly fields to be initialized
so that when we start adding nullable annotations not all of them have to be nullable

### DIFF
--- a/src/Elastic.Apm/Report/PayloadSenderV2.cs
+++ b/src/Elastic.Apm/Report/PayloadSenderV2.cs
@@ -96,7 +96,7 @@ namespace Elastic.Apm.Report
 				_metadata.Labels.Add(globalLabelKeyValue.Key, globalLabelKeyValue.Value);
 			_cachedActivationMethod = _metadata.Service?.Agent.ActivationMethod;
 
-			if (isEnabled)
+			if (_isEnabled)
 				ResetActivationMethodIfKnownBrokenApmServer();
 
 			if (configuration.MaxQueueEventCount < configuration.MaxBatchEventCount)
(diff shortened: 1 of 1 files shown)


src/Elastic.Apm/Report/PayloadSenderV2.cs
```diff
@@ -73,27 +74,30 @@ public PayloadSenderV2(
...
 			_serverInfoCallback = serverInfoCallback;
+
 			var process = ProcessInformation.Create();
 			_metadata = new Metadata { Service = service, System = System, Process = process };
 			foreach (var globalLabelKeyValue in configuration.GlobalLabels)
 				_metadata.Labels.Add(globalLabelKeyValue.Key, globalLabelKeyValue.Value);
 			_cachedActivationMethod = _metadata.Service?.Agent.ActivationMethod;
-			ResetActivationMethodIfKnownBrokenApmServer();
+
+			if (isEnabled)
```
stevejgordon, line 99: nit: for consistency, should we prefer the `_isEnabled` field rather than the parameter here? We use the parameter above.

src/Elastic.Apm/Report/PayloadSenderV2.cs
```diff
@@ -123,7 +127,9 @@ public PayloadSenderV2(
 			_eventQueue = new BatchBlock<object>(configuration.MaxBatchEventCount);
 
 			SetUpFilters(TransactionFilters, SpanFilters, ErrorFilters, apmServerInfo, logger);
-			StartWorkLoop();
+
+			if (isEnabled)
```
stevejgordon, line 131: nit: for consistency, should we prefer the `_isEnabled` field rather than the parameter here? We use the parameter above.

stevejgordon (approved):
One nit, but otherwise, this LGTM.

stevejgordon (approved):
LGTM

PR #2437, opened 2024-09-09
github-action: use ephemeral tokens with the required permissions

v1v:
## Details

⚠️ This PR was created by an automated tool. Please review the changes carefully. ⚠️ 

### What

Use https://github.com/tibdex/github-app-token to generate ephemeral tokens with the required
permissions only

This is the alternative to moving away from finer-grained GitHub tokens and reducing the
cumbersome of rotating them as we do nowadays.

### Implementaiton details

We have used the same GitHub action in other places.

If there are any questions, please reach out to the @elastic/observablt-ci


reakaleek (approved):

v1v:
CLA checker will cause some disruptions - I'm working on it with the relevant CLA owners. for now, I'll keep this draft to avoid surprises.

.github/workflows/updatecli.yml
```diff
@@ -17,6 +17,18 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Get token
```
Mpdreamz, line 20:
```
permissions:
  contents: read
```

Do we need to update this instead and rely on `GITHUB_TOKEN`?

v1v: Unfortunately, GitHub actions don't trigger any action if the `Actor` is `Github Actions bot`. 

See https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#about-the-github_token-secret

> Commits pushed by a GitHub Actions workflow that uses the GITHUB_TOKEN do not trigger a GitHub Pages build.

That's the reason we use a Service Machine account, and we are moving to a GitHub app to help us moving away from finer-tokens and rotating them.

PR #2444, opened 2024-09-13
No longer parse request cookies, but ensure they are still redacted in the `Cookie` header string

stevejgordon:
We've identified that sending transactions with cookies parsed to `Request.Cookies` can result in failures to index those transactions when any of the cookies include periods (which is common for ASP.NET core by default). 

This PR removes the extraction from the header onto `Request.Cookies`. Instead, we now modify the `Cookie` header value to redact the cookie items based on the configured list of field names to sanitize. For .NET targets, we optimize this to avoid allocations as much as possible. Unfortunately, for the .NET Framework, we can't be as efficient.


src/Elastic.Apm/Api/Request.cs
```diff
@@ -27,11 +27,6 @@ public class Request
 		/// </summary>
 		public Dictionary<string, string> Headers { get; set; }
 
-		/// <summary>
```
Mpdreamz, line 30: Breaking?

src/Elastic.Apm/Helpers/CookieHeaderRedacter.cs
```diff
@@ -0,0 +1,206 @@
...
+
+		var redactedCookieHeader = string.Empty;
+
+#if NETFRAMEWORK
+		// The use of `Span<T>` in NETFX can cause runtime assembly loading issues due to our friend,
+		// binding redirects. For now, we take a slightly less allocation efficient route here, rather
+		// than risk introducing runtime issues for consumers. Technically, this should be "fixed" in
+		// NET472+, but during testing I surprisingly still reproduced an exception. Elastic.APM depends on
+		// `System.Diagnostics.DiagnosticSource 5.0.0` which itself depends on `System.Runtime.CompilerServices.Unsafe`
+		// which is where the exception occurs. 5.0.0 is marked as deprecated so we could look to prefer
```
Mpdreamz, line 49: Can you create an issue for this? Would love to bump SDD to 6.0.0 if we can :)

src/Elastic.Apm/Helpers/CookieHeaderRedacter.cs
```diff
@@ -0,0 +1,206 @@
...
+		// a new version but we have special handling for the ElasticApmAgentStartupHook
+		// zip file version. For now, we decided not to mess with this as it's hard to test all scenarios.
+
+		var sb = new StringBuilder(cookieHeaderValue.Length);
```
Mpdreamz, line 53: Nit but can we split this into 2 implementation methods?

Mpdreamz (changes requested):
Few nitpicks mostly LGTM :)

src/Elastic.Apm/Helpers/CookieHeaderRedacter.cs
stevejgordon: Yeah, makes sense.

src/Elastic.Apm/Helpers/CookieHeaderRedacter.cs
stevejgordon: https://github.com/elastic/apm-agent-dotnet/issues/2445

src/Elastic.Apm/Api/Request.cs
stevejgordon: Yep, will document this in our release notes.

src/Elastic.Apm/Helpers/CookieHeaderRedacter.cs
stevejgordon: I've rearranged the compiler pre-processor directives to create two versions of the `Redact` method: one for the .NET Framework and one for everywhere else.

Mpdreamz (approved):

PR #2455, opened 2024-09-30
Improve OTel bridge compatibility with existing Azure instrumentation

stevejgordon:
This ensures that the Azure SDK instrumentation doesn't use the existing `Activity` span ID, which, when combined with the OTel bridge, can lead to spans being parented by themselves.

This also enhances the logic of the OTel bridge so that we skip span creation for activities coming from the Azure SDK libraries if our Azure instrumentation assemblies are loaded. This ensures we don't create essentially repeated child spans.

Fixes #2454

### DIFF
--- a/src/Elastic.Apm/Model/Transaction.cs
+++ b/src/Elastic.Apm/Model/Transaction.cs
@@ -27,6 +27,10 @@ internal class Transaction : ITransaction
 {
 	internal static readonly string ApmTransactionActivityName = "ElasticApm.Transaction";
 
+#if NET
+	private static readonly ActivitySource ElasticApmActivitySource = new("Elastic.Apm");
+#endif
+
 	internal readonly TraceState _traceState;
 
 	internal readonly ConcurrentDictionary<SpanTimerKey, SpanTimer> SpanTimings = new();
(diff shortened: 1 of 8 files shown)


src/Elastic.Apm/Model/Transaction.cs
```diff
@@ -27,6 +27,10 @@ internal class Transaction : ITransaction
 {
 	internal static readonly string ApmTransactionActivityName = "ElasticApm.Transaction";
 
+#if NET
+	internal static readonly ActivitySource ElasticApmActivitySource = new("Elastic.Apm");
```
Mpdreamz, line 31: 👍

Mpdreamz: Should we bump our DiagnosticSource reference from 5.0.0? 

We can do in a follow up PR.

Mpdreamz (dismissed):
LGTM

src/Elastic.Apm/Model/Transaction.cs
stevejgordon: Yeah, I'll open an issue to track that if we don't have one. I think we should, but I'm not sure about the implications, so we'll need to spike it and see if anything breaks!

Mpdreamz (approved):

PR #2461, opened 2024-10-07
Limit attribute count and truncate values in ElasticActivityListener

Mpdreamz:
Implement limits to comply with OpenTelemetry specs, capping attributes at 128 and truncating string values to 10,000 characters. This ensures payload size remains manageable and adheres to the defined standards.

https://opentelemetry.io/docs/specs/otel/common/#attribute-limits


stevejgordon (approved):
LGTM

PR #2469, opened 2024-10-23
github-actions: use ephemeral github tokens

v1v:
Use the temporary GitHub tokens generated with the GitHub app.

See https://github.com/elastic/apm-agent-dotnet/pull/2464

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,7 +7,6 @@ on:
 permissions:
   contents: write
   issues: write
-  pull-requests: write
 
 env:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -5,9 +5,7 @@ on:
     types: [published]
 
 permissions:
-  contents: write
-  issues: write
-  pull-requests: write
+  contents: read
```
v1v, line 8: These permissions are not needed anymore - the other jobs have defined their required permissions

Mpdreamz (approved):
Awesome thanks @v1v !

reakaleek (approved):

PR #2470, opened 2024-10-23
Remove internal newtonsoft json and move over to system.text.json

Mpdreamz:
This is opened more as exploration if we could and to discuss if we should @stevejgordon 

There's 9 failing unit tests to go over currently.


stevejgordon:
Closing as unplanned

PR #2498, opened 2024-11-25
Remove net 6.0 targets

stevejgordon:


Mpdreamz (approved):
Lgtm, one thing that comes to mind is to double check all our `.dotnet-tools.json` dependencies run on .net 8.0. 

We might get bit by that during release. 

Starting with .net 9 we can control this strict sdk pinning for running dotnet tools by specifying it's safe to run this tool on a later sdk (99.99÷) of the time.

stevejgordon:
> Lgtm, one thing that comes to mind is to double check all our `.dotnet-tools.json` dependencies run on .net 8.0.
> 
> We might get bit by that during release.
> 
> Starting with .net 9 we can control this strict sdk pinning for running dotnet tools by specifying it's safe to run this tool on a later sdk (99.99÷) of the time.

Good shout! `assembly-differ` has a specific `net8.0` target, so I think that's fine, and `minver-cli` has the `RollForward` MSBuild property. As that's been working on the .NET 8 SDK already, I think that should be ok.

PR #2503, opened 2024-11-27
Phase one logger optimisations

### LINKED ISSUE #2500: Notable increase in memory usage from 1.25.x to 1.30.x
Hi there,

Initial Update and Issue Discovery On November 24, 2024, at 1:26:34 AM GMT+3:30, I updated our gateway to: 

- .NET 9 from 8
- Yarp 2.2 from 2.1
- Elastic.Apm.AspNetCore 1.30.1 from 1.23.0

Approximately 120,000 TPM (Transactions Per Minute)
Observed Issue: Notable Memory Usage Increase Following the update, a significant rise in memory usage was detected. For reference, our service's memory usage pattern is illustrated below on the morning of discovery. Fortunately, our 4GB RAM per pod capacity temporarily mitigated the immediate impact.

Image

Troubleshooting Steps:

Rollback and Partial Update: I rolled back the changes and only updated Yarp and dThe outcome is shown below.
Image

Testing Elastic.Apm.AspNetCore Versions:

Version 1.25: Although this version exhibited errors (detailed below), memory usage was more efficient compared to version 1.23.
Image
But Kibana was empty in this duration

Version 1.29.0 (Current): Memory usage increases at a gentler rate. However, version 1.25 was preferable in terms of memory efficiency, despite its error profile and the absence of data in Kibana.
Image

Empty APM with version 1.25
Image

Additional Metrics:

Ratio: 0.1
Throughput: Illustrated below for context.
Image

stevejgordon:
This PR dramatically reduces the overhead of scoped logging, even for scenarios where the configured log level verbosity was low (e.g., Warning). The most critical improvement is caching scoped loggers within Spans, Traces, and Errors. We lose convenience in no longer having the span ID in the scope, but it's generally available from the Span argument. 

Previously, a new `ScopedLogger` was created for every span and transaction, and a new `ConditionalWeakTable` was also allocated. Due to the use of Finalisers on the CWT, this had an even more significant GC impact. We now maintain a cache of scoped loggers (of which there are relatively few).

The ID creation for the Spans has been switched to `ActivitySpanId.CreateRandom().ToString()` from the MS diagnostics library, as this has been more heavily optimised than our original version and achieves the same goal.

In `Tracer.End()`, we used an `Assertion` with a large interpolated string. This was being interpolated for each transaction as the `Assertion` path was enabled. However, the condition we assert is generally not met. Therefore, we now check this manually to only create the message if we are in an invalid state.

I've also switched two structs to record structs, as these were used as keys in dictionaries and ended up being boxed due to a lack of `IEquality` implementations.

These are phase one of optimisations for the logging implementation and address the general production scenario of using a less verbose log level.

### Profiling

Profiling was achieved using a .NET 9 ASP.NET Core minimal API project with 20 concurrent connections and 100,000 requests.

**Before Memory Profile**
image

**After Memory Profile**
image

**Before Top Retained**
image

**After Top Retained**
image

Closes #2500


src/Elastic.Apm/Logging/IApmLoggingExtensions.cs
```diff
@@ -12,13 +13,87 @@
...
+		LogScoped(logger, scope, LogLevel.Warning, message, null, args);
+
+	internal static void LogScopedWarningWithException(this IApmLogger? logger, ScopeName scope, Exception exception, string message, params object?[]? args) =>
+		LogScoped(logger, scope, LogLevel.Warning, message, exception, args);
+
+	internal static void LogScopedError(this IApmLogger? logger, ScopeName scope, string message, params object?[]? args) =>
+		LogScoped(logger, scope, LogLevel.Error, message, null, args);
+
+	internal static void LogScopedErrorWithException(this IApmLogger? logger, ScopeName scope, Exception exception, string message, params object?[]? args) =>
+		LogScoped(logger, scope, LogLevel.Error, message, exception, args);
```
gregkalapos, line 65: The idea behind the `?.` notation was to avoid evaluation of the log statement when the log level of the given line is not matching the current log level. That is because sometimes in e.g. debug logs we access things that may be expensive to get (some getters trigger crazy things), so we only wanted to access those when we really need them. By getting rid of the `?.` notation and using these methods we'd loose this. 

Not the best example, but e.g. here is one that shows this

In `Span.cs` `_logger.Trace()?.LogException` is replaced by `_logger.LogScopedTraceWithException`. I did a test and added back the `?.` notation before the `_logger.Trace()?.LogException` call to compare the result.
Below is the decompiled version to C#1 (IL is the real truth, but this is easier to read :) ):

Screenshot 2024-11-27 at 17 27 51

As you can see, it results in an if block, which does not run when the log level is not matching, so e.g. no getter is accessed on the HTTP context. Same for all the other things in similar scenarios - we fetch and concatenate a lot of things there that'd only want to evaluate when we really print those things.

gregkalapos (commented):
Overall I totally agree with the direction. 👍 

Below one thing about replacing the `?.` notation to check log levels. Less relevant from the allocation's perspective, but I think it'd be still worth keeping that.

ArminShoeibi:
There is a gotcha here, we should use lock like dotnet framework for AddOrUpdate

https://github.com/elastic/apm-agent-dotnet/blob/3386ba62c316b311cf1da30278ce927a6735f2ce/src/Elastic.Apm/Logging/IApmLoggingExtensions.cs#L151
image


for more details:

https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.addorupdate?view=net-9.0

stevejgordon:
@ArminShoeibi Formatters is `ConditionalWeakTable<string, LogValuesFormatter>`. I need to review that, but I think it's okay here.

ArminShoeibi:
> @ArminShoeibi Formatters is `ConditionalWeakTable<string, LogValuesFormatter>`. I need to review that, but I think it's okay here.

My bad, it has a lock itself

https://github.com/dotnet/runtime/blob/1d1bf92fcf43aa6981804dc53c5174445069c9e4/src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/ConditionalWeakTable.cs#L126

stevejgordon:
No worries. It's hard to keep track!

src/Elastic.Apm/Logging/IApmLoggingExtensions.cs
stevejgordon: I've refactored back to a smaller phase one change to account for the above. Indeed we incurred some overhead for the log arguments with the changes, so we're now seeing an even bigger overall improvement.

gregkalapos (approved):
Thanks for the updates.

lgtm

gregkalapos (approved):

PR #2510, opened 2024-12-02
github-actions: dependabot cannot use the id-token:write

v1v:
For security reason, we don't access secrets when running any dependabot PR. For the same reason, the `id-token: write` is not an option when the actor is dependabot.


stevejgordon (approved):
LGTM

reakaleek (approved):

PR #2562, opened 2025-03-04
Fix detection of api key and secret token

monty241:
When the `ApiKey == string.Empty`, a header will be added. This seems unlikely to be intended. Dito for `SecretToken`.

We have had some problems due to this handling in Summer of 2024.

Now we are making improvement, inline with other code in this file such as:

```
if (!string.IsNullOrEmpty(service.Name))
```


stevejgordon:
run docs-build

PR #2557, opened 2025-02-26
[docs] Migrate docs from AsciiDoc to Markdown

colleenmcginnis:
Migrate docs from AsciiDoc to Markdown. The preview can be built after #2556  is merged.

PR #2569, opened 2025-03-13
[docs] Remove AsciiDoc changelog file

colleenmcginnis:
Follow up to #2557 

Removes `CHANGELOG.asciidoc`. Use `docs/release-notes/index.md` instead.

Notes:
* **Findability**: I removed this to prevent anyone from editing this file. If you're concerned about users expecting a changelog file at the top level of this repo, let me know and I can also add a Markdown file like [this one](https://github.com/elastic/apm-server/blob/main/CHANGELOG.md) we added in /apm-server.
* **Automation**:  It looks like [`.github/workflows/release.yml`](https://github.com/elastic/apm-agent-dotnet/blob/main/.github/workflows/release.yml#L204) expects a `CHANGELOG.asciidoc` file. I'm not exactly sure how this works, but let me know if I can help at all.


stevejgordon:
@colleenmcginnis, on further reflection, having a markdown CHANGELOG does make sense, I think.

stevejgordon (changes requested):
Let's add the links to the new change notes, just in case people are looking for a CHANGELOG.md in the root.

PR #2572, opened 2025-03-19
ci: pin actions to specific commits

kruskall:
replace mutable tag with commit hash to improve security and reproducibility

### DIFF
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   docs-preview:
-    uses: elastic/docs-builder/.github/workflows/preview-build.yml@main
+    uses: elastic/docs-builder/.github/workflows/preview-build.yml@99b12f8bf7a82107ffcf59dacd199d00a965e9db # main
     with:
       path-pattern: docs/**
     permissions:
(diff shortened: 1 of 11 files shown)


v1v (commented):
Let's talk about this first.

There are many implications here for instance, @reakaleek might need to be involved.

I'm not saying we should not, I'm saying let's talk about this first

.github/workflows/docs-build.yml
```diff
@@ -9,7 +9,7 @@ on:
 
 jobs:
   docs-preview:
-    uses: elastic/docs-builder/.github/workflows/preview-build.yml@main
+    uses: elastic/docs-builder/.github/workflows/preview-build.yml@99b12f8bf7a82107ffcf59dacd199d00a965e9db # main
```
reakaleek, line 12: This is currently not possible for us.

We are assembling the docs from multiple repos and currently, we are just using the latest version.

As of now we can't handle different source repos with different versions.

Keeping this to `main` ensures that the docs are verified with the latest version of our docs-builder.

cc @elastic/docs-engineering

reakaleek: Sorry, I wrote too fast.

We might be able to do this because the worfklow downloads the latest docs-builder binary anyway.

But we to verify this first and if there are any other implications.

PR #2574, opened 2025-03-20
Update labeler.yml

ismisepaul:
Pinning action to a full length commit SHA [see](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions)


v1v (approved):

v1v:
> not sure why others are exclude, is that intended ?

As far as I know , they are pinned per specific case, one PR per GH action

PR #2580, opened 2025-04-01
update MongoDB.Driver PackageReference to 3.3

badge2033:
update MongoDB.Driver PackageReference to 3.3
in this version package MongoDB.Driver.Core was merged with the main package and so was removed


stevejgordon:
We need to consider this one a bit more. The issue is, for consumers using older packages, this may cause conflicts and breaking changes, which we need to manually test and validate.

badge2033:
> We need to consider this one a bit more. The issue is, for consumers using older packages, this may cause conflicts and breaking changes, which we need to manually test and validate.

i think this reasoning -while valid- is counter productive.
multiple projects are locked from a needed update for the mongo packages because of the desynchronised dependencies on the elastic packages.
But conservative project managers who are not keen on speedy updating, will stay on elastic old packages version, so they won't be affected either way.
please reconsider this update as very urgent.

stevejgordon:
run docs-build

PR #2613, opened 2025-06-20
Avoid shallow exception thrown already existing activity parent

JeremyBessonElastic:


src/Elastic.Apm/Model/Transaction.cs
```diff
@@ -270,10 +270,14 @@ internal Transaction(
 				// This is so that the distributed tracing data will flow to any child activities
 				try
 				{
-					_activity.SetParentId(
-						ActivityTraceId.CreateFromString(distributedTracingData.TraceId.AsSpan()),
-						ActivitySpanId.CreateFromString(distributedTracingData.ParentId.AsSpan()),
-						distributedTracingData.FlagRecorded ? ActivityTraceFlags.Recorded : ActivityTraceFlags.None);
+					// Only if the parent does not already exist
+					if (_activity.ParentId != null && _activity.Parent != null)
```
stevejgordon, line 274: Should the logic here be inverted? I think we only want to try to set this if `ParentId == null`.

stevejgordon (changes requested):

src/Elastic.Apm/Model/Transaction.cs
JeremyBessonElastic: ho, yes

Thanks for the catch

stevejgordon (approved):
LGTM

Issue #2031, opened 2023-03-05
[BUG] Cannot connect to APM server which supports only TLSv1.3


VMelnalksnis:
### APM Agent version
1.20.0

### Environment

**Operating system and version:**
Ubuntu 20.04.5 LTS

**.NET Framework/Core name and version** (e.g. .NET 4.6.1, NET Core 3.1.100) :
.NET SDK 7.0.102

**Application Target Framework(s)** (e.g. net461, netcoreapp3.1):
net7.0

### Describe the bug
After upgrading to 1.20.0 nothing is sent to APM server, because it fails the TLS handshake:
```
{"log.level":"error","@timestamp":"2023-03-05T11:03:29.598+0200","log.logger":"beater.http","log.origin":{"file.name":"http/server.go","file.line":3195},"message":"http: TLS handshake error from 10.10.12.162:33834: tls: client offered only unsupported versions: [303 302 301]","service.name":"apm-server","ecs.version":"1.6.0"}
```

### To Reproduce

1. APM server configuration:
```yaml
ssl:
    enabled: true
    certificate: '/path/to/certificate.crt'
    key: '/path/to/key.key'
    key_passphrase: '${KEY_PASSPHRASE}'
    supported_protocols: [TLSv1.3]
```
2. `appsettings.json`
```json
        "ElasticApm": {
                "Enabled": true,
                "CentralConfig": false,
                "ServiceName": "Foo",
                "ServerUrl": "https://apm.some.fully.qualified.name.com:8200"
        }
```
3. `Startup.cs`
```csharp
application.UseElasticApm(
	_configuration,
	new AspNetCoreDiagnosticSubscriber(),
	new HttpDiagnosticsSubscriber(),
	new EfCoreDiagnosticsSubscriber(),
	new ElasticsearchDiagnosticsSubscriber());
```

After switching `supported_protocols: [TLSv1.3]` to `supported_protocols: [TLSv1.3,TLSv1.2]` the connection is established successfully and the APM agent works as expected. I'm guessing this is related to #1926.

#### Expected behavior
TLSv1.3 works same as it did in 1.19.0

#### Actual behavior
TLSv1.3 does not work

kuanpak:
I have the same issue after upgraded Elastic.Apm to 1.20+

monty241:
A possible solution might be to change in BackendCommUtils.cs from:
```
#if !NET462
[...]

PR #2625, opened 2025-07-14
Add TLSv1.3 support for .NET 3 or higher

JeremyBessonElastic:
For issue: https://github.com/elastic/apm-agent-dotnet/issues/2031

Add the proposed fix

### DIFF
--- a/src/Elastic.Apm/BackendComm/BackendCommUtils.cs
+++ b/src/Elastic.Apm/BackendComm/BackendCommUtils.cs
@@ -230,7 +230,6 @@ namespace Elastic.Apm.BackendComm
 			ServicePointManager.SecurityProtocol |= SecurityProtocolType.Tls12;
 			logger.Info()?.Log("CreateHttpClientHandler - SslProtocols: {SslProtocols}", ServicePointManager.SecurityProtocol);
 #endif
-
 #if !NET462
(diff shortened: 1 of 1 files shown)


Mpdreamz (approved):

src/Elastic.Apm/BackendComm/BackendCommUtils.cs
```diff
@@ -216,19 +216,27 @@ private static HttpClientHandler CreateHttpClientHandler(IConfiguration configur
 			// and runtime exceptions on .NET Framework versions <4.7.2, we don't attempt to set the certificate
 			// validation callback or SSL protocols on net462, which should also apply to .NET Framework <4.7.2 runtimes
 			// which resolve to that target.
+#if NET3_0_OR_GREATER
+			httpClientHandler.SslProtocols |= SslProtocols.Tls12;
```
Mpdreamz, line 220: I do wonder if we should just let this default to `SslProtocols.None` at least for .NET 8 and up. 

The defaults are good out of the box in .NET 

https://learn.microsoft.com/en-us/dotnet/api/system.security.authentication.sslprotocols?view=net-9.0

stevejgordon: Makes sense to me. @JeremyBessonElastic - Would you be okay to update and then I think we can merge.

stevejgordon (approved):

Issue #2312, opened 2024-03-21
[BUG]  Exceptions handled by IExceptionHandler<T> implementations are shown as Unhandled in APM UI


ben-wallis:
### APM Agent version
1.26.0

### Environment

**Operating system and version:**
N/A

**.NET Framework/Core name and version** 
.NET 8.0.2

**Application Target Framework(s)** (e.g. net462, netcoreapp3.1): 
net80

### Describe the bug
In .NET 8 ASP.NET Core applications, the new `IExceptionHandler<T>` exception handlers result in exceptions showing as `Unhandled` in the APM UI even when they return `true` to indicate the exception was handled.

### To Reproduce

A reproduction is available here: https://github.com/ben-wallis/apm-agent-dotnet/commit/be1ac9d63d9fda0055e87284e6b0572e6e2d0b2c

Even though `ArithmericExceptionHandler` in the above reproduction correclty handles the exception and returns `true` the exception is still shown on the Errors tab as Unhandled in APM:

![image](https://github.com/elastic/apm-agent-dotnet/assets/11904214/626a404a-88e8-4881-805e-9efbeae3da4d)

Exceptions handled by `IExceptionHandler<T>` implementations emit the `Microsoft.AspNetCore.Diagnostics.HandledException` diagnostic event, whereas those that are truly unhandled emit the `Microsoft.AspNetCore.Diagnostics.UnhandledException` diagnostic event. Despite this difference, both diagnostic events result in the exception being captured as unhandled in `AspNetCoreDiagnosticListener.cs` due to both of these diagnostic events being handled in the same code path: https://github.com/elastic/apm-agent-dotnet/blame/a82dba1e52b20f303c16a772c3fdfdbffcdbfb76/src/integrations/Elastic.Apm.AspNetCore/DiagnosticListener/AspNetCoreDiagnosticListener.cs#L77

#### Expected behavior
Handled exceptions are not shown as Unhandled in the APM UI

#### Actual behavior
Handled exceptions are shown as Unhandled in the APM UI

stevejgordon:
Thanks for raising this and including the repro @ben-wallis. We'll take a look at amending the logic to account for this.

JeremyBessonElastic:
Fixed with the new version

PR #2626, opened 2025-07-14
Add isHandled=true for handled exceptions

JeremyBessonElastic:
Have isHandled=true for handled exceptions
Have isHandled=false for unhandled exceptions

Issue: https://github.com/elastic/apm-agent-dotnet/issues/2312


Mpdreamz (approved):
Eek, bit of an oversight. Nice fix.

PR #2643, opened 2025-08-20
Add proxy apmserver

JeremyBessonElastic:
Add the possibility to add a server proxy for the APM server using new configurations: `ProxyUrl`, `ProxyUserName` and `ProxyPassword`.

### DIFF
--- a/src/Elastic.Apm/BackendComm/BackendCommUtils.cs
+++ b/src/Elastic.Apm/BackendComm/BackendCommUtils.cs
@@ -238,6 +238,9 @@ namespace Elastic.Apm.BackendComm
 			// apply to the whole AppDomain and that may not be desired. A consumer can set this themselves if they
 			// need custom validation behaviour.
 #endif
+
+			ConfigureHttpProxy(httpClientHandler, configuration);
+
 			return httpClientHandler;
 		}
 
@@ -288,6 +291,23 @@ namespace Elastic.Apm.BackendComm
 
 			// Replace invalid characters by underscore. All invalid characters can be found at
 			// https://github.com/dotnet/corefx/blob/e64cac6dcacf996f98f0b3f75fb7ad0c12f588f7/src/System.Net.Http/src/System/Net/Http/HttpRuleParser.cs#L41
+		}
+
+		private static void ConfigureHttpProxy(HttpClientHandler httpClientHandler, IConfiguration configuration)
+		{
+			var proxyOption = configuration.ProxyOption;
+
+			if (proxyOption.IsEnabled)
+			{
+				var proxy = new WebProxy(proxyOption.Url)
+				{
+					Credentials = new NetworkCredential(proxyOption.UserName, proxyOption.Password),
+
+				};
+
+				httpClientHandler.UseProxy = true;
+				httpClientHandler.Proxy = proxy;
+			}
 		}
 
 		private static string GetUserAgent(Service service)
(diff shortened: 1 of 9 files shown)


JeremyBessonElastic:
Doesn't seem worthwhile to add some additional unit-tests.

Probably some docs should be updated with the new config but I am not sure where.

src/Elastic.Apm/Config/AbstractConfigurationReader.cs
```diff
@@ -338,6 +338,21 @@ protected Uri ParseServerUrl(ConfigurationKeyValue kv)
 			return DefaultValues.ServerUri;
 		}
 
+		protected Uri ParseProxyServerUrl(ConfigurationKeyValue kv)
```
stevejgordon, line 341: I think we also need parser for username and password strings similar to `ParseServerCert` to check for null or empty strings.

test/Elastic.Apm.Tests/ConstructorTests.cs
```diff
@@ -155,6 +155,12 @@ private class LogConfiguration : IConfiguration, IConfigurationDescription
 
 		public bool UsePathAsTransactionName => ConfigConsts.DefaultValues.UsePathAsTransactionName;
 
+		public Uri ProxyUrl => ConfigConsts.DefaultValues.ProxyUrl;
```
stevejgordon, line 158: Rather than constants I think we could have a readonly autoproperty for these? e.g.

```suggestion
		public Uri ProxyUrl { get; }
```

src/Elastic.Apm/Config/ConfigConsts.cs
```diff
@@ -54,6 +54,9 @@ public static class DefaultValues
 			public const bool UsePathAsTransactionName = true;
 			public const bool VerifyServerCert = true;
 			public const string TraceContinuationStrategy = "continue";
+			public const Uri ProxyUrl = null;
```
stevejgordon, line 57: Not sure we need these constants if we update test/Elastic.Apm.Tests/ConstructorTests.cs

stevejgordon (changes requested):
A few small things to check/update, but this looks pretty good, thanks.

stevejgordon:
We should probaly also add to the test `ReadConfingsFromEnvVarsViaIConfig` in `ApmConfigurationTests` to validate the proxy settings are read correctly.

src/Elastic.Apm/BackendComm/BackendCommUtils.cs
```diff
@@ -290,6 +293,23 @@ internal static HttpClient BuildHttpClient(IApmLogger loggerArg, IConfiguration
...
 
+		private static void ConfigureHttpProxy(HttpClientHandler httpClientHandler, IConfiguration configuration)
+		{
+			var proxyUrl = configuration.ProxyUrl;
+
+			if (proxyUrl != null)
+			{
+				var proxy = new WebProxy(proxyUrl)
+				{
+					Credentials = new NetworkCredential(configuration.ProxyUserName, configuration.ProxyPassword),
```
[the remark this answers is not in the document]
JeremyBessonElastic: yes, done

src/Elastic.Apm/Config/ConfigConsts.cs
JeremyBessonElastic: yes we don't need. Just tried to follow a pattern but not the right one

test/Elastic.Apm.Tests/ConstructorTests.cs
JeremyBessonElastic: yes, done

stevejgordon (changes requested):
It's complaining about whitespace in docs/reference/configuration-on-asp-net-core.md, but other than that, this looks good, thanks.

PR #2653, opened 2025-09-02
docs: update apmschema url to use apm-data

kruskall:
apm schema was moved to apm-data a long time ago and the apm-server files are just a mirror

update the script to use apm-data files to we can stop mirroring them in apm-server

Related to https://github.com/elastic/apm-server/issues/17707


stevejgordon (approved):
LGTM, failing test can be ignored

PR #2656, opened 2025-09-04
Ensure we always start an Activity even when OTelBridge disabled

### LINKED ISSUE #2654: [BUG] Transaction `Activity` not created when OTel bridge is disabled.
Transaction `Activity` is not created on `NET` targets after #2455, when OTel bridge is disabled.

See https://github.com/elastic/apm-agent-dotnet/pull/2455#pullrequestreview-3176065835

stevejgordon:
The comments and previous implementation always started an activity for the transaction, even when the bridge is disabled. This enures we fix the regression in the behaviour.

This also ensures we only stop the Activity is we started it, which logically feels correct.

Fixes #2654


JeremyBessonElastic (approved):
LGTM

Issue #2301, opened 2024-02-23
Support for auto-instrumentation of Confluent Kafka .NET Client 2.x


stevejgordon:
The profiler-based instrumentation is currently enabled for v1.4.0 and newer, but only in the 1.x major version. We need to review the API for breaking changes and evaluate support for v2.x clients.

djosemartine:
Hi @stevejgordon, I wonder if there is any plan to support this shortly.

PR #2677, opened 2025-10-06
Support for auto-instrumentation of Confluent Kafka .NET Client 2.x

JeremyBessonElastic:
See title

#issue: https://github.com/elastic/apm-agent-dotnet/issues/2301


JeremyBessonElastic:
Documentation added

docs/reference/supported-technologies.md
```diff
@@ -93,6 +93,7 @@ We support automatic instrumentation for the following messaging systems
 | Azure.Messaging.ServiceBus 7.0.0+ |
 | Azure Queue Storage {applies_to}`apm_agent_dotnet: ga 1.10` | Azure.Storage.Queues 12.6.0+ | [NuGet package](/reference/setup-azure-storage.md) |
 | Kafka {applies_to}`apm_agent_dotnet: ga 1.12` | See profiler documentation | [Profiler auto instrumentation](/reference/setup-auto-instrumentation.md) |
+| Kafka {applies_to}`apm_agent_dotnet: ga v1.34.2` | Confluent.Kafka 2.11.1 | [NuGet package](/reference/setup-kafka.md) |
```
stevejgordon, line 96: We need to think about the ga version for this as technically, this NuGet approach doesn't depend on the most recent version of the APM agent. We might want to map this to the nearest release after they introduced that observability package?

stevejgordon (changes requested):
I think we still need to update the version info for Kafka in `setup-auto-instrumentation.md` as well as we now support 2.x there.

PR #2679, opened 2025-10-06
Add `netstandard2.1` target and update `DiagnosticSource` dependency

### LINKED ISSUE #2445: Update dependency of `System.Diagnostics.DiagnosticSource` to at least 6.0.0
We have encountered binding redirect issues on .NET Framework when using things like `Span<T>` in the library. We currently use 5.0.0, which is deprecated, and we should look to depend on a more recent version, which may unlock some of the binding redirect issues that would enable span use on .NET Framework paths, such as cookie redaction.

### LINKED ISSUE #2657: Ensure `NET8_0_OR_GREATER` features work on legacy runtimes (where possible)
We currently do not ship a `netstandard2.1` package for `Elastic.Apm`. This leads to issues where some features, such as OpenTelemetryBridge, which are behind a `NET8_0_OR_GREATER` pragama become unavailable on older runtimes when the library is updated. While those runtimes are out of support, we need to do our best not to break existing consumers who may choose to update the `Elastic.Apm` package to get bugs fixes.

I propose we introduce the `netstandard2.1` target which older runtimes (e.g. .NET 6) will resolve. When possible code compiled behind `#if NET_8_OR_GREATER` so be switched to `#if NET8_0_OR_GREATER || NETSTANDARD2_1`.

stevejgordon:
Fixes #2657
Fixes #2445

This adds a `netstandard2.1` target so that features such as OTel bridge are available on older runtimes, even though those are out of support. This prevents an update of our package, without a runtime update, breaking/removing features. Features may still not be 100% compatible with older unsupported runtimes, but should work in general.

As part of this, I updated the minimum dependency for `System.Diagnostics.DiagnosticSource` which is no generally 8.0.0. We still ship a version of `Elastic.Apm` and loader for the Startup hooks zip file which are built against 6.0.0. These support older unsupported runtimes although with no official testing.

Compiler pre-processor directives now prefer `NET` over a specific .NET version. We also add `NETSTANDARD2_1` so that features continue to work on older runtimes without a specific .NET target, i.e. 6 and 7.


stevejgordon:
I updated to the .NET 9 SDK and that I suspect is triggering the new formatting warnings, but I'm not entirely sure why. Will work on resolving those.

src/Elastic.Apm/OpenTelemetry/ElasticActivityListener.cs
```diff
@@ -135,21 +133,23 @@ private bool CreateTransactionForActivity(Activity activity, long timestamp, Lis
...
 					timestamp, true, activity.SpanId.ToString(),
-					distributedTracingData: dt, links: spanLinks, current: activity);
+					distributedTracingData: dt, links: spanLinks.Count > 0 ? spanLinks : null, current: activity);
 			}
 			else if (activity.ParentId == null)
 			{
 				transaction = _tracer.StartTransactionInternal(activity.DisplayName, "unknown",
 					timestamp, true, activity.SpanId.ToString(),
-					activity.TraceId.ToString(), links: spanLinks, current: activity);
+					activity.TraceId.ToString(), links: spanLinks.Count > 0 ? spanLinks : null, current: activity);
```
JeremyBessonElastic, line 142: It was intentional to change the code here?

src/Elastic.Apm/OpenTelemetry/ElasticActivityListener.cs
```diff
@@ -135,21 +133,23 @@ private bool CreateTransactionForActivity(Activity activity, long timestamp, Lis
...
 
-			if (transaction == null) return false;
+			if (transaction == null)
+				return false;
 
 			transaction.Otel = new OTel { SpanKind = activity.Kind.ToString() };
 
 			if (activity.Id != null)
-				_activeTransactions.AddOrUpdate(activity, transaction);
+				_activeTransactions.Add(activity, transaction);
```
JeremyBessonElastic, line 151: It was intentional to change the code here?

src/Elastic.Apm/OpenTelemetry/ElasticActivityListener.cs
```diff
@@ -159,15 +159,16 @@ private void CreateSpanForActivity(Activity activity, long timestamp, List<SpanL
 			if (_tracer.CurrentSpan == null)
 			{
 				newSpan = (_tracer.CurrentTransaction as Transaction)?.StartSpanInternal(activity.DisplayName, "unknown",
-					timestamp: timestamp, id: activity.SpanId.ToString(), links: spanLinks, current: activity);
+					timestamp: timestamp, id: activity.SpanId.ToString(), links: spanLinks.Count > 0 ? spanLinks : null, current: activity);
```
JeremyBessonElastic, line 162: It was intentional to change the code here?

src/Elastic.Apm/OpenTelemetry/ElasticActivityListener.cs
```diff
@@ -178,7 +179,7 @@ private void CreateSpanForActivity(Activity activity, long timestamp, List<SpanL
 			}
 
 			if (activity.Id != null)
-				_activeSpans.AddOrUpdate(activity, newSpan);
+				_activeSpans.Add(activity, newSpan);
```
JeremyBessonElastic, line 182: It was intentional to change the code here?

src/Elastic.Apm/OpenTelemetry/ElasticActivityListener.cs
stevejgordon: Good spot! In this case, yes, it's not related to the main issues but I noticed while tested we serialise an emtpy array which is unneccessary when span links (not frequently used) are not set.

src/Elastic.Apm/OpenTelemetry/ElasticActivityListener.cs
```diff
@@ -159,15 +159,16 @@ private void CreateSpanForActivity(Activity activity, long timestamp, List<SpanL
...
 			{
 				newSpan = (_tracer.CurrentTransaction as Transaction)?.StartSpanInternal(activity.DisplayName, "unknown",
-					timestamp: timestamp, id: activity.SpanId.ToString(), links: spanLinks, current: activity);
+					timestamp: timestamp, id: activity.SpanId.ToString(), links: spanLinks.Count > 0 ? spanLinks : null, current: activity);
 			}
 			else
 			{
 				newSpan = (_tracer.CurrentSpan as Span)?.StartSpanInternal(activity.DisplayName, "unknown",
-					timestamp: timestamp, id: activity.SpanId.ToString(), links: spanLinks, current: activity);
+					timestamp: timestamp, id: activity.SpanId.ToString(), links: spanLinks.Count > 0 ? spanLinks : null, current: activity);
```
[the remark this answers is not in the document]
stevejgordon: Good spot! In this case, yes, it's not related to the main issues but I noticed while tested we serialise an emtpy array which is unneccessary when span links (not frequently used) are not set.

stevejgordon:
Working on the test failures. The IIS ones suggest we might still have to ship an older DiagnosticsSource build too, although I need to investigate why exactly based on our packaging process.

PR #2712, opened 2026-03-02
Add *.ttf to default ignore list

wast:


stevejgordon (commented):
Looks good, thanks. We should probably cover otf too. Would you mind adding that here. If not I can make it a follow up.

Apologies for the long delay on getting to this!

stevejgordon (dismissed):

stevejgordon (approved):

trentm (approved):

PR #2752, opened 2026-05-21
Include targets for net472 in the profiler zip

### LINKED ISSUE #2530: [BUG] Profiler auto-instrumentation on .NET Framework 4.7.2+ doesn't honour the `VerifyServerCert` configuration
Our agent targets different versions of .NET, and specific code paths apply to each target for certain functionality. One of these is the handling for `ELASTIC_APM_VERIFY_SERVER_CERT`, which behaves differently on older .NET Framework versions (< 4.6.2) due to limitations there. 

Our profiler zip asset only ships the base `net462` libraries. Therefore, when using auto-instrumentation with the profiler, some features, including the verify server certificate configuration option, are not applied, even when the instrumented application runs on .NET 4.7.2 or newer.

We need to update the build scripts and release workflows to ensure that the `net472` libraries are included in the zip file and resolved by the managed loader.

stevejgordon:
Fixes #2530

### DIFF
--- a/build/scripts/Build.fs
+++ b/build/scripts/Build.fs
@@ -124,7 +124,7 @@ module Build =
     /// When running on Windows and not CI, also runs MSBuild Build on .NET Framework
     let Build () =
         // On Windows, pre-build this first to prevent a parallel build race condition: both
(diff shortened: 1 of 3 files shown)


hegerchr (approved):
LGTM

PR #2749, opened 2026-05-20
Update docs to reflect actual behaviour

stevejgordon:
Fixes #2488

PR #2757, opened 2026-05-21
Update docs with programatic configuration

stevejgordon:


stevejgordon:
Depends on #2749 merging so we can rebase and get this to build.

LikeTheSalad (approved):

PR #2770, opened 2026-06-18
fix: reuse serialization buffer in PayloadSenderV2 to eliminate LOH fragmentation

nimanikoo:
We've been seeing steady process memory growth in long-running deployments (visible after ~2 weeks uptime).
The root cause: `ProcessQueueItems` was allocating a `new MemoryStream(1024)` on every flush batch.
A `MemoryStream` doubles its internal byte array as it grows, and for batches with many spans
the final buffer crosses the 85 KB Large Object Heap threshold. LOH is not compacted during
ordinary GC cycles, so each flush permanently fragments heap and over millions of batches
the process RSS just keeps climbing.

**The fix** is simple: allocate one `MemoryStream` as a field, and reset it with `SetLength(0)`
before each batch instead of creating a new one. `SetLength(0)` resets the logical length and
write cursor without releasing the underlying buffer, so the same allocation is reused forever.
This is safe because `ProcessQueueItems` is always called from a single dedicated background
thread (`ElasticApmPayloadSenderV2`), so there's no concurrency concern.

One subtlety caught during testing: `StreamContent` disposes its underlying stream when the
`using` block exits. The fix wraps the reusable buffer in a lightweight, non-owning
`MemoryStream` view before handing it to `StreamContent`, so the shared buffer is never
closed between batches.

**Benchmark results (Apple M4, .NET 8, BenchmarkDotNet):**

| Batch size | Allocated (old) | Allocated (new) | Gen1 GC (old) | Gen1 GC (new) |
|------------|-----------------|-----------------|---------------|---------------|
| 5 spans    | 28.4 KB         | 21.2 KB         | 0.11          | 0.05          |
| 20 spans   | 76.0 KB         | 60.8 KB         | 0.52          | 0.15          |
| 50 spans   | 171.2 KB        | 140.0 KB        | **1.71**      | **0.00**      |

At production-scale batch sizes, Gen1 (and with it, LOH-adjacent promotion pressure) drops
to zero. In a high-throughput service flushing thousands of batches per minute, this is the
difference between stable and ever-growing resident memory.

**Testing:** added `SequentialBatches_SerializationBufferIsIsolated` to guard that buffer reuse
never bleeds content from one batch into the next.

### DIFF
--- a/benchmarks/Elastic.Apm.Benchmarks/PayloadSenderSerializationBenchmarks.cs
+++ b/benchmarks/Elastic.Apm.Benchmarks/PayloadSenderSerializationBenchmarks.cs
@@ -1,0 +1,168 @@
+// Licensed to Elasticsearch B.V under one or more agreements.
+// Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information
+
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using BenchmarkDotNet.Attributes;
+using Elastic.Apm.Api;
+using Elastic.Apm.Logging;
+using Elastic.Apm.Metrics;
+using Elastic.Apm.Model;
+using Elastic.Apm.Report.Serialization;
+using Elastic.Apm.Tests.Utilities;
+
+namespace Elastic.Apm.Benchmarks;
+
+/// <summary>
+/// Measures the allocation cost of the serialization buffer strategy used by
+/// <see cref="Elastic.Apm.Report.PayloadSenderV2"/> when flushing a batch of APM events.
+///
+/// <para>
+/// <b>Old approach</b>: <c>new MemoryStream(1024)</c> allocated on every batch.
+/// As the stream grows it doubles its internal <c>byte[]</c>, producing a chain of
+/// short-lived arrays.  For batches larger than 85 KB the final array lands in the
+/// Large Object Heap (LOH), which is not compacted during ordinary GC cycles.
+/// Over a long-running deployment this causes LOH fragmentation and sustained
+/// process-level memory growth.
+/// </para>
+///
+/// <para>
+/// <b>New approach</b>: a single <c>MemoryStream</c> is allocated once and reset with
+/// <c>SetLength(0)</c> before each batch.  The internal buffer is reused, so no new
+/// heap objects are created per batch.
+/// </para>
+///
+/// <para>
+/// Key metrics to watch in the results:
+/// <list type="bullet">
+///   <item><b>Allocated</b> – bytes per operation.
+///         OldApproach ≈ serialised-payload-size (plus MemoryStream overhead).
+///         NewApproach ≈ 0 (no buffer allocation).</item>
+///   <item><b>Gen0 / Gen1 / Gen2</b> – GC pressure.
+///         For <c>SpanCount = 50</c> the old approach generates Gen2 collections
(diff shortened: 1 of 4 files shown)


test/Elastic.Apm.Tests/BackendCommTests/PayloadSenderTests.cs
```diff
@@ -558,6 +558,73 @@ await payloadSender.EnqueueEventInternal(
 			sendCount.Should().Be(1, "FlushAsync must return only after the HTTP POST completes");
 		}
 
+		/// <summary>
+		/// Regression guard for the serialization-buffer reuse optimisation.
+		///
```
stevejgordon: nit - remove the blank lines in this XML doc comment

benchmarks/Elastic.Apm.Benchmarks/PayloadSenderSerializationBenchmarks.cs
```diff
@@ -0,0 +1,168 @@
...
+using Elastic.Apm.Model;
+using Elastic.Apm.Report.Serialization;
+using Elastic.Apm.Tests.Utilities;
+
+namespace Elastic.Apm.Benchmarks;
+
+/// <summary>
+/// Measures the allocation cost of the serialization buffer strategy used by
+/// <see cref="Elastic.Apm.Report.PayloadSenderV2"/> when flushing a batch of APM events.
+///
```
stevejgordon: nit - remove the blank lines in this XML doc comment (several places)

stevejgordon (changes requested):
Thanks @nimanikoo for identifying and fixing this. The reasoning is sound and the fix looks appropriate. Code-wise, I'm happy to take this as-is. My slight concern is keeping the large buffer if it grows which may mean one large payload causes a large retained allocation than we generally need. But that concern is likely small and I'll review if we want to make any followup changes to resize. 

A couple of minor nits on the styling of the XML docs comments and this is good to merge once CI checks pass.

EDIT: Looks like you may need to run `dotnet format` over this to fix some of the whitespace etc.

test/Elastic.Apm.Tests/BackendCommTests/PayloadSenderTests.cs
nimanikoo: thanks
resolved , commit : d2d7548d661a31a60d967d226ba42af9d6e556ca

benchmarks/Elastic.Apm.Benchmarks/PayloadSenderSerializationBenchmarks.cs
nimanikoo: thanks
resolved , commit : 832267a640995d27f04867cc511c19cafd0a78e3

nimanikoo:
@stevejgordon 
Thanks for the thorough review and the valuable feedback!

I appreciate you taking the time to look into both the implementation and the potential memory implications. I'll address the XML doc styling comments and run `dotnet format` to fix the remaining formatting issues.

I'm looking forward to contributing more meaningful improvements to the project after this PR . This was my first contribution and more of a starting point, and I'm excited to continue getting involved and helping wherever I can.

Thanks again

stevejgordon (changes requested):
Thanks again @nimanikoo. Looks good, but could you please exclude the LICENSE file from the commit. It gets regenerated and often shows as a change we don't really need.

stevejgordon (approved):

stevejgordon:
run docs-build

PR #2781, opened 2026-07-09
Fix : insert span link duplicate and double copy

nimanikoo:
### Problem

`InsertSpanLinkInternal` is called by the Azure Service Bus instrumentation when a received message batch carries distributed tracing headers. Two issues were found:

---

#### Bug — `Span.InsertSpanLinkInternal` (missing `else`)

```csharp
// Old — BUG: no else, falls through and adds items a second time
if (Links == null || !Links.Any())
    Links = spanLinks;         // sets Links = [A, B, C]
// always executes:
var newList = new List<SpanLink>(Links);  // [A, B, C]
newList.AddRange(spanLinks);             // [A, B, C, A, B, C] ← duplicates
Links = new List<SpanLink>(newList);     // second unnecessary copy
```

When `Links` was initially `null` (the common first-call case), every link was recorded twice. This means a span consuming from an Azure Service Bus batch would report duplicate upstream message links to APM Server, corrupting the distributed trace view.

---

#### Waste — `Transaction.InsertSpanLinkInternal` (double copy)

The `else` branch was logically correct but built `newList` and then immediately wrapped it in `new List<SpanLink>(newList)` — a second allocation that copied the list for no reason.

---

### Fix

The fix is identical for both: early return on the null/empty path (eliminating the fall-through entirely), and direct assignment of the newly built list on the append path:

```csharp
internal void InsertSpanLinkInternal(IEnumerable<SpanLink> links)
{
    if (Links == null || !Links.Any())
    {
        Links = links as SpanLink[] ?? links.ToArray();
        return;
    }
    var newList = new List<SpanLink>(Links);
    newList.AddRange(links);
    Links = newList;
}
```

---

### Benchmark results

> Apple M4 · .NET 8.0.19 · BenchmarkDotNet v0.13.5

| Method | LinkCount | Mean | Ratio | Allocated | Alloc Ratio |
|---|---|---:|---:|---:|---:|
| Old: Span first-call (null Links) — duplicates | 10 | 62.444 ns | base | 536 B | base |
| New: Span first-call — correct, zero alloc | 10 | 3.957 ns | -94% | — | -100% |
| Old: Transaction append — unnecessary second copy | 10 | 56.729 ns | base | 320 B | base |
| New: Transaction append — single List, no copy | 10 | 40.557 ns | -35% | 176 B | -67% |

The Span path is 94% faster with zero allocations because real callers (Azure Service Bus) already pass `SpanLink[]` arrays — the early return simply stores the reference.

---

### Changes

| File | Description |
|---|---|
| `src/Elastic.Apm/Model/Span.cs` | Fix missing `else` (bug fix) |
| `src/Elastic.Apm/Model/Transaction.cs` | Remove pointless second list copy |
| `test/Elastic.Apm.Tests/ApiTests/ApiTests.cs` | 4 regression tests covering null/empty first-call and append for both `Span` and `Transaction` |
| `benchmarks/Elastic.Apm.Benchmarks/SpanLinkInsertBenchmarks.cs` | New benchmark |

**Test suite: 2032 passed, 0 failed**


nimanikoo:
docs-build

test/Elastic.Apm.Tests/ApiTests/ApiTests.cs
```diff
@@ -1276,6 +1276,101 @@ public void MultipleSpanLinksOnChildSpan()
...
+		public void Span_InsertSpanLinkInternal_WhenLinksInitiallyNull_DoesNotDuplicate()
+		{
+			var payloadSender = new MockPayloadSender();
+			using var agent = new ApmAgent(new TestAgentComponents(payloadSender: payloadSender));
+
+			var transaction = agent.Tracer.StartTransaction("tx", "test");
+			var span = transaction.StartSpan("op", "test") as Model.Span;
+			span.Should().NotBeNull();
+
+			var link1 = new SpanLink("aaaa0000000000000000000000000001", "bbbb000000000000000000000000000000000001");
```
stevejgordon: The spanId values here are 32 hex chars and the traceId values are 40 hex chars, but the W3C trace-context spec requires 16 and 32 respectively. SpanLink has no format validation today so the tests pass, but they'll break if validation is ever added.

src/Elastic.Apm/Model/Span.cs
```diff
@@ -672,13 +672,14 @@ private bool TryToCompressComposite(Span sibling)
 
 		internal void InsertSpanLinkInternal(IEnumerable<SpanLink> links)
 		{
-			var spanLinks = links as SpanLink[] ?? links.ToArray();
 			if (Links == null || !Links.Any())
-				Links = spanLinks;
-
+			{
+				Links = links as SpanLink[] ?? links.ToArray();
```
stevejgordon: I think we should probably `links.ToArray()` unconditionally to ensure we always have a defensive copy in case the incoming array is mutated.

src/Elastic.Apm/Model/Span.cs
```diff
@@ -672,13 +672,14 @@ private bool TryToCompressComposite(Span sibling)
 
 		internal void InsertSpanLinkInternal(IEnumerable<SpanLink> links)
```
stevejgordon, line 705: We now have identical copies of this in two files (which is why the original bug slipped in). Could this be refactored into a shared helper to avoid future drift?

stevejgordon (changes requested):
Sorry for the delay. A couple of minor things to suggest. The CLA issue is because one of the commits includes `Co-authored-by: Copilot Autofix powered by AI <175728472+Copilot@users.noreply.github.com>`. Can you rebase to avoid that commit, or amend it's message at least. I think that will make CLA happy.

test/Elastic.Apm.Tests/ApiTests/ApiTests.cs
nimanikoo: https://github.com/elastic/apm-agent-dotnet/pull/2781/changes/7c0056cf885ed8f5c56e9f6208ef56bcf94b22bf

thanks , fixed

src/Elastic.Apm/Model/Span.cs
nimanikoo: https://github.com/elastic/apm-agent-dotnet/pull/2781/commits/f985fa0dacc387e967985d9359440bfacf6d830d

fixed

src/Elastic.Apm/Model/Span.cs
nimanikoo: https://github.com/elastic/apm-agent-dotnet/pull/2781/commits/dc2bd53879b056b309a8c17297e8b985ae22ebb2

Done. I extracted the shared merge logic into `ExecutionSegmentCommon.InsertSpanLinks` and updated both `Span` and `Transaction` to use it.

This keeps the defensive-copy behavior in one place and prevents the two implementations from drifting apart again. Verified with clean builds and the span-link tests passing (8/8).


@stevejgordon

src/Elastic.Apm/Model/ExecutionSegmentCommon.cs
```diff
@@ -365,5 +366,23 @@ internal static void CaptureErrorLog(ErrorLog errorLog, IPayloadSender payloadSe
...
 			payloadSender.QueueError(error);
 		}
+
+		/// <summary>
+		/// Merges <paramref name="additionalLinks" /> into <paramref name="existingLinks" /> and returns the value to
+		/// assign back to the span's or transaction's <c>Links</c> property.
+		/// When there are no existing links the input is defensively copied (the caller may keep and later mutate the
+		/// array it passed in); otherwise a single combined list is produced.
+		/// Shared by <see cref="Span.InsertSpanLinkInternal" /> and <see cref="Transaction.InsertSpanLinkInternal" /> so
+		/// the two implementations cannot drift apart.
```
stevejgordon: I'm not sure we need this detail about the implementation and usage as that too can drift.

benchmarks/Elastic.Apm.Benchmarks/SpanLinkInsertBenchmarks.cs
stevejgordon, line 1: I think we can delete this benchmark as it's served it's purpose and may drift from real code in the future.

stevejgordon (changes requested):
A couple of other items and then we do need to rebase to remove the CoPilot author affecting CLA. I can help with that if you want as I should be able to push on this forked branch?

nimanikoo:
The branch was already rebased onto the latest main and the Copilot-Autofix commit removed (force-pushed a while back) , the current 4 commits are all authored by me with no Co-authored-by trailers. Could you re-trigger the CLA check? It looks like it's still evaluating an old commit SHA. Happy to push an empty commit to kick it if that helps. And yes, send over the other review items.

@stevejgordon

benchmarks/Elastic.Apm.Benchmarks/SpanLinkInsertBenchmarks.cs
nimanikoo: https://github.com/elastic/apm-agent-dotnet/pull/2781/changes/c73a6d68263abbcbe2206c1c9a1a3b7dcd9c4a1d

Thanks , done both:

Shared helper doc: trimmed to just the contract (appends the links; the returned collection never aliases the caller's input). Dropped the per-branch walkthrough and the caller list, since those duplicate code that can drift.
SpanLinkInsertBenchmarks: removed. The bug and the fix are now covered by the _InsertSpanLinkInternal_ API tests, so the benchmark's job is done and there's no near-copy of the logic left to drift.

src/Elastic.Apm/Model/ExecutionSegmentCommon.cs
nimanikoo: https://github.com/elastic/apm-agent-dotnet/pull/2781/changes/c73a6d68263abbcbe2206c1c9a1a3b7dcd9c4a1d

 Thanks , done both:

Shared helper doc: trimmed to just the contract (appends the links; the returned collection never aliases the caller's input). Dropped the per-branch walkthrough and the caller list, since those duplicate code that can drift.
SpanLinkInsertBenchmarks: removed. The bug and the fix are now covered by the _InsertSpanLinkInternal_ API tests, so the benchmark's job is done and there's no near-copy of the logic left to drift.

stevejgordon:
run docs-build

nimanikoo:
@stevejgordon 
Thanks for the review and for all the feedback

I’ve fixed all the issues and comments you pointed out, and I’ve also updated the branch with the latest changes. I believe it’s ready for another review now.

If there’s anything else you have in mind or any other points you’d like me to address, please let me know and I’ll fix them.

stevejgordon:
@nimanikoo - Thanks for taking the effort and time to submit this PR. Unfortunately, we cannot accept this at this time and therefore I will close this PR.

PR #2791, opened 2026-08-16
Perf: Replace inefficient LINQ usage in WildcardMatcher

nkindberg:
The wildcard matcher has inefficient usages of LINQ `ElementAt`. These calls result in enumeration instead of indexing directly in the string. They also cause significant garbage collection pressure by allocating `CharEnumerator` instances on every invocation on hot paths.


stevejgordon:
run docs-build

stevejgordon (approved):
Looks good! Thanks for the contribution.

